PR #11, opened 2017-10-26
add recipes for HPolygon

mforets:
need to consider about the save to file feature that we used. how does it work with Plots? document it.


docs/src/lib/representations.md
```diff
@@ -44,6 +42,13 @@ LinearConstraint
 Line
 ```
 
+## Hyperrectangle
```
mforets: (this is irrelevant for this issue, srry)

schillic:
PR11 and PR12 look good from the outside (I did not check them out).
Please do the merges.

Issue #19, opened 2017-11-01
Change type unions to common supertype


schillic:
Quite often we have the following code patterns:
```julia
d::Union{Vector{Float64}, SparseVector{Float64,Int64}}
M::Union{Matrix{Float64}, SparseMatrixCSC{Float64,Int64}}
```
We should use instead the common supertype:
```julia
d::AbstractVector{Float64}
M::AbstractMatrix{Float64}
```
This is easier to read and gives more flexibility (e.g., allow `SubArray`s if not already mentioned).
I do not expect negative performance impact; we should of course validate this.

mforets:
yep, i agree that we should do this! please go ahead with these changes

mforets:
you can either push to master directly, or we get used to filing a new PR each time, so we can detect if sth was broken

schillic:
There will be a merge conflict with #18, so I would like to finish that one first.

PR #21, opened 2017-11-02
#19 - change type unions to common supertype

schillic:
This also fixes the recent additional unit test for `LinearMap` (I do not know why it did not work; when I tried it locally it worked; I must have done another change).

I also took the liberty and replaced `Array{X, 1}` by `Vector{X}` everywhere.

### DIFF
--- a/src/Ball2.jl
+++ b/src/Ball2.jl
@@ -26,7 +26,7 @@ We evaluate the support vector in a given direction:
 
 ```julia
 julia> σ(ones(5), B)
-5-element Array{Float64,1}:
+5-element Vector{Float64}:
 0.06742
 0.13484
 0.20226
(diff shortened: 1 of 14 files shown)


src/Ball2.jl
```diff
@@ -26,7 +26,7 @@ We evaluate the support vector in a given direction:
 
 ```julia
 julia> σ(ones(5), B)
-5-element Array{Float64,1}:
+5-element Vector{Float64}:
```
mforets: well if this was a doctest (instead of putting `julia`, write `@jldoctest` as explained in Documenter.jl, then it should be exactly the same as in julia REPL's output, i guess.. i haven't checked, though, we can leave it like that and see when we add it as a doctest.

schillic: Ah, now I understand your point here.
I agree, this will probably break.
I fixed that directly in master.

PR #22, opened 2017-11-02
type SparseMatrixExp in ExponentialMap.jl

mforets:


src/ExponentialMap.jl
```diff
@@ -22,26 +22,26 @@ This class is provided for use with very large and very sparse matrices. The
 evaluation of the exponential matrix action over vectores relies on the
 Expokit package. 
 """
-mutable struct SparseMatrixExp
-    M::SparseMatrixCSC{Float64,Int64}
+mutable struct SparseMatrixExp{Float64}
+    M::SparseMatrixCSC{Float64, 2}
```
schillic: Should it not be something like that?
```julia
mutable struct SparseMatrixExp{T}
    M::SparseMatrixCSC{T, 2}
```

schillic:
I commented the commit.

src/ExponentialMap.jl
mforets: or even 

```julia
mutable struct SparseMatrixExp{T} where {T<:AbstractFloat}
    M::SparseMatrixCSC{T, 2}
end
```
 ?

mforets: `mutable struct` yes or no?

schillic: Yes, we should use `where {T<:AbstractFloat}`.

> mutable struct yes or no?

I think "no", unless we need to replace the array.

schillic: The correct syntax is:
```julia
struct SparseMatrixExp{T<:AbstractFloat}
```

schillic: On a side note, what is the second argument for `SparseMatrixCSC`?
You changed it to `2`.
Sometimes it is `Int64`.
Sometimes it does not have a second type argument.
Should a matrix not be 2D by definition?

schillic: > what is the second argument for `SparseMatrixCSC`?

My question is answered [here](https://docs.julialang.org/en/stable/manual/arrays/#man-csc-1).
I presume that we should always use `Int64`?

mforets: ye, though we could probably use `Int32` as well. 

thanks fr helping out with this, i've pushed the updates.

PR #24, opened 2017-11-04
save dim field in polyhedron

mforets:
(cf. gl issue 137)


mforets:
oops, an unrelated commit for the docs also got in here, srry.

PR #26, opened 2017-11-08
add andrew scan

nikos-kekatos:


src/andrew_scan.jl
```diff
@@ -0,0 +1,42 @@
+points = [[0.9,0.2], [0.4,0.6], [0.2,0.1], [0.1,0.3], [0.3,0.28]] # as a list of lists
```
mforets: we shall remove this line,  but add this to the tests

src/andrew_scan.jl
```diff
@@ -0,0 +1,42 @@
...
+- `points` -- array of vectors containing the 2D coordinates of the points
+
+### Algorithm
+
+This function implements Andrew's monotone chain convex hull algorithm, that constructs
+the convex hull of a set of ``n`` 2D points in ``O(n log n)`` time.
+For further details see:
+[Monotone chain](https://en.wikibooks.org/wiki/Algorithm_Implementation/Geometry/Convex_hull/Monotone_chain)
+"""
+function andrew_monotone_chain(points::Vector{Vector{T}}) where {T<:AbstractFloat}
```
mforets: add this function to the docs and add a doctest

src/andrew_scan.jl
```diff
@@ -0,0 +1,42 @@
...
+
+    points = [sortrows(hcat(points...)')[i, :] for i in 1:length(points)]
+
+    # Build lower hull
+    lower = Vector{Vector{Float64}}()
+    for i in 1:length(points)
```
mforets: since we don't need `i`, now we shall be able to do `for p in points`

mforets:
also we need to include this function in `LazySets`

PR #27, opened 2017-11-09
Revise HPolygon, HPolygonOpt, VPolygon

mforets:


src/VPolygon.jl
```diff
@@ -0,0 +1,112 @@
...
+- `P` -- a polygon given in vertex representation
+
+### Output
+
+List of vertices as an array of vertex pairs, Vector{Vector{Float64}}.
+"""
+function vertices_list(P::VPolygon)::Vector{Vector{Float64}}
+    return P.vertices_list
+end
+#=
```
schillic: Why is this kept here? It is already added in `HPolygon.jl`.
Also: Should we write `plot_polygon` everywhere (with lowercase "p")?

mforets: why not? it's the `vertices_list` of a `VPolygon`.

the commented code is garbage; i'll fix it.

yep, `plot_polygon` everywhere.

PS: still i haven't worked out the new support function for this `VPolygon` type.

schillic: Yes, I meant the code that follows, Github shows the three preceding lines...

PR #33, opened 2017-11-12
Special set types

mforets:


schillic:
Do we really want to have the `DummySet` in this library?

mforets:
i'm not sure -- ok let's move on with EmptySet and ZeroSet here and leave that one in stand-by, i'll put it in another branch

src/CartesianProduct.jl
```diff
@@ -59,6 +59,54 @@ function *(X::LazySet, Y::LazySet)::CartesianProduct
...
+```
+
+Right multiplication of a set by an empty set.
+
+### Input
+
+- `X` -- a convex set
+- `E` -- an empty set
+
+## Output
```
schillic: only two "#" (also below and in `MinkowskiSum.jl`, probably copy-paste)...

schillic:
I was thinking that, since all instances of `EmptySet` are identical, we could have one central instance and in the constructor always return this instance (aka [singleton pattern](https://en.wikipedia.org/wiki/Singleton_pattern)).

mforets:
ok. notice that `\emptyset` returns an instance, see eg. the doctests. does it do the job?

schillic:
> ok. notice that \emptyset returns an instance, see eg. the doctests. does it do the job?

Yes, perfect 👍

src/MinkowskiSum.jl
```diff
@@ -77,6 +77,10 @@ Minkowski sum.
 """
 +(∅::EmptySet, X::LazySet) = EmptySet()
```
schillic: Return `∅` instead of creating a new instance.

schillic: There are four occurrences of `EmptySet()` in this file and eight occurrences in `CartesianProduct.jl`.

schillic: In fact the Julia compiler seems to unify all instances of this type automatically, but it is still much faster to use `∅`:
```julia
julia> @time EmptySet();
  0.000056 seconds (4 allocations: 160 bytes)
julia> @time ∅;
  0.000003 seconds (4 allocations: 160 bytes)
```

schillic: I am beginning to wonder if `∅` is really a constant.
```julia
julia> expand(∅)
LazySets.EmptySet()
```

schillic: `EmptySet()` always returns the same instance as well.
```julia
object_id(EmptySet())
0x34e36c91d4a34027
object_id(EmptySet())
0x34e36c91d4a34027
```
Then I do not understand why `∅` is faster...

schillic:
`MethodError`: ambiguous methods for `∅+∅`, `∅*∅`, `ZeroSet(2)+ZeroSet(2)`

schillic:
No further complaints from my side 👍

PR #34, opened 2017-11-13
add funcs for norms, radius, diameter

mforets:


mforets:
i'd like to add more doctests for some of these functions

schillic:
Looks good.
I noticed that we could write `p::Real=Inf` (this is what `?norm` uses).
And maybe we want to return the value of `p` in the error message: "... for `p`=$p is not implemented")

mforets:
ok!
last todo: add tests for norm, radius, diameter of `Hyperrectangle` and `BallInf`

Issue #9, opened 2017-10-13
Accept parametric number types


mforets:
It can be interesting to accept parametric number types so that we can work with:

- rationals (slower but exact)
- single precision (less precise but faster)
- complex numbers (eg. for [complex zonotopes](https://sites.google.com/site/arvind23adi/research/complex-zonotope))

schillic:
This is also suggested in a [Julia design style guide section](https://docs.julialang.org/en/stable/manual/style-guide/#Avoid-using-floats-for-numeric-literals-in-generic-code-when-possible-1).

schillic:
Then again, this is frowned upon in [the performance tips](https://docs.julialang.org/en/stable/manual/performance-tips/#Avoid-containers-with-abstract-type-parameters-1).

mforets:
> Then again, this is frowned upon in the performance tips.

that's probably about

```julia
struct Ball2 <: LazySet
    center::Vector{AbstractFloat}
    radius::AbstractFloat
    Ball2(center, radius) = radius < 0. ? throw(DomainError()) : new(center, radius)
end
```
(bad idea), versus

```julia
struct Ball2{T} <: LazySet where {T<:Real}
    center::Vector{T}
    radius::T
    Ball2(center, radius) = radius < 0. ? throw(DomainError()) : new(center, radius)
end
```
which won't impact performance and still be generic.

see also: [julia gotchas handle](http://www.stochasticlifestyle.com/7-julia-gotchas-handle/), and [package example](https://github.com/ChrisRackauckas/ExamplePackage.jl/blob/master/src/output_string.jl)

schillic:
What would be a good type abbreviation? For instance `N`umber or `F`loat?
We are not bound to a single letter, but I think it is convenient.

mforets:
i've seen a lot of `N` for number types. for a case such as say, `Ball2` or `BallInf` we can use `N <: Real`: this includes `AbstractFloat` and also `Rational`'s.

for operators like `CartesianProduct`, `ConvexHull` we don't need to add anything at all, right? seems like there are other operators such as `LinearMap` where we should better use `AbstractMatrix{N}` and `where{N}` on the type's signature, that is without saying anything (`Any`). this is because i have in mind linear maps which contains interval coefficients (or even `LazySets`).

schillic:
Correct, for now I only add the type argument to base set types.
And I chose `N`, very good :)

[...]

PR #37, opened 2017-11-17
#9 - add parametric number types

schillic:
This PR is not finished yet and works just as discussion base.

### DIFF
--- a/src/ConvexHull.jl
+++ b/src/ConvexHull.jl
@@ -49,7 +49,7 @@ Return the support vector of a convex hull in a given direction.
 - `d`  -- direction
 - `ch` -- the convex hull of two sets
 """
-function σ(d::AbstractVector{<:Real}, ch::ConvexHull)::Vector{<:Real}
+function σ(d::AbstractVector{<:Real}, ch::ConvexHull)::AbstractVector{<:Real}
     σ1 = σ(d, ch.X)
     σ2 = σ(d, ch.Y)
     ρ1 = dot(d, σ1)
--- a/src/ConvexHull.jl
+++ b/src/ConvexHull.jl
@@ -18,7 +18,9 @@ Type that represents the convex hull of the union of two convex sets.
 struct ConvexHull{T1<:LazySet,T2<:LazySet} <: LazySet
     X::T1
     Y::T2
-    ConvexHull{T1,T2}(X::T1, Y::T2) where {T1<:LazySet,T2<:LazySet} = dim(X) != dim(Y) ? throw(DimensionMismatch) : new(X, Y)
+
+    ConvexHull{T1,T2}(X::T1, Y::T2) where {T1<:LazySet,T2<:LazySet} =
+        dim(X) != dim(Y) ? throw(DimensionMismatch) : new(X, Y)
 end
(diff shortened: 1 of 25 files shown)


schillic:
We still need to
- add the missing code as suggested [here](https://github.com/JuliaReach/LazySets.jl/issues/9#issuecomment-345074278),
- add some test cases,
- review calls such as mentioned [here](https://github.com/JuliaReach/LazySets.jl/issues/9#issuecomment-345048415).

src/ConvexHull.jl
```diff
@@ -47,15 +49,10 @@ Return the support vector of a convex hull in a given direction.
 - `d`  -- direction
 - `ch` -- the convex hull of two sets
 """
-function σ(d::AbstractVector{Float64}, ch::ConvexHull)::Vector{Float64}
+function σ(d::AbstractVector{<:Real}, ch::ConvexHull)::Vector{<:Real}
```
mforets: here why not `function σ(d::AbstractVector, ch::ConvexHull)` ?

schillic: Counter question:
Why not check that the input is of real number type and the output is a) a `Vector` and b) of real number type?

mforets: alright, we *don't check* that input/output number types are the same here, for that we would need to use `N` as well. 

let's go with `<:Real` then

mforets: for the output type, i just saw that if we want to allow sparse vectors for input (`d`) then it should also be a `AbstractVector`.

mforets: e.g. we want that `σ(sparsevec([1],[1.0],2), BallInf(ones(2),1.0)` works -- returning a sparsevector in some cases. maybe not in all cases

schillic: Yes, I was also thinking about that. Sometimes we have `AbstractVector` and sometimes we have `Vector`.
The user can of course still expect an `AbstractVector`.

If the implementation ensures that the type is always a `Vector`, everything will work, but I still prefer to keep the return type more general so that a caller does not rely on that type.

On a side note, when I click on the file, then it shows me an `AbstractVector` oO

mforets: yes. the implementation was indeed not working if you passed a sparse vector to the `BallInf` constructor (because the output type would also be an abstract vector, and in that case it didn't match the `Vector` output annotation).

mforets: what i was saying above is that it's not true in all cases that we respect the type of the input direction. either rational -> float64 or sparse -> dense conversions may occur. this is a case by case study that we should see but can do it in separate issues.

mforets: please if u agree [with the note just added](https://github.com/JuliaReach/LazySets.jl/pull/37/commits/5c3b4f2df57c00537286a239637d274c68b2e713). this is to say that the rational computations are currently not reliable -- the output support vector is *not* a vertex of the given set! (assuming it's exact)

schillic: agreed

mforets: let's use `AbstractFloat` instead of real in that case, then it just say "i don't know" if you throw it rational.

PR #39, opened 2017-11-18
add low and high methods of Hyperrectangle

mforets:


src/Hyperrectangle.jl
```diff
@@ -177,3 +177,37 @@ A real number representing the diameter.
...
+### Input
+
+- `H` -- a hyperrectangle
+
+### Output
+
+A vector with the higher coordinates of the hyperrectangle, one entry per dimension.
+"""
+function high(H::Hyperrectangle)
+    return H.center + H.radius
```
schillic: Can you do `.+` here? Also below...
Not tested...

mforets: yep, i did it and added tests. thanks :)

Issue #13, opened 2017-10-30
Add support vector for VPolygon


mforets:
given a set of 2d vertices, the support vector of a `VPolygon <:LazySet` can be computed by:

-  sorting vertices by polar angle (see [function <=](https://github.com/JuliaReach/LazySets.jl/blob/master/src/helper_functions.jl#L72))
- performing binary search for the given direction

a convex hull preprocessing maybe needed, in general.

in a similar manner to `HPolygonOpt`, there can be a `VPolygonOpt` for warm-starting.

mforets:
> a convex hull preprocessing maybe needed, in general.

we perform the convex hull on the construction of each `VPolygon`, by default.

we plan to do a straightfoward implementation first and leave the binary search approach for later (see [this geomalgorithms entry](http://geomalgorithms.com/a14-_extreme_pts.html#polyMax_2D())).

mforets:
done in PR #41

PR #41, opened 2017-11-20
#13 - Implement support vector for VPolygon

mforets:


mforets:
Remains to add doctests and the docstring entry.

Issue #36, opened 2017-11-15
Automatically check for existence of required subtype specific methods


schillic:
We have a lot of subtypes, e.g., of `LazySet`.
We want all of them to provide a method for the function `σ`.
Hence we should check that those methods exist to prevent new types that may break things.

We can get all subtypes with the help of `subtypes(LazySet)` and use the `method_exists` function to check for existence:
```julia
for S in subtypes(LazySet)
    println(method_exists(σ, (Vector{Float64}, S)))
end
```

I propose that we add such tests for all abstract types, also in `Reachability`.

We could even create a function that automates this process if we have several common methods:
Add in `LazySets.jl`:
```julia
function require()::Vector{Function}
    return [
        S -> (σ, (Vector{Float64}, S)),
        S -> (my_func1, (S)),
        S -> (my_func2, (S, Real, Int))
    ]
end
```
Then in the unit tests we just have to go through this vector and apply the test to each of them.
We probably need a mechanism to allow several `require` methods for different abstract types, e.g., by changing the name.

(I have not tested the second one, the first one works.)

schillic:
We want to have this, but do not need the `require` function.
Just add those functions that you want into a unit test.

Currently `VPolygon` does not work, so we wait a bit here.

mforets:
> Currently VPolygon does not work, so we wait a bit here.

it does now.

PR #45, opened 2017-11-22
#36 - test existence of API methods

schillic:
The test is for `σ(::Vector{Float64}, ::S)` where `S<:LazySet`.
I also added a test for `dim(S)` because it worked 😎

@mforets:
For some subtypes the `σ` test breaks with `AbstractVector`, with `Real`, and with `<:Real` (I do not know the difference to `Real`, but apparently there is one because there are less cases working).
Should we fix that?

Also, should we additionally check for the same function with `Rational`?
Since we know that for `BallInf` it will not work, we could exclude `BallInf`.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -2,6 +2,11 @@
 using LazySets, Base.Test
 
 # =======================================
+# Testing common API of all LazySet types
+# =======================================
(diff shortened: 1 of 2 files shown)


schillic:
I also do not know if this is, strictly speaking, a unit test.
To me this is rather a workaround for something that other programming languages can guarantee.
But I still called the file `unit_LazySet.jl`.

Issue #46, opened 2017-11-24
Implement unit ball in the p-norm


mforets:
for support function computations, see: [Reachability analysis of linear systems with linear continuous dynamics](https://tel.archives-ouvertes.fr/tel-00422569v1/document), page 19.

schillic:
@mforets: Can this be closed?

PR #47, opened 2017-11-25
#46 - add Ballp and docs

mforets:


schillic:
Can we remove `Ball2` then?

mforets:
i think ball2 is sufficiently "special" to keep it separated; this general p-norm ball is more rarely used. note that it falls back to ball2 whenever `p=2` is chosen

mforets:
i didn't add yet unit tests for this file. some computation by hand can be done for `p=1`.

schillic:
The only difference is that we do not store `p` and the computation of `σ`.
Why is `σ` different?
Also I think it will crash if `||d|| = 0``. `Ball2` has a case split for this.

schillic:
Is `p::Real` correct? Does not make sense to me...

mforets:
>  Why is σ different?

in [Le Guernic's thesis, Appendix A page 125](https://tel.archives-ouvertes.fr/tel-00422569v1/document) there is the proof for the p-norm in the special case `c=0` and `r=1`. as written in the new docstring of `Ballp`, i used an affine transformation to extend it for any `c` and `r`.

>  Also I think it will crash if ||d|| = 0. Ball2 has a case split for this.

yes, i can add a case for this (we should return 0 as well).

mforets:
shoud fix `p<= 1`. thanks :)

mforets:
actually i'll go for creating a new type `Ball1`. the reason is that this one is "special" similarly to the 2 and oo cases.

docs/make.jl
```diff
@@ -1,7 +1,7 @@
 using Documenter, LazySets
 
 makedocs(
-#   doctest = false,  # uncomment locally for skipping doctests (saves time!)
```
schillic: I would prefer to keep the `false` uncommented.
By default it is active, so ne need to write this explicitly.

mforets: yes, this was a distraction, thanks

mforets: well, actually i had written `true` in the right hand side :)

schillic: You need to set it to `false`;)

Issue #49, opened 2017-11-25
Add unit test for `d=0`


mforets:
We should review that for `d=0` the support vector returns a point in the set, or to consider the special cases that raise an error otherwise.

schillic:
I reviewed this in #73.

`d=0` should work as expected for the following base types:
 `BallInf`, `Ballp`, `HPolygon`, `HPolygonOpt`, `VPolygon`, `Hyperrectangle`, `Singleton`, `Zonotope`
(and of course for all "operation" types)

For the following types we have to revise:
- `Ball1` uses the center but changes one coordinate to 0. We should guard this change by checking if the direction entry is 0.
- `Ball2` returns the origin. Instead we should return the center.

@mforets: Do you agree with these changes?

For the unit tests it will be helpful to have #68.

mforets:
In some set operations we specifically say that the results depends on the wrapped sets, e.g. 

https://github.com/JuliaReach/LazySets.jl/blob/2423c0d096a2206aea2beda306610cdc7c04be81/src/LazyOperations/CartesianProduct.jl#L147

In other cases the check is explicit and an assertion error is returned, e.g. for Bloating,

https://github.com/JuliaReach/LazySets.jl/blob/2423c0d096a2206aea2beda306610cdc7c04be81/src/LazyOperations/Bloating.jl#L92

---

In other cases as mentioned in [this comment](https://github.com/JuliaReach/LazySets.jl/issues/49#issuecomment-350470455) the operation returns a hard-coded value eg. the origin.

mforets:
I think consistency across different types for this corner case would be nice for the next major release, so I'm adding this to the v2.0 milestone.

PR #73, opened 2017-12-09
mixed improvements

schillic:
- added output types to all functions
- turned shorthand function style into `function ... end` blocks (except for constructors)
- renamed `is_contained` -> `∈`
- checked #49
- minor doc fixes


mforets:
some comments:

- we can have `iscontained` (or `is_contained` ?) as an alias.

- some functions enter in 80 characters => they fit nicely as one-liners.. is there any doc-related reason for the change? just aesthetics?

- the Output block of `sign_cadlag` is wrong, should say: `1 if x≥0, -1 otherwise.` shall we update it here? maybe add too:

```
### Notes

This is the sign function right-continuous at zero (see [cadlag function](https://en.wikipedia.org/wiki/C%C3%A0dl%C3%A0g)). It can be used with vector-valued arguments via the dot operator.
```

schillic:
> we can have iscontained (or is_contained ?) as an alias.

We already have `in`, I think this suffices.

> some functions enter in 80 characters => they fit nicely as one-liners

Citation needed :) I saw one case in `convex_hull_algorithms.jl`.
I prefer to break documentation sentences after the period to avoid moving everyhing around with a later change.

schillic:
> the Output block of sign_cadlag is wrong

I fixed it here.

mforets:
> We already have in, I think this suffices.

ok. it's the same as `Base.\in` ? or `in` should be declared too? i didn't find it in this commit

> Citation needed :)

i don't have one, no problem :)

schillic:
> >  We already have in, I think this suffices.
>
> ok. it's the same as Base.\in ? or in should be declared too? i didn't find it in this commit

Sorry, I should have said that. It seems to work automatically. It just aliases `∈` and then calls the right function.

Issue #68, opened 2017-12-08
Containment check


schillic:
Add a containment check `∈(x, S)` for all sets.

For some we already have `is_contained`. I prefer `in` instead (the Julia alias for `∈`).

mforets:
where `x` is a point? for polygons we should check [point in polygon](https://en.wikipedia.org/wiki/Point_in_polygon) algorithms.

agreed for overloading `in`

mforets:
.. still we can use `is_contained` in the mathematical sense, as for #69

schillic:
Yes, `x` is a point. Should it be a `Vector` or a `Singleton`?
We can provide a convenience conversion function for the other combination.

mforets:
we can implement it with sets, i.e. using `Singleton`, and also provide the shortcut for `x` being a Vector.

schillic:
I thought about it and I think it is more efficient to unwrap a `Singleton` (which is anyway what has to be done) in one case than to wrap a vector in a `Singleton` in the other case.
So I want to implement it for `AbstractVector` and provide a conversion function for `Singleton`.

schillic:
> for polygons we should check point in polygon algorithms.

Thanks for this pointer. "Ray casting" sounds simple for our sorted `VPolygon`, so I will implement this one.

The algorithms in this link work for nonconvex polygons, and at the bottom it is said "Simpler algorithms are possible for [...] convex polygons [...]."
We also already have an implementation for `HPolygon` using support vectors. Not sure if this is "optimal", though.
We can always add better algorithms later, for now I just want *one* solution.

schillic:
We are not 100% sure that the Zonotope implementation works in all cases, but keep it for now.

Most types are now supported by #74. Eventually we can continue in #77 with the missing ones.

PR #74, opened 2017-12-09
containment methods

schillic:
Discussion in #68.

Already available: HPolygons, CartesianProduct

Added:
* all Balls, Hyperrectangle, Singleton, VPolygon, LinearMap, Zonotope, ExponentialMap, ZeroSet, EmptySet
* conversion function for `Singleton` instead of `AbstractVector`

Missing and dismissed (outsourced to #77):
* MinkowskiSum (I have thought about it and it seems very complicated. One could subtract one set from the point and check for nonempty intersection with the other set. There is literature for the case that the sum consists of polytopes only, this could be interesting.)
* ConvexHull (I have thought about it and it seems too complicated. A sufficient test is to check containment in both generating sets, but for negative answers one has to continue, e.g., checking ray intersection from some vertex of one set in the other set.)
* ExponentialProjectionMap (I am not familiar with this type.)
* Polyhedron (We currently do not support them, so I will ignore them.)

### DIFF
--- a/src/Ball1.jl
+++ b/src/Ball1.jl
@@ -1,3 +1,5 @@
+import Base.∈
+
 export Ball1
 
 """
@@ -9,7 +11,7 @@ norm.
 It is defined as the set
 
 ```math
-\\mathcal{B}_1^n(c, r) = \\{ x ∈ \\mathbb{R}^n : ∑_{i=1}^n |x_i| ≤ r \\},
+\\mathcal{B}_1^n(c, r) = \\{ x ∈ \\mathbb{R}^n : ∑_{i=1}^n |c_i - x_i| ≤ r \\},
 ```
 where ``c ∈ \\mathbb{R}^n`` is its center and ``r ∈ \\mathbb{R}_+`` its radius.
 
@@ -67,7 +69,7 @@ function dim(B::Ball1)::Int
 end
 
 """
-    σ(d::AbstractVector{N}, B::Ball1)::AbstractVector{N} where {N<:AbstractFloat}
+    σ(d::AbstractVector{N}, B::Ball1)::AbstractVector{N} where {N<:Real}
 
 Return the support vector of a `Ball1` in a given direction.
 
@@ -86,4 +88,52 @@ function σ(d::AbstractVector{N},
     imax = indmax(abs.(d)) 
     res[imax] = sign(d[imax]) * B.radius
     return res
+end
+
+"""
+    ∈(x::AbstractVector{N}, B::Ball1{N})::Bool where {N<:Real}
+
+Check whether a given point is contained in a ball in the 1-norm.
+
+### Input
+
+- `x` -- point/vector
+- `B` -- ball in the 1-norm
+
+### Output
+
+`true` iff ``x ∈ B``.
+
+### Notes
+
+This implementation is worst-case optimized, i.e., it is optimistic and first
+computes (s. below) the whole sum before comparing to the radius.
+In applications where the point is typically far away from the ball, a fail-fast
+implementation with interleaved comparisons could be more efficient.
+
+### Algorithm
+
+Let ``B`` be an ``n``-dimensional ball in the 1-norm with radius ``r`` and let
+``c_i`` and ``x_i`` be the ball's center and the vector ``x`` in dimension
+``i``, respectively.
+Then ``x ∈ B`` iff ``∑_{i=1}^n |c_i - x_i| ≤ r``.
+
+### Examples
+
+```jldoctest
+julia> B = Ball1([1., 1.], 1.);
+
+julia> ∈([.5, -.5], B)
+false
+julia> ∈([.5, 1.5], B)
+true
+```
+"""
+function ∈(x::AbstractVector{N}, B::Ball1{N})::Bool where {N<:Real}
+    @assert length(x) == dim(B)
+    sum = zero(N)
+    for i in eachindex(x)
+        sum += abs(B.center[i] - x[i])
+    end
+    return sum <= B.radius
 end
--- a/src/Ball1.jl
+++ b/src/Ball1.jl
@@ -107,7 +107,7 @@ Check whether a given point is contained in a ball in the 1-norm.
 ### Notes
 
 This implementation is worst-case optimized, i.e., it is optimistic and first
-computes (s. below) the whole sum before comparing to the radius.
+computes (see below) the whole sum before comparing to the radius.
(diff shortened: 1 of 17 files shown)


src/VPolygon.jl
```diff
@@ -132,3 +132,62 @@ List containing a singleton for each vertex.
...
+                     apply_convex_hull=false);
+
+julia> ∈([44//10, 34//10], P)  #  with rational numbers the answer is correct
+true
+```
+"""
+function ∈(x::AbstractVector{N}, P::VPolygon{N})::Bool where {N<:Real}
+    @assert length(x) == 2
+
+    @inline function side(x, u, v)
```
mforets: we may use the existent `right_turn(x, u, v)`  here

src/Ball1.jl
```diff
@@ -9,7 +11,7 @@ norm.
 It is defined as the set
 
 ```math
-\\mathcal{B}_1^n(c, r) = \\{ x ∈ \\mathbb{R}^n : ∑_{i=1}^n |x_i| ≤ r \\},
+\\mathcal{B}_1^n(c, r) = \\{ x ∈ \\mathbb{R}^n : ∑_{i=1}^n |c_i - x_i| ≤ r \\},
```
mforets: good catch

src/Ball1.jl
```diff
@@ -87,3 +89,51 @@ function σ(d::AbstractVector{N},
...
+- `B` -- ball in the 1-norm
+
+### Output
+
+`true` iff ``x ∈ B``.
+
+### Notes
+
+This implementation is worst-case optimized, i.e., it is optimistic and first
+computes (s. below) the whole sum before comparing to the radius.
```
mforets: s. below -> see below   (?)

src/Ball1.jl
```diff
@@ -87,3 +89,51 @@ function σ(d::AbstractVector{N},
...
+true
+```
+"""
+function ∈(x::AbstractVector{N}, B::Ball1{N})::Bool where {N<:Real}
+    @assert length(x) == dim(B)
+    sum = zero(N)
+    for i in eachindex(x)
+        sum += abs(B.center[i] - x[i])
+    end
+    return sum <= B.radius
```
mforets: i tend to prefer the more synthetic `return @. sum(abs(B.center - x)) <= B.radius` if possible. in julia we can have both with the same performance 👍 

the assertion is not "needed" since it is raised whenever dims do not match for the `-` operator, giving `ERROR: DimensionMismatch("dimensions must match")` (in other words i see no difference of having our assertion error rather than this DimensionMismatch error).

src/LinearMap.jl
```diff
@@ -124,3 +124,38 @@ that ``σ(d, L) = M⋅σ(M^T d, S)`` for any direction ``d``.
...
+julia> ∈([5.0, 1.0], lm)
+false
+julia> ∈([3.0, 1.0], lm)
+true
+```
+"""
+function ∈(x::AbstractVector{N},
+           lm::LinearMap{<:LazySet, N})::Bool where {N<:Real}
+    @assert length(x) == dim(lm)
+    return ∈(inv(lm.M) * x, lm.sf)
```
mforets: what happens if `M` is not invertible?

it's preferable to divide via `lm.M \ x` instead of computing the full matrix inverse explicitly with `inv`, which is not needed.

src/VPolygon.jl
schillic: I do not immediately see the mapping and if these functions do the same.

src/Ball1.jl
schillic: ok

src/LinearMap.jl
schillic: Do you mean `return ∈(lm.M \ x, lm.sf)`?

src/VPolygon.jl
mforets: `side(x, u, v) = right_turn(u, x, v)`

src/Ball1.jl
schillic: `return @. sum(abs(B.center - x)) <= B.radius` does not work because the output is a Boolean array. When I wrap it in an `all(...)` block, the performance is poor.
```julia
julia> @time alternative([.5, 1.5], B)
  0.000023 seconds (9 allocations: 4.578 KiB)
julia> @time ∈([.5, 1.5], B)
  0.000011 seconds (5 allocations: 256 bytes)
```

src/VPolygon.jl
schillic: Should I put `right_turn` to `helper_functions`?

src/LinearMap.jl
mforets: yes. this covers the invertible case, but doesn't answer the problem generally i think. let

```julia
julia> B = BallInf(zeros(4), 1.);
julia> M = [1. 0 0 0; 0 1 0 0]/2;
julia> x = ones(4.);
```

then the point `[0.5, 0.5]` is in the image set of the linear map of `B` under `M`. however the check of above will fail in `M \ x`.

schillic: on a side note:
```julia
julia> dim(M*B)
2
```

mforets: if you want we can say in the docstring that this method applicable to the invertible case for now.

src/Ball1.jl
mforets: did i miss some parentheses? like `@.(sum(abs(B.center - x)))<= B.radius` 
it will have the same performance as the loops, no?

schillic: Now I used `sum(@. abs(B.center - x)) <= B.radius`
```julia
julia> @time alternative([.5, 1.5], B)
  0.000010 seconds (6 allocations: 352 bytes)
```
Still not as good as the explicit form.

mforets: > Still not as good as the explicit form.

at this scale we'll get some better comparison with `@btime` from `BenchmarkTools`

schillic: I'm only looking at the memory.

mforets: what does `@allocated`  say, and did you try declaring the variable `B` as `const`? recall that [measuring in the global scope can give misleading results](https://discourse.julialang.org/t/allocations-in-a-copy-construct/7323/2?u=mforets)

schillic: I tried it only in the REPL. There I cannot declare as `const`.

schillic:
```julia
julia> @allocated ∈([.5, -.5], B)
96
julia> @allocated alternative([.5, -.5], B)
192
```

Issue #69, opened 2017-12-08
Subset check


schillic:
Add a subset check `⊆(S1, S2)`, in particular for special types of `S1`.
In particular, for polytopes (and "subtypes"), since `S2` is convex, it should suffice to check containment of all corners (see #68).

Also check if Julia has an alias for `⊆`, otherwise add one such as `isincluded` or `contains`.

schillic:
Possibly helpful: [Boolean operations on polygons](https://en.wikipedia.org/wiki/Boolean_operations_on_polygons)

schillic:
I would really like to have #58 here.
Now we have to provide
- `Hyperrectangle ⊆ Hyperrectangle`
- `Hyperrectangle ⊆ BallInf`
- `BallInf ⊆ Hyperrectangle`
- `BallInf ⊆ BallInf`
and the algorithm is almost exactly the same except that `BallInf`'s radius is a scalar.
If instead we had a common supertype, we could have a function `radius(dim)` that returns the radius in a given dimension and thus write **one function for all four cases** (with no performance penalty).

schillic:
The following table shows straight-forward implementations for most basic set types.
There might be more efficient implementations, but for the moment I am happy with *any* solution.
"v_i ∈ S" means "check containment of every vertex in set S"; here we need `vertices_list` and `∈`.
"done" means that this is already implemented.

| ↓ ⊆ →   | Ball1 | Ball2 | Ballp | Hyperrectangle/BallInf | Polygon | Singleton |
|---------|-------|-------|-------|------------------------|---------|-----------|
|Ball1    | v_i ∈ S | v_i ∈ S | v_i ∈ S | v_i ∈ S          | v_i ∈ S | c = S & r = 0 |
|Ball2    |  |  |  | interval hull |  | c = S & r = 0 |
|Ballp    |  |  |  | interval hull |  | c = S & r = 0 |
|HR/BI    | v_i ∈ S | v_i ∈ S | v_i ∈ S | done             | v_i ∈ S | c = S & r = 0 |
|Polygon  | v_i ∈ S | v_i ∈ S | v_i ∈ S | v_i ∈ S          | v_i ∈ S |  |
|Singleton| x ∈ S | x ∈ S | x ∈ S | x ∈ S | x ∈ S | x = S   |

schillic:
[...]

PR #75, opened 2017-12-11
#69 - subset check

schillic:
Discussion in #69.


mforets:
what is the status of this branch? have you rebased it with `master`?

schillic:
Yes and no, it is rebased on `schillic/63` which is rebased on `master`.
So maybe we first merge #91.

mforets:
conflicts ahead

schillic:
Yeah, I forgot to rebase with the latest changes. Now it should work.

src/AbstractHyperrectangle.jl
```diff
@@ -159,3 +160,35 @@ function ∈(x::AbstractVector{N},
...
 end
+
+"""
+    ⊆(H1::AbstractHyperrectangle, H2::AbstractHyperrectangle)::Bool
+
+Check whether a given hyperrectangle is contained in another hyperrectangle.
+
+### Input
+
+- `H1` -- first hyperrectangle (containee?)
```
mforets: "inner hyperrectangle" vs. "outer hyperrectangle"   ??

src/AbstractHyperrectangle.jl
```diff
@@ -159,3 +160,35 @@ function ∈(x::AbstractVector{N},
...
+- `H2` -- second hyperrectangle (container?)
+
+### Output
+
+`true` iff ``H1 ⊆ H2``.
+
+### Algorithm
+
+``H1 ⊆ H2`` iff ``c_1 + r_1 ≤ c_2 + r_2 ∧ c_1 - r_1 ≥ c_2 - r_2`` iff
+``r_1 - r_2 ≤ c_1 - c_2 ≤ -(r_1 - r_2)``.
```
mforets: "... where \leq is taken component-wise."

src/AbstractHyperrectangle.jl
schillic: Well, this is what the function has to find out. I did not find a good way to express this.

mforets:
i think we should mention `issubset` in one example. that's enough for a quick search in the docs to give something useful.

src/Ball2.jl
```diff
@@ -167,3 +167,44 @@ function ∈(x::AbstractVector{N}, B::Ball2{N})::Bool where {N<:Real}
...
+### Output
+
+* If `witness` option is deactivated: `true` iff ``B ⊆ S``
+* If `witness` option is activated:
+  * `(true, [])` iff ``B ⊆ S``
+  * `(false, v)` iff ``B \\not\\subseteq S`` and ``v ∈ B \\setminus S``
+"""
+function ⊆(B::Ball2{N}, S::AbstractSingleton{N}, witness::Bool=false
+          )::Union{Bool,Tuple{Bool,Vector{N}}} where {N<:Real}
+    result = B.center == element(S) && B.radius == 0
```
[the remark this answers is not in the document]
schillic: I think Julia automatically converts the 0 to the correct type. I do not know which one is better/faster or if they are both doing the same thing.
Do you have preferences?

schillic:
> i think we should mention issubset in one example. that's enough for a quick search in the docs to give something useful.

Where exactly do you mean? I think there are not examples yet.
Maybe we should add a new doc page for these new set operations (∈, ⊆, ∩) where we can also include the tables? But then let's do this in a new issue.

test/unit_Ball2.jl
```diff
@@ -53,3 +53,13 @@ d = [0., -1.]
...
+
+# subset
+b1 = Ball2([1., 2.], 2.)
+b2 = Ball2([1., 2.], 0.)
+s = Singleton([1., 2.])
+subset, point = ⊆(b1, s, true)
+@test !⊆(b1, s) && !subset && point ∈ b1 && !(point ∈ s)
+@test ⊆(b2, s) && ⊆(b2, s, true)[1]
+@test !⊆(b1, BallInf([1., 2.], 1.))
+@test ⊆(b2, BallInf([1., 2.], 2.))
```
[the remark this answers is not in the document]
schillic: Okay, let me add this.

mforets:
> Where exactly do you mean? 

it can go as an `Example` block in some function eg. [this one](https://github.com/JuliaReach/LazySets.jl/pull/75/files#diff-65f50ad7749bd6509c1335e07bc63b13R223). Ok for another issue, and the tables too.

schillic:
I prefer to not hide this in a random one. I'll create a new issue.

src/Ball2.jl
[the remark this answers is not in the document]
mforets: yes, it seems so. in that case i would prefer just `0` :)

schillic:
@mforets: Do you want to review the last three commits? Then we can merge.

src/Ball2.jl
```diff
@@ -209,3 +209,43 @@ function ⊆(B::Ball2{N}, S::AbstractSingleton{N}, witness::Bool=false
...
+            return (result, N[])
+        end
+    else
+        return result
+    end
+
+    # compute a witness 'v'
+    v = B1.center .+ B1.radius * (B1.center .- B2.center)
+    return (false, v)
+end
```
mforets: this would be in shortened form

```julia
v = result ? N[] : B1.center .+ B1.radius * (B1.center .- B2.center)
return witness ? (result, v) : result
```

although it has the drawback of computing `v` also when this is not really asked for.

schillic: Yes, I definitely want to avoid this because I think in usual applications you do not want a witness.

mforets:
ok, let it be. thx for fiding and updating the removed method too.

PR #78, opened 2017-12-12
initial section for interval hulls

mforets:


docs/src/man/interval_hulls.md
```diff
@@ -0,0 +1,104 @@
...
+DocTestSetup = quote
+    using LazySets, Plots, LazySets.Approximations
+end
+```
+
+## Balls and singleton plots
+
+Consider a ball in the 2-norm:
+
+```@example
```
schillic: I need an additional `using LazySets` here, otherwise I get an error. Seems that the `DocTestSetup` above is not executed for `@example` blocks. Does it have to be an `@example` block here?

docs/src/man/interval_hulls.md
```diff
@@ -0,0 +1,104 @@
...
+```
+ 
+For instance, compute the support vector in the south-east direction:
+ 
+```julia
+σ([1., -1.], S)
+2-element Array{Float64,1}:
+  1.5
+ -1.5
+ ```
```
schillic: Here and below there is a space in front of the "```" which makes the text in between be rendered in Julia style.

docs/src/man/interval_hulls.md
```diff
@@ -0,0 +1,104 @@
...
+plot!(Singleton(σ([1., 1], X)))
+plot!(Singleton(X.center), markershape=:x)
+```
+
+## Symmetric interval hull
+
+The symmetric interval hull is defined in the `Approximations` module:
+
+```julia
+julia> S = Approximations.symmetric_interval_hull(X)
```
schillic: No need to write `Approximations.` since you included it above. I would make it consistent: either include it visibly for the user once and omit it then, or write it out all the time, but then do not include it above.

docs/make.jl
```diff
@@ -1,7 +1,7 @@
 using Documenter, LazySets
 
 makedocs(
-    doctest = true,  # use this flag to skip doctests (saves time!)
+    doctest = false,  # use this flag to skip doctests (saves time!)
```
schillic: !

docs/src/man/interval_hulls.md
mforets: correct; either that or i can add a label to this block, i'll use the latter.

docs/src/man/interval_hulls.md
mforets: ok. unfortunately, it seems that the `@meta` environment affects doctests but not `@example` blocks..

mforets:
this PR is ready from my side

schillic:
Okay, I can have a look later today.

docs/src/man/interval_hulls.md
```diff
@@ -0,0 +1,187 @@
...
+numbers. Other numeric types (such as lower precision floating point, or rational)
+can be defined with the proper argument types in the `Ball2` constructor.
+
+```@example example_ih
+using LazySets, Plots
+
+X = Ball2(ones(2), 0.5)
+```
+
+To plot a lazy set, we use the `plot` function. Recall that, by design, lazy sets
```
schillic: "recall" from where?

docs/src/man/interval_hulls.md
```diff
@@ -0,0 +1,187 @@
...
+borders, use the accuracy as a second argument.
+
+```@example example_ih
+plot(X, 1e-3, aspectratio=1)
+```
+
+To add plots to the same pair of axes we use `plot!`. Let's add some points of
+the set which are farthest in some given directions. Single points can be plotted
+using the `Singleton` type.
+In the third line we plot the center of the ball picking a custom cross marker.
```
schillic: maybe better "line of code"

docs/src/man/interval_hulls.md
```diff
@@ -0,0 +1,187 @@
...
+plot!(Singleton(σ([1., 0], X)))
+plot!(Singleton(σ([1., 1], X)))
+plot!(Singleton(X.center), markershape=:x)
+```
+
+!!! note
+    To see the list of available plot keyword arguments, use the `plotattr([attr])`
+    function, where `attr` is the symbol `:Plot`, `:Series`, `:Axis` or `:Subplot`.
+
+For the remaining of this section we define another ball in the 2-norm and their
```
schillic: "remaining section" or "remainder of this section"

docs/src/man/interval_hulls.md
```diff
@@ -0,0 +1,187 @@
...
+!!! note
+    The `interval_hull` function is an alias for the `box_approximation` function.
+    The nomenclature for approximation functions is `*_approximation_*`. To see a
+    list of all approximation functions, either search in the docs or type
+    `names(LazySets.Approximations)`.
+
+## Symmetric interval hull
+
+Contrary to the previous approximations, the symmetric interval hull is centered
+around the origin. It is defined in the `Approximations` module as well.
```
schillic: You did not mention the `Approximations` module before.

docs/src/man/interval_hulls.md
```diff
@@ -0,0 +1,187 @@
...
+These functions apply generally to any `LazySet`. For some types, specialized
+methods are triggered automatically through multiple-dispatch.
+
+The *norm* of a convex set is the norm of the enclosing ball (of the given norm)
+of minimal volume. For instance:
+
+```@example example_ih
+import LazySets.Approximations: norm, radius, diameter
+
+norm(X), norm(Y), norm(Z)
```
schillic: Maybe recall in the text that you defined `X`,`Y`,`Z` above (it is a new section).

docs/src/man/interval_hulls.md
```diff
@@ -0,0 +1,187 @@
...
+
+Finally, it is sometimes convenient to ask directly the diameter of the set,
+defined as twice the radius:
+
+```@example example_ih
+diameter(X), diameter(Y), diameter(Z)
+```
+
+In the previous examples, we have used the infinity norm (default). It is possible
+to pass the required `p`-norm as a second argument.
```
schillic: Maybe add a short example here.

mforets: good catch; none of them is implemented yet for p not Inf, hence i removed this sentence.

schillic:
LGTM

PR #80, opened 2017-12-14
rename ∈ -> ⊆ for Singleton

schillic:
A singleton is a set, so the function name did not make sense.


mforets:
good!

we can use `in` instead of `\in`; do we have an alias `isincluded`, or `contains`? shall this be added here or globally so that it affects all instances of `\subseteq`?

schillic:
I will add an alias in #69, thanks for the reminder.

PR #81, opened 2017-12-14
Update decompose example

schillic:
@mforets: Do you agree? I also removed the image files from the repo.


schillic:
ready from my side

schillic:
Looks good, don't know why the build was failing.

Issue #83, opened 2017-12-14
∈ implementation crashing/not correct for VPolygon with <= 2 vertices


schillic:

schillic:
There are still problems for 1 vertex:
```julia
∈([1., 1.], VPolygon([[0., 0.]]))
true
∈([1., 1.], VPolygon([[0., 0.], [0.5, 0.]]))
false
```
This can be fixed by catching this case and comparing the vector to the vertex. Is there another way?

mforets:
Ok

schillic:
I did not say that you should fix it.

PR #87, opened 2017-12-15
#83 - VPolygon 1 and 2 vertices corner cases

mforets:


schillic:
Just to say, #83 was not concerned with the convex hull.

mforets:
does this fix #83 ? if not do you have a breaking example? that we can add here?

schillic:
```julia
∈([1.,2.], VPolygon([[1.,2.]]))
```

mforets:
why is this wrong?

```
julia> ∈([1.,2.], VPolygon([[1.,2.]]))
true
```

schillic:
Ha, it's fixed. I don't understand why, but 👍

mforets:
the `VPolygon` constructor, through convex_hull, was buggy (returning a 0-element set in these corner cases)

Issue #85, opened 2017-12-15
Line intersection throws errors


schillic:
This happens if the lines are parallel or identical.

```julia
l1 = Line([1., 1.], 1.)
l2 = Line([1., 1.], 1.)
intersection(l1, l2) # Base.LinAlg.SingularException(2)
l3 = Line([1., 0.], 1.)
l4 = Line([1., 0.], 1.)
intersection(l3, l4) # Base.LinAlg.LAPACKException(2)
```

mforets:
should they return an ∅ ?

mforets:
through a `try`..`catch` block?

schillic:
I do not know, but we should at least document the behavior in these cases.
`∅` is not much better because it is not a `Vector` - an empty `Vector` would make more sense.
We could also return two values, a `Bool` and a `Vector if the `Bool` is `true.

schillic:
We agreed that we `try..catch` and return a 0-vector for now, and continue later in #57.

PR #88, opened 2017-12-15
#85 - output empty vector for parallel Line intersection

schillic:
see #85


schillic:
All unit tests and doctests pass on my local end...

Issue #63, opened 2017-12-07
Add function that returns a point inside a set


schillic:
* A default implementation could just pick any direction and ask for a support vector.
* The empy set should throw an exception.
* What would be a good name for this function? `element`? CC: @mforets

schillic:
I will use `element`, which is consistent with the function of `AbstractSingleton`.

mforets:
+1 `element`, and +1.5 for `an_element`

schillic:
@mforets: Should we add a default implementation for any `LazySet` as follows?
```julia
function an_element(S::LazySet)::Vector{<:Real}
    return σ(ones(dim(S)), S)
end
```
Problem: the direction has type `Float`.
If yes, where do we put this? Create a new file `LazySet_functions.jl`? Or a file `LazySet.jl` and move the type definition there as well?
There are already `norm`/`radius`/`diameter` in `box_approximations.jl`, so maybe we want to move those to this file, too.

mforets:
> Problem: the direction has type Float.

this could be worked around if #44 gets in.

> If yes, where do we put this? 

also note the file `support_function.jl`, which just defines generic support vector for any `LazySet`. a file `LazySet.jl` for this group of definitions is not a bad idea.

mforets:
.. or a file `AbstractLazySet.jl` , and in that case also export the type `AbstractLazySet` as an alias for `LazySet`. i'm ok with either way, what do you think?

schillic:
> this could be worked around if # 44 gets in.

Good, then let us fix that one first.

> a file `LazySet.jl` for this group of definitions is not a bad idea.

I also like this.

> or a file `AbstractLazySet.jl`

What is an argument for this? (We can also rename the file later and just need to change one line.)

mforets:
ok. 

> What is an argument for this? 

only that it goes in line with other files for abstract types such as `Abstractxyz`.

for the implementation we can take as well `σ([1.; fill(0, dim(S))], S)` which has more zeros :stuck_out_tongue:

PR #91, opened 2017-12-18
#63 - point inside a set

schillic:
See #63.
This PR ready.


mforets:
in the doctests what do you think about checking with `\in` instead of `==`

schillic:
"instead of" or "additionally"?
You are right that the contract of this method only requires `in`, and maybe we will change the implementation.

PR #96, opened 2017-12-20
#3 - lazy symmetric interval hull

### LINKED ISSUE #3: Symbolic symmetric interval hull
The symmetric interval hull is computed explicitly in [Approximations.box_approximation_helper](https://github.com/mforets/LazySets.jl/blob/master/src/Approximations.jl#L226), through 2n support function evaluations.

An application of a new lazy (total or partial) type, symmetric interval hull, is to speedup the discretization phase; the explicit form can take up to 97% of the time for n>10000.

schillic:
Closes #3.

### DIFF
--- a/src/SymmetricIntervalHull.jl
+++ b/src/SymmetricIntervalHull.jl
@@ -104,9 +104,10 @@ This function computes the symmetric interval hull explicitly.
 function radius_hyperrectangle(sih::SymmetricIntervalHull{N}
                               )::Vector{N} where {N<:Real}
     dim = dim(sih)
+    one_N = one(N)
     for i in 1:dim
         if !sih.known[i]
-            compute_radius(sih, i, dim)
+            compute_radius(sih, i, dim, one_N)
         end
     end
     return sih.radius
@@ -188,12 +189,13 @@ function σ(d::AbstractVector{N}, sih::SymmetricIntervalHull
 
     svec = similar(d)
     zero_N = zero(N)
+    one_N = one(N)
     for i in eachindex(d)
         if d[i] == zero_N
             svec[i] = zero_N
         else
             if !sih.known[i]
-                compute_radius(sih, i, len)
+                compute_radius(sih, i, len, one_N)
             end
             svec[i] = sign(d[i]) * sih.radius[i]
         end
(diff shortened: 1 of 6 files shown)


schillic:
Reminder: This type needs to be adapted to #122.

mforets:
let's rebase when #122 gets merged

schillic:
Done, but this has not advanced and is not ready.

schillic:
Was quite easy in the end 😀
@mforets: Can you have a look?

schillic:
The breaking build is due to outakes in Travis CI. We can restart when they are resolved.

mforets:
i think the auxiliary type should be a single one, maybe a `Dict`, that has tuple values `(index, radius)`? the type and main loop become a bit more compact and you avoid having to choose `radius = [0,0..0]` for a freshly initialized set

mforets:
should we use a cache for `length(d)` in `σ(d::AbstractVector{N}, sih::SymmetricIntervalHull)`, similarly to `zero_N` and `one_N`? (BTW i don't know if `d[i] == zero_N` the same as `d[i] == 0`)

schillic:
A `Dict` does not have constant-time access, and it only hides from you that it also allocates and initializes memory.
Would a `Vector{Tuple}` or `Tuple{Vector, Vector}` be ok?
Let me add that this is an implementation detail. The user does not have to work with these variables (and if Julia had the concept of private variables, these would definitely be private).

schillic:
I assume that most vector implementations don't actually compute anything for `length`, so the compiler might inline this. But ok, we can store it, does not cost much 👍

schillic:
I rebased with master and squashed my old commits. Now I will add the subtyping from `Hyperrectangle` that you proposed.

src/SymmetricIntervalHull.jl
```diff
@@ -0,0 +1,238 @@
...
+
+### Algorithm
+
+We ask for the support vector of the underlying set for both the positive and
+negative unit vector in the dimension `i`.
+"""
+function compute_radius(sih::SymmetricIntervalHull,
+                        i::Int,
+                        dimension::Int=dim(sih))
+    one_N = one(N)
```
mforets: the type `N` will not be known there. we can pass it?

schillic: Yes, I will fix this.

src/SymmetricIntervalHull.jl
```diff
@@ -0,0 +1,238 @@
...
+    right_bound = σ(sparsevec([i], [one_N], dimension), sih.X)
+    left_bound = σ(sparsevec([i], [-one_N], dimension), sih.X)
+    sih.radius[i] = max(right_bound[i], abs(left_bound[i]))
+    sih.known[i] = true
```
mforets: should we invert these last two lines? such that it returns the computed radius which is more interesting.

schillic: Actually, the docs said that it returns the radius...

schillic:
This one is ready again, but I did not change the auxiliary fields yet. What should we do here?

PR #98, opened 2017-12-26
Remove MkDocs deps from docs/make.jl

mortenpi:
Since the docs are rendered with Documenter's native HTML renderer, MkDocs-related deps are not necessary. Just something that caught my eye.


mforets:
Thanks for having a look and fixing! 👍

Issue #71, opened 2017-12-09
Failing doctests should fail the build


schillic:
@mforets: Currently there is a failing doctest in master:

```julia
> File: .../LazySets/src/Ball1.jl

> Code block:                                                                                                       

\`\`\`jldoctest
julia> σ([0.,1], B)
1-element Array{Float64,1}:
 1.0
 0.0
\`\`\`

> Subexpression:                                                                                                    

    σ([0.,1], B)

> Output Diff:                                                                                                      

1-element 2-element Array{Float64,1}:
 1.0
 0.00.0
 1.0 
```
However, Travis does not complain.
See [here](https://github.com/JuliaDocs/Documenter.jl/issues/527) how to potentially implement this.

schillic:
In my recent commits I did not include unit tests but only doctests. Assuming this issue is resolved, they should be equivalent from the automatic builds view (fails iff one of the tests breaks).

`+` Get doc examples for free.
`-` Usually you only run the unit tests locally; then you do not see if you broke something.

I do not have a strong opinion here. We could also have both, but then one of the tests is obviously redundant.

schillic:
@mforets asked about this in the Documenter repo.

mforets:
In #677 i set the `strict` flag build to `false` by default.  The doctests were passing locally but there still the docs was not working.

I'm not proposing this measure to be definitive, but i'm reopening this issue to keep it in mind. Maybe we should add the `strict` option to be read in the command line, like

```julia
$ julia --color=yes docs/make.jl --strict=true
```

schillic:
There is now a [new documentation for using build stages](https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#.travis.yml-Configuration-1). After the jump to v1.0, we should try this again.

PR #102, opened 2017-12-28
[WIP #71] - Use build stages for docs deployment

mforets:


schillic:
Yeah, it worked!

mforets:
let me try again.. it failed because of a `ERROR: Documenter is not an installed package` in the in the new "deploy docs" stage.

schillic:
Damn, now it worked?

schillic:
@mforets: Since you asked back in the other thread: I think their use case was to fail the build when building the docs fails. This is different from the doc**tests**.

mforets:
then i don't know what else to try  😢

schillic:
My guess is that Documenter just prints warnings. Travis does not recognize them as an error.

Maybe this can help?
```
Documenter: skipping docs deployment.
You can set DOCUMENTER_DEBUG to "true" in Travis to see more information.
```

schillic:
I just found [this](https://github.com/JuliaDocs/Documenter.jl/wiki/Documenter.jl-roadmap#use-travis-build-stages). Not sure if this is what you did...

mforets:
i would suggest to close this in favor of #467

schillic:
Yes. Do we want to use the idea with build stages there as well?

mforets:
i never got the build stages to work as expected as evidenced in our comments above  :/

schillic:
Okay, then I will close.

Issue #61, opened 2017-12-05
Illustrative examples for the manual


mforets:
let's create a new `examples` folder with some illustrative examples and pictures:

- [x] polygon overapproximations => iterative refinement, see #108
- [x] convex hulls
- [x] interval hulls
- [x] reachability algorithm for linear systems using zonotopes

PR #30, opened 2017-11-11
Fix approximate

mforets:
See #6.

PR #108, opened 2017-12-30
WIP - docs iterative refinement

mforets:
See #61.


schillic:
I do not understand this PR. Can you explain?

mforets:
this PR will solve add docs example for iterative refinement and is a warm up for PR #30.

the first stage (done) was to move some docstrings from the Manual to the Library section, as it is the case for the other types and method definitions.

the next step is to add a simple illustrative example of iterative refinement in 2D.

Issue #106, opened 2017-12-30
Add type parameter to overapproximate/decompose


schillic:
This is part of [this issue](https://github.com/JuliaReach/Reachability.jl/issues/24) in `Reachability.jl`.

The idea is to allow more output types than just `HPolygon`.

PR #110, opened 2017-12-30
#106 - type parameter for overapproximate/decompose

schillic:
See #106.

### DIFF
--- a/src/Approximations/2D_approximations.jl
+++ b/src/Approximations/2D_approximations.jl
@@ -1,0 +1,84 @@
+"""
+    overapproximate(S::LazySet, ::Type{HPolygon})::HPolygon
+
+Return an approximation of a given 2D convex set as a box-shaped polygon.
+
+### Input
+
+- `S` -- convex set, assumed to be two-dimensional
+- `HPolygon` for dispatch
+
+### Output
+
+A box-shaped polygon in constraint representation.
+"""
+function overapproximate(S::LazySet, ::Type{HPolygon})::HPolygon
+    @assert dim(S) == 2
+    pe, pn, pw, ps = box_bounds(S)
+    constraints = Vector{LinearConstraint{eltype(pe)}}(4)
+    constraints[1] = LinearConstraint(DIR_EAST, dot(pe, DIR_EAST))
+    constraints[2] = LinearConstraint(DIR_NORTH, dot(pn, DIR_NORTH))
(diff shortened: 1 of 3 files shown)


test/unit_decompose.jl
```diff
@@ -9,7 +9,7 @@ Y = [1.92664 1.00674 1.0731 -0.995149;
     -2.43723 0.782825 -3.99255 3.93324]
 Y = Y * CPA([BallInf([0.767292, 0.936613], 0.1), BallInf([0.734104, 0.87296], 0.1)])
 Ω0 = CH(X, Y)
-dec = Approximations.decompose(Ω0)
+dec = Approximations.decompose(Ω0, HPolygon)
```
mforets: can you make the second argument optional?

src/Approximations/2D_approximations.jl
```diff
@@ -0,0 +1,79 @@
...
+### Input
+
+- `S` -- convex set, assumed to be two-dimensional
+- `HPolygon` for dispatch
+
+### Output
+
+A box-shaped polygon in constraint representation.
+"""
+function overapproximate(S::LazySet, ::Type{HPolygon})::HPolygon
```
mforets: i don't understand, where is the second argument used?

mforets:
ok, i see. i would suggest to make this change an optional keyword argument, eg. `"type"` or `"set_type"`, being `HPolygon` the default one.

src/Approximations/2D_approximations.jl
mforets: nevermind. i think i what i wanted to suggest is defining

```julia
overapproximate(S::LazySet) = overapproximate(S::LazySet, ::Type{HPolygon})
```

schillic: > where is the second argument used?

It's only used by Julia to find the correct method.

schillic: > maybe define
> `overapproximate(S::LazySet) = overapproximate(S, HPolygon)`

Good idea, I will try.

test/unit_decompose.jl
schillic: Okay, same as below.

schillic:
I rebased with `master` and incorporated your suggestion with the default value. This also saved some changes to unit tests and plot recipes 👍

mforets:
failing checks are a Travis CI timeout => positive review

PR #112, opened 2018-01-06
tutorial on iterative refinement

mforets:


docs/src/man/iterative_refinement.md
```diff
@@ -11,61 +12,213 @@ Depth = 3
...
+addapproximation!(Ω, pw, DIR_WEST, ps, DIR_SOUTH)
+addapproximation!(Ω, ps, DIR_SOUTH, pe, DIR_EAST)
+
+plot!(Approximations.tohrep(Ω), alpha=0.2, color="orange")
+```
+
+Next we refine the first approximation of the list.
+
+```@example example_iterative_refinement
+import LazySets.Approximations.refine
```
schillic: Maybe do `import LazySets.Approximations: refine, tohrep`?

docs/src/man/iterative_refinement.md
```diff
@@ -11,61 +12,213 @@ Depth = 3
...
+p3 = plot!(p0, overapproximate(b, 0.01), alpha=0.4, aspectratio=1)
+
+plot(p1, p2, p3, layout=(1, 3))
+```
+
+We can check that the error is getting smaller with `ε` in each case:
+
+```@example example_iterative_refinement
+f = x -> (minimum(x), maximum(x))
+g = ε ->  f([ai.err for ai in Approximations.approximate(b, ε).approx_list])
```
schillic: Maybe add `import LazySets.Approximations.approximate` to remove `Approximations.approximate` twice?

docs/src/man/iterative_refinement.md
```diff
@@ -11,61 +12,213 @@ Depth = 3
...
+1. *Initialization*. The approximation is initialized with box directions,
+   i.e. it starts with four `LocalApproximation` objects. Let `i=1`.
+2. *Refinement loop*. If the local approximation at index `i` has an error greater
+   than the threshold `ε`, then refine. Otherwise, increment `i <- i+1`.
+3. *Redundancy check*. Insert the refined right approximation at position `i`,
+   and check whether the left approximation is redundant or not with respect to the one
+   at position `i+1`. Checking for redundancy amounts to checking for overlap of both
+   `p1` and `q`. Then, either substitute at `i+1` or insert (keeping the approximation
+    at `i+1`) depending on the redundancy check.
+4. *Stopping criteria*. If the index `i` is smaller or equal than the current length of
```
schillic: "criterion"?
Currently it is rather a criterion when to *not* stop. What about:
```
*Stopping criterion*. Terminate if the index `i` exceeds the current length of the approximations list; otherwise continue with step 2.
```

docs/src/man/iterative_refinement.md
```diff
@@ -11,61 +12,213 @@ Depth = 3
...
+h(1.), h(0.1), h(0.01)
+```
+
+!!! note
+    Actually, the plotting function for an arbitrary `LazySet` `plot(...)`,
+    called *recipe* in the context of [Plots.jl](https://github.com/JuliaPlots/Plots.jl),
+    is such that it receives a numeric argument `ε` and the routine itself calls
+    `overapproximate`. However, some sets such as abstract polygons have
+    their own plotting recipe hence don't require the error threshold, since
+    they are plotted exactly as the convex hull of its vertices.
```
schillic: "its" → "their"

schillic:
Reading the tutorial I realized that all this `insert!` is not very efficient (in practice it might be *okay* because you start at the front).
What about having two arrays, one for finished refinements and one for unfinished ones?

```julia
initialize with box direction approximations
[] [a b c d]
turn around initial array
[] [d c b a]
pick a
[] [d c b]
refine a to (d, e)
add them in the opposite order
[] [d c b e d]
pick d
assume it is not refinable, so add it to the left
[d] [d c b e]
pick e
[d] [d c b]
assume it is not refinable, so add it to the left
[d e] [d c b]
...
```

docs/src/man/iterative_refinement.md
mforets: yes this is better, thx.

docs/src/man/iterative_refinement.md
mforets: yes

docs/src/man/iterative_refinement.md
mforets: done

docs/src/man/iterative_refinement.md
```diff
@@ -11,61 +12,213 @@ Depth = 3
...
-The basic idea is to add new supporting directions whenever the approximation
-error is still bigger than ``ɛ``.
-
 ## Local approximations
 
-The approximation is represented by a list of local approximations or refinements.
-Each refinement describes a set with one angle.
+The polygonal approximation of an arbitrary lazy convex set `S` is represented by a
+list of local approximations or refinements. Each refinement is used to define a
+linear constraint, and the union of all such local approximations constitute an
```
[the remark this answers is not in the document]
mforets: yes, it defines two support lines but we pick just the one that is normal to `d1`. i found "a set with one angle" too vague. i've refactored the text a bit.

docs/src/man/iterative_refinement.md
```diff
@@ -11,61 +12,213 @@ Depth = 3
...
-Then we just iterate through all refinement instances and check if the error is
-bigger than the threshold individually.
-If so, we refine the instance by splitting into two more precise refinement
-instances and apply the checks recursively.
+plot(b, 1e-3, aspectratio=1, alpha=0.3)
+
+plot!(Singleton([1.0, 0.0]), annotations=(1.1, 0.08, text("p1")), color="green")
+plot!(Singleton([0.0, 1.0]), annotations=(0.1, 1.1, text("p2")), color="green")
+plot!(Singleton([1.0, 1.0]), annotations=(1.09, 1.06, text("q")))
+plot!(annotations=(1.5, 0.1, text("d1")))
```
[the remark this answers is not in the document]
mforets: edited, could you please check it looks better?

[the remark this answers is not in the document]
mforets: ooops, i moved the singleton as well, let me fix..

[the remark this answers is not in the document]
schillic: "p1" and "d1" still have different `y` coordinates.

docs/src/man/iterative_refinement.md
schillic: But also remove the two `Approximations.tohrep` 😉

mforets:
> Reading the tutorial I realized that all this insert! is not very efficient (in practice it might be okay because you start at the front).

i don't know if it's optimal, but yes it goes sequentially and asks each approximation if  its `.refinable` only once because it doesn't mix up the order of the approximations.

schillic:
> i don't know if it's optimal, but yes it goes sequentially and asks each approximation if its `.refinable` only once because it doesn't mix up the order of the approximations.

True. I will still create an issue, not meant to be resolved here.

Issue #104, opened 2017-12-30
Fix Singleton examples


mforets:
`plot([Singleton([0.0, 0.0]), Singleton([1., 0]), Singleton([0.5, .5])])` does no longer work because of the more general parametric `Vector{<:LazySet}`. it requires that the given input vector is itself of `Singleton` type.

schillic:
The vector is of `Singleton` type.
```julia
typeof([Singleton([0.0, 0.0]), Singleton([1., 0]), Singleton([0.5, .5])])
Array{LazySets.Singleton{Float64},1}
```
I think the problem is rather the numeric type parameter. The function expects a `Vector{Singleton}`.
```julia
[Singleton([0.0, 0.0]), Singleton([1., 0]), Singleton([0.5, .5])] isa Vector{Singleton}
false
[Singleton([0.0, 0.0]), Singleton([1., 0]), Singleton([0.5, .5])] isa Vector{Singleton{Float64}}
true
```
Here is my solution:
```julia
[Singleton([0.0, 0.0]), Singleton([1., 0]), Singleton([0.5, .5])] isa Vector{<:Singleton}
true
```
So I suggest to add `<:` to all vectors of the input parameters.

schillic:
Side comment: Do you actually need those specific vector functions? Can you not define just one function for `LazySet` and then call the plot function for each element inside?
```julia
@recipe function plot_vector(arr::Vector{<:LazySet}; ...)
```
(This might be completely wrong - I do not understand this Plots.jl thing.)

mforets:
changing the function's signature 

```julia
@recipe function plot_singleton(arr::Vector{<:AbstractSingleton};
                                color="blue", label="", grid=true, legend=false)
```
does the job, good observation! let me make this a PR.

i think the primary use is to do a `plot(S)` where `S` is a lazy set. the vector argument is more of a short cut for a chain of `plot -> plot! -> plot!` etc. i'm not sure that we want to mix these.

schillic:
From [Plots.jl](http://docs.juliaplots.org/latest/basics/):
If the argument is a "matrix-type", then each column will map to a series, cycling through columns if there are fewer columns than series. In this sense, a vector is treated just like an "nx1 matrix".

mforets:
that would suggest that i can safely remove the arguments with `Vector{...}`. i just tried that for the Singleton case and it didn't plot anything.

schillic:
[...]

PR #119, opened 2018-01-06
#104 - Singleton vector plot

mforets:


schillic:
👍 for using `AbstractSingleton`

Issue #44, opened 2017-11-22
Parametric numeric type of a LazySet 


mforets:
In some cases we'd like to have `eltype(X)` where `X` is a `LazySet`, to return the underlying numeric type. Eg. `eltype(BallInf(ones(2), 0.1))` would be `Float64`.

In principle  this would just require

```
abstract type LazySet{N<:Number} end
```

But we should see if this makes sense for all LazySets, both set representations and operations.

schillic:
Maybe better `numtype`.

schillic:
Related: We should get rid of the `Int64` types at some places, e.g., in `HPolygonOpt`.
AFAIK this restricts the usage to 64-bit machines.

schillic:
I tried to add type parameters to an abstract type today and it did not work.
Maybe I did a mistake; otherwise it just does not work in Julia.
We can still have this function for every subtype that does the trick, though.

schillic:
There are TODOs for this for the box directions (`DIR_EAST` etc.).

PR #122, opened 2018-01-07
#44 - Parametric numeric type of a LazySet 

schillic:
See #44.

This took way longer than I expected, and unfortunately there is a caveat:
I needed to introduce an additional numeric type parameter to the composite types (`CartesianProduct`, `MinkowskiSum`, `ConvexHull`, `LinearMap`, `ExponentialMap`) because I did not find a way how to write the signature with nested type parameters (I searched **a lot** in documentations) like this:
```julia
CartesianProduct(s::S1 where S1<:LazySet{N<:Real}, s::S2 where S2<:LazySet{N}) <: LazySet{N}
```
The problem is that `N` is not visible outside. The outer `where` syntax is only meant for functions.

@mforets: There were hidden problems in the unit tests. Let `B` be a `BallInf`. Then `2.*B` is interpreted as `2 .* B` instead of the intended `2. * B`. I changed this in a separate commit, please note this for the future.

**EDIT: All known errors are fixed now.**
The unit tests were priceless. This typing stuff is hard to do right (at least for me 🤕). We might want to test a bit more. We should also run the reachability examples for additional testing.

I also took the liberty to:
* remove the `<: LazySet`from `UnionSet`
* remove some redundant function definitions
* remove some output types for simple functions where they are crystal clear
* fix the plot recipes for vectors of types different from `AbstractSingleton` (this was recently fixed for `AbstractSingleton` in #119)
* fix subtype problems in the recent #110 (not sure why they did not trigger before)
* fix a type conversion bug in `LinearMap` with a constant (was always a matrix of type `Int`)
* add type parameters to the doc files (creates more warnings, but should be the correct way to do it; for some this was even required due to the changes here)

### DIFF
--- a/src/LinearMap.jl
+++ b/src/LinearMap.jl
@@ -61,7 +61,7 @@ Return a linear map of a convex set by a scalar value.
 The linear map of the convex set.
 """
 function *(a::Real, S::LazySet)::LinearMap
-    return LinearMap(sparse(a*I, dim(S)), S)
+    return LinearMap(a * speye(typeof(a), dim(S)), S)
 end
 
 """
--- a/src/LinearMap.jl
+++ b/src/LinearMap.jl
@@ -3,7 +3,7 @@ import Base: *, ∈
 export LinearMap
 
 """
-    LinearMap{S<:LazySet, N<:Real} <: LazySet
+    LinearMap{N<:Real, S<:LazySet{N}} <: LazySet{N}
 
 Type that represents a linear transformation ``M⋅S`` of a convex set ``S``.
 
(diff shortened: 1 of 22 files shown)


src/LinearMap.jl
```diff
@@ -61,7 +61,7 @@ Return a linear map of a convex set by a scalar value.
 The linear map of the convex set.
 """
 function *(a::Real, S::LazySet)::LinearMap
-    return LinearMap(sparse(a*I, dim(S)), S)
+    return LinearMap(a * speye(typeof(a), dim(S)), S)
```
mforets: how are left and right different? is this related to:

> fix a type conversion bug in LinearMap with a constant (was always a matrix of type Int)

schillic: Yes, I separated this in a single commit. I can try again, but I think it had the `Int` type previously.

schillic: Hm, you are right, I cannot reproduce the problem anymore...

schillic:
```julia
sparse(1*I, 5)
5×5 SparseMatrixCSC{Int64,Int64}
a=1.; a * speye(typeof(a), 5)
5×5 SparseMatrixCSC{Float64,Int64}
```
This is what I meant.

PR #131, opened 2018-01-10
add size assertion to linear map

mforets:


src/LinearMap.jl
```diff
@@ -37,6 +37,7 @@ Return the linear map of a convex set.
 If the matrix is null, a `ZeroSet` is returned; otherwise a lazy linear map.
```
schillic: Is "null" a common term?

src/LinearMap.jl
```diff
@@ -37,6 +37,7 @@ Return the linear map of a convex set.
...
 """
 function *(M::AbstractMatrix, X::LazySet)
+    @assert dim(X) == size(M, 2) "a linear map of size $(size(M)) cannot be applied to a set of dimension of $(dim(X))"
```
schillic: Is "of dimension of" correct?

src/LinearMap.jl
mforets: as a matrix whose entries are all zero, see the wikipedia article [Zero matrix](https://en.wikipedia.org/wiki/Zero_matrix). so i could use that, or "If all the entries of the matrix are zero, ..."

src/LinearMap.jl
mforets: thanks, it's a typo 👍

src/LinearMap.jl
schillic: I am fine with "null".

PR #140, opened 2018-01-11
#139 - Intersection of two sets

### LINKED ISSUE #139: Intersection of two sets
Add a lazy binary operator type `Intersection` (and maybe an n-ary variant) representing the intersection of two sets `S1` and `S2`.
Also have a convenience alias `∩(S1, S2)`, for which Julia's built-in alias is `intersect`.

Related but simpler: #67

schillic:
Closes #139.

Missing:
- [x] support vector (+docs for zero vector case) // we do not implement the support vector for the moment
- [x] include file
- [x] docs
- [x] tests

### DIFF
--- a/docs/src/lib/operations.md
+++ b/docs/src/lib/operations.md
@@ -101,6 +101,14 @@ ProjectionSparseMatrixExp
 *(::ProjectionSparseMatrixExp{Float64}, ::LazySet{Float64})
 ```
 
+## Intersection
+
+```@docs
+Intersection
+dim(::Intersection{Float64, LazySet{Float64}, LazySet{Float64}})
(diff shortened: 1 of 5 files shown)


mforets:
have you considered to run `is_intersection_empty` in the constructor? (it could be an option). thus if the intersection is empty you return the empty set, otherwise keep the symbolic intersection.

schillic:
This function is not implemented for most types, so this will usually break.

mforets:
my point was that we can remove the `intersection` field, since it really gets info from `is_intersection_empty`, so it's really the same missing functionality (yet). unless the user knows in advance if the intersection is empty or not, and wants to put this info in the type, but maybe that's not the common use case.

schillic:
Okay, the motivation was that this is stored for future queries, but we don't store things for other lazy types, so we can also just remove it.

Issue #144, opened 2018-01-16
Left margin of some figures is cut


mforets:
Some left margins are cut in figures of `@example` blocks, such as the last of iterative refinement example.

schillic:
I also do not have this problem locally.

The other two plots on above mentioned figure are also pressed to the left such that they partially overlap with the next plot. So I think instead of font size decrease you should rather add some space to the left. A hack might be to add a "y" axis label (empty/invisible if possible).

The problem *might* be with `GR` and that the Travis computers do not have a display. See the following warning from the build output:
```
[...]
Documenter: expanding markdown templates.
GKS: can't open display on ""
     Is your DISPLAY environment variable set correctly?
     Did you enable X11 and TCP forwarding?
[...]
```
But this warning could also be completely unrelated. In particular, this does not explain why we still get an almost-correct plot output...

schillic:
Hot trace: Issue 1182 [here](https://github.com/JuliaPlots/Plots.jl).

schillic:
The following pages were affected:
* [x] iterative refinement
* [ ] interval hulls
* [x] convex hulls
* [x] a reachability algorithm

#154 solved three of them. For the last one we might just reduce the image size a bit.

PR #154, opened 2018-01-18
#144 - set GR to 'no output'

schillic:
See #144.


schillic:
I will just merge and check the plots on master...

PR #158, opened 2018-01-18
#103 - Implement convex hull for an array of sets

### LINKED ISSUE #103: Implement convex hull array
Implement the symbolic convex hull array operator such that

```julia
julia> C = ConvexHullArray([Singleton([1.0, 0.5]), Singleton([1.1, 0.2]), Singleton([1.4, 0.3]), Singleton([1.7, 0.5]), Singleton([1.4, 0.8])])
```

works (alias: `CHArray`). The choice `xyzArray` follows the nomenclature used previously for `MinkowskiSum` and `CartesianProduct`.

mforets:
Closes #103.

### DIFF
--- a/src/ConvexHull.jl
+++ b/src/ConvexHull.jl
@@ -143,12 +143,35 @@ Alias for `ConvexHullArray`.
 """
 const CHArray = ConvexHullArray
 
-ConvexHullArray(X::LazySet{N}, ::EmptySet{N}) where {N<:Real} = X
-ConvexHullArray(::EmptySet{N}, X::LazySet{N}) where {N<:Real} = X
-(ConvexHullArray(∅::E, ::E)) where {E<:EmptySet} = ∅
-dim(ch::ConvexHullArray) = dim(ch.array[1])
+"""
+    dim(cha::ConvexHullArray)::Int
 
-function σ(d::AbstractVector{<:Real}, ch::ConvexHullArray)::AbstractVector{<:Real}
+Return the dimension of the convex hull of a finite number of convex sets.
+
+### Input
+
+- `cha` -- convex hull array
+
+### Output
+
+The ambient dimension of the convex hull of a finite number of convex sets.
+"""
+function dim(cha::ConvexHullArray)::Int
+    @assert !isempty(cha)
+    return dim(cha.array[1])
+end
+
+"""
+    σ(d::AbstractVector{<:Real}, cha::ConvexHullArray)::AbstractVector{<:Real}
+
+Return the support vector of a convex hull array in a given direction.
+
+### Input
+
+- `d`   -- direction
+- `cha` -- convex hull array
+"""
+function σ(d::AbstractVector{<:Real}, cha::ConvexHullArray)::AbstractVector{<:Real}
     s = r = similar(d)
     for (i, ci) in enumerate(ch.array)
        s[i] = σ(d, ci)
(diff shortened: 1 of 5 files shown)


src/ConvexHull.jl
```diff
@@ -106,3 +107,52 @@ function σ(d::AbstractVector{<:Real}, ch::ConvexHull)::AbstractVector{<:Real}
...
+
+### Fields
+
+- `array` -- array of sets
+"""
+struct ConvexHullArray{N<:Real, S<:LazySet{N}} <: LazySet{N}
+    array::Vector{S}
+end
+# type-less convenience constructor
+ConvexHullArray(a::Vector{S}) where {S<:LazySet{N}} where {N<:Real} = ConvexHullArray{N, S}(a)
```
schillic: Is this needed when there is no inner constructor? I remember that sometimes you need it...

src/ConvexHull.jl
```diff
@@ -106,3 +107,52 @@ function σ(d::AbstractVector{<:Real}, ch::ConvexHull)::AbstractVector{<:Real}
...
+(ConvexHullArray(∅::E, ::E)) where {E<:EmptySet} = ∅
+dim(ch::ConvexHullArray) = dim(ch.array[1])
+
+function σ(d::AbstractVector{<:Real}, ch::ConvexHullArray)::AbstractVector{<:Real}
+    s = r = similar(d)
+    for (i, ci) in enumerate(ch.array)
+       s[i] = σ(d, ci)
+       r[i] = dot(d, s[i])
+    end
+    return s[indmax(r)]
```
schillic: Are you sure about this? What about negative directions?

src/ConvexHull.jl
mforets: yes, its' needed, just checked

mforets:
i changed the algorithm for σ(d, cha) (only keep greatest value of support vector), and added some examples. the idea of the second one was the plot (but i don't want to add `using Plots`)

```julia
b = [Ball2([2*pi*i/100, sin(2*pi*i/100)], 0.01) for i in 1:100];
c = ConvexHullArray(b);
plot(c, 1e-3, alpha=0.2, color="red")
plot!(b, 1e-3, alpha=0.2, color="orange")
```

maybe i'll propose to add it to the [Convex hulls](https://juliareach.github.io/LazySets.jl/latest/man/convex_hulls.html) manual section.

mforets:
done, this is ready to go..

schillic:
It is a bit inconvenient that we do not have a function symbol for the convex hull. Now if you want to add a new set to a `ConvexHullArray`, you have to access the underlying array. Should we say that `CH(cha::ConvexHullArray, X::LazySet) = cha.array.add(X)` (+ the symmetric case + the *merge* case for two `ConvexHullArray`s)?

src/ConvexHull.jl
```diff
@@ -164,6 +164,18 @@ Alias for `ConvexHullArray`.
 """
 const CHArray = ConvexHullArray
 
+CH(cha::ConvexHullArray, ∅::EmptySet) = cha
+CH(cha::EmptySet, ::ConvexHullArray) = cha
```
schillic: `cha` should be the *second* argument.

src/ConvexHull.jl
```diff
@@ -164,6 +164,18 @@ Alias for `ConvexHullArray`.
...
+
+function CH(cha::ConvexHullArray, X::LazySet)
+    push!(cha.array.X)
```
schillic: How does this work? Is it not `push!(cha.array, X)`?

mforets:
the `CH(cha::ConvexHullArray, X::LazySet)` requires i think to change the def of `CH` so i'm leaving this outside this issue, see #165.

Issue #57, opened 2017-12-02
Revisit LinearConstraints.jl


schillic:
1. Rename the file to the singular form?
2. Outsource `Line` to a new file?
3. Should `Line`'s field `a::Vector{N}` become `a::Tuple{N, N}` because the constructor requires 2D input?
On the other hand, code-wise I do not see any problem with a `Line` having more than two dimensions. We only use it for 2D plots, but why should we restrict the general type? If we want to keep it 2D, I would rename it to `Line2` or so.
4. Should we promote `Line` to `LazySet`?
5. (related) Should we have a `LazySet` type for a half-space? This is inspired by the doc for `LinearConstraint`.

Action points here:
* [x] rename `LinearConstraints.jl` to `Line.jl` and promote `Line` to `LazySet`
* [x] add `LinearConstraint` as alias to `HalfSpace` and remove the type
* [x] outsource `intersection` function to a new file `concrete_intersection.jl`

schillic:
From a discussion with @mforets:
This type is only used in `HPolygon(Opt).jl` and `iterative_refinement.jl` so far.
Maybe change the return type to `LazySet` which can here be either a `Singleton`, an `EmptySet` or a `Line` again.
We might even want to add a type `Point2D <: AbstractSingleton` of fixed size for this purpose.

mforets:
1. ok
2. ok
3. `Line` is understood for 2d (or 3d). but for higher dims we should better use `Hyperplane` >> let's create a new issue?
4. ok -- what to return for its support vector?
5. ok -- new issue at #149

schillic:
3. This is #150.

schillic:
4. I have a similar problem for the half-space and the hyperplane. My thinking is that a solution only exists if the direction is either the zero vector or a normal of the line/plane. In those cases we can return any point `x` on the line/plane, i.e., such that `a⋅x = b`, as follows: Assuming that `a` is not the zero vector, we can choose `x[j] = 0` for all but one dimension `i` such that `a[i] ≠ 0`. For dimension `i` we can choose `x[i]` such that `x[i] = b / a[i]`. In other words, we return a scaled unit vector.
(Maybe we should check in the constructor that `a` is not the zero vector...)

schillic:
Should we get rid of `LinearConstraint` and `Line` because they mean exactly the same as `HalfSpace` and `Hyperplane`?
We could have an alias `LinearConstraint` because it sounds more natural sometimes.

[...]

PR #160, opened 2018-01-18
#149/#150 - HalfSpace/Hyperplane

### LINKED ISSUE #149: Add `HalfSpace` set type
This is part of the discussion in #57.

### LINKED ISSUE #150: Add `Hyperplane` set type
It seems that this type has almost identical behavior (and implementation) to a `HalfSpace` (#160). I only see a difference in `∈` and in the fact that for a hyperplane the normal vector can also be additively inverted.

Proposal: Put most of the functions from `HalfSpace` in this type and create an auxiliary `Hyperplane` instance in the `HalfSpace` functions.

schillic:
Closes #149.
Closes #150.

The implementation idea is sketched [here](https://github.com/JuliaReach/LazySets.jl/issues/57#issuecomment-358692135).

### DIFF
--- a/docs/src/lib/representations.md
+++ b/docs/src/lib/representations.md
@@ -260,6 +260,16 @@ singleton_list(::Zonotope{Float64})
 order(::Zonotope{Float64})
 ```
 
+## Hyperplane
+
+```@docs
+Hyperplane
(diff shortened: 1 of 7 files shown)


schillic:
From my side this is ready.

Issue #113, opened 2018-01-06
Add manual for set operations


schillic:
Add a manual page for the new set operations (`∈`, `⊆`, `is_intersection_empty`).

* [ ] operation names with syntax, description, aliases
* [ ] table with available combinations (see [here](https://github.com/JuliaReach/LazySets.jl/issues/69#issuecomment-350864515))
* [ ] examples

mforets:
and also union `∪` that needs to be added (the symbol) would go in the same page.

schillic:
What would be the meaning of `∪`?

schillic:
I was thinking if we want to move the code of the binary functions to new files (`subset.jl`/`intersection.jl`). The reason is that it is not clear where `⊆(::T1, ::T2)` should be, in `T1.jl` or in `T2.jl`. It would also simplify the type files a bit and you can find existing implementations faster (currently you have to search in all files).

mforets:
> What would be the meaning of ∪?

set union. we've got the file `UnionSet.jl`, but it's a rather hidden one, since it is not even included in `LazySets.jl` !

schillic:
>> What would be the meaning of ∪?
>
>set union. we've got the file `UnionSet.jl`, but it's a rather hidden one, since it is not even included in `LazySets.jl` ! 

I see, two remarks:
1. I do not want to have `UnionSet.jl` being a `LazySet` because it is not convex and thus breaks some assumptions.
2. The function `∩` does *not* compute the intersection. It checks if the intersection is empty. Maybe we should rename it, in particular if we want to have a function that computes the intersection.

mforets:
1. agreed. since this is a running assumption in LazySets, we may in the future open another project repo to handle non-convex sets, like an extension of LazySets.

2. yes, we should rename it. we'd need to compute the intersection for applications.

schillic:
1. Programmatically this would be simple: Just define a supertype of `LazySet`. But we should probably instead have an abstract type `ConvexSet` in between, otherwise it will be confusing.
2. Should we create a new issue for this or do it in this one?

mforets:
+1 for new issue

schillic:
added #117

schillic:
[...]

PR #163, opened 2018-01-20
#113 - outsource binary functions

schillic:
This is one step from #113.
Basically moving functions around...
I also merged two identical implementations for `Ball2`/`Ballp`.


schillic:
I found at least one combination where there are method ambiguities:
`⊆(::AbstractSingleton, ::AbstractHyperrectangle)`
`is_intersection_empty(::AbstractSingleton, ::AbstractHyperrectangle)`
They should be resolved.

The unit test code does not work fully correctly yet, and it does not test non-interface types such as `EmptySet`. The idea is that this test code detects method ambiguities.

PR #164, opened 2018-01-20
#113 - docs for set operations

### LINKED ISSUE #113: Add manual for set operations

schillic:
Closes #113.

- [x] tables
- [x] examples
- [x] clarify `norm` documentation
- [x] outsource docs to new `lib` file


schillic:
I found at least one combination where there are method ambiguities:
`⊆(::AbstractSingleton, ::AbstractHyperrectangle)`
`is_intersection_empty(::AbstractSingleton, ::AbstractHyperrectangle)`
They should be resolved.

The unit test code does not work fully correctly yet, and it does not test non-interface types such as `EmptySet`. The idea is that this test code detects method ambiguities.

docs/src/man/set_operations.md
```diff
@@ -0,0 +1,166 @@
...
+In this section we show which typical set operations this library supports.
+
+```@contents
+Pages = ["set_operations.md"]
+Depth = 3
+```
+
+## Unary operations
+
+The following table lists all operations (apart from `dim` and ``, which must be
```
mforets: is σ missing in the parentheses?

docs/src/man/set_operations.md
```diff
@@ -0,0 +1,166 @@
...
+| `LinearMap`                  |        |          |      |            | x |
+| `MinkowskiSum`               |        |          |      |            |   |
+| `MinkowskiSumArray`          |        |          |      |            |   |
+| `SymmetricIntervalHull`      |  (x)   |   (x)    | (x)  |    (x)     |(x)|
+
+
+### `radius`
+
+This function returns the radius of a set.
+It is defined as the enclosing ball (of the given norm) of minimal volume.
```
mforets: "as the radius of" ?

docs/src/man/set_operations.md
```diff
@@ -0,0 +1,166 @@
...
+
+### `diameter`
+
+This function returns the diameter of a set.
+It is defined as the enclosing ball (of the given norm) of minimal volume.
```
mforets: "as the diameter of" ?

mforets:
> The unit test code does not work fully correctly yet (hence deactivated), and it does not test non-interface types such as EmptySet

if you prefer we can comment out the WIP unit test, while merging this issue for the new manual section.

schillic:
Okay for splitting. Then in the new issue I would also resolve the method ambiguities.

Do you think this manual is sufficient for merging already?
* I wanted to add some examples, one for `true`/`false` with and without witness production.
* Should we move the documentation of the different implementations here (and put the file into the `lib` folder)?

mforets:
some comments:

> Should we move the documentation of the different binary implementations here (and put the file into the lib folder)?

we were writing in the `lib` the `@docs` blocks, and in the `man` those more high level explanation, what do you think?

> Should I add dim and σ to the unary table? 

no, i think that wouldn't add new interesting info, or we can say this in a sentence but outside the table.

> I would clarify that norm is the radius of the ball centered in the origin in all function docs.

👍

schillic:
> > Should we move the documentation of the different binary implementations here (and put the file into the lib folder)?
>
> we were writing in the lib the @docs blocks, and in the man those more high level explanation, what do you think?

Let me rephrase: Where should we put the `@docs` blocks? Currently they are with the type documentations, but this does not work because for `⊆(T1, T2)` we have to decide whether to add it to `T1` or to `T2`. I think we should have a new section or file in `lib` for binary operations.

mforets:
ok, thanks for the remark. indeed a new file `lib` file for binary operations seems appropriate.

schillic:
I outsourced the documentation and also already used the new name introduced in the parallel #172.

From my side this is ready to go.

PR #169, opened 2018-01-21
Fix bug in check_method_implementation

### LINKED ISSUE #168: Implement vertices_list for HPolytope
We have `HPolytope <: AbstractPolytope`, but this interface requires an implementation of `vertices_list`. Consequently, some operations break.

There is a bug in the interface checking function, hence we did not detect that this function is missing.

Actually, if we have `vertices_list`, we might as well add `VPolytope` because we get `tovrep` for free (we can leave `tohrep` open for now).

schillic:
* [x] fix bug in `check_method_implementation`
* [x] implement `vertices_list` for `HPolytope` >> see #189

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -7,3 +7,5 @@ Expokit
 MathProgBase
 GLPKMathProgInterface
 Requires
+Polyhedra
+CDDLib
--- a/src/helper_functions.jl
+++ b/src/helper_functions.jl
@@ -132,7 +132,9 @@ function check_method_implementation(interface::Type,
                                      args_funcs::AbstractVector{Function};
                                      print_results::Bool=false
                                     )::Bool
+    has_subtypes = false # NOTE: 'isleaftype' does not work (type parameters)
     for subtype in subtypes(interface)
+        has_subtypes = true
         found = false
         for args_func in args_funcs
             if method_exists(func_name, args_func(subtype))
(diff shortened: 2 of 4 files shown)


schillic:
@mforets: How should we implement `vertices_list`?
One idea: Implement the intersection for two `Hyperplane`s (result: a `Hyperplane` of lower dimension). Then intersect all defining hyperplanes of the polytope as long as possible until you arrive at 1D. Is this necessary for every (unordered) pair of constraints?

mforets:
does the vertices list functionality depend on  the "check_method_implementation"? how does that work?

schillic:
It just checks if the method exists. When #189 is merged, it should work (we should still try).

mforets:
do you refer to this test?

```julia
@test check_method_implementation(AbstractPolytope, vertices_list,
                                  Function[S -> (S{Float64},)])
```

schillic:
Yes. I just rebased with master. If it gets green, you can merge.

mforets:
the access of the method `vertices_list` requires Polyhedra (it's hidden under a `@require` macro, since we don't want it to be loaded at startup). Maybe we needs `using Polyhedra` before the test of this file?

schillic:
Indeed:
```julia
julia> method_exists(vertices_list, (HPolytope{Float64},))
false

julia> using Polyhedra

julia> method_exists(vertices_list, (HPolytope{Float64},))
true
```

schillic:
Unfortunately, when adding `using Polyhedra`, I get an error with `Halfspace`, see #194.

schillic:
For now my fix is to load this test set last. We should still investigate #194 another time.

PR #173, opened 2018-01-22
Polyhedra interface

mforets:


schillic:
I wonder if would be better to store `A`/`b` instead of the constraints. It seems to me that for the interesting operations we need to convert to `A`/`b` before we can do something.

mforets:
it could be, although for support vector computations it's more natural to use the list of constraints. FWIW in `Polyhedra.jl` the `hreps` iterator also gives the list of hyperplanes and halfspace objects.

for the rest of this PR let me propose to:

- [ ] create a `HRep` from a `HPolytope`
- [ ] the converse
- [ ] add concrete ops:  convex hull and cartesian product

and add the `VPolytope` in a separate PR.

schillic:
>  and add the VPolytope in a separate PR.

Do we need it for now at all?

mforets:
only for / it is related to #169

schillic:
> it could be, although for support vector computations it's more natural to use the list of constraints.

But in our implementation we construct a matrix and a vector from the constraints. (This code is old.) So we have to convert for every query.

mforets:
so far as the functionalities are concerned this is done. i'll move on to the vertices list in #169

src/HPolytope.jl
```diff
@@ -164,3 +179,138 @@ function constraints_list(P::HPolytope{N}
...
+    for hi in hreps(P)
+        push!(constraints, LinearConstraint(hi.a, hi.β))
+    end
+    return HPolytope(constraints)
+end
+
+"""
+    polyhedron(P::HPolytope{N}, backend=CDDLib.CDDLibrary()) where {N}
+
+### Input
```
schillic: one-line comment missing

src/HPolytope.jl
```diff
@@ -164,3 +179,138 @@ function constraints_list(P::HPolytope{N}
...
+function HPolytope(P::HRep{N, T}, backend=CDDLib.CDDLibrary()) where {N, T}
+    constraints = LinearConstraint{T}[]
+    for hi in hreps(P)
+        push!(constraints, LinearConstraint(hi.a, hi.β))
+    end
+    return HPolytope(constraints)
+end
+
+"""
+    polyhedron(P::HPolytope{N}, backend=CDDLib.CDDLibrary()) where {N}
```
schillic: According to the rules, `backend` should be in brackets in all the docs here.

src/HPolytope.jl
```diff
@@ -164,3 +179,138 @@ function constraints_list(P::HPolytope{N}
...
+
+An `HRep` polyhedron.
+"""
+function polyhedron(P::HPolytope{N}, backend=CDDLib.CDDLibrary()) where {N}
+    return polyhedron(SimpleHRepresentation(tosimplehrep(P)...), backend)
+end
+
+"""
+    intersect(P1::HPolytope{N}, P2::HPolytope{N},
+                           backend=CDDLib.CDDLibrary())::HPolytope where {N<Real}
```
schillic: `prunefunc` is missing (and should be in brackets).

PR #174, opened 2018-01-22
Concrete operations for zonotopes

### LINKED ISSUE #148: Concrete set operations for zonotopes
Start with a `minkowski_sum` function that computes the concrete Minkowski sum of two `Zonotope`s.
Maybe extend to the product (see also [here](https://github.com/mforets/Zonotopes.jl/blob/master/src/Zonotopes.jl)).

mforets:
Closes #148.

### DIFF
--- a/docs/src/lib/representations.md
+++ b/docs/src/lib/representations.md
@@ -258,6 +258,9 @@ center(::Zonotope{Float64})
 vertices_list(::Zonotope{Float64})
 singleton_list(::Zonotope{Float64})
 order(::Zonotope{Float64})
+minkowski_sum(Z1::Zonotope, Z2::Zonotope)
+linear_map(M::AbstractMatrix, Z::Zonotope)
+scale(α::Real, Z::Zonotope)
 ```
 
 ## Hyperplane
--- a/src/Zonotope.jl
+++ b/src/Zonotope.jl
@@ -1,7 +1,10 @@
 import Base.∈
 
 export Zonotope,
-       order
+       order,
+       minkowski_sum,
+       linear_map,
+       scale
 
 """
     Zonotope{N<:Real} <: AbstractPointSymmetricPolytope{N}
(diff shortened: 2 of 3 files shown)


schillic:
Should we have these concrete functions (not in this PR) for any `LazySet` where they just apply the lazy type? (This would also mean removing the different constructors and instead adding different instances of these functions.)

mforets:
thanks for the review.

yes, i think it makes sense to add concrete functions to other types as well. for instance, the polytope is a good candidate. could you expand on "This would also mean removing the different constructors" ?

schillic:
Question: Why did you not add a `MinkowskiSum` constructor for `Zonotope`s?
Answer: Because you wanted to reserve the `MinkowskiSum` constructor for the lazy operation.

I think that currently the design is inconsistent. For some lazy operations we actually do a concrete operation (like `LinearMap(lm::LinearMap)`, `MinkowskiSum(E::EmptySet, S::LazySet)`) and for some concrete operations and only some types we have another function (like `minkowski_sum`). Why should the user know what to use? In particular in an application where you do not know the set type in advance!

I was thinking if we should have this:
```julia
minkowski_sum(S1::LazySet, S2::LazySet)
    return MinkowskiSum(S1, S2)
end
minkowski_sum(Z1::Zonotope, Z2::Zonotope)
    # what you did here
end
minkowski_sum(E::EmptySet, S::LazySet)
    return S
end
minkowski_sum(T1::Type1, T2::Type2) # other type combinations where concrete implemenations are simple
    ...
end
```
Then we could remove the specific `MinkowskiSum` constructors like `MinkowskiSum(E::EmptySet, S::LazySet)`. The user would use this new function instead and "under the hood" we decide what happens (lazy or concrete).

But now I realized that this is a design decision and I do not have an opinion here. Previously you had to call the constructor `MinkowskiSum` and it either created a lazy `MinkowskiSum` object or, in special cases, skipped it (`EmptySet`). Now we have the mixture (sometimes you have to call the constructor and sometimes it is concrete; sometimes you have to call this function to get the concrete implementation).

The question is if we want to give the user the freedom to choose between lazy and concrete set operations. I see three options for who decides between lazy and concrete operations:
1. *We* decide: We only provide one interface (namely the `MinkowskiSum` constructor). In that case we should remove the function you wrote and add a respective constructor.
2. *The user* decides: The constructor `MinkowskiSum` is always lazy; the function `minkowski_sum` is always concrete if possible (and lazy otherwise).
3. *The user* **can** decide but *we* provide also a default choice.

I prefer the most general answer (3.), but it requires some discussion. Example implementation:
The constructor `MinkowskiSum` is always lazy; the function `minkowski_sum` gets a new argument from the domain {"lazy", "concrete", "default"} with default value "default".
Alternative: Have three functions `minkowski_sum_lazy`, `minkowski_sum_concrete`, and `minkowski_sum`.

We could move this to a new issue if you want.

Issue #132, opened 2018-01-11
Add unit tests for different numeric types


schillic:
Currently all unit tests are for `Float64`.
Add tests for all functions to check if they preserve the numeric type; most importantly `Rational`, maybe also `Float32`.

schillic:
There are quite problems that prevent using different numeric types.
 * [ ] `Ball2` - `σ`
* [ ] `Hyperrectangle` - `≈`??
* [ ] `Polygon` - `<=`
* [ ] `ConvexHull` - Overflow??
* [ ] `ExponentialMap` - `σ`
  There are also some commented lines in the test file...
* [ ] `CartesianProduct` - `<=`
* [ ] `overapproximation2D` - `overapproximate`
* [ ] `box_approximation` - `box_approximation`
* [ ] `radiusdiameter` - `box_approximation`
* [ ] `decompose` - `decompose`

PR #183, opened 2018-01-23
ExponentialMap parametric N

### LINKED ISSUE #178: Add numeric type parameter for ExponentialMap
See #132.

mforets:
Closes #178.

### DIFF
--- a/src/ExponentialMap.jl
+++ b/src/ExponentialMap.jl
@@ -281,7 +281,8 @@ function dim(eprojmap::ExponentialProjectionMap)::Int
 end
 
 """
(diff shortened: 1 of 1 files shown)


mforets:
notes: 

- i've removed the `<: Real` dependency because this is not really a restriction
- the unit test in different numeric types will be added in #132

schillic:
> i've removed the <: Real dependency because this is not really a restriction

This is a bit inconsistent now. We can also remove them from all other set types again if you want.

mforets:
i removed it from the lazy mexp constructors, since you could want complex (we could use `Number` ?) ,but i'm not proposing to remove it from all other set types.

schillic:
Okay, merge?

PR #184, opened 2018-01-23
Add polynomial zonotope

mforets:
cc: @nikos-kekatos


src/PolynomialZonotope.jl
```diff
@@ -0,0 +1,118 @@
...
+       dim,
+       polynomial_order,
+       order
+
+"""
+    AbstractNonConvexSet
+
+Abstract type for non-convex sets.
+"""
+abstract type AbstractNonConvexSet{N} end
```
schillic: Let us do this properly in a separate PR:
1. Move to a new file (and maybe rename to `AbstractSet` or even `Set` or `LazySet`).¹
2. Make `LazySet` inherit from this new type and rename to `AbstractConvexSet` or just `ConvexSet`.¹ This basically means changing every file, so better merge most PRs first.
3. Make `UnionSet` inherit from this new type and include it (there are already some unit tests).

Since this is a breaking change, we should probably do it *before* releasing 1.1.0.

¹This will be something we and the user will write a lot, so a short name would be preferable.

src/PolynomialZonotope.jl
```diff
@@ -0,0 +1,118 @@
...
+"""
+    PolynomialZonotope{N<:Real} <: AbstractNonConvexSet{N}
+
+Type that represents a polynomial zonotope.
+
+### Fields
+
+- `c`  -- starting point
+- `E`  -- multi-indexed generators such that *all* indices have the same value
+- `F`  -- multi-indexed generators such that *not all* indices are the same value
```
schillic: "are" → "have"

src/PolynomialZonotope.jl
```diff
@@ -0,0 +1,118 @@
...
+- `G`  -- single-indexed generators
+
+### Notes
+
+Polynomial zonotope were introduced by M. Althoff in *Reachability analysis of nonlinear
```
schillic: "zonotope" → "zonotopes"

src/PolynomialZonotope.jl
```diff
@@ -0,0 +1,118 @@
...
+
+"""
+    AbstractNonConvexSet
+
+Abstract type for non-convex sets.
+"""
+abstract type AbstractNonConvexSet{N} end
+
+"""
+    PolynomialZonotope{N<:Real} <: AbstractNonConvexSet{N}
```
schillic: Here you have `N<:Real`, but not in the actual struct definition.

src/PolynomialZonotope.jl
```diff
@@ -0,0 +1,118 @@
...
+"""
+    dim(pz::PolynomialZonotope)
+
+Return the ambient dimension of a polynomial zonotope.
+
+### Input
+
+- `pz` -- polynomial zonotope
+"""
+dim(pz::PolynomialZonotope) = length(pz.c)
```
schillic: Add output docs and output type?

src/PolynomialZonotope.jl
```diff
@@ -0,0 +1,118 @@
...
+### Input
+
+- `pz` -- polynomial zonotope
+
+## Output
+
+The polynomial order, defined as the maximal power of the scalar factors ``β_i``.
+Usually denoted ``η``.
+"""
+polynomial_order(pz::PolynomialZonotope) = length(pz.E)
```
schillic: Add output type?

src/PolynomialZonotope.jl
```diff
@@ -0,0 +1,118 @@
...
+
+Order of a polynomial zonotope.
+
+### Input
+
+- `pz` -- polynomial zonotope
+
+## Output
+
+The polynomial order, defined as the number of generators divided by the ambient
```
schillic: Drop "polynomial".

src/PolynomialZonotope.jl
```diff
@@ -0,0 +1,118 @@
...
+dimension.
+"""
+function order(pz::PolynomialZonotope)
```
schillic: Add output type?

src/PolynomialZonotope.jl
schillic: outsourced to #185

mforets: > Since this is a breaking change, we should probably do it before releasing 1.1.0.

2.0 iiuc since that would be an incompatible API change (https://semver.org/)

mforets:
To-do:

- [ ] `zonotope_overapproximation`
- [ ] `polytope_overapproximation`
- [ ] convex hull
- [ ] cartesian product
- [ ] linear map

src/PolynomialZonotope.jl
mforets: ok i've updated to `<: LazySet{N}`, what i should have done in the 1st place, at least 'till we resolve the approach for #185.

schillic: #185 was created for this issue (only later). You could not have done that before ;)

src/PolynomialZonotope.jl
```diff
@@ -0,0 +1,228 @@
...
+end
+
+"""
+    scale(α::Number, pz::PolynomialZonotope)
+
+Return a polynomial zonotope modified by a scalar factor.
+
+### Input
+
+- `α` -- polynomial zonotope
```
schillic: scalar factor?

src/PolynomialZonotope.jl
```diff
@@ -0,0 +1,228 @@
...
+- `pz` -- polynomial zonotope
+
+## Output
+
+Polynomial zonotope such that its center and generators are multiples of those
```
schillic: Above you called the center "starting point".

src/PolynomialZonotope.jl
```diff
@@ -0,0 +1,228 @@
...
+### Input
+
+- `pz` -- polynomial zonotope
+- `z`  -- usual zonotope
+
+## Output
+
+Polynomial zonotope.
+"""
+function minkowski_sum(pz::PolynomialZonotope, z::Zonotope)
```
schillic: Add the symmetric method? It can fall back to this one.

test/unit_PolynomialZonotope.jl
```diff
@@ -0,0 +1,17 @@
...
+G = diagm(N[0.3, 0.3]);
+p = PolynomialZonotope(c, E, F, G)
+
+@test dim(p) == 2
+@test order(p) == 7//2
+@test polynomial_order(p) == 2
+
+# type-specific concrete methods
+scale(N(0.5), p)
+linear_map(N[1.0 2.0; 2.0 5.0], p)
```
schillic: Add a unit test for `minkowski_sum`?

schillic:
I added "support vector" to the open tasks list above - do not forget it ;)

src/PolynomialZonotope.jl
mforets: [scale factor](https://en.wikipedia.org/wiki/Scale_factor) 👍

src/PolynomialZonotope.jl
mforets: good. i imagine some macro for lazy people (like us) used as:

```julia
@symmetric function minkowski_sum(pz::PolynomialZonotope, z::Zonotope)
    c = pz.c + z.center
    G = [G z.generators]
    return PolynomialZonotope(c, pz.E, pz.F, G)
end
```

and builds the new symmetric function using this one.

test/unit_PolynomialZonotope.jl
mforets: ok. i plan to add proper `@test`'s when with #230 gets in

PR #189, opened 2018-01-26
Vertices_list for HPolytope

### LINKED ISSUE #168: Implement vertices_list for HPolytope

mforets:
Closes #168.


src/HPolytope.jl
```diff
@@ -313,4 +313,30 @@ function cartesian_product(P1::HPolytope, P2::HPolytope; backend=CDDLib.CDDLibra
     return HPolytope(Pcp)
 end
 
+"""
+    vertices_list(P::HPolytope{N})::Vector{Vector{N}} where {N<:Real}
```
schillic: arguments missing (put in brackets)

PR #190, opened 2018-01-26
reorder docs common set reps

mforets:


mforets:
>  Alphabetic sorting?

yes, i think it looks better aesthetically.

> Actually they do not fit in this file.

i agree, i never knew where to put those docs :) also because support function and support vector make sense for operations and set types.

mforets:
actually isn't the place [here](https://juliareach.github.io/LazySets.jl/latest/lib/interfaces.html#LazySet-1)?

schillic:
This is also ok. But we should rename the file anyway!

PR #211, opened 2018-01-31
Zonotope order reduction (Girard's algorithm)

### LINKED ISSUE #195: Order reduction for Zonotopes

mforets:
Closes #195.

### DIFF
--- a/src/Zonotope.jl
+++ b/src/Zonotope.jl
@@ -338,7 +338,7 @@ Integer representing the number of generators.
 ngens(Z::Zonotope)::Int = size(Z.generators, 2)
 
 """
-    reduce_order(Z::Zonotope, r::Int)
+    reduce_order(Z::Zonotope, r::Int)::Zonotope
 
 Reduce the order of a zonotope by overapproximating with a zonotope with less
 generators.
@@ -357,30 +357,30 @@ A new zonotope with less generators, if possible.
 This function implements the algorithm described in A. Girard's
 *Reachability of Uncertain Linear Systems Using Zonotopes*, HSCC. Vol. 5. 2005.
 """
-function reduce_order(Z::Zonotope, r::Int)
+function reduce_order(Z::Zonotope{N}, r::Int)::Zonotope{N} where {N<:Real}
     c, G = Z.center, Z.generators
-    p, d = ngens(Z), dim(Z)
-    h = zeros(p)
-    for i in 1:p
-        h[i] = vecnorm(G[:, i], 1) - vecnorm(G[:, i], Inf)
-    end
-    ind = sortperm(h)
+    d = dim(Z)
 
     if r * d >= p
         # do not reduce
         return Z
     end
+    p = ngens(Z)
+    h = zeros(N, p)
+    for i in 1:p
+        h[i] = vecnorm(G[:, i], 1) - vecnorm(G[:, i], Inf)
+    end
+    ind = sortperm(h)
+
     m = p - floor(Int, d * (r - 1))
-    rg = G[:, ind[1:m]]
+    rg = G[:, ind[1:m]] # reduced generators
 
     # interval hull computation of reduced generators
-    d = sum(abs.(rg), 2)
     Gbox = diagm(sum(abs.(rg), 2)[:])
     if m > p
         Gred = [G[:, ind[m+1]:end] Gbox]
     else
         Gred = Gbox
     end
-    Zred = Zonotope(c, Gred)
-    return Zred
+    return Zonotope(c, Gred)
 end
--- a/src/Zonotope.jl
+++ b/src/Zonotope.jl
@@ -4,7 +4,9 @@ export Zonotope,
        order,
        minkowski_sum,
        linear_map,
-       scale
+       scale,
+       ngens,
+       reduce_order
 
 """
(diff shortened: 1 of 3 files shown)


src/Zonotope.jl
```diff
@@ -319,3 +321,66 @@ function scale(α::Real, Z::Zonotope)
...
+### Output
+
+A new zonotope with less generators, if possible.
+
+### Algorithm
+
+This function implements the algorithm described in A. Girard's
+*Reachability of Uncertain Linear Systems Using Zonotopes*, HSCC. Vol. 5. 2005.
+"""
+function reduce_order(Z::Zonotope, r::Int)
```
schillic: output type?

src/Zonotope.jl
```diff
@@ -319,3 +321,66 @@ function scale(α::Real, Z::Zonotope)
...
+    c, G = Z.center, Z.generators
+    p, d = ngens(Z), dim(Z)
+    h = zeros(p)
```
schillic: `zeros(N, p)` (you need to define `N` first).

src/Zonotope.jl
```diff
@@ -319,3 +321,66 @@ function scale(α::Real, Z::Zonotope)
...
+    for i in 1:p
+        h[i] = vecnorm(G[:, i], 1) - vecnorm(G[:, i], Inf)
+    end
+    ind = sortperm(h)
+
+    if r * d >= p
```
schillic: Pull up as far as possible? No need to compute `h` in that case!

src/Zonotope.jl
```diff
@@ -319,3 +321,66 @@ function scale(α::Real, Z::Zonotope)
...
+        # do not reduce
+        return Z
+    end
+    m = p - floor(Int, d * (r - 1))
+    rg = G[:, ind[1:m]]
+
+    # interval hull computation of reduced generators
+    d = sum(abs.(rg), 2)
```
schillic: What does this line do? `d` is never used again...

src/Zonotope.jl
```diff
@@ -319,3 +321,66 @@ function scale(α::Real, Z::Zonotope)
...
+    Gbox = diagm(sum(abs.(rg), 2)[:])
+    if m > p
+        Gred = [G[:, ind[m+1]:end] Gbox]
+    else
+        Gred = Gbox
+    end
+    Zred = Zonotope(c, Gred)
+    return Zred
```
schillic: return immediately?

src/Zonotope.jl
schillic: I checked it: For `Rational` it returns `Float64`, but for `Float32` it returns `Float32`. So let us keep it as is (`Float32` is automatically promoted to `Float64`, I checked it).

src/Zonotope.jl
```diff
@@ -319,3 +321,66 @@ function scale(α::Real, Z::Zonotope)
...
+
+This function implements the algorithm described in A. Girard's
+*Reachability of Uncertain Linear Systems Using Zonotopes*, HSCC. Vol. 5. 2005.
+"""
+function reduce_order(Z::Zonotope, r::Int)
+    c, G = Z.center, Z.generators
+    p, d = ngens(Z), dim(Z)
+    h = zeros(p)
+    for i in 1:p
+        h[i] = vecnorm(G[:, i], 1) - vecnorm(G[:, i], Inf)
```
schillic: For vectors `norm` and `vecnorm` should be equivalent. Any reason why you used `vecnorm` here?

src/Zonotope.jl
mforets: the sum is used just below, to build Gbox

src/Zonotope.jl
mforets: not particular, no. i tried to use `vecnorm` column-wise in the matrix `G` with a broadcast but it didn't work, thus the for loop.

src/Zonotope.jl
schillic: Yes, but this line can be removed I think (or use `d` below instead).

src/Zonotope.jl
```diff
@@ -319,3 +321,66 @@ function scale(α::Real, Z::Zonotope)
...
+        return Z
+    end
+    m = p - floor(Int, d * (r - 1))
+    rg = G[:, ind[1:m]]
+
+    # interval hull computation of reduced generators
+    d = sum(abs.(rg), 2)
+    Gbox = diagm(sum(abs.(rg), 2)[:])
+    if m > p
+        Gred = [G[:, ind[m+1]:end] Gbox]
```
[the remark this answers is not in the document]
mforets: `ind[m+1]` points to a number (index), so this chooses from that index to the last one

src/Zonotope.jl
mforets: done

src/Zonotope.jl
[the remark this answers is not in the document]
mforets: but `ind` is a vector of length `p` thus the check is nonsense (we need `m+1 <= p`), updated.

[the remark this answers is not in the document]
schillic: equivalent to `m < p` for integers ;)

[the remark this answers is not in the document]
mforets: actually, `ind[m+1:end]` because these is the set of non reduced generators, well done.

PR #216, opened 2018-02-02
#63 - refactoring LazySet & an_element default implementation

### LINKED ISSUE #63: Add function that returns a point inside a set

schillic:
Closes #63.


mforets:
the functions `σ` and `dim` also make sense for non-convex. 

how about creating a new abstract function `isconvex`?

schillic:
Yes, I did not want to introduce the non-abstract type in this PR.

Issue #188, opened 2018-01-25
AbstractPolytope list of constraints


mforets:
proposing to extend:

```
Every concrete `AbstractPolytope` must define the following functions:
- `vertices_list(::AbstractPolytope{N})::Vector{Vector{N}}` -- return a list of
    all vertices
```

such that it also defines a list of constraints (`constraints_list`).

both for the vertices and the constraints case, it can be done with and without using iterators.

schillic:
* For `AbstractHPolygon` we already have this function.
* For `HPolytope` it is be trivial.
* For `VPolygon` it is also rather simple.
* Remaining: `VPolytope`, `AbstractHyperrectangle`, `Zonotope`, `Ball1`.

mforets:
I've updated the list with the issues for each set type that is missing `constraints_list`. Also adapted the title of this issue because it is only about `constraints_list` (no iterators yet).

schillic:
Before we close this one, let me add `constraints_list` to the interface, including a check that the function is implemented by all subtypes.

Issue #191, opened 2018-01-26
HPolygon and HPolytope


mforets:
It would be convenient to have a `HPolytope` constructor from a `HPolygon` (and conversely). Also, note that the their fields are `constraints_list` and `constraints` respectively; we should decide for one of these names..

schillic:
Is the motivation to 1. use the external library also for `HPolygon`s or 2. for the user to switch between them?
We could even introduce a `Polytope` function that creates an `HPolygon` instance for 2D sets and an `HPolytope` instance otherwise.

👍 for unifying the name; I would go for `constraints`.

PR #220, opened 2018-02-06
change type field constraints_list -> constraints

mforets:
This point was discussed in #191


schillic:
LGTM

schillic:
Does this close #188?

mforets:
> Does this close #188?

not really, but thx for the reminder -- we still need to add the iterators

PR #227, opened 2018-02-08
[WIP #191] HPolygon/HPolytope conversion

### LINKED ISSUE #191: HPolygon and HPolytope

schillic:
Closes #191.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -25,6 +25,7 @@ makedocs(
         "Set Interfaces" => "lib/interfaces.md",
         "Common Set Representations" => "lib/representations.md",
         "Common Set Operations" => "lib/operations.md",
+        "Conversion between set representations" => "lib/conversion.md",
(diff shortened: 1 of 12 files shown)


schillic:
@mforets: I went for a `convert` strategy. If you agree with this solution, I will add the documentation and also move the recent `Zonotope` to the new file.

mforets:
yes, i agree with overloading `convert`, as long as these conversions are straightforward and exact. if there are other complications or overapproximations i would put them on the target set's file (or in the Approximations module, e.g. overloading`overapproximate`)

schillic:
Where should we put the documentation? Create a new section `conversion.md`?

mforets:
ok, looks good!

Issue #162, opened 2018-01-20
Add macros to automatically create boilerplate code


schillic:
I want to try [metaprogramming in Julia](https://docs.julialang.org/en/stable/manual/metaprogramming/#Code-Generation-1) for repeated patterns that we have such as defining algebraic laws like neutral/absorbing elements for operations. I imagine something like:
```julia
@commutative_monoid(ConvexHull, EmptySet)
@commutative_monoid_absorbing(MinkowskiSum, ZeroSet, EmptySet)
```
These macros are supposed to create the three (resp. six) constructors for the different combinations.
I think this is easier to both read and write, less error-prone, and easier to adapt to changes.

The only downside I see is that we cannot write documentation for such constructors anymore. But we should probably just mention this in the type's documentation anyway.

schillic:
After merging #228 we still need to address [this](https://github.com/JuliaReach/LazySets.jl/pull/228#pullrequestreview-95613688) comment.

schillic:
What should be the behavior of the following functions?
```julia
1. MinkowskiSum(X::MinkowskiSum, Y::MinkowskiSum)
2. MinkowskiSum(X::MinkowskiSumArray, Y::LazySet)
3. MinkowskiSum(X::MinkowskiSumArray, Y::MinkowskiSumArray)
4. MinkowskiSumArray(X::MinkowskiSum, Y::MinkowskiSum)
5. MinkowskiSumArray(X::MinkowskiSumArray, Y::LazySet)
6. MinkowskiSumArray(X::MinkowskiSumArray, Y::MinkowskiSumArray)
```
Note that 1.-3. and 4.-6. are identical up to the function name. Should they have the same behavior?
Currently this is inconsistent: We usually modify the argument set, except for `CH(cha1::ConvexHullArray, cha2::ConvexHullArray) = ConvexHullArray(vcat(cha1.array, cha2.array))` which creates a new array. I think for efficiency reasons we should rather append (this is why we have these array types), but it can be confusing that the argument changes.

mforets:
i would suggest (i think this corresponds to what you are proposing) that `MinkowskiSumArray` plays well with `MinkowskiSum`, since the former is just another lower level implementation of the latter. however, i wouldn't instantiate `MinkowskiSumArray`'s with elements of type `MinkowskiSum`, because this is an optional optimization.

if we follow this line:

- no , 1 would not append or create array, thus behave differently that the implementation that uses arrays,`MinkowskiSumArray`
[...]

Issue #145, opened 2018-01-16
Move algebraic definitions from functions to constructors


schillic:
For `MinkowskiSum` and `CartesianProduct` we have defined neutral/absorbing elements only for the `+`/`*` functions, but not for the constructor. It would be more consistent to define them using new constructors so that `A + B === MinkowskiSum(A, B)` in all cases.

Note that one has to add type parameters to avoid ambiguities. We saw that for `ConvexHull` where we already have this (because there is no separate function symbol).

Issue #197, opened 2018-01-27
Add check method implementation to docs


mforets:
does it belong to `src/lib/utils.md`?

schillic:
If anywhere, yes. But I wondered if we should rather totally remove the code from the `src` folder and put it inside the `test` folder. Same for the other check method `check_method_ambiguity_binary`.

schillic:
We agreed to move the functions to the test folder.

mforets:
+1  . we should do some work for getting their docstring inside the rendered html docs. maybe we can just `include` the relevant test files in `src/lib/utils.md`

schillic:
Oh you still want to have them? Okay 👍

mforets:
if it's not too complicated i'd add them to the docs (they are not in the docs now)

schillic:
I tried and did not succeed. It seems to me that Documenter just searches for files in the given modules. We could move those test files to a new submodule `src/Test`.

schillic:
@mforets: Should we close this one (without the documentation)?

PR #228, opened 2018-02-08
#162 - Add macros to automatically create boilerplate code

schillic:
See #162.
Also resolves most of #145 (had to be done in one PR) (only `ConvexHull` missing).

### DIFF
--- a/docs/src/lib/utils.md
+++ b/docs/src/lib/utils.md
@@ -10,4 +10,5 @@ end
 ```@docs
 sign_cadlag
 @commutative_neutral
+@commutative_absorbing
 ```
--- a/docs/src/lib/utils.md
+++ b/docs/src/lib/utils.md
@@ -11,4 +11,5 @@ end
 sign_cadlag
 @commutative_neutral
 @commutative_absorbing
+@commutative_neutral_absorbing
 ```
--- a/docs/src/lib/utils.md
+++ b/docs/src/lib/utils.md
@@ -9,5 +9,6 @@ end
 
 ```@docs
 sign_cadlag
-@commutative_neutral_absorbing
+@neutral
+@absorbing
 ```
--- a/docs/src/lib/utils.md
+++ b/docs/src/lib/utils.md
@@ -9,4 +9,5 @@ end
 
 ```@docs
 sign_cadlag
+@commutative_neutral
 ```
--- a/docs/src/lib/utils.md
+++ b/docs/src/lib/utils.md
@@ -9,7 +9,5 @@ end
 
 ```@docs
 sign_cadlag
-@commutative_neutral
-@commutative_absorbing
 @commutative_neutral_absorbing
 ```
--- a/src/CartesianProduct.jl
+++ b/src/CartesianProduct.jl
@@ -50,7 +50,10 @@ CartesianProduct(Xarr::Vector{S}) where {S<:LazySet{N}} where {N<:Real} =
                                    CartesianProduct(Xarr[2:length(Xarr)])))
 
 # EmptySet is the absorbing element for CartesianProduct
-@commutative_absorbing(CartesianProduct, EmptySet)
+absorbing(::Type{CartesianProduct}) = EmptySet
+
+# create absorbing element functions
+@commutative_neutral_absorbing(CartesianProduct)
 
 """
 ```
@@ -245,8 +248,11 @@ function CartesianProductArray(cpa1::CartesianProductArray,
     return cpa1
 end
 
-# EmptySet is the absorbing element for CartesianProduct
-@commutative_absorbing(CartesianProductArray, EmptySet)
+# EmptySet is the absorbing element for CartesianProductArray
+absorbing(::Type{CartesianProductArray}) = EmptySet
+
+# create absorbing element functions
+@commutative_neutral_absorbing(CartesianProductArray)
 
 """
     array(cpa::CartesianProductArray{N, S})::Vector{S} where {N<:Real, S<:LazySet{N}}
--- a/src/CartesianProduct.jl
+++ b/src/CartesianProduct.jl
@@ -50,10 +50,7 @@ CartesianProduct(Xarr::Vector{S}) where {S<:LazySet{N}} where {N<:Real} =
                                    CartesianProduct(Xarr[2:length(Xarr)])))
 
 # EmptySet is the absorbing element for CartesianProduct
-absorbing(::Type{CartesianProduct}) = EmptySet
-
-# create absorbing element functions
-@commutative_neutral_absorbing(CartesianProduct)
+@absorbing(CartesianProduct, EmptySet)
 
 """
 ```
@@ -249,10 +246,7 @@ function CartesianProductArray(cpa1::CartesianProductArray,
 end
 
 # EmptySet is the absorbing element for CartesianProductArray
-absorbing(::Type{CartesianProductArray}) = EmptySet
-
-# create absorbing element functions
-@commutative_neutral_absorbing(CartesianProductArray)
+@absorbing(CartesianProductArray, EmptySet)
 
 """
     array(cpa::CartesianProductArray{N, S})::Vector{S} where {N<:Real, S<:LazySet{N}}
--- a/src/CartesianProduct.jl
+++ b/src/CartesianProduct.jl
@@ -20,6 +20,10 @@ The Cartesian product of three elements is obtained recursively.
 See also `CartesianProductArray` for an implementation of a Cartesian product of
 many sets without recursion, instead using an array.
 
+The `EmptySet` is the absorbing element for `CartesianProduct`.
+
+Constructors:
+
 - `CartesianProduct{N<:Real, S1<:LazySet{N}, S2<:LazySet{N}}(X1::S1, X2::S2)`
   -- default constructor
 
@@ -45,6 +49,9 @@ CartesianProduct(Xarr::Vector{S}) where {S<:LazySet{N}} where {N<:Real} =
                 : CartesianProduct(Xarr[1],
                                    CartesianProduct(Xarr[2:length(Xarr)])))
 
+# EmptySet is the absorbing element for CartesianProduct
+@commutative_absorbing(CartesianProduct, EmptySet)
+
 """
 ```
     *(X::LazySet, Y::LazySet)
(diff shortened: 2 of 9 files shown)


src/helper_functions.jl
```diff
@@ -1,6 +1,8 @@
 export sign_cadlag,
        check_method_implementation,
-       check_method_ambiguity_binary
+       check_method_ambiguity_binary,
+       @commutative_neutral,
```
mforets: i'm not sure that we want to export these names, since this is more of an internals of the package (same comment applies to the 2 `check_...` methods of above)

src/helper_functions.jl
```diff
@@ -228,3 +230,122 @@ function check_method_ambiguity_binary(op;
...
+- `SET`  -- set type
+- `NEUT` -- set type of the neutral element
+
+### Output
+
+Three function definitions.
+
+### Examples
+
+`@commutative_neutral(ConvexHull, EmptySet)` creates the following functions:
```
mforets: you can more synthetically write the equations, eg. "it creates the following functions for any set ``X``: ``CH(X, ∅) = CH( ∅, X) = X``  and ``CH(X, ∅) = X`` .

src/helper_functions.jl
```diff
@@ -228,3 +230,122 @@ function check_method_ambiguity_binary(op;
...
+- `NEUT` -- set type of the neutral element
+- `ABS` -- set type of the absorbing element
+
+### Output
+
+Eight function definitions.
+
+### Examples
+
+`@commutative_neutral_absorbing(MinkowskiSum, ZeroSet, EmptySet))` creates the
```
mforets: extra parentheses in `EmptySet`

src/CartesianProduct.jl
```diff
@@ -180,32 +150,34 @@ Type that represents the Cartesian product of a finite number of convex sets.
 
 ### Notes
 
-- `CartesianProductArray(array::Vector{<:LazySet})` -- default constructor
+The `EmptySet` is the absorbing element for `CartesianProductArray`.
 
-- `CartesianProductArray()` -- constructor for an empty Cartesian product
+Constructors:
```
mforets: while we are at this line, a general thought is that i find the illustration of the different constructors to be more friendly if it is given as a proper example, rather than a copy paste of the source code -- if one wants that level of detail, then it's better to just look the code, by clicking on "view source", no? (or use Julia's `@edit` macro). that is to say that for me we can delete this part and instead put some examples of the different uses (in another PR).

src/CartesianProduct.jl
```diff
@@ -242,52 +216,15 @@ from the left.
 
 The modified Cartesian product of a finite number of convex sets.
 """
-function *(S::LazySet, cpa::CartesianProductArray)::CartesianProductArray
-    push!(cpa.array, S)
-    return cpa
+function CartesianProductArray(S::LazySet,
```
mforets: ok, this change is to save the symbol `*` for the binary operator only, in other words these two that used to be the same are now different:

```julia
julia>  b = Ball2(zeros(2), 1.);

julia> c = CartesianProductArray([b, b])

julia> CartesianProductArray(b, c)
LazySets.CartesianProductArray{Float64,LazySets.Ball2{Float64}}(LazySets.Ball2{Float64}[LazySets.Ball2{Float64}([0.0, 0.0], 1.0), LazySets.Ball2{Float64}([0.0, 0.0], 1.0), LazySets.Ball2{Float64}([0.0, 0.0],
 1.0)])

julia> b * c
LazySets.CartesianProduct{Float64,LazySets.Ball2{Float64},LazySets.CartesianProductArray{Float64,LazySets.Ball2{Float64}}}(LazySets.Ball2{Float64}([0.0, 0.0], 1.0), LazySets.CartesianProductArray{Float64,Laz
ySets.Ball2{Float64}}(LazySets.Ball2{Float64}[LazySets.Ball2{Float64}([0.0, 0.0], 1.0), LazySets.Ball2{Float64}([0.0, 0.0], 1.0), LazySets.Ball2{Float64}([0.0, 0.0], 1.0)]))
```

i doubt a bit, but maybe it is more sensible.

src/helper_functions.jl
```diff
@@ -228,3 +230,122 @@ function check_method_ambiguity_binary(op;
...
+following functions:
+* `MinkowskiSum(X::LazySet{N}, ::ZeroSet{N}) where {N<:Real} = X`
+* `MinkowskiSum(::ZeroSet{N}, X::LazySet{N}) where {N<:Real} = X`
+* `MinkowskiSum(Y::ZeroSet{N}, ::ZeroSet{N}) where {N<:Real} = Y`
+* `MinkowskiSum(::LazySet{N}, Y::EmptySet{N}) where {N<:Real} = Y`
+* `MinkowskiSum(Y::EmptySet{N}, ::LazySet{N}) where {N<:Real} = Y`
+* `MinkowskiSum(::ZeroSet{N}, Y::EmptySet{N}) where {N<:Real} = Y`
+* `MinkowskiSum(Y::EmptySet{N}, ::ZeroSet{N}) where {N<:Real} = Y`
+"""
+macro commutative_neutral_absorbing(SET, NEUT, ABS)
```
mforets: i was wondering if it's possible to work it around such that the extra work done by `commutative_neutral_absorbing` is done on the function `commutative_absorbing`, *provided that* the neutral element of the given set has been defined. we can do this by adding a function `neutral(::set_type)::neutral_type` and a check `isdefined(neutral($SET))`.

src/helper_functions.jl
schillic: Good point.

But I think we need to export the check methods for usage in the unit tests (not tested). Somewhere I proposed that we remove them from the code base and move them to the unit tests. Any opinion?

src/helper_functions.jl
schillic: `∅` might be misleading, but I get your point.

src/CartesianProduct.jl
schillic: I added #232.

src/CartesianProduct.jl
schillic: Actually I wanted to discuss this.
Previously `*` and `CartesianProduct` were not identical. I found this inconsistent, because `*(S, T)` was said to be an alias for `CartesianProduct(S, T)`.
Now it is identical, but we get what you observed.

I propose to additionally add a similar macro relating `CartesianProduct` and `CartesianProductArray` (and similarly for the other two) such that we also get the old behavior.
If you agree, should I do it here or in a follow-up PR?

src/helper_functions.jl
schillic: Hm, we have three of these cases, one with only neutral, one with only absorbing, and one with both neutral and absorbing. So I do not see why `neutral` should play a different role.

But I like the general idea. I would rather add two functions `neutral` and `absorbing` as follows.
The default implementation is `neutral(set) = nothing` and for the respective types we override with `neutral(MinkowskiSum) = ZeroSet`. Then we can get rid of two of the three macros. What would be the name of the remaining macro then?

src/helper_functions.jl
mforets: even if it is the name is not exported, one can use the function by writing the module's name, as in `LazySets.check_method_implementation`, or it's used in many places do a `import LazySets.check_method_implementation`.

but i agree to move those check methods to the tests folder 👍

schillic: > one can use the function by writing the module's name

You are right, I forgot that.

> but i agree to move those check methods to the tests folder 

Good, I changed #197 accordingly.

src/CartesianProduct.jl
mforets: >>  follow-up PR

src/helper_functions.jl
mforets: >  Then we can get rid of two of the three macros. 

how is that? i had in mind to substitute the call to `commutative_neutral_absorbing` to the calls `commutative_neutral` and `commutative_absorbing`.  but one wants to do this such that the order of the calls doesn't matter :/

schillic: You just call this new function (call it `fun`) and internally it calls the `neutral` and `absorbing` functions to see what it should do. Does it make sense?

src/CartesianProduct.jl
schillic: Okay, I took a note in the related issue #162.

src/helper_functions.jl
schillic: I understood now that what you proposed was very similar, sorry for the confusion.

schillic:
Ready from my side.

src/helper_functions.jl
mforets: with only 1 macro, the name `commutative_neutral_absorbing` is not the most intuitive IMO since `EmptySet` has nothing to do with `_neutral_` of `CartesianProduct`, while we have to write:

```julia
# EmptySet is the absorbing element for CartesianProduct
absorbing(::Type{CartesianProduct}) = EmptySet

# create absorbing element functions
@commutative_neutral_absorbing(CartesianProduct)
```

idea:

```julia
# EmptySet is the absorbing element for CartesianProduct
@absorbing(CartesianProduct) = EmptySet
```
(or statement like: `Emptyset @absorbs CartesianProduct` or `@is_absorbing_for`)

that unrolls to `absorbing(::Type{CartesianProduct}) = EmptySet` and the function definitions macro. but for a case like M-sum one does:

```julia
@absorbing(MinkowskiSum) = EmptySet
@neutral(MinkowskiSum) = ZeroSet
```

but should also work:

```julia
@neutral(MinkowskiSum) = ZeroSet
@absorbing(MinkowskiSum) = EmptySet
```

schillic: I admit that this is a very nice solution from the reader's perspective. Note that we will not have documentation for those generated functions, but they are anyway not interesting. Let me see if it works out 👍

schillic: > `@absorbing(MinkowskiSum) = EmptySet`

I think you cannot do assignments in macros. So I will introduce `@absorbing(MinkowskiSum, EmptySet)`

schillic:
I updated again.

Side note: This macro stuff can be complicated at times. In the end I used a bit of copy-paste because I lost patience.

docs/src/lib/utils.md
```diff
@@ -9,4 +9,7 @@ end
 
 ```@docs
 sign_cadlag
+@neutral
+@absorbing
+@neutral_and_absorbing
```
mforets: is `@neutral_and_absorbing` a remainder?

schillic: oops, fixed

schillic:
Thanks for the help and reviews, I think it is much better now.

PR #233, opened 2018-02-10
#57 - Refactor LinearConstraints.jl

### LINKED ISSUE #57: Revisit LinearConstraints.jl

schillic:
Closes #57.


schillic:
Only thing missing: Where to put `intersection`?

PR #236, opened 2018-02-11
#162 #145 #165 - add macro to restore old array type behavior

### LINKED ISSUE #145: Move algebraic definitions from functions to constructors

### LINKED ISSUE #162: Add macros to automatically create boilerplate code

### LINKED ISSUE #165: Extend CHArray with a CH
In #103 we implemented the convex hull array. One would like that `CH(a, b)` where `a` is a `CHArray` and `b` another lazy set, extends the `.array` field of `a` with the set `b`.

schillic:
Closes #162 and closes #145 and closes #165.

### DIFF
--- a/docs/src/lib/utils.md
+++ b/docs/src/lib/utils.md
@@ -11,4 +11,13 @@ end
 sign_cadlag
 @neutral
 @absorbing
+@declare_array_version
+```
+
+## Helper functions for internal use only
+
+```@docs
+@neutral_absorbing
+@array_neutral
+@array_absorbing
 ```
--- a/src/CartesianProduct.jl
+++ b/src/CartesianProduct.jl
@@ -248,6 +248,9 @@ end
 # EmptySet is the absorbing element for CartesianProductArray
 @absorbing(CartesianProductArray, EmptySet)
 
+# add functions connecting CartesianProduct and CartesianProductArray
+@declare_array_version(CartesianProduct, CartesianProductArray)
+
 """
     array(cpa::CartesianProductArray{N, S})::Vector{S} where {N<:Real, S<:LazySet{N}}
 
--- a/src/CartesianProduct.jl
+++ b/src/CartesianProduct.jl
@@ -174,77 +174,6 @@ function CartesianProductArray(n::Int=0, N::Type=Float64)::CartesianProductArray
     return CartesianProductArray(arr)
 end
 
-"""
-```
-    CartesianProductArray(cpa::CartesianProductArray,
-                          S::LazySet)::CartesianProductArray
-```
-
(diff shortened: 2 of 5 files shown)


mforets:
i fail to see what is the logic for these blocks:

```julia
...
        # if the array set type has already been defined, create combinations
        if isdefined(:array_constructor) &&
                method_exists(array_constructor, (Type{$SET},))
            @eval(@array_neutral($(esc(SET)),
                                 $(esc(NEUT)),
                                 array_constructor($(esc(SET)))))
        elseif isdefined(:is_array_constructor) &&
                method_exists(is_array_constructor, (Type{$SET},))
            @eval(@array_neutral($(esc(SET)),
                                 $(esc(NEUT)),
                                 $(esc(SET))))
        end
```

why is there an `if` clause? what is the difference between `array_constructor` and `is_array_constructor`?

schillic:
This is needed so that you can use the macros `@neutral`, `@absorbing` and `@declare_array_version` in any order. The if clauses make sure that the functions are defined only once.

mforets:
> This is needed so that you can use the macros @neutral, @absorbing and @declare_array_version in any order. 

ok

> The if clauses make sure that the functions are defined only once.

i guess my confusion arises from the fact that both `array_constructor` and `is_array_constructor` are defined in the same place (`declare_array_version`). how is that we have two `@eval`'s? isn't it used only the first one? thanks

schillic:
Okay, I admit that it is overengineered :)

We have `@neutral` two times, once for the normal type and once for the array type.
```julia
@neutral(MinkowskiSum, ZeroSet)
@neutral(MinkowskiSumArray, ZeroSet)
```
Those calls to the macro should have a slightly different behavior. I use the two functions to detect which one.

PR #243, opened 2018-02-14
add overapproximate specialization

### LINKED ISSUE #229: Convex hull of two zonotopes
If `Z1` and `Z2` are zonotopes, their convex hull is a polytope but not a zonotope, in general. For instance, let:

```julia
using LazySets, Plots
Z1 = Zonotope(ones(2), [[1., 0.], [0., 1.], [1., 1.]])
Z2 = Zonotope(-ones(2), [[.5, 1.], [-.1, .9], [1., 4.]])
Y = ConvexHull(Z1, Z2)

plot([Z1, Z2], alpha=.1)
plot!(Y, 1e-2, alpha=.1)
```

screen shot 2018-02-09 at 22 02 00

The convex hull `Y` was computed above using a polygonal approximation. As discussed in [Reachability of Uncertain Linear Systems Using Zonotopes, A. Girard, HSCC 2005], the set `Y` can be overapproximated with a zonotope as follows:

```julia
center = (Z1.center+Z2.center)/2
generators = [(Z1.generators .+ Z2.generators) (Z1.center - Z2.center) (Z1.generators .- Z2.generators)]/2
Z3 = Zonotope(center, generators)
plot!(Z3, alpha=0.1)
```

screen shot 2018-02-09 at 22 02 05

It is also noted that `Z3` is not necessarily  the minimal enclosing zonotope, which is in general expensive in high dimensions and it is investigated in [Zonotopes as bounding volumes, L. J. Guibas et al, PRoc. of Symposium on Discrete Algorithms, pp 803-812].

---

We could define:

```julia
function overapproximate(S::ConvexHull{N, Zonotope{N}, Zonotope{N}}, ::Type{<:Zonotope})::Zonotope where {N<:Real}
    Z1, Z2 = S.X, S.Y
    center = (Z1.center+Z2.center)/2
    generators = [(Z1.generators .+ Z2.generators) (Z1.center - Z2.center) (Z1.generators .- Z2.generators)]/2
    return Zonotope(center, generators)
end
```

that implements this function. What do you think?

Another approach for the API is to introduce some sort of new name `convex_hull_overapproximation`.. but i think that it is better to exploit the existent ones (so long as we are coherent).

mforets:
Closes #229

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -35,7 +35,7 @@ include("to_N.jl")
 # =================================================================
 # Algorithms for approximation of convex sets using support vectors
 # =================================================================
-@time @testset "LazySets.Approximations.overapproximation" begin include("unit_overapproximatejl") end
+@time @testset "LazySets.Approximations.overapproximation" begin include("unit_overapproximate.jl") end
 @time @testset "LazySets.Approximations.box_approximation" begin include("unit_box_approximation.jl") end
 @time @testset "LazySets.Approximations.ballinf_approximation" begin include("unit_ballinf_approximation.jl") end
 @time @testset "LazySets.Approximations.radiusdiameter" begin include("unit_radiusdiameter.jl") end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -35,7 +35,7 @@ include("to_N.jl")
 # =================================================================
 # Algorithms for approximation of convex sets using support vectors
 # =================================================================
-@time @testset "LazySets.Approximations.overapproximation" begin include("unit_overapproximation2D.jl") end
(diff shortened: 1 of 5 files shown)


test/runtests.jl
```diff
@@ -35,7 +35,7 @@ include("to_N.jl")
 # =================================================================
 # Algorithms for approximation of convex sets using support vectors
 # =================================================================
-@time @testset "LazySets.Approximations.overapproximation" begin include("unit_overapproximation2D.jl") end
+@time @testset "LazySets.Approximations.overapproximation" begin include("unit_overapproximatejl") end
```
schillic: missing period

PR #245, opened 2018-02-14
[WIP #185 #240] Convex set interface

### LINKED ISSUE #185: Add abstract type for non-convex sets
To support non-convex sets, we need to do:

1. Add new type `AbstractNonConvexSet` (or maybe `AbstractSet` or even `Set` or `LazySet`).¹
2. Make `LazySet` inherit from this new type and rename to `AbstractConvexSet` or just `ConvexSet`.¹ This basically means changing every file.

We can then also add `UnionSet` (let it inherit from the non-convex set type) (there are already some unit tests, also include them).

¹This will be something we and the user will write a lot, so a short name would be preferable.

### LINKED ISSUE #240: Update `abstract type LazySet` docstring
because:

- we there is a substype of LazySet which is not necessarily convex (polynomial zonotope),
- i think that calling `ConvexHull` "convex sets" can be confusing at first

i think we could rework the docstring of `abstract type LazySet` in the lines of:

> Abstract type for set representations and binary and n-ary operations between sets. Most subtypes of `LazySet` correspond to convex sets, i.e., sets characterized by a (possibly infinite)
intersection of halfspaces, or equivalently, sets ``S`` such that for any two
elements ``x, y ∈ S`` and ``0 ≤ λ ≤ 1`` it holds that ``λ x + (1-λ) y ∈ S``. However, convexity is not a restriction for subtyping `LazySet`.

schillic:
Closes #185 and closes #240.

* [ ] review: Can we still compute the support vector as before for all types? in particular (those docs have not been changed yet):
  * [ ] `LinearMap`
  * [ ] `ExponentialMap`
  * [ ] `iterative_refinement`
* [x] remove "*convex*" from documentation
* [x] remove "*convex*" from manual

After merging:
* include `UnionSet` (let us just call it `Union`)
* change the description of the repository


schillic:
This PR is blocked until we resolve above-mentioned problems (discussion in #185).

schillic:
Plan changed:
* We keep this library clean of non-convex sets, but provide the `AbstractConvexSet` type and the `isconvex` function.
* We remove `UnionSet.jl`.
* We outsource `PolynomialZonotope.jl`.

PR #249, opened 2018-02-18
#126 - Optional tolerance argument for containment checks

### LINKED ISSUE #126: Optional tolerance argument for membership check
In #68 we added containment checks. For floating-point numbers this check might fail due to precision issues. I would like to have an optional argument `tolerance::N=zero(N)`.

schillic:
Closes #126.

* [x] default implementation
* [x] `AbstractHyperrectangle`
* [x] `AbstractHPolygon`

The specialized implementations are **not** correct in the sense that at the corners the tolerance is bigger than allowed. I create a polytope of the same shape where all facets are pushed to the outside. On the other hand, a "usual" bloating with a Euclidean ball would produce rounded corners instead.

Is there a simple way to express the "usual" bloating? I can live with the not-correct bloating, but we should find a way to make it consistent.

### DIFF
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -52,6 +52,7 @@ support_function
 
 ```@docs
 an_element(S::LazySet{Float64})
+∈(::AbstractVector{Float64}, ::LazySet{Float64}, ::Float64)
 ```
 
 ## Point symmetric set
--- a/src/AbstractHPolygon.jl
+++ b/src/AbstractHPolygon.jl
@@ -142,7 +142,9 @@ function an_element(P::AbstractHPolygon{N})::Vector{N} where {N<:Real}
 end
 
 """
-    ∈(x::AbstractVector{N}, P::AbstractHPolygon{N})::Bool where {N<:Real}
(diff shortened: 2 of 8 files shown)


mforets:
some comp geometry softwares that can be useful: [CGAL](https://www.cgal.org/) or [geometric tools](https://www.geometrictools.com/Source/Mathematics.html), and the notes in [GTEngine](https://www.geometrictools.com/Documentation/ArbitraryPrecision.pdf)

mforets:
these are somehow related thoughts:

1) I think we can use consistently achieve containment checks with tolerance using interval arithmetic. (for higher dimensions we would need to do https://github.com/JuliaReach/LazySets.jl/issues/302)

2) it is not currently the case in all sets because of `N<:AbstractFloat`, but in principle we can define sets with intervals that contain the real entries, such as   `BallInf([IntervalArithmetic.Interval(0.0)], IntervalArithmetic.Interval(1.0))`

schillic:
Thanks for the comment. This PR is really old, and I do not like it. Not sure what to do with it. Should we close but keep the branch?

Issue #136, opened 2018-01-11
Generalize decomposition into general block structures


schillic:
Currently we decompose into 2D blocks only.

Related: #135 and #242

schillic:
The PR should have resolved this issue.

PR #259, opened 2018-02-27
#136 - Generalize decomposition into general block structures

schillic:
See #136.

### DIFF
--- a/src/Approximations/decompositions.jl
+++ b/src/Approximations/decompositions.jl
@@ -70,12 +70,12 @@ function decompose(S::LazySet{N};
 end
 
 """
-    function project(S::LazySet{N},
-                     block::AbstractVector{Int},
-                     set_type::Type{<:LazySet},
-                     [n]::Int=dim(S),
-                     [ɛ]::Real=Inf
-                    )::LazySet{N} where {N<:Real}
+    project(S::LazySet{N},
+            block::AbstractVector{Int},
+            set_type::Type{<:LazySet},
+            [n]::Int=dim(S),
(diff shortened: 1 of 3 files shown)


schillic:
While I was doing this, I realized: We could implement *all* project-approximate methods as follows:
```julia
M = sparse([1, 2], [block[1], block[2]], [one(N), one(N)], 2, n)
return overapproximate(M * S, ɛ)
```
This is the implementation for epsilon-close polygon approximation, but it should also work for other types.

Advantage: No extra code in `decompositions.jl` for new set types is needed. All the stuff happens in `overapproximate.jl`.

src/Approximations/decompositions.jl
```diff
@@ -1,145 +1,176 @@
...
-    pn = σ(sparsevec([2*bi], [one(N)], n), S)
-    pw = σ(sparsevec([2*bi-1], [-one(N)], n), S)
-    ps = σ(sparsevec([2*bi], [-one(N)], n), S)
-    return (pe, pn, pw, ps)
-end
 
 """
-    decompose(S::LazySet{N}, ɛi::Vector{<:Real})::CartesianProductArray where {N<:Real}
+    function project(S::LazySet{N},
+                     block::AbstractVector{Int},
```
mforets: call it `blocks`? (same below)

mforets: ok i see, this is a block of size 2.

mforets:
-> add new functions to Approximations docs

mforets:
test pass and docs build here.

PR #261, opened 2018-02-28
#260 - AssertionError in Iterative refinement

### LINKED ISSUE #260: AssertionError in Iterative refinement
`AssertionError: R.refinable`

In lines 71-73 we do:
```julia
approx_error = min(norm(q - σ(ndir, S)), dot(ndir, q - p1))
refinable = (approx_error > TOL(N)) && !(norm(p1-q, 2) < TOL(N) || norm(q-p2, 2) < TOL(N))
ap = LocalApproximation{N}(p1, d1, p2, d2, q, refinable, approx_error)
```
In line 185 we check:
```julia
if Ω.approx_list[i].err <= ɛ
```
Now it can happen that `refinable == false`, although `Ω.approx_list[i].err`, which is `approx_error` above, is `<= ɛ`.
Instead, we should check:
```julia
approx = Ω.approx_list[i]
if !approx.refinable || approx.err <= ɛ
```

@mforets: Do you agree?

schillic:
Closes #260.

### DIFF
--- a/src/Approximations/iterative_refinement.jl
+++ b/src/Approximations/iterative_refinement.jl
@@ -47,7 +47,8 @@ PolygonalOverapproximation{N}(S::LazySet) where {N<:Real} =
     PolygonalOverapproximation(S::LazySet, LocalApproximation{N}[])
 
 """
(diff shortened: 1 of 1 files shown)


schillic:
I did a mistake here and fixed it now. Unit tests pass on my end.

PR #268, opened 2018-03-01
#250 - Inplace array n-ary constructors

### LINKED ISSUE #250: Inplace array n-ary constructors
Example: to introduce `ConvexHull!(a, b)` such that if `a` is a set and `b` is a `ConvexHullArray`, it modifies in-place the array of `b`.

Similarly, introduce a `MinkowskiSum!` as a shortcut of the lazy operation (`+`) followed by modifying the array.

schillic:
Closes #250.

EDIT: We *do not* need to adapt `Reachability.jl` after merging.

### DIFF
--- a/src/CartesianProduct.jl
+++ b/src/CartesianProduct.jl
@@ -2,6 +2,7 @@ import Base:*, ×, ∈
 
 export CartesianProduct,
        CartesianProductArray,
+       CartesianProduct!,
        array
 
 """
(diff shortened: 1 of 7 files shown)


schillic:
Note that I changed the following:
`CH(::CHA, ::LazySet)` returned a new `CHA` instance. Was this a bug or was this intended?

mforets:
> Was this a bug or was this intended?

i agree to be strict such that the mutation is reserved to `!` operator. 👍

PR #271, opened 2018-03-01
#242 - revise overapproximate

### LINKED ISSUE #242: Revise `overapproximate`
- if `X` is any `LazySet`, we can do: `overapproximate(Y)` that calls `overapproximate(S, HPolygon)` and this is in turn a box overapproximation

- there is `overapproximate(Y, 1e-3)` that overapproximates with a polygon for the given tolerance

- we started to add specialized functions `overapproximate(S, set_type)` (eg. convex hull of two zonotopes)

proposal: add `overapproximate(Y, HPolygon, error_tolerance)`

schillic:
Closes #242.


mforets:
possible change: to make `overapproximate(S)` default to an `Hyperrectangle` return type.

PR #283, opened 2018-03-06
Interval related doctest fixes

mforets:
some fixes and proposal to modify the `Interval` types signature having `N<:Real`.

### DIFF
--- a/src/Interval.jl
+++ b/src/Interval.jl
@@ -12,7 +12,7 @@ export Interval, IA,
        low, high, vertices_list
 
 """
-    Interval{N, IN <: IA.AbstractInterval{N}} <: LazySet{N}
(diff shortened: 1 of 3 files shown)


src/Interval.jl
```diff
@@ -26,38 +26,39 @@ Unidimensional intervals are symbolic representations of a real closed interval.
...
 
 ```jldoctest interval_constructor
-julia> x = LazySets.Interval(0.0, 1.0)
+julia> x = LazySets.Interval([0.0, 1.0])
 ```
-or from a 2-vector:
+
+Note that if the package `IntervalArithmetic` is loaded in the current scope,
+have to prepend the `LazySets` to the interval type, since there is
```
schillic: we/you

mforets:
thanks for the review.

PR #294, opened 2018-03-13
#293 - Revise output type annotations

### LINKED ISSUE #293: Investigate possible slowdowns
Follow-up of #290.
We realized that
`function σ(d::AbstractVector{N}, lm::LinearMap)::AbstractVector{N} where {N}`
is a lot slower than
`function σ(d::AbstractVector{N}, lm::LinearMap)::Vector{N} where {N}`
or even
`function σ(d::V, lm::LinearMap)::V where {N<:Real, V<:AbstractVector{N}}`.

The task here is to investigate other places with output type `AbstractVector` and see if the same trick applies. Find them with `grep -lr ")::AbstractVector" src/`.

It is helpful to have a small benchmarking script as we had in [#290](https://github.com/JuliaReach/LazySets.jl/issues/290#issuecomment-372541377).

mforets:
Closes #293

### DIFF
--- a/src/Ball1.jl
+++ b/src/Ball1.jl
@@ -121,7 +121,7 @@ Return the support vector of a ball in the 1-norm in a given direction.
 
 Support vector in the given direction.
 """
-function σ(d::AbstractVector{N}, B::Ball1{N})::AbstractVector{N} where {N<:Real}
+function σ(d::V, B::Ball1{N})::V where {N<:Real, V<:AbstractVector{N}}
     res = copy(B.center)
     imax = indmax(abs.(d))
     res[imax] += sign(d[imax]) * B.radius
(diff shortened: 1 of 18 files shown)


src/Ball1.jl
```diff
@@ -121,7 +121,7 @@ Return the support vector of a ball in the 1-norm in a given direction.
 
 Support vector in the given direction.
 """
-function σ(d::AbstractVector{N}, B::Ball1{N})::AbstractVector{N} where {N<:Real}
```
schillic: The docs have not been updated.

schillic:
This branch actually made some of the benchmarks slightly slower:
* Heat: 12 vs. 9.15 seconds
* ISS: 2.47 vs. 1.94 seconds
* Beam: 33 vs. 28 seconds
* MNA1: 100 vs. 90 seconds
* FOM: 7.45 vs. 6.64 seconds

mforets:
I'll remove the type annotations then.

PR #297, opened 2018-03-14
#292 - Remove zero matrix check from LinearMap

### LINKED ISSUE #292: Remove zero matrix check from LinearMap
Outsourced from #290.
In `LinearMap.jl` we do this:
```julia
function *(M::AbstractMatrix, X::LazySet)
    @assert dim(X) == size(M, 2) "..."
    if findfirst(M) != 0
        return LinearMap(M, X)
    else
        return ZeroSet(dim(X))
    end
end
```
We should not always perform this check.

Idea: We could have some macro which goes one step further and sees that if the matrix is null obviously the outcome is a `ZeroSet`, e.g., `@simplify M * B` or even `@concrete` if we cannot do anything else with concrete `linear_map`.

mforets:
Closes #292.

### DIFF
--- a/src/LinearMap.jl
+++ b/src/LinearMap.jl
@@ -51,7 +51,7 @@ Return the linear map of a convex set.
 
 ### Output
 
-If the matrix is null, a `ZeroSet` is returned; otherwise a lazy linear map.
+A lazy linear map, i.e. a `LinearMap` instance.
 """
 *(M::AbstractMatrix, X::LazySet) = LinearMap(M, X)
 
--- a/src/LinearMap.jl
+++ b/src/LinearMap.jl
@@ -22,11 +22,20 @@ type `N`, where `N` is a floating point number type such as `Float64`.
 struct LinearMap{NM, N} <: LazySet{N}
     M::AbstractMatrix{NM}
     X::LazySet{N}
+
+    # default constructor with dimension match check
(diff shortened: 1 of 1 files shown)


src/LinearMap.jl
```diff
@@ -22,12 +22,21 @@ type `N`, where `N` is a floating point number type such as `Float64`.
 struct LinearMap{NM, N} <: LazySet{N}
     M::AbstractMatrix{NM}
     X::LazySet{N}
+
+    # default constructor with dimension match check
+    function LinearMap{NM, N}(M::AbstractMatrix{NM}, X::LazySet{N}) where {NM, N}
+        @assert dim(X) == size(M, 2) "a linear map of size $(size(M)) cannot be applied to a set of dimension $(dim(X))"
+        return new(M, X)
```
schillic: Do you not need `new{NM, N}(M, X)`?

src/LinearMap.jl
```diff
@@ -44,14 +53,8 @@ Return the linear map of a convex set.
 
 If the matrix is null, a `ZeroSet` is returned; otherwise a lazy linear map.
```
schillic: update docs

src/LinearMap.jl
schillic: Apparently not :hushed:

src/LinearMap.jl
mforets: oops, fixed.

PR #307, opened 2018-03-29
#306 - Revise `SparseMatrixExp`

### LINKED ISSUE #306: warn in `SparseMatrixExp` constructor that the type is sparse
the lazy matrix exponential is only defined for sparse matrices; we could have a better error message than:

```julia
julia> A = SparseMatrixExp(rand(2, 2))
ERROR: MethodError: Cannot `convert` an object of type Array{Float64,2} to an object of type LazySets.SparseMatrixExp
This may have arisen from a call to the constructor LazySets.SparseMatrixExp(...),
Stacktrace:
 [1] LazySets.SparseMatrixExp(::Array{Float64,2}) at ./sysimg.jl:77
```

something along the lines:

```julia
SparseMatrixExp(M::Matrix{N}) where {N} =
        error("only sparse matrices can be used to create a `SparseMatrixExp`")
```

an example in the docstring would also help.

mforets:
Closes #306.

- [x]  extend SparseMatrixExp examples
- [x] add constructor error for dense
- [x] add an example to `ExponentialMap`

### DIFF
--- a/src/ExponentialMap.jl
+++ b/src/ExponentialMap.jl
@@ -23,7 +23,7 @@ Type that represents the matrix exponential, ``\\exp(M)``, of a sparse matrix.
 
 ### Examples
 
-Take for exammple a random sparse matrix;
+Take for exammple a random sparse matrix:
 
 ```jldoctest SparseMatrixExp_constructor
 julia> A = sprandn(100, 100, 0.1);
@@ -45,7 +45,7 @@ julia> get_column(E, 10); # compute E[:, 10]
 
 julia> get_rows(E, [10]); # same as get_row(E, 10) but a 1x100 matrix is returned
 
-julia> get_columns(E, [10]); # same as get_row(E, 10) but a 100x1 matrix is returned
+julia> get_columns(E, [10]); # same as get_column(E, 10) but a 100x1 matrix is returned
 ```
 
 ### Notes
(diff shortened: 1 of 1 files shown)


src/ExponentialMap.jl
```diff
@@ -21,6 +21,33 @@ Type that represents the matrix exponential, ``\\exp(M)``, of a sparse matrix.
 
 - `M` -- sparse matrix
 
+### Examples
+
+Take for exammple a random sparse matrix;
```
schillic: Should this be a colon (`:`)?

src/ExponentialMap.jl
```diff
@@ -21,6 +21,33 @@ Type that represents the matrix exponential, ``\\exp(M)``, of a sparse matrix.
...
+they return row and column vectors (or matrices). For example:
+
+```jldoctest SparseMatrixExp_constructor
+julia> get_row(E, 10); # compute E[10, :]
+
+julia> get_column(E, 10); # compute E[:, 10]
+
+julia> get_rows(E, [10]); # same as get_row(E, 10) but a 1x100 matrix is returned
+
+julia> get_columns(E, [10]); # same as get_row(E, 10) but a 100x1 matrix is returned
```
schillic: "same as get_column"

mforets:
thanks for the review.

Issue #194, opened 2018-01-27
Resolve namespace conflicts with other packages


schillic:
There might be no solution here, because name conflicts are in general unavoidable.
We could think about not exporting some of the symbols, but that sounds stupid.

* `Polyhedra` warnings
```julia
julia> using LazySets, Polyhedra

# press TAB
julia> WARNING: both Polyhedra and LazySets export "HalfSpace"; uses of it in module Main must be qualified
WARNING: both Polyhedra and LazySets export "Interval"; uses of it in module Main must be qualified
WARNING: both Polyhedra and LazySets export "Line"; uses of it in module Main must be qualified
WARNING: both Polyhedra and LazySets export "dim"; uses of it in module Main must be qualified
```

* `IntervalArithmetic` warnings
Note that the warnings do not appear when only typing `using LazySets` (because `IntervalArithmetic` is only loaded internally). So I think we can ignore this.
```julia
julia> using LazySets, IntervalArithmetic

# press TAB
julia> WARNING: both LazySets and IntervalArithmetic export "Interval"; uses of it in module Main must be qualified
WARNING: both IntervalArithmetic and LazySets export "isdisjoint"; uses of it in module Main must be qualified
WARNING: both LazySets and IntervalArithmetic export "radius"; uses of it in module Main must be qualified
WARNING: both LazySets and IntervalArithmetic export "∅"; uses of it in module Main must be qualified
```

* `Plots` warning
```julia
julia> using LazySets, Plots

# press TAB
julia> WARNING: both Plots and LazySets export "center"; uses of it in module Main must be qualified
```

schillic:
This even makes the unit tests crash if `Polyhedra` is loaded first.

schillic:
Related but not that scary:
```julia
julia> using LazySets, Plots; center
WARNING: both Plots and LazySets export "center"; uses of it in module Main must be qualified
```

mforets:
i think we can close this?

schillic:
I restructured the first post to collect everything we have here.
I think we cannot solve the namespace conflicts, but we should at least make the unit tests robust against known problems.

PR #316, opened 2018-04-03
#300 - Correct usage of IntervalArithmetic

### LINKED ISSUE #300: Correct usage of IntervalArithmetic
Follow-up of #270.
We should use [`@interval`](https://juliaintervals.github.io/IntervalArithmetic.jl/latest/usage.html#Creating-intervals-1) instead of `IntervalArithmetic.Interval` to create intervals.

mforets:
Closes #300.

### DIFF
--- a/test/unit_decompose.jl
+++ b/test/unit_decompose.jl
@@ -1,4 +1,3 @@
-# using IntervalArithmetic
 import LazySets.Approximations.decompose
 
 for N in [Float64, Float32] # TODO Rational{Int}
@@ -42,8 +41,8 @@ for N in [Float64, Float32] # TODO Rational{Int}
     @test d.array[1] isa HPolygon && test_directions(d.array[1])
 
     if N == Float64
-    d = decompose(b, set_type=LazySets.Interval, blocks=ones(Int, 6))
-    @test d.array[1] isa LazySets.Interval &&
+    d = decompose(b, set_type=Interval, blocks=ones(Int, 6))
+    @test d.array[1] isa Interval &&
         σ(N[1], d.array[1])[1] == one(N) && σ(N[-1], d.array[1])[1] == -one(N)
     end
 
@@ -72,10 +71,8 @@ for N in [Float64, Float32] # TODO Rational{Int}
         @test ai isa Hyperrectangle
     end
     # 1D intervals
-    if N == Float64
-        d = decompose(b, set_type=LazySets.Interval)
-        for ai in array(d)
-            @test ai isa LazySets.Interval
-        end
+    d = decompose(b, set_type=Interval)
+    for ai in array(d)
+        @test ai isa Interval
     end
 end
--- a/test/unit_decompose.jl
+++ b/test/unit_decompose.jl
@@ -40,11 +40,9 @@ for N in [Float64, Float32] # TODO Rational{Int}
     d = decompose(b, set_type=HPolygon, ɛ=to_N(N, 1e-2))
     @test d.array[1] isa HPolygon && test_directions(d.array[1])
 
-    if N == Float64
     d = decompose(b, set_type=Interval, blocks=ones(Int, 6))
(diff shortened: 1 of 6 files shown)


mforets:
This section illustrates why before we didn't have problems in the tests and now i had to change them, since are using `@interval` instead of `IntervalArithmetic.Interval`.

```julia
julia> using IntervalArithmetic

julia> Interval(Float32(1.0), Float32(2.0))

[1, 2]

julia> typeof(ans)
IntervalArithmetic.Interval{Float32}

julia> @interval(Float32(1.0), Float32(2.0))
ERROR: MethodError: Cannot `convert` an object of type IntervalArithmetic.Interval{Float64} to an object of type Float32
This may have arisen from a call to the constructor Float32(...),
Stacktrace:
 [1] Float32(::IntervalArithmetic.Interval{Float64}) at ./sysimg.jl:77

julia> @interval(1.0, 2.0)
[1, 2]

julia> typeof(ans)
IntervalArithmetic.Interval{Float64}

julia> setprecision(Interval, precision(Float32))
24

julia> @interval(1.0, 2.0)
[1, 2]₂₄

julia> typeof(ans)
IntervalArithmetic.Interval{BigFloat}
```

test/unit_overapproximate.jl
```diff
@@ -90,10 +90,16 @@ for N in [Float64, Float32] # TODO Rational{Int}
...
 
     # Interval approximation
-    b = Ball1(N[0.], N(1.))
-    p = overapproximate(b, LazySets.Interval)
-    for d in to_N(N, [[1.], [-1.]])
-        @test σ(d, p)[1] ≈ σ(d, b)[1]
+    # for use with other precision, use
+    # setprecision(IntervalArithmetic.Interval, precision(N))
+    # however, this will use BigFloat's of given number of bits, which is a
+    # different numeric type than d, so this will triggeer a MethodError
```
schillic: "triggeer"

test/unit_decompose.jl
```diff
@@ -40,9 +40,12 @@ for N in [Float64, Float32] # TODO Rational{Int}
     @test d.array[1] isa Hyperrectangle && test_directions(d.array[1])
     d = decompose(b, set_type=HPolygon, ɛ=to_N(N, 1e-2))
     @test d.array[1] isa HPolygon && test_directions(d.array[1])
+
+    if N == Float64
     d = decompose(b, set_type=LazySets.Interval, blocks=ones(Int, 6))
```
schillic: indent?

schillic:
Actually, the idea of `@interval` is to create an interval from numerals (as in the tests). In the source code, it does not help because we already have floating point numbers. But let us keep it.

Does that mean that we cannot create `Interval`s of type `Rational{Int}` anymore? Can we provide constructors for different types at least? We should add a note to the docs of the `Interval` type.

:+1: for using `import`. This could (I did not try) have removed the warnings mentioned in #194.

mforets:
> 👍 for using import. This could (I did not try) have removed the warnings mentioned in #194.

yes, i think so.

---

> Does that mean that we cannot create Intervals of type Rational{Int} anymore?

Hmmm...

```julia
julia> Interval(0, 1)

julia> Interval(0//1, 2//1)
LazySets.Interval{Float64,IntervalArithmetic.Interval{Float64}}([0, 2])
```

This is not great. At least as you say we should provide constructors. I will ask `IntervalArithmetic` developers. (Since yesterday, there is `#intervals` in `https://julialang.slack.com/`).

test/unit_decompose.jl
```diff
@@ -42,8 +41,8 @@ for N in [Float64, Float32] # TODO Rational{Int}
     @test d.array[1] isa HPolygon && test_directions(d.array[1])
 
     if N == Float64
```
schillic: Can be removed again, right?

PR #318, opened 2018-04-04
[WIP #315] - Add cache to SparseMatrixExp

### LINKED ISSUE #315: Add cache to ExponentialMap and related types
In the `lazy_expm` mode of `Reachability.jl`, we compute parts of the matrix (some block rows) and then multiply the matrix with another lazy matrix (in fact the same as from the beginning). This is all repeated inside a loop.
Question: Can we cache the computed numbers (block rows) and use them later (matrix multiplication) instead of recomputing?

schillic:
Closes #315.

```julia
julia> A = spzeros(3,3); @time S = SparseMatrixExp(A);
  0.000010 seconds (19 allocations: 1.078 KiB)

julia> @time get_column(S, 1)
  0.000083 seconds (70 allocations: 5.609 KiB)
3-element Array{Float64,1}:
 1.0
 0.0
 0.0

julia> @time get_column(S, 1)
  0.000010 seconds (8 allocations: 496 bytes)
3-element Array{Float64,1}:
 1.0
 0.0
 0.0
```

### DIFF
--- a/src/ExponentialMap.jl
+++ b/src/ExponentialMap.jl
@@ -53,12 +53,24 @@ julia> get_columns(E, [10]); # same as get_column(E, 10) but a 100x1 matrix is r
 This type is provided for use with very large and very sparse matrices.
 The evaluation of the exponential matrix action over vectors relies on the
 [Expokit](https://github.com/acroy/Expokit.jl) package.
+
+Once computed, this type stores the computed results in a cache.
 """
 struct SparseMatrixExp{N} <: AbstractMatrix{N}
     M::SparseMatrixCSC{N, Int}
+    cache::SparseMatrixCSC{N, Int}
+    valid_rows::SparseVector{Bool, Int}
+    valid_cols::SparseVector{Bool, Int}
+
+    SparseMatrixExp{N}(M::SparseMatrixCSC{N, Int}) where {N} =
+        new{N}(M, spzeros(N, size(M, 1), size(M, 2)), spzeros(Bool, size(M, 1)),
+            spzeros(Bool, size(M, 2)))
 end
 
-SparseMatrixExp(M::Matrix) =
+# type-less convenience constructor
+SparseMatrixExp(M::SparseMatrixCSC{N, Int}) where {N} = SparseMatrixExp{N}(M)
+
+SparseMatrixExp(M::AbstractMatrix) =
         error("only sparse matrices can be used to create a `SparseMatrixExp`")
 
 Base.eye(spmexp::SparseMatrixExp) = SparseMatrixExp(spzeros(size(spmexp.M)...))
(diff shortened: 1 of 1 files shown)


schillic:
Note that if the cache matrix was fully computed using `get_row`/`get_rows`, the current implementation will recompute entries for `get_column`/`get_columns` because of how the validity check (two separate vectors for rows and columns) is implemented. So in the worst case we have to compute each entry twice. I think in practice it is very rare that we have computed the full matrix. See also [this comment](https://github.com/JuliaReach/LazySets.jl/issues/315#issuecomment-378516032).

schillic:
@mforets: Should we change other types (`ExponentialMap`, `ProjectionSparseMatrixExp`, `ExponentialProjectionMap`) as well?

schillic:
Runtime comparison (in seconds):

| Model | `master` | this branch |
|-|-|-|
| FOM | 4.36 | 2.81 |

schillic:
The plot for MNA5 looks very different, so there must be a bug somewhere.

src/ExponentialMap.jl
```diff
@@ -75,52 +87,78 @@ function size(spmexp::SparseMatrixExp, ax::Int)::Int
...
+            res[:, count] = column
+        end
         count += 1
     end
-    return ans
+    return res
 end
 
 function get_row(spmexp::SparseMatrixExp{N}, i::Int)::RowVector{N} where {N}
+    if spmexp.valid_rows[j]
```
mforets: `spmexp.valid_rows[i]` ?

src/ExponentialMap.jl
```diff
@@ -75,52 +87,78 @@ function size(spmexp::SparseMatrixExp, ax::Int)::Int
...
+        return spmexp.cache[i, :]
+    end
     n = size(spmexp, 1)
     aux = zeros(N, n)
     aux[i] = one(N)
-    return transpose(expmv(one(N), spmexp.M.', aux))
+    row = transpose(expmv(one(N), spmexp.M.', aux))
+    spmexp.cache[i, :] = row
+    spmexp.valid_rows[j] = true
```
mforets: same here? `spmexp.valid_rows[i] = true`

mforets:
> Runtime comparison (in seconds):

i don't quite understand the improvement, did you check if it is in the decomposition or in the iteration? in the latter `ϕpowerk_πbi` changes inside each loop in `k`, so i don't see how storing the values of a given `k` help.

schillic:
> i don't quite understand the improvement, did you check if it is in the decomposition or in the iteration? in the latter ϕpowerk_πbi changes inside each loop in k, i don't see how storing the values of a given k can help.

The decomposition time is the same.

You are right, here is the mistake:
```julia
ϕpowerk.M .= ϕpowerk.M + ϕ.M
```
We modify the matrix `ϕpowerk.M`, but we do not reset the validity of `ϕpowerk`'s entries. But if we reset, then we do not gain anything.
Long story short, in our application this approach does not help. Should we just close this PR then?

mforets:
> Should we just close this PR then?

agreed 👍

PR #341, opened 2018-05-17
#336 - Revise conversion from HPolytope to HPolygon

### LINKED ISSUE #336: Revise conversion from `HPolytope` to `AbstractHPolygon`
The current code

```julia
function convert(::Type{HPOLYGON},
                 P::HPolytope) where {HPOLYGON<:AbstractHPolygon}
    if dim(P) != 2
        error("polytope must be 2D for conversion")
    end
    return HPOLYGON(P.constraints)
end
```

Is not safe in the sense that it doesn't check that the constraints in the given `HPolytope` are sorted in counter-clockwise fashion. I think this should be at least the (default, true) option done here. It is easy eg.

```julia
function convert(::Type{HPOLYGON},
                 P::HPolytope) where {HPOLYGON<:AbstractHPolygon}
    if dim(P) != 2
        error("polytope must be 2D for conversion")
    end
    H = HPOLYGON()
    for ci in constraints_list(P)
        add_constraint!(H, p)
    end 
    return H
end
```

mforets:
Closes #336.

### DIFF
--- a/src/HPolygonOpt.jl
+++ b/src/HPolygonOpt.jl
@@ -44,8 +44,15 @@ HPolygonOpt(constraints::Vector{LinearConstraint{N}},
             ind::Int=1) where {N<:Real} =
     HPolygonOpt{N}(constraints, ind)
 
+# constructor for an HPolygon with no constraints
+HPolygonOpt{N}() where {N<:Real} = HPolygonOpt{N}(Vector{LinearConstraint{N}}(0), 1)
+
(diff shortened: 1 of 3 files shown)


schillic:
We also need to pass the numeric type explicitly to the constructor of `HPOLYGON` (take the one from `P`).

mforets:
Thanks!  (I also had to add the constructor for empty HPolygonOpt).

PR #354, opened 2018-05-27
#350 - Add forget_sets! function to CacheMinkowskiSum

### LINKED ISSUE #350: Light Minkowski sum array
#349 adds caching for the M-sum array, quoting from @schillic's comment:

> Possible future work: Currently the sum stores all sets, which takes a lot of memory. In applications where we always query the same support vector(s) and then add a new set in a loop, we could also provide a variant that only stores the newest sets.

schillic:
Closes  #350.

### DIFF
--- a/docs/src/lib/operations.md
+++ b/docs/src/lib/operations.md
@@ -105,6 +105,7 @@ CacheMinkowskiSum
 array(::CacheMinkowskiSum{Float64, LazySet{Float64}})
 dim(::CacheMinkowskiSum{Float64, LazySet{Float64}})
 σ(::AbstractVector{Float64}, ::CacheMinkowskiSum{Float64, LazySet{Float64}})
+forget_sets!(::CacheMinkowskiSum)
 ```
 
 ## Maps
--- a/src/MinkowskiSum.jl
+++ b/src/MinkowskiSum.jl
@@ -1,10 +1,11 @@
-import Base.+
+import Base:+, getindex
 
 export MinkowskiSum, ⊕,
        MinkowskiSumArray,
        MinkowskiSum!,
        array,
-       CacheMinkowskiSum
+       CacheMinkowskiSum,
+       forget_sets!
 
 """
     MinkowskiSum{N<:Real, S1<:LazySet{N}, S2<:LazySet{N}} <: LazySet{N}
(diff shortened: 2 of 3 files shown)


src/MinkowskiSum.jl
```diff
@@ -343,10 +368,65 @@ function σ(d::AbstractVector{<:Real}, cms::CacheMinkowskiSum)::Vector{<:Real}
...
+
+Tell a caching Minkowski sum to forget the stored sets.
+Only those sets are forgotten such that for each cached direction the support
+vector has been computed before.
+
+### Input
+
+- `cms` -- caching Minkowski sum
+
+# Output
```
mforets: (minor)  `### Output`

src/MinkowskiSum.jl
```diff
@@ -343,10 +368,65 @@ function σ(d::AbstractVector{<:Real}, cms::CacheMinkowskiSum)::Vector{<:Real}
...
     end
-    cache[d] = (l, svec)
+    cache[d] = CachedPair(l, svec)
     return svec
 end
 
+"""
+    forget_sets!(cms::CacheMinkowskiSum)::Int
+
+Tell a caching Minkowski sum to forget the stored sets.
```
mforets: ... without actually deleting the cache, if i understand correctly?

src/MinkowskiSum.jl
```diff
@@ -343,10 +368,65 @@ function σ(d::AbstractVector{<:Real}, cms::CacheMinkowskiSum)::Vector{<:Real}
...
+The number of sets that have been forgotten.
+
+### Notes
+
+This function should only be used under the assertion that no new directions are
+queried in the future; otherwise such support vector results will be incorrect.
+
+This implementation is optimistic and first tries to remove all sets.
+However, it verifies that for all cached directions the support vector has been
+computed before.
```
mforets: i don't quite follow this sentence. isn't it necessary to have computed a support vector, in order to cache a direction? is there a use case where there is a cached direction for which the support vector has *not* been computed?

docs/src/lib/operations.md
```diff
@@ -105,6 +105,7 @@ CacheMinkowskiSum
 array(::CacheMinkowskiSum{Float64, LazySet{Float64}})
 dim(::CacheMinkowskiSum{Float64, LazySet{Float64}})
 σ(::AbstractVector{Float64}, ::CacheMinkowskiSum{Float64, LazySet{Float64}})
+forget(::CacheMinkowskiSum)
```
mforets: i think this would be `forget_sets!`

src/MinkowskiSum.jl
schillic: Okay, okay, this is not clearly explained :smirk_cat:
You can add two sets to the sum, then compute the support vector, and then add another three sets to the sum. Now the cache only contains the vector for the first two sets.
The function `forget_sets!` wants to remove all five sets, but then detects that there is at least one direction which is only cached for the first two sets. Thus it only "forgets" the first two sets.

src/MinkowskiSum.jl
schillic: Yep, of course :grin:

src/MinkowskiSum.jl
mforets: Ah, got it now. Sounds good.

PR #358, opened 2018-05-29
Fix 0.7 deprecations for linear algebra

mforets:


src/Approximations/Approximations.jl
```diff
@@ -21,6 +21,17 @@ export approximate,
        BoxDiagDirections,
        OctDirections
 
+# see also compat.jl in LazySets module
```
schillic: Can you not say something like `import..compat.jl`?

src/compat.jl
```diff
@@ -3,6 +3,16 @@ This file defines internal functions for compatibility across
 different Julia versions.
 =#
 
+using Compat
+import Compat.String
+
+if VERSION < v"0.7-"
+    import Base.LinAlg:norm, checksquare
+else
+    using SparseArrays, LinearAlgebra
```
schillic: Out of curiosity: What is `SparseArrays` used for?

mforets: for example in julia 0.7 `sparse` or `sparsevec` needs to do `using SparseArrays`.

but i can avoid having to write these `using` each time the REPL is started with the startup script that should be in `.julia/config`: 

```
$ pwd
/Users/forets/.julia
$ cat config/startup.jl
using LinearAlgebra,
      Pkg,
      SparseArrays,
      Test,
      BenchmarkTools
```

schillic: No, it's fine!

src/Approximations/Approximations.jl
mforets: i'll check and merge afterwards if it works

mforets: done

PR #373, opened 2018-07-05
#371 -  Add concrete linear map for abstract hyperrectangles

### LINKED ISSUE #371: Add concrete linear map operations for abstract hyperrectangles
Each of these can have its own `linear_map(M, X)` implementation:

```julia
julia> subtypes(AbstractHyperrectangle)
4-element Array{Union{DataType, UnionAll},1}:
 LazySets.AbstractSingleton
 LazySets.BallInf
 LazySets.Hyperrectangle
 LazySets.SymmetricIntervalHull
```

mforets:
Closes #371.

### DIFF
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -76,6 +76,12 @@ Note that there is a special interface combination
 AbstractPolytope
 ```
 
+This interface defines the following functions:
+
+```@docs
+linear_map(M::AbstractMatrix, P::AbstractPolytope{Float64})
+```
+
 ### Polygon
 
 A polygon is a two-dimensional polytope.
@@ -121,5 +127,4 @@ This interface defines the following functions:
 
 ```@docs
 linear_map(M::AbstractMatrix, S::AbstractSingleton{Float64})
-linear_map(M::AbstractMatrix, H::AbstractHyperrectangle{Float64})
 ```
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -121,4 +121,5 @@ This interface defines the following functions:
 
 ```@docs
 linear_map(M::AbstractMatrix, S::AbstractSingleton)
+linear_map(M::AbstractMatrix, H::AbstractHyperrectangle{Float64})
 ```
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -116,3 +116,9 @@ A singleton is a special hyperrectangle consisting of only one point.
 ```@docs
 AbstractSingleton
 ```
+
+This interface defines the following functions:
+
+```@docs
+linear_map(M::AbstractMatrix, S::AbstractSingleton)
+```
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -120,6 +120,6 @@ AbstractSingleton
 This interface defines the following functions:
 
 ```@docs
-linear_map(M::AbstractMatrix, S::AbstractSingleton)
+linear_map(M::AbstractMatrix, S::AbstractSingleton{Float64})
 linear_map(M::AbstractMatrix, H::AbstractHyperrectangle{Float64})
 ```
--- a/src/AbstractHyperrectangle.jl
+++ b/src/AbstractHyperrectangle.jl
@@ -184,14 +184,14 @@ function ∈(x::AbstractVector{N},
 end
 
 """
-    linear_map(M::AbstractMatrix, H::AbstractHyperrectangle)
+    linear_map(M::AbstractMatrix, H::AbstractHyperrectangle{N}) where {N<:Real}
 
 Concrete linear map of an abstract hyperrectangle.
 
 ### Input
 
 - `M` -- matrix
-- `S` -- abstract hyperrectangle
+- `H` -- abstract hyperrectangle
 
 ### Output
 
@@ -199,7 +199,7 @@ The polytope in V-representation obtained by applying the linear map ``M`` to
 the set ``H``. If the given hyperrectangle is two-dimensional, a polygon instead
 of a general polytope is returned. 
 """
-function linear_map(M::AbstractMatrix, H::AbstractHyperrectangle{N}) where {N}
+function linear_map(M::AbstractMatrix, H::AbstractHyperrectangle{N}) where {N<:Real}
     if dim(H) == 2
         T = VPolygon
     else
--- a/src/AbstractHyperrectangle.jl
+++ b/src/AbstractHyperrectangle.jl
@@ -182,33 +182,3 @@ function ∈(x::AbstractVector{N},
     end
     return true
 end
-
-"""
-    linear_map(M::AbstractMatrix, H::AbstractHyperrectangle{N}) where {N<:Real}
-
-Concrete linear map of an abstract hyperrectangle.
-
-### Input
-
-- `M` -- matrix
-- `H` -- abstract hyperrectangle
-
-### Output
-
-The polytope in V-representation obtained by applying the linear map ``M`` to
-the set ``H``. If the given hyperrectangle is two-dimensional, a polygon instead
-of a general polytope is returned. 
-"""
-function linear_map(M::AbstractMatrix, H::AbstractHyperrectangle{N}) where {N<:Real}
-    if dim(H) == 2
-        T = VPolygon
-    else
-        T = VPolytope
-    end
-    vlist = vertices_list(H)
-    new_vlist = Vector{Vector{N}}(length(vlist))
-    @inbounds for vi in vlist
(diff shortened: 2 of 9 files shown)


mforets:
Hmm i think that it's slightly better to use the linearity in the formula for vertices list and apply the `M` inside . . .

mforets:
bad idea, forget it 😜

branch is ready from my side.

docs/src/lib/interfaces.md
```diff
@@ -116,3 +116,10 @@ A singleton is a special hyperrectangle consisting of only one point.
 ```@docs
 AbstractSingleton
 ```
+
+This interface defines the following functions:
+
+```@docs
+linear_map(M::AbstractMatrix, S::AbstractSingleton)
+linear_map(M::AbstractMatrix, H::AbstractHyperrectangle{Float64})
```
schillic: Is there a reason why there is a numeric type for `AbstractHyperrectangle` and not for `AbstractSingleton`?

src/AbstractHyperrectangle.jl
```diff
@@ -181,3 +182,32 @@ function ∈(x::AbstractVector{N},
...
+
+"""
+    linear_map(M::AbstractMatrix, H::AbstractHyperrectangle)
+
+Concrete linear map of an abstract hyperrectangle.
+
+### Input
+
+- `M` -- matrix
+- `S` -- abstract hyperrectangle
```
schillic: `H`

src/AbstractHyperrectangle.jl
```diff
@@ -181,3 +182,32 @@ function ∈(x::AbstractVector{N},
...
+        T = VPolygon
+    else
+        T = VPolytope
+    end
+    vlist = Vector{Vector{N}}()
+    for vi in vertices_list(H)
+        push!(vlist, M * vi)
+    end
+    return T(vlist)
+end
```
schillic: line break

src/AbstractHyperrectangle.jl
```diff
@@ -181,3 +182,32 @@ function ∈(x::AbstractVector{N},
...
+The polytope in V-representation obtained by applying the linear map ``M`` to
+the set ``H``. If the given hyperrectangle is two-dimensional, a polygon instead
+of a general polytope is returned. 
+"""
+function linear_map(M::AbstractMatrix, H::AbstractHyperrectangle{N}) where {N}
+    if dim(H) == 2
+        T = VPolygon
+    else
+        T = VPolytope
+    end
```
schillic: very nice solution :+1:

src/AbstractHyperrectangle.jl
```diff
@@ -181,3 +182,32 @@ function ∈(x::AbstractVector{N},
...
+    vlist = Vector{Vector{N}}()
```
schillic: For the moment it would be more efficient to use the size of `vertices_list` for initialization. But once we have the iterator (instead of the list) available, we may replace it anyway, so we can also just keep it.

src/AbstractSingleton.jl
```diff
@@ -109,6 +110,25 @@ function vertices_list(S::AbstractSingleton{N}
...
+- `S` -- abstract singleton
+
+### Output
+
+The abstract singleton of the same type of ``S`` obtained by applying the
+linear map to the element in ``S``.
+"""
+function linear_map(M::AbstractMatrix, S::AbstractSingleton)
+    T = typeof(S)
+    return T(M * element(S))
```
schillic: Where are the unit tests? :stuck_out_tongue_closed_eyes:

This will not work for `ZeroSet`. And if we ever implement a `OneSet` or whatever, it will also fail. We should at least have a custom implementation for `ZeroSet` for now. And maybe we should change `AbstractSingleton` to `Singleton` here (but I am not sure).
(We might run into this issue more often if we only work with abstract types. I did not see this issue before.)

schillic:
We should also consider an implementation with `HPolytope`, which should be more efficient (because the number of constraints is linear rather than exponential (number of vertices) for rectangles in >2D). But this could be a follow-up issue.
For this I would suggest to have an optional parameter flag and two separate implementations for when the flag is active or not.

docs/src/lib/interfaces.md
mforets: In the former case i needed the numeric type `N`, in

```julia
...
    if dim(H) == 2
        T = VPolygon
    else
        T = VPolytope
    end
...
```
while in the latter, i didn't use `N`.

mforets: it will compile to the same thing, but let's add the `N`.

on the other hand, annotating the output type is not obvious to me since it really depends on the concrete subtype of `AbstractSingleton`. hence, i would suggest not to write anything.

src/AbstractHyperrectangle.jl
mforets: thanks

docs/src/lib/interfaces.md
schillic: Ah, you are right, I was only looking at the docs here. You cannot remove the `N` in the code.
But maybe you can remove it for the docs (not sure what happens). If it works, dropping the type for the docs would be good because it avoids the warning.

src/AbstractHyperrectangle.jl
mforets: > For the moment it would be more efficient to use the size of vertices_list for initialization.

But `vertices_list` does the actual computation, ie. if i do `sizehint!(vlist, length(vertices_list(H))` it will already compute the list of vertices. Or you have sth else in mind?

src/AbstractHyperrectangle.jl
mforets: fixed. some problem with my new configuration of Atom, i'll be more careful.

src/AbstractHyperrectangle.jl
schillic: You call `vertices_list` below. Just call it above and cache this value.

mforets:
> We should also consider an implementation with HPolytope

hmm, something like $M A^{-1} b$ could work in some cases, but it needs $A$ to be invertible; i'm not sure that something better can be done with just the constraint representation, in general..

src/AbstractSingleton.jl
mforets: ah, good catch! then let's implement the custom implementation for `ZeroSet`, and leave this one generic as a "fallback"

src/ZeroSet.jl
```diff
@@ -126,3 +127,22 @@ function ∈(x::AbstractVector{N}, Z::ZeroSet{N})::Bool where {N<:Real}
...
+
+- `M` -- matrix
+- `Z` -- zero set
+
+### Output
+
+The zero set whose dimension matches the output dimension of the given matrix.
+"""
+function linear_map(M::AbstractMatrix, Z::ZeroSet{N}) where {N<:Real}
+    m, n = size(M)
```
schillic: You can say `m = size(M, 1)`.

schillic:
>>   We should also consider an implementation with HPolytope
>
> hmm, something like M A^{-1} b could work in some cases, but it needs A to be invertible; i'm not sure > that something better can be done with just the constraint representation, in general..

1. To be honest, I did not think about *how* to implement it (see 2. below).
Why is it not `M A b`? Maybe I am missing something.

2. I was very unclear. The idea was only for actual `Hyperrectangle`s. Just create the (axis-aligned) constraints, and rotate them. This should always work, right? My motivation was that the current implementation based on the vertices is very inefficient for high-dimensional `Hyperrectangle`s.

src/ZeroSet.jl
mforets: got it, but i'll change this line anyway in #382

mforets:
>  Why is it not M A b? Maybe I am missing something.

Actually if $y = Mx$ and if $M$ is invertible, then $x = inv(M) y$ and $Ax <= b$ transforms into $Ainv(M) y <= b$. Otherwise if i can do change $Ax <= b$ into $x <= inv(A) b$ then $y = Mx$ is $y <= Minv(A) b$ (these transformations make sense if i don't change the sign in the inequalities)

mforets:
> I was very unclear. The idea was only for actual Hyperrectangles. Just create the (axis-aligned) constraints, and rotate them. This should always work, right?

Yes, that's a different story, seems to me that this will work in general for linear maps of axis aligned hyperrectangles  👍

schillic:
> Actually if $y = Mx$ and if $M$ is invertible, then $x = inv(M) y$ and $Ax <= b$ transforms into $Ainv(M) y <= b$. Otherwise if i can change $Ax <= b$ into $x <= inv(A) b$ then $y = Mx$ is $y <= Minv(A) b$ (these transformations make sense if i don't change the sign in the inequalities..)

Got it, thanks for this detailed answer. Should we create a method where we try to invert one of the matrices, and if that fails, the method falls back to the current implementation (in a new issue)?

PR #374, opened 2018-07-05
#254 - Implement an_element function for LinearMap

### LINKED ISSUE #254: Implement an_element function for LinearMap
We have a default implementation for this function, but it uses a support vector query.
For a `LinearMap` this should be simple: multiply with the (EDIT: wrong: inverse; right: original) map and call `an_element` for the wrapped set. This can be more efficient.

mforets:
Closes #254.


src/LinearMap.jl
```diff
@@ -202,3 +203,21 @@ true
...
+- `lmap` -- linear map
+
+### Output
+
+An element in the linear map. It relies on the `an_element` function of the wrapped
+set.
+"""
+function an_element(lm::LinearMap)
+    return lm.M * an_element(lm.X)
+end
```
schillic: line break

test/unit_LinearMap.jl
```diff
@@ -52,10 +52,10 @@ for N in [Float64, Rational{Int}, Float32]
     @test lm1_copy.M == lm1.M
     @test lm1_copy.X == lm1.X
 
-    # an_element function (default implementation)
+    # an_element function
     lm = N(2.) * BallInf(N[0., 0.], N(1.))
     an_element(lm)
-#     @test an_element(lm) ∈ lm # TODO results in an error for Rational
```
schillic: :+1:

mforets:
Thanks for the review!

PR #376, opened 2018-07-08
#5 - Implement hrep from vrep for VPolygon

### LINKED ISSUE #5: Implement hrep from vrep for VPolygon
We have hrep -> vrep, but we are missing the converse.

schillic:
Closes #5.

* [x] docs
* [x] tests
* [x] manual handling of 2-vertices case (the current solution is unbounded)


schillic:
```julia
using LazySets
points_all = [[0.9,0.2], [0.4,0.6], [0.2,0.1], [0.1,0.3], [0.3,0.28]]
for i in [0, 1, 2, 5]
    points = i == 0 ? Vector{Vector{Float64}}() : points_all[1:i]
    V = VPolygon(points, apply_convex_hull=i > 0)
    H1 = tohrep(V)
    H2 = HPolygon{Float64}() # check that constraints are sorted correctly
    for c in H1.constraints
        addconstraint!(H2, c)
    end
    println(V)
    println(H1)
    println(H1.constraints == H2.constraints)
    println()
end
```
Output:
```julia
LazySets.VPolygon{Float64}(Array{Float64,1}[])
LazySets.HPolygon{Float64}(LazySets.HalfSpace{Float64}[])
true

LazySets.VPolygon{Float64}(Array{Float64,1}[[0.9, 0.2]])
LazySets.HPolygon{Float64}(LazySets.HalfSpace{Float64}[LazySets.HalfSpace{Float64}([1.0, 1.0], 1.1), LazySets.HalfSpace{Float64}([-1.0, 0.0], -0.9), LazySets.HalfSpace{Float64}([0.0, -1.0], -0.2)])
true

LazySets.VPolygon{Float64}(Array{Float64,1}[[0.9, 0.2], [0.4, 0.6]])
LazySets.HPolygon{Float64}(LazySets.HalfSpace{Float64}[LazySets.HalfSpace{Float64}([0.4, 0.5], 0.46), LazySets.HalfSpace{Float64}([-0.4, -0.5], -0.46)])
true

LazySets.VPolygon{Float64}(Array{Float64,1}[[0.1, 0.3], [0.2, 0.1], [0.9, 0.2], [0.4, 0.6]])
LazySets.HPolygon{Float64}(LazySets.HalfSpace{Float64}[LazySets.HalfSpace{Float64}([0.4, 0.5], 0.46), LazySets.HalfSpace{Float64}([-0.3, 0.3], 0.06), LazySets.HalfSpace{Float64}([-0.2, -0.1], -0.05), LazySets.HalfSpace{Float64}([0.1, -0.7], -0.05)])
true
```

schillic:
Before I add the tests, what do you think?

mforets:
> Before I add the tests, what do you think?

i didn't check the details yet, but i would suppose that we want to exploit that fact that a `VPolygon` has its vertices ordered. Hence we can go through the vertices in, say, counter clockwise-order and for each pair of vertices add the corresponding constraint.

schillic:
> i would suppose that we want to exploit that fact that a VPolygon has its vertices ordered

Yes, I do that. Even better, I start with the vertex pair that corresponds to the first constraint in the resulting `HPolygon`. This avoids additional sorting.

schillic:
Thanks for the documentation.

> we can equally choose `q` instead of `p`

No, that is why it is called `halfspace_left` :wink:

src/HalfSpace.jl
```diff
@@ -116,3 +116,33 @@ We just check if ``x`` satisfies ``a⋅x ≤ b``.
...
+### Algorithm
+
+The implementation is simple: the half-space ``a⋅x ≤ b`` is calculated as
+`a = [dy, -dx]`, where ``d = (dx, dy)`` denotes the line segment
+`pq`, that is, ``\\vec{d} = \\vec{p} - \\vec{q}``, and `b = dot(p, a)` because
+`p` is on this line (we can equally choose `q` instead of `p`). 
+"""
+function halfspace_left(p::AbstractVector{N},
+                        q::AbstractVector{N})::HalfSpace{N} where {N<:Real}
+    @assert p != q "the points must not be equal"
```
mforets: this notion of "left", i think, only makes sense in 2D, but `HalfSpace` is general so i think we should make this check here and throw an error otherwise. i can add it and see if you agree.

mforets:
> No, that is why it is called halfspace_left

My formulation was not the best, my point being that for choosing `b` (not `a`) one can pick either `p` or `q`, since `dot(p, a) = dot(q, a)` for the given `a`.

mforets:
That said, i think that it looks better without that part.

mforets:
I wonder how does this `tohrep` function compare to the (for me simpler):

```julia
constraints_list = vcat([halfspace_left(P.vertices_list[i+1], P.vertices_list[i]) for i in 1:length(P.vertices_list)-1],
                        [halfspace_left(P.vertices_list[end], P.vertices_list[1])])
return HPolygon(constraints_list)
```

src/HalfSpace.jl
schillic: 1. I do not remember if I just chose the name from the 2D view or if I did not think about the nD case. But now that I do, the function is general, right? So maybe one day we want to use it in nD. Should we just find a better name?
2. Do you think we should also add a version from a `LineSegment` (which is just the one-liner which unwraps the two vertices and calls this function here)?

src/HalfSpace.jl
```diff
@@ -137,11 +138,45 @@ describes the left-hand side of the directed line segment `pq`.
...
 The implementation is simple: the half-space ``a⋅x ≤ b`` is calculated as
 `a = [dy, -dx]`, where ``d = (dx, dy)`` denotes the line segment
-`pq`, that is, ``\\vec{d} = \\vec{p} - \\vec{q}``, and `b = dot(p, a)` because
-`p` is on this line (we can equally choose `q` instead of `p`). 
+`pq`, that is, ``\\vec{d} = \\vec{p} - \\vec{q}``, and `b = dot(p, a)`.
+
+### Examples
+
+The left half-space of the "east" and "west" directions in two-dimensions are the
+upper and lower half-spaces:
```
schillic: very good!

schillic:
> I wonder how does this tohrep function compare to the (for me simpler):
>
> ```julia
>constraints_list = vcat([halfspace_left(P.vertices_list[i], P.vertices_list[i+1]) for i in 1:length(P.vertices_list)-1],
>                        [halfspace_left(P.vertices_list[end], P.vertices_list[1])])
>return HPolygon(constraints_list)
>```

Good thinking, but you need to start at the index that was found before (`i` in my code). Basically it should do what you did but with three lists and different start index. Maybe like this (not tested):

```julia
vl = P.vertices_list
constraints_list = vcat(
    [halfspace_left(vl[j], vl[j+1]) for j in i:length(vl)-1],
    [halfspace_left(vl[end], vl[1])],
    [halfspace_left(vl[j], vl[j+1]) for j in 1:i-1])
return HPolygon(constraints_list)
```

My usual way of thinking is to write the code myself, but I prefer your version. Do you want to add it?

mforets:
> My usual way of thinking is to write the code myself, but I prefer your version. Do you want to add it?

ok. it took me a while to understand what was wrong with my version 😄

src/HalfSpace.jl
```diff
@@ -181,3 +181,29 @@ function halfspace_left(p::AbstractVector{N},
...
+describes the right-hand side of the directed line segment `pq`.
+
+### Algorithm
+
+See the documentation of `halfspace_left`. 
+"""
+function halfspace_right(p::AbstractVector{N},
+                         q::AbstractVector{N})::HalfSpace{N} where {N<:Real}
+    return halfspace_right(q, p)
+end
```
schillic: add a line

src/LineSegment.jl
```diff
@@ -144,3 +144,7 @@ function ∈(x::AbstractVector{N}, L::LineSegment{N})::Bool where {N<:Real}
     return min(L.p[1], L.q[1]) <= x[1] <= max(L.p[1], L.q[1]) &&
            min(L.p[2], L.q[2]) <= x[2] <= max(L.p[2], L.q[2])
 end
+
+halfspace_left(L::LineSegment) = halfspace_left(L.p, L.q)
+
+halfspace_right(L::LineSegment) = halfspace_right(L.p, L.q)
```
schillic: add a line

schillic:
The build fail with Julia v0.6 is the same as in #386. It could also be a recent bug in `IntervalArithmetic.jl` that only occurs with v6.0?

mforets:
> The build fail with Julia v0.6 is the same as in #386. It could also be a recent bug in IntervalArithmetic.jl that only occurs with v6.0?

Hmm in my machine the tests pass (v0.6.1) and 

```julia
julia> Pkg.installed("IntervalArithmetic")
v"0.13.0+"
```

mforets:
Hmm we can always pick the version that we use in `REQUIRE`, such as `v0.13`.

Sooner or later we should also switch to 0.7 .. 1.0

schillic:
Here (v0.6.2) they pass as well (I also updated the package). Note that the Travis builds for v0.6.1 and v0.6.1 pass; only v0.6 fails. The thing is that as soon as we merge, the green badge on the front page will become red :disappointed:

schillic:
Note to myself: I want to do [this](https://github.com/JuliaReach/LazySets.jl/pull/376#issuecomment-406894108) before merging. :warning: :stop_sign:

PR #391, opened 2018-07-23
#389 - Change Travis build versions

### LINKED ISSUE #389: Change Travis build versions
Result of a discussion with @mforets:
* We want to have build tests for the following Julia versions: v0.6, v0.6.2, v0.6.4, v0.7 (beta)
* We only want to allow failures for v0.7. Previously the builds randomly broke, but we hope that they are more stable now.

schillic:
Closes #389.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ language: julia
 os:
   - linux
 julia:
-  - 0.6
-  - 0.6.1
+  - 0.6.0
   - 0.6.2
+  - 0.6.4
   - nightly
 notifications:
   email: false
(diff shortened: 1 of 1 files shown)


schillic:
I saw that we send coverage data to `Coveralls` but never retrieve that data. Should we change this or add a badge on the main page?

mforets:
+1 
I think we are fine with `codecov` which already works

mforets:
actually, `- 0.6` installs the same as `- 0.6.4`

schillic:
Ah, this explains why it fell back to v0.6.4. Does `- 0.6.0` work?

mforets:
i didn't try, but i think it would work.

schillic:
Yep, it worked.

PR #401, opened 2018-07-24
#400 - Documenter error with `convert`

### LINKED ISSUE #400: Documenter error with `convert`
We need to add concrete types to `convert` in `conversion.md`. Otherwise `Documenter` tries to document *all* methods with the name `convert`.
As a side note, currently there is an error with one of them (which is why I found this issue).

schillic:
Closes #400.

### DIFF
--- a/docs/src/lib/conversion.md
+++ b/docs/src/lib/conversion.md
@@ -16,5 +16,9 @@ end
 ```
 
 ```@docs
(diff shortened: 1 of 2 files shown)


mforets:
seems good to me -- although i cannot yet build the docs successfully because of #400.

PR #408, opened 2018-07-26
#405 - Change intersection of two lines to set operation

### LINKED ISSUE #405: Change intersection of two lines to set operation
We have this concrete intersection function: [`intersection(L1::Line{N}, L2::Line{N})::Vector{N}`](https://github.com/JuliaReach/LazySets.jl/blob/master/src/concrete_intersection.jl#L34).
The return value (a vector) is inconsistent with intersection operations for other set types.
Proposal: Return a `Singleton` (resp. an `EmptySet`).

schillic:
Closes #405.

### DIFF
--- a/src/AbstractHPolygon.jl
+++ b/src/AbstractHPolygon.jl
@@ -96,11 +96,11 @@ function vertices_list(P::AbstractHPolygon{N},
         return points
     end
     @inbounds for i in 1:n-1
-        points[i] = intersection(Line(P.constraints[i]),
-                                 Line(P.constraints[i+1]))
+        points[i] = element(intersection(Line(P.constraints[i]),
+                                         Line(P.constraints[i+1])))
     end
-    points[n] = intersection(Line(P.constraints[n]),
-                             Line(P.constraints[1]))
+    points[n] = element(intersection(Line(P.constraints[n]),
+                                     Line(P.constraints[1])))
     return apply_convex_hull ? convex_hull(points) : points
 end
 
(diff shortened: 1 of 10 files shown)


schillic:
okay :+1:

src/AbstractHPolygon.jl
```diff
@@ -96,11 +96,11 @@ function vertices_list(P::AbstractHPolygon{N},
...
     @inbounds for i in 1:n-1
-        points[i] = intersection(Line(P.constraints[i]),
-                                 Line(P.constraints[i+1]))
+        points[i] = element(intersection(Line(P.constraints[i]),
+                                         Line(P.constraints[i+1])))
     end
-    points[n] = intersection(Line(P.constraints[n]),
-                             Line(P.constraints[1]))
+    points[n] = element(intersection(Line(P.constraints[n]),
+                                     Line(P.constraints[1])))
```
mforets: ... and (later) we can move one step further and add a `convex_hull` that deals with a list of singletons (points), thus not needing these `element(..)`. i think it would make the code a bit cleaner.

schillic: Hm, we could add an iterator type over Singletons that returns their vectors, say, `Singleton2Vector`, and pass `Singleton2Vector(points)` to `convex_hull`. However, if no convex hull shall be applied, we still need to unwrap them.

mforets:
... ouch, that was unintended.

mforets:
can be undone with `$ git reset --hard HEAD~1`

mforets:
i lost your newer commits, sorry 😱

schillic:
Hm? What did you do?

mforets:
`git push -f` (-f for forced), i did this inadvertently, since i really wanted to `pull` your branch.

schillic:
So this was only the fix of the doctests, right?

schillic:
Yep, I force-pushed again.

schillic:
Can we remove the "Changes requested" review without the "Dismiss review" button?

mforets:
in https://github.com/JuliaReach/LazySets.jl/pulls it now showsup as Approved.
i didnt click on dismiss Review. i think that i clicked on the Files Changed and then sent the (new) review.

PR #421, opened 2018-07-29
#420 - SymmetricIntervalHull's radius_hyperrectangle function is broken

### LINKED ISSUE #420: SymmetricIntervalHull's radius_hyperrectangle function is broken

schillic:
Closes #420.

### DIFF
--- a/src/SymmetricIntervalHull.jl
+++ b/src/SymmetricIntervalHull.jl
@@ -28,14 +28,10 @@ bounds in the `cache` field.
 """
 struct SymmetricIntervalHull{N<:Real, S<:LazySet{N}} <: AbstractHyperrectangle{N}
     X::S
-    cache::Vector{Tuple{N, Bool}}
+    cache::Vector{N}
 
     function SymmetricIntervalHull{N, S}(X::S) where {S<:LazySet{N}} where {N<:Real}
-        cache = Vector{Tuple{N, Bool}}(dim(X))
-        zero_N = zero(N)
-        for i in 1:length(cache)
-            cache[i] = (zero_N, false)
-        end
+        cache = fill(-one(N), dim(X))
         return new{N, S}(X, cache)
     end
 end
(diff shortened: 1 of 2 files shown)


src/SymmetricIntervalHull.jl
```diff
@@ -28,13 +28,13 @@ bounds in the `cache` field.
...
+    cache::Vector{N}
 
     function SymmetricIntervalHull{N, S}(X::S) where {S<:LazySet{N}} where {N<:Real}
-        cache = Vector{Tuple{N, Bool}}(dim(X))
-        zero_N = zero(N)
+        cache = Vector{N}(dim(X))
+        minus_one_N = -one(N)
         for i in 1:length(cache)
-            cache[i] = (zero_N, false)
+            cache[i] = minus_one_N
```
mforets: (the same but shorter: `fill!(cache, -one(N)`, or even `cache = fill(-one(N), dim(X))`.)

src/SymmetricIntervalHull.jl
```diff
@@ -106,14 +106,14 @@ This function computes the symmetric interval hull explicitly.
...
                               )::Vector{N} where {N<:Real}
-    dim = dim(sih)
+    n = dim(sih)
     one_N = one(N)
-    for i in 1:dim
-        if !sih.cache[i][2]
-            compute_radius(sih, i, dim, one_N)
+    for i in 1:n
+        if sih.cache[i] == -1
+            compute_radius(sih, i, n, one_N)
```
mforets: better if we call it `compute_radius!`, since it modifies `sih` in-place

mforets: i would suggest that `compute_radius!` also returns the computed result, if not only for testing/debugging purposes

```julia
    sih.cache[i] = max(right_bound[i], abs(left_bound[i]))
end
```

schillic: Good suggestions. I renamed `compute_radius!` to `get_radius!` and added the cache look-up inside this function, which made the code simpler.

Issue #402, opened 2018-07-24
Documenter crashes when building the docs


schillic:
(not related to #400)
```julia
ERROR: LoadError: MethodError: no method matching convert(::Type{AssertionError}, ::String)
Closest candidates are:
  convert(::Type{Any}, ::ANY) at essentials.jl:28
  convert(::Type{T}, ::T) where T at essentials.jl:29
Stacktrace:
 [1] AssertionError(::String) at ./coreimg.jl:14
 [2] walk(::Function, ::Dict{Symbol,Any}, ::Array{Any,1}) at .../.julia/v0.6/Documenter/src/Documents.jl:460
 ...
```

EDIT: The error can be reproduced as follows:
* Remove every file from the folder `docs/src`.
* In `docs/make.jl` remove every line of `pages = [ ... ]`.
* Add the file `docs/src/index.md` with the following content:
````julia
[here](there)

```@example
using Plots
```
````
* Remove every file from the `src` folder (source code) and add `LazySets.jl` with this content:
```julia
module LazySets
import IntervalArithmetic
end
```
* Run `julia docs/make.jl`.

Conclusion: The error is unrelated to us and seems to occur with the combination of hyperlinks and `@example` blocks with `Plots` and `IntervalArithmetic`...

mforets:
did you report the bug upstream?

schillic:
> did you report the bug upstream?

No. Maybe it is still related to our code. So I would have to create a new project and test there. I guess I just hoped that it gets fixed automatically :smiley:

mforets:
i get this error even commenting out all our pages... strange

```julia
ENV["GKSwstype"] = "100"  # set 'GR environment' to 'no output' (for Travis CI)
using Documenter, LazySets

makedocs(
    doctest = false,  # use this flag to skip doctests (saves time!)
    modules = [LazySets, Approximations],
    format = :html,
    assets = ["assets/juliareach.css"],
    sitename = "LazySets.jl",
    pages = [
        #"Home" => "index.md",
        #"Manual" => Any[
        #"Getting Started" => "man/getting_started.md",
        #"Polyhedral Approximations" => "man/polyhedral_approximations.md",
        #"Decomposing an Affine Map" => "man/decompose_example.md",
        #"Fast 2D LPs" => "man/fast_2d_LPs.md",
        #"Iterative Refinement" => "man/iterative_refinement.md",
        #"Interval Hulls" => "man/interval_hulls.md",
[...]

PR #432, opened 2018-07-30
Clean-up .travis.yml

schillic:
We added two packages that are already in our REQUIRE list, so they are already installed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,8 @@ matrix:
 #script:
 #  - julia -e 'Pkg.clone(pwd()); Pkg.build("LazySets"); Pkg.test("LazySets"; coverage=true)'
 after_success:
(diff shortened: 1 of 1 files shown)


mforets:
about adding GR there, i remember something like it was needed to render the plots in the automatically generated docs using GR (otherwise it was using other default). if you want we can try and see if anything changed.

schillic:
I guess we did not have it in REQUIRE at that time. Because what would `Pkg.add("GR")` do in addition to trying to install it again now? It is executed on the same machine.

mforets:
yeah, i don't know, let's try

schillic:
Where do you see this? The `gh-pages` branch has not been updated since several weeks (probably due to #402).

mforets:
ah, i was looking at the plots. how about if we write `Pkg.pin("Documenter", "0.18.0")` in `.travis.yml`?

schillic:
See #439.

Issue #72, opened 2017-12-09
Fix 'docstrings potentially missing' warnings


schillic:
Creating the documentation warns about potentially missing docstrings.
However, they are defined and occur properly in the final documentation.

Most notably:
* `LazySets.σ`
* `LazySets.apply_recipe`

For the non-`apply_recipe` functions the problem is most probably related to type parameters.

schillic:
I had a look at the `Documenter.jl` code [where this warning is printed](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/DocChecks.jl#L27).

Use case: `σ` for `BallInf`:
We have this function signature in the source code:
```julia
σ(d::AbstractVector{<:Real}, B::BallInf)
```
and this doc signature:
```julia
σ(::AbstractVector{Float64}, ::BallInf)
```
In `Documenter.jl` in the above-mentioned function this results in the following `object`:
```julia
object == LazySets.σ-Tuple{AbstractArray{Float64,1},LazySets.BallInf}
object.binding == LazySets.σ
object.signature == Tuple{AbstractArray{Float64,1},LazySets.BallInf}
```
So far so good. Now let us see what  signature is stored in `signatures`:
```julia
signatures == Set(Type[ ..., Tuple{AbstractArray{#s9,1} where #s9<:Real,LazySets.BallInf}, ...])
```
(The identifier `#s9` changes with each run.)

`Documenter.jl` cannot match this. So let us change the doc signature the way it seems to be expected.
```julia
σ(::AbstractVector{N} where N<:Real, ::BallInf)
```
This still does not help. It seems that `Documenter.jl` does not understand `where` phrases.
However, the following works:
```julia
σ(::AbstractVector{<:Real}, ::BallInf)
```
Partial success, this will resolve many warnings! 🎉

schillic:
Still open: How can we resolve the warnings for shared type parameters such as in `HPolygon.jl`?
```julia
addconstraint!(P::HPolygon{N}, constraint::LinearConstraint{N}) where {N<:Real}
```
Current docs signature:
```julia
addconstraint!(::HPolygon{Float64}, ::LinearConstraint{Float64})
```
The following does *not* work because it is not even found in an earlier phase of `Documenter.jl`.
```julia
addconstraint!(::HPolygon{<:Real}, ::LinearConstraint{<:Real})
```

schillic:
Another comment:
The reason why we do not see this warning for some other functions such as
```julia
[...]

PR #436, opened 2018-07-30
#159, #72 - Restructuring of docs

### LINKED ISSUE #159: Remove duplicate inherited functions from docs
Only define function docs at the interface level.

schillic:
Closes #159. Workaround for #72.

* [x] Add missing documentation.
* [x] Add workaround for #72. The only warnings that are still printed are the ones for functions only defined in presence of `Polyhedra.jl`. Should we remove them as well?
* [x] Have interface documentation at one place and link to it (#159).
  * There are still some problems with `Documenter` in case of multiple methods. As a workaround, I omit the hyperlinks when there are conflicts.
  * To elaborate:
    * default values (e.g., `p::Real=Inf`): This simply does not work.
    * type parameters: On the lowest interface level we can use an instantiation of the type parameters, like `an_element(::LazySet{Real})`; for this case one can also create a hyperlink reference. On higher (interface/type) levels, one has to use the `where` syntax in order to use the correct method; for this case there are no hyperlinks.
* [x] Remove duplicate documentation for concrete types.
  * See above (`where` syntax).
  * I did not find a way to make the `array` function for array types (probably due to nested type parameters) and the functions with default values (`norm`/`radius`/`diameter`) work.
* [x] Check that each implementation occurs in the docs.
* [x] Check that workaround for #72 is not created by default.
  * The page is not created, but the warnings are absorbed.
* [x] Fix broken doctests.
  * I have never seen those broken ones so far. Apparently up to now the doctests never executed for `VPolytope.jl` and `plot_recipes.jl`, probably because the documentation was not used before.

### DIFF
--- a/docs/src/lib/methods_fix.md
+++ b/docs/src/lib/methods_fix.md
@@ -1,0 +1,40 @@
+# Functions with several methods
+
+```@meta
+CurrentModule = LazySets
+DocTestSetup = quote
+    using RecipesBase, LazySets
+end
+```
+
+```@docs
+RecipesBase.apply_recipe
+addconstraint!
+an_element
+array
+center
+constraints_list
+convert
+diameter
+dim
+element
+halfspace_left
+halfspace_right
+high
+intersection
+low
+norm
+is_intersection_empty
+linear_map
+radius
+radius_hyperrectangle
+tohrep
+tovrep
+vertices_list
+*
++
+<=
+σ
+∈
+⊆
+```
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -20,7 +20,8 @@ makedocs(
         "Operations on Sets" => "man/set_operations.md",
         "A Reachability Algorithm" => "man/reach_zonotopes.md",
         "A Hybrid Reachability Algorithm" => "man/reach_zonotopes_hybrid.md",
-        "Concrete Polyhedra" => "man/concrete_polyhedra.md"],
+        "Concrete Polyhedra" => "man/concrete_polyhedra.md",
+        ],
         "Library" => Any[
         "Set Interfaces" => "lib/interfaces.md",
         "Common Set Representations" => "lib/representations.md",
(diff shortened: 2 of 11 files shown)


docs/src/index.md
```diff
@@ -189,6 +189,10 @@ Depth = 2
 
 ## Library Outline
 
+```@docs
+LazySets
```
mforets: just a comment: many people do not add a docstring to `module LazySets`, in which case this `@docs` is not needed -- by default julia will print the README.md of the package. i like this idea because it has some more info such as links to online documentation that our current `Main module for `LazySets.jl` -- a Julia package for calculus with convex sets.`

docs/src/lib/methods_fix.md
```diff
@@ -0,0 +1,41 @@
+# Functions with several methods
```
mforets: what is the purpose of the file `methods_fix.md`?

docs/src/index.md
schillic: Okay, then I vote for removing this `@docs`. Otherwise we get a warning.

docs/src/lib/methods_fix.md
schillic: If you add it, all the `Documenter` warnings about missing documentation is gone. Of course this is dangerous, so I commented this line. But it can help find missing docs at least for new function/type names quickly. In particular, I finally managed to exclude the warnings about `apply_recipe`!

docs/src/index.md
mforets: okay. can you do that?

docs/src/lib/methods_fix.md
mforets: hooray 🙌 

most if not all of these functions with several methods should be documented, such as `constraints_list`. why does Documenter complain? since we include some (maybe not all) the variations of `constraints_list` in the docs.

schillic: No, the reason is described in #72. `Documenter` just does not recognize methods with default values or `where` syntax. Currently there is just no other way than manually searching for those docs at the moment. That is why I said that it is dangerous :worried:

docs/src/index.md
schillic: done

schillic:
Yes, these are only there with `Polyhedra`.
It is really cheating here. I just define a file that absorbs all other warnings. I am really not sure if this is a good idea or not. At least we can spot problems with *new* functions again. To me the previous state was at least as useless...

PR #439, opened 2018-07-31
#402 - Use old version of Documenter for Travis build

### LINKED ISSUE #402: Documenter crashes when building the docs

schillic:
Closes #402.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,8 @@ matrix:
 #script:
 #  - julia -e 'Pkg.clone(pwd()); Pkg.build("LazySets"); Pkg.test("LazySets"; coverage=true)'
 after_success:
-  # documentation
-  - julia -e 'Pkg.add("Documenter"); cd(Pkg.dir("LazySets")); include(joinpath("docs", "make.jl"))'
+  # documentation (restrict Documenter to release version 0.18.0 due to problems, see #402)
+  - julia -e 'Pkg.add("Documenter", v"0.18.0", v"0.19.0"); cd(Pkg.dir("LazySets")); include(joinpath("docs", "make.jl"))'
   # code coverage (for both Coveralls and Codecov)
   - julia -e 'Pkg.add("Coverage")'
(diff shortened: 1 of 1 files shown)


schillic:
To explain, I wanted to avoid a `Pkg.add("Documenter"); Pkg.pin("Documenter", v"0.18.0")`.
`Pkg.add("Documenter", v"0.18.0")` did not work for some reason, but with two arguments it worked. However, the `v"0.19.0"` is a version that already causes the crash. It just so happens that, at least for me, `Pkg` uses the first version.

mforets:
Great. This is a temporary workaround until the bug is solved in the dependent packages.

schillic:
Do you agree that this closes #402?

mforets:
Yes and no. I agree to close it, although i suggest to have a new issue that highlights the fact that we cannot build docs with v0.19.

PR #450, opened 2018-08-01
#55 - Revise diameter function

### LINKED ISSUE #55: Pull up `diameter` from specific types to `LazySet`
Instead of

```julia
diameter(B::BallInf, p::Real=Inf)
diameter(H::Hyperrectangle, p::Real=Inf)
...
```

just define it once

```julia
diameter(S::LazySet, p::Real=Inf)
```

(In fact we have this, but it throws an error.)

@mforets: Any objection?

schillic:
Closes #55.

@mforets: Please check that these changes are correct:
1. The diameter is *always* two times the radius (of the respective norm). So there is no need to throw an error, because it will be thrown by `radius`.
2. The diameter of a `Hyperrectangle` is defined in the same way, so we can just use the default implementation.

### DIFF
--- a/src/AbstractHyperrectangle.jl
+++ b/src/AbstractHyperrectangle.jl
@@ -126,31 +126,6 @@ function radius(H::AbstractHyperrectangle, p::Real=Inf)::Real
 end
 
 """
-    diameter(H::AbstractHyperrectangle, [p]::Real=Inf)::Real
-
-Return the diameter of a hyperrectangular set.
-
-### Input
-
-- `H` -- hyperrectangular set
-- `p` -- (optional, default: `Inf`) norm
-
-### Output
-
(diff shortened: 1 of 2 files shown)


mforets:
> 1. The diameter is always two times the radius (of the respective norm).

yes. but in some particular case it could be useful to implement directly the diameter, instead of the radius.

> 2. The diameter of a hyperrectangle ...

yes

---

should we rather keep just the "error" here in LazySet, and instead define `radius` inside `BallInf`?

```
function radius(S::LazySet, p::Real=Inf)
    if p == Inf
        return radius(Approximations.ballinf_approximation(S)::BallInf, p)
    else
        error("the radius for this value of p=$p is not implemented")
    end
end
```

PR #464, opened 2018-08-02
#461 - Change VPolytope constructor from VPolygon

### LINKED ISSUE #461: VPolytope constructor from VPolygon shares vertices
Is this intended? I would say that this is not a good default behavior; we should copy the array instead.
We could add another parameter that controls sharing/copying.

schillic:
Closes #461.

By default, the vertices are now copied and not shared anymore.


schillic:
There is [a test](https://github.com/JuliaReach/LazySets.jl/blob/c67d7e13ab4e7e0497a83426151378e1705fb0df/test/unit_Polytope.jl#L55) already.
I think we wanted to describe constructors in the docstring of the type itself. And this is not a new feature, just a fix, so I am not legally bound to write it :grin:

Issue #130, opened 2018-01-10
Increase code coverage


schillic:

mforets:
what would be the target? 80, 90, 95% ?

schillic:
I would say ultimately 100% (which is realistic with this size of code), but we do not have to achieve it immediately. We can have several PRs.

schillic:
We should exclude `compat.jl` from the coverage computation.

mforets:
> Can we define different test suites?

we can have an argument that is read from `runtests.jl` , for example 
to be run either as

```julia
$ julia --color=yes test/runtests.jl
$ julia --color=yes test/runtests.jl --normal
```

or

```julia
$ julia --color=yes test/runtests.jl --long
```

with

```julia
# in runtests.jl

using Base.Test

if (length(ARGS) == 0) || (ARGS[1] == "--normal")
    # NORMAL TESTS
    include("continuous.jl")
    include("discrete.jl")
    include("inputs.jl")
elseif ARGS[1] == "--long"
    # LONG TESTS
    include("xxx.jl")
    include("yyy.jl")
end
```

mforets:
After #130 there are 4 remaining "red" regions: HPolytope and VPolytope (which are somehow tied to features from Polyhedra), apply_recipes (to which your comment of above applies), and ExponentialMap (that i don't know why we are missing those tests).

schillic:
And some small parts that I forgot. I want to add them in the near future.

PR #467, opened 2018-08-02
#71 - Strict doctests

### LINKED ISSUE #71: Failing doctests should fail the build

schillic:
Closes #71.

### DIFF
--- a/docs/make_doctests_only.jl
+++ b/docs/make_doctests_only.jl
@@ -1,0 +1,8 @@
+using Documenter, LazySets
+
+makedocs(
+    doctest = true,
+    modules = Module[LazySets, Approximations],
+    source = "src/lib",
+    strict = true
+)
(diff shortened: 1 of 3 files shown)


schillic:
It seems to work.
@mforets: What do you think?
We should probably incorporate [your comment](https://github.com/JuliaReach/LazySets.jl/issues/130#issuecomment-408680648) such that we do not always have to run these expensive tests.
We could also try to combine with the build stages idea from #102.

PR #469, opened 2018-08-02
#468 - Fix interface unit tests

### LINKED ISSUE #468: Fix interface unit tests
1. The tests do not work recursively as intended.
2. The tests for `AbstractSingleton` are missing.

schillic:
Closes #468.

I had to do some minor changes to make the tests pass again.

### DIFF
--- a/src/HPolygon.jl
+++ b/src/HPolygon.jl
@@ -41,7 +41,7 @@ HPolygon(S::LazySet) = convert(HPolygon, S)
 
 
 """
-    σ(d::AbstractVector{<:Real}, P::HPolygon{N})::Vector{N} where {N<:Real}
+    σ(d::AbstractVector{N}, P::HPolygon)::Vector{N} where {N<:Real}
 
 Return the support vector of a polygon in a given direction.
 
@@ -61,7 +61,7 @@ norm zero, any vertex is returned.
 Comparison of directions is performed using polar angles; see the overload of
 `<=` for two-dimensional vectors.
 """
-function σ(d::AbstractVector{<:Real}, P::HPolygon{N})::Vector{N} where {N<:Real}
+function σ(d::AbstractVector{N}, P::HPolygon)::Vector{N} where {N<:Real}
     n = length(P.constraints)
     @assert n > 0 "the polygon has no constraints"
     k = 1
--- a/src/VPolygon.jl
+++ b/src/VPolygon.jl
@@ -143,7 +143,7 @@ end
 
 
 """
-    σ(d::AbstractVector{<:Real}, P::VPolygon{N})::Vector{N} where {N<:Real}
+    σ(d::AbstractVector{N}, P::VPolygon)::Vector{N} where {N<:Real}
 
 Return the support vector of a polygon in a given direction.
 
@@ -171,8 +171,7 @@ have been sorted in counter-clockwise fashion.
 In that case a binary search algorithm can be used that runs in ``O(\\log n)``.
 See issue [#40](https://github.com/JuliaReach/LazySets.jl/issues/40).
 """
-function σ(d::AbstractVector{<:Real},
-           P::VPolygon{N})::Vector{N} where {N<:Real}
+function σ(d::AbstractVector{N}, P::VPolygon)::Vector{N} where {N<:Real}
     @assert !isempty(P.vertices) "the polygon has no vertices"
     i_max = 1
     @inbounds for i in 2:length(P.vertices)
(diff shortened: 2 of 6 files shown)


src/HPolygon.jl
```diff
@@ -61,7 +61,7 @@ norm zero, any vertex is returned.
 Comparison of directions is performed using polar angles; see the overload of
 `<=` for two-dimensional vectors.
 """
-function σ(d::AbstractVector{<:Real}, P::HPolygon{N})::Vector{N} where {N<:Real}
+function σ(d::AbstractVector{N}, P::HPolygon)::Vector{N} where {N<:Real}
```
mforets: alright; maybe add a line of this behavior in the output, or with an example .. a priori we could have matched the output numeric type with that of the input set (as it was before)..

mforets:
tests pass and docs build in my machine 👍

src/VPolygon.jl
```diff
@@ -171,8 +171,7 @@ have been sorted in counter-clockwise fashion.
 In that case a binary search algorithm can be used that runs in ``O(\\log n)``.
 See issue [#40](https://github.com/JuliaReach/LazySets.jl/issues/40).
 """
-function σ(d::AbstractVector{<:Real},
-           P::VPolygon{N})::Vector{N} where {N<:Real}
+function σ(d::AbstractVector{N}, P::VPolygon)::Vector{N} where {N<:Real}
```
mforets: for the "example" i mean something as simple as

```julia
julia> N = Rational{BigInt}
julia> P = VPolygon([N[1.0, 1.0], N[1.0, -1.0], N[-1.0, -1.0], N[-1.0, 1.0]])

julia> σ(N[1.0, 1.0], P)
2-element Array{Rational{BigInt},1}:
 1//1
 1//1

julia> σ([1.0, 1.0], P)
2-element Array{Float64,1}:
 1.0
 1.0
```

src/HPolygon.jl
mforets: .. hmm why don't we keep the signature 

```julia
σ(d::AbstractVector{N}, P::HPolygon{N})::Vector{N} where {N<:Real}
```

schillic: The problem is that for the interface tests I need consistent *input* types.
In this case I went for the minimal change, which was to remove the numeric type for the second argument for the polygon types.
The current signature is `σ(::AbstractVector{<:Real}, ::LazySet)`.

I am not against adding the type for the second argument again, but then we also need to add it for all other set types. Actually, I would prefer that.

What do you think?

mforets: > I am not against adding the type for the second argument again, but then we also need to add it for all other set types. 

the motivation for `HPolygon` here could be to avoid the "unsafe" case where you pass an inexact vector `d` to an exact polygon `P`. 

i suppose that this type of observation holds for all other set types. but i also remember that we had some performance issues so i would be careful of changing everything; at least do it gradually.

mforets: for instance in

```julia
function σ(d::V, lm::LinearMap) where {N<:Real, V<:AbstractVector{N}}
    return lm.M * σ(_At_mul_B(lm.M, d), lm.X)
end
```

we don't strictly fix that the numeric type of the `lm` is `N`, on purpose (the discussion is somewhere.)

schillic: I think the problem was with the *output* type `AbstractVector{N}`, which we have dropped. I do not see a possible problem with the input type.

schillic: Then, should I do the changes in this PR or in a new one? If in a new one, merge this one first or wait for the other one?

mforets: > I think the problem was with the output type AbstractVector{N}, which we have dropped.

OK. Typically we have `σ(d::V, B::Ball1{N}) where {N<:Real, V<:AbstractVector{N}}` which is "safe" in the context of the current discussion. I guess that the types that need to be adapted are not that many after all.

> Then, should I do the changes in this PR or in a new one? If in a new one, merge this one first or wait for the other one?

I don't think that #468 is representative of the proposed change, so, if you want we can merge this branch and continue with #478.

src/VPolygon.jl
mforets: BTW, this example would produce an error when the new behavior in #478 is implemented.

schillic: What exactly would break? Can you try in #480?

schillic: Ah, now I see. Yes, this should not work, but is this not what we want?

mforets: yes, this is what we want, i thought we can add a test with the (wrong) usage, or at least mention in the docstring

schillic: Hm, in which file? This basically holds for every type...

mforets: True. I thought adding it only on the `VPolygon.jl` and/or hpolygon. But you're right, a section of the manual with some examples would be good. I've opened #483

PR #472, opened 2018-08-03
#451 - Revise type parameters of LinearMap

schillic:
Closes #451.

* [x] reorder type parameters
* [x] add type parameter for set
* [x] add type parameter for matrix

PR #480, opened 2018-08-04
#478 - "Safely" querying support vectors 

### LINKED ISSUE #478: "Safely" querying support vectors
Proposal: check that in support vector queries the numeric type of the given direction matches that of the lazy set, eg. as in `σ(d::V, B::Ball1{N}) where {N<:Real, V<:AbstractVector{N}}`.

schillic:
Closes #478.

* [x] all types, including `LinearMap`
* [x] update `@docs` in `docs/src/lib/*` and add a note in `LazySet.jl` about the numeric types


schillic:
Still missing: The change for `LinearMap`. First I would like to merge #472 and then rebase.

schillic:
ready from my side

PR #485, opened 2018-08-06
#484 - Fix membership test of ExponentialMap

### LINKED ISSUE #484: Membership test of ExponentialMap is incorrect
Example:
```julia
julia> n = 3;
julia> x = ones(n)
3-element Array{Float64,1}:
 1.0
 1.0
 1.0
julia> b = BallInf(ones(n), 0.1);

julia> sme = SparseMatrixExp(speye(n))
3×3 LazySets.SparseMatrixExp{Float64}:
 2.71828  0.0      0.0    
 0.0      2.71828  0.0    
 0.0      0.0      2.71828
julia> ex = sme * x
3-element Array{Float64,1}:
 2.71828
 2.71828
 2.71828
julia> em = sme * b;

julia> x ∈ b
true
```
Now since `x ∈ b`, we should expect that `sme * x ∈ sme * b`, or in short `ex ∈ em`.
However:
```julia
julia> sme * x ∈ sme * b
false
julia> ex ∈ em
false
```

schillic:
Closes #484.

* [x] fix code
* [x] add unit test
* [x] fix doctest

### DIFF
--- a/src/ExponentialMap.jl
+++ b/src/ExponentialMap.jl
@@ -253,7 +253,7 @@ This follows from ``\\exp(-M)⋅\\exp(M) = I`` for any ``M``.
 julia> em = ExponentialMap(SparseMatrixExp(SparseMatrixCSC([2.0 0.0; 0.0 1.0])),
                            BallInf([1., 1.], 1.));
 
-julia> ∈([5.0, 1.0], em)
+julia> ∈([-1.0, 1.0], em)
 false
 julia> ∈([1.0, 1.0], em)
 true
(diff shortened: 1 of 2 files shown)


src/ExponentialMap.jl
```diff
@@ -253,15 +253,15 @@ This follows from ``\\exp(-M)⋅\\exp(M) = I`` for any ``M``.
 julia> em = ExponentialMap(SparseMatrixExp(SparseMatrixCSC([2.0 0.0; 0.0 1.0])),
                            BallInf([1., 1.], 1.));
 
-julia> ∈([5.0, 1.0], em)
+julia> ∈([-1.0, 1.0], em)
```
mforets: usually we need a space before each `julia> ` for the test to work. does it mean that we are not doctesting 

```julia
∈(x::AbstractVector{N}, em::ExponentialMap{<:LazySet{N}})::Bool where {N<:Real}
```

schillic: No, the lines below each command must be equal to the output in the REPL. If the command ends with a `;`, the REPL just prints an empy line. Hence you need an empty line *after* each `julia>` that ends in `;`. The doctests worked here.

PR #487, opened 2018-08-06
Revise set type constructors

schillic:
Our convenience constructors are sometimes redundant, and in Julia v0.7 they even override the default constructors (see #121). Further reading [here](https://docs.julialang.org/en/latest/manual/constructors/#Parametric-Constructors-1).


schillic:
The last commit fixes all remaining v0.7 warnings *during precompilation* (there are still tons of warnings when running the unit tests, even some errors). The unit tests were priceless for this PR :+1:

mforets:
> and in Julia v0.7 they even override the default constructors

good catch.

PR #508, opened 2018-08-08
#121 - Fix imports in v0.7

schillic:
See #121.

For v0.7 the imports need a stricter format.

### DIFF
--- a/test/unit_decompose.jl
+++ b/test/unit_decompose.jl
@@ -1,8 +1,7 @@
-import LazySets.Approximations,
-       Approximations.decompose,
-       Approximations.BoxDirections,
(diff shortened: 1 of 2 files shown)


mforets:
the idea of the that code was to omit one import, maybe only this works now..

```julia
import LazySets.Approximations
import Approximations.decompose ...
```

schillic:
I do not know why the old version does not work anymore.

schillic:
I think this is the related remark from their release news:

"The syntax `using A.B` can now only be used when `A.B` is a module, and the syntax `using A: B` can only be used for adding single bindings"

PR #512, opened 2018-08-11
More efficient alternative Hyperrectangle constructor for v0.7

schillic:
This is a proposal to remove one of the keyword argument constructors for `Hyperrectangle` to make the other one more efficient in Julia v0.7. We do not need the removed constructor because it just imitates the default constructor in an inefficient way.

Old version in v0.6:
```julia
julia> @time Hyperrectangle(high=[1.], low=[1.])
  0.000283 seconds (72 allocations: 3.703 KiB)
LazySets.Hyperrectangle{Float64}([1.0], [0.0])
```
New version in v0.7:
```julia
julia> @time Hyperrectangle(high=[1.], low=[1.])
  0.000022 seconds (15 allocations: 848 bytes)
Hyperrectangle{Float64}([1.0], [0.0])
```

### DIFF
--- a/src/Hyperrectangle.jl
+++ b/src/Hyperrectangle.jl
@@ -15,6 +15,27 @@ Cartesian product of one-dimensional intervals.
 - `center` -- center of the hyperrectangle as a real vector
 - `radius` -- radius of the ball as a real vector, i.e., half of its width along
               each coordinate direction
+
+### Examples
+
+There is also a constructor from lower and upper bounds with keyword arguments
+`high` and `low`.
+The following two constructions are equivalent:
+
+```jldoctest
+julia> c = ones(2);
+
+julia> r = [0.1, 0.2];
+
+julia> l = [0.9, 0.8];
+
+julia> h = [1.1, 1.2];
+
+julia> Hyperrectangle(c, r)
+LazySets.Hyperrectangle{Float64}([1.0, 1.0], [0.1, 0.2])
+julia> Hyperrectangle(low=l, high=h)
+LazySets.Hyperrectangle{Float64}([1.0, 1.0], [0.1, 0.2])
+```
 """
 struct Hyperrectangle{N<:Real} <: AbstractHyperrectangle{N}
     center::Vector{N}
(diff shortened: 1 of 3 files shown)


src/Hyperrectangle.jl
```diff
@@ -34,56 +55,34 @@ end
...
+end # @eval
+else # if VERSION < v"0.7-"
+@eval begin
+function Hyperrectangle(;
+                        high::AbstractVector{N},
+                        low::AbstractVector{N}) where {N<:Real}
+    # compute center and radius from high and low vectors
+    center = (high .+ low) ./ 2
+    radius = abs.(high .- center)
+    return Hyperrectangle(center, radius)
```
mforets: does it "help" the compiler to write `Hyperrectangle{N}(center, radius)` here?

test/unit_Hyperrectangle.jl
```diff
@@ -68,22 +68,18 @@ for N in [Float64, Rational{Int}, Float32]
...
-    @test H1.radius == H2.radius
-    @test H2.radius ≈ H3.radius
-    @test_throws ArgumentError Hyperrectangle(xyz="zyx")
+    H2 = Hyperrectangle(low=l, high=h)
+    @test H1.center ≈ H2.center
+    @test H1.radius ≈ H2.radius
 
     # Test low and high methods for a hyperrectangle
-    H = Hyperrectangle(center=to_N(N, [-2.1, 5.6, 0.9]), radius=fill(to_N(N, 0.5), 3))
+    H = Hyperrectangle(to_N(N, [-2.1, 5.6, 0.9]), fill(to_N(N, 0.5), 3))
```
mforets: although not pertinent to this PR, i think we can do

```julia
H = Hyperrectangle(N[-2.1, 5.6, 0.9], fill(N(0.5), 3))
```

src/Hyperrectangle.jl
```diff
@@ -34,56 +55,34 @@ end
...
+        radius = abs.(high .- center)
         return Hyperrectangle(center, radius)
     end
-    throw(ArgumentError("invalid arguments for Hyperrectangle: Use either " *
-        "'center' and 'radius' or 'high' and 'low'."))
+    throw(ArgumentError("invalid arguments for Hyperrectangle: " *
+        "use 'high' and 'low'"))
+end
+end # @eval
+else # if VERSION < v"0.7-"
```
mforets: this was supposed to be `else # if VERSION > v"0.7-"` ?

should we keep only this version?

mforets:
to be honest i don't understand the change. there are no keyword arguments if we use the constructor

```julia
# convenience constructor without type parameter
Hyperrectangle(center::Vector{N}, radius::Vector{N}) where {N<:Real} =
    Hyperrectangle{N}(center, radius)
```

src/Hyperrectangle.jl
schillic: I am not completely sure, but `Hyperrectangle(center, radius)` falls back to `Hyperrectangle{N}(center, radius)`, so I hope that this step is skipped by the compiler.

test/unit_Hyperrectangle.jl
schillic: Sometimes this does not work with `Rational`s, although here with only one decimal place it should work.

src/Hyperrectangle.jl
schillic: No, this just a comment to which `if` this `else` belongs (because of the many nestings). I see now that this can be confusing. Should I remove the comment?

schillic:
> to be honest i don't understand the change. there are no keyword arguments if we use the constructor

The point is that instead of a generic dictionary we require two type-specific arguments. This is now possible in v0.7. The compiler can create much more efficient code out of this.
We never needed the constructor with keyword arguments for `center` and `radius` (as you observed), so in my opinion we can safely remove it.

src/Hyperrectangle.jl
schillic: ?

src/Hyperrectangle.jl
mforets: you're right. both versions compile to the same code

```julia
julia> @code_warntype Hyperrectangle2(high=rand(2), low=rand(2))
Body::Hyperrectangle{Float64}
 1 ─ %1  = (Base.getfield)(#temp#, :high)::Array{Float64,1}                                                                                        │╻     getindex
 │   %2  = (Base.getfield)(#temp#, :low)::Array{Float64,1}                                                                                         ││    
 │   %3  = (Base.slt_int)(0, 1)::Bool                                                                                                              ││╻╷╷╷  iterate
 └──       goto #3 if not %3                                                                                                                       │││┃│    iterate
 2 ─       goto #4                                                                                                                                 ││││┃     iterate
 3 ─       invoke Base.getindex(()::Tuple{}, 1::Int64)                                                                                             │││││ 
 └──       $(Expr(:unreachable))                                                                                                                   │││││ 
 4 ┄       goto #5                                                                                                                                 ││││  
 5 ─       goto #6                                                                                                                                 ││╻     iterate
 6 ─       goto #7                                                                                                                                 ││    
 7 ─       nothing                                                                                                                                 │     
 │   %12 = invoke Main.:(#Hyperrectangle2#4)(%1::Array{Float64,1}, %2::Array{Float64,1}, _3::Function)::Hyperrectangle{Float64}                    │     
 └──       return %12  
```

PR #517, opened 2018-08-11
#121 - Add package InteractiveUtils for v0.7 (for 'subtypes' function)

schillic:
See #121.

### DIFF
--- a/test/check_method_implementation.jl
+++ b/test/check_method_implementation.jl
@@ -1,3 +1,7 @@
+if VERSION >= v"0.7-"
+    @eval using InteractiveUtils
+end
+
 """
     check_method_implementation(interface::Type,
                                 func_name,


mforets:
sorry, i just saw that `subtypes`  appears many times in `/src` as well. 

should we add the import in compat? should we import only `import InteractiveUtils.subtypes`?

schillic:
Hm, there are more problems.
When I import, I get this warning:
```julia
Warning: Package LazySets does not have InteractiveUtils in its dependencies:
│   added InteractiveUtils as a dependency but haven't updated your primary
└ Loading InteractiveUtils into LazySets from project dependency, future warnings for LazySets are suppressed.
```
Is there a way to `REQUIRE` a package only with v0.7?

PR #521, opened 2018-08-11
#121 - Replace 'start/next/done' iterators -> 'iterate'

schillic:
See #121.

Unfortunately, v0.7 deprecates the 'start/next/done' iterators, so now we have to support both versions.


mforets:
in my machine, the tests for `unit_template_directions.jl` pass in 0.7

PR #523, opened 2018-08-11
#522 - Illegal subtyping syntax in concrete intersection of hyperrectangles

### LINKED ISSUE #522: Illegal subtyping syntax in concrete intersection of hyperrectangles
in v0.7:
```julia
using LazySets
for N in [Float64]
    B1 = BallInf(N[1.0], N(1.0))
    B2 = BallInf(N[1.0], N(1.0))
    intersection(B1, B2)
end
```
This crashes with a crazy error message:
`Unreachable reached at 0x7ff07c15ed18`
`signal (4): Illegal instruction`
But if I run it outside of the loop, it is fine:
```julia
using LazySets
N = Float64
B1 = BallInf(N[1.0], N(1.0))
B2 = BallInf(N[1.0], N(1.0))
intersection(B1, B2)
```
There is an open issue `#25430` in the Julia repo that is probably related.

For us the fix is in `concrete_intersection.jl` in the return type of the hyperrectangle method:
`::Union{<:Hyperrectangle{N}, EmptySet{N}}`
The `<:` does not make sense (subtyping of `Hyperrectangle` is not possible).

I have no idea why this leads to this crazy error, but anyway :relieved:

schillic:
Closes #522.

### DIFF
--- a/src/concrete_intersection.jl
+++ b/src/concrete_intersection.jl
@@ -78,7 +78,7 @@ Otherwise the result uses these borders in each dimension.
 """
 function intersection(H1::AbstractHyperrectangle{N},
                       H2::AbstractHyperrectangle{N}
-                     )::Union{<:Hyperrectangle{N}, EmptySet{N}} where {N<:Real}
+                     )::Union{Hyperrectangle{N}, EmptySet{N}} where {N<:Real}
     n = dim(H1)
     c1 = center(H1)
     c2 = center(H2)


schillic:
I do not understand your point. `return EmptySet{N}()` definitely does not result in a `Hyperrectangle{N}`.

mforets:
i mean the last line of the function. let's do it in a separate PR though..

schillic:
This is strange because [the respective constructor](https://github.com/JuliaReach/LazySets.jl/blob/c57d320afe748c3b4cc8d70ca9d73fc288c0b7a6/src/Hyperrectangle.jl#L74) does not have a type annotation. In #512 I also suggest to change this code, but there will be no type annotation as well.

schillic:
To answer your initial question: This does not have anything to do with union-splitting. Union-splitting happens in the caller code, and in this case it should happen regardless of what the last line looks like, because we annotate the return type here. What I can imagine is that we should annotate the return type of that specific `Hyperrectangle` constructor accordingly, so that the compiler can see that the type is indeed parametric in `N` and eliminate that check. (I might be wrong, though.)

PR #525, opened 2018-08-11
#121 - Replace 'full' -> 'Matrix'

schillic:
See #121.

### DIFF
--- a/test/unit_ExponentialMap.jl
+++ b/test/unit_ExponentialMap.jl
@@ -68,7 +68,7 @@ for N in [Float64, Float32]
     σ(sparsevec(d), emap)
 
     # check consistency with respect to explicit computation of the matrix exponential
-    svec_explicit = σ(d, expm(full(m)) * b)
+    svec_explicit = σ(d, expm(Matrix(m)) * b)
     if N == Float64
         # precision with Float32 is not sufficient
         @test svec ≈ svec_explicit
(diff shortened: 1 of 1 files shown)


test/unit_ExponentialMap.jl
```diff
@@ -68,7 +68,7 @@ for N in [Float64, Float32]
     σ(sparsevec(d), emap)
 
     # check consistency with respect to explicit computation of the matrix exponential
-    svec_explicit = σ(d, expm(full(m)) * b)
+    svec_explicit = σ(d, expm(Matrix(m)) * b)
```
mforets: in julia 1.0 the matrix exponential function is also called `exp`. (should we use `Compat.expm` then?)


```julia
julia> expm(rand(2,2))
┌ Warning: `expm` is deprecated, use `exp` instead.
│   caller = top-level scope at none:0
└ @ Core none:0
2×2 Array{Float64,2}:
 2.35497  0.636294
 1.77983  1.93737 
```

mforets: or `Compat.exp`?

mforets: hmm i don't find it [here](https://github.com/JuliaLang/Compat.jl#new-functions-macros-and-methods), so i guess we have to touch our `compat.jl`..

schillic: Yes, this was supposed to be the next PR. It is not related to this one. One warning at a time :snail:

PR #527, opened 2018-08-11
#121 - Replace 'expm' -> 'Compat.expm'

schillic:
See #121.


mforets:
does `Compat.expm` work for you?

```julia
$ julia7
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: https://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.7.0 (2018-08-08 06:46 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-apple-darwin14.5.0

julia> using LinearAlgebra, Compat

julia> Compat.expm([1.0 1.0; 1.0 1.0])
WARNING: importing deprecated binding Base.expm into Compat.
WARNING: Base.expm is deprecated: it has been moved to the standard library package `LinearAlgebra`.
Add `using LinearAlgebra` to your imports.
 in module Compat
┌ Warning: `expm` is deprecated, use `exp` instead.
│   caller = top-level scope at none:0
└ @ Core none:0
2×2 Array{Float64,2}:
 4.19453  3.19453
 3.19453  4.19453
```

schillic:
Strange, I could swear that there was no warning. But yes, I also get it.

schillic:
This is really nasty, because `Compat.expm` and `Compat.exp` behave differently in v0.7. Since I could not find that in their code, I imagine that it just falls back to the original definition. So we need to go with a custom function.

schillic:
Is it okay to define this in `compat.jl`? I do not expect that we plan to use these functions in the source code.

mforets:
> Is it okay to define this in compat.jl? I do not expect that we plan to use these functions in the source code.

yes, i think so. in a later stage if we drop support for 0.6 we can remove `compat.jl` altogether..

PR #535, opened 2018-08-12
Update .travis.yml for 0.7

mforets:


schillic:
I am not a fan of adding more failing versions, even if we add them to the allowed failures. This just makes more machines run for nothing.
We should first at least make sure that the unit tests work locally. Then we may need to fix some code in the Travis script.

mforets:
i rebase w/master to see if it works now - and it does 🙌

schillic:
Cool! But it still fails for the nightly build. So it is good that you added this additional build :+1:

mforets:
I haven't checked why the nightly fails (it could be just dependencies failing. i think we can accept this).

mforets:
merge?

PR #536, opened 2018-08-12
Update Interval.jl docstring

mforets:
the comment is old, since

```julia

julia> using LazySets

julia> Interval(0.0, 1.0)
```

### DIFF
--- a/src/Interval.jl
+++ b/src/Interval.jl
@@ -28,8 +28,6 @@ library for representation of intervals and arithmetic operations.
 ### Examples
 
 Unidimensional intervals are symbolic representations of a real closed interval.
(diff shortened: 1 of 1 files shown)


schillic:
The iterators can be written as follows:
```julia
Base.iterate(input::ConstantInput, state::Void=nothing) = (input.U, state)

Base.iterate(input::VaryingInput, state::Int=1) =
    state > length(input.U) ? nothing : (input.U[state], state + 1)
```

mforets:
> The iterators can be written as follows:

Thank for the help! Although the comment got in the wrong issue/repository :smile:

schillic:
Oh :stuck_out_tongue_winking_eye:

PR #540, opened 2018-08-15
#539 - Separate test suite for Polyhedra

### LINKED ISSUE #539: Separate test suite for Polyhedra
Suggestion: Create a new test suite to run tests that require `Polyhedra.jl`. Then we should remove the respective tests in `unit_interfaces.jl` (`vertices_list`) from the default test suite. This makes our test suite run in Julia v0.7.
We can also use the new test suite for #130 to add unit tests for `Polyhedra`-related functionality, which we currently lack.

schillic:
Closes #539.

### DIFF
--- a/test/check_method_implementation.jl
+++ b/test/check_method_implementation.jl
@@ -1,7 +1,8 @@
 """
     check_method_implementation(interface::Type,
                                 func_name,
-                                args_funcs::AbstractVector{Function},
+                                args_funcs::AbstractVector{Function};
+                                [ignore_types]::Vector{Type}=Type[],
                                 [print_results]::Bool=false
                                )::Bool
 
--- a/test/check_method_implementation.jl
+++ b/test/check_method_implementation.jl
@@ -13,6 +13,8 @@ Check that a given (interface) function is implemented by all subtypes.
 - `func_name`     -- function name
 - `args_funcs`    -- list of functions that each map a type to an argument
                      signature tuple
+- `ignore_types`  -- (optional, default: `Type[]`) list of types that should be
+                     ignored
 - `print_results` -- (optional, default: `false`) flag for printing intermediate
                      results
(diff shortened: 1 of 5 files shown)


schillic:
I think we can now remove `Polyhedra` and `CDDLib` from our dependencies. Should we do that?

mforets:
i like the idea. 

we may need  another `REQUIRE` for the tests which includes them (an for building the docs these are needed too)

test/check_method_implementation.jl
```diff
@@ -13,6 +13,8 @@ Check that a given (interface) function is implemented by all subtypes.
 - `func_name`     -- function name
 - `args_funcs`    -- list of functions that each map a type to an argument
                      signature tuple
+- `ignore_types`  -- (optional, default: `Type[]`) list of types that should be
```
mforets: `ignore_types` is missing from the docstring signature

mforets:
should we make some adaptations in `unit_Polytope.jl`? for instance,

```julia
    # singleton list (only available with Polyhedra library)
    if !isdefined(@__MODULE__, :Polyhedra) && (polyhedra-in-testset)
        @test_throws MethodError singleton_list(p)
    end
```

schillic:
> we may need another REQUIRE for the tests which includes them (an for building the docs these are needed too)

I think for now we do not need the package for the default test suite.
And for the docs I am not sure - where would we need them?

schillic:
> should we make some adaptations in unit_Polytope.jl? for instance,

Hm, why? If the module is loaded for some reason we did not think about, the test will not throw an error, so I think the current check is exactly what we want.

mforets:
> Hm, why? If the module is loaded for some reason we did not think about, the test will not throw an error, so I think the current check is exactly what we want.

ok.

PR #546, opened 2018-08-16
#545 - Add StaticArrays before building the docs

### LINKED ISSUE #545: Add StaticArrays before building the docs
In `man/convex_hulls.md`, we have a line `using StaticArrays`. I have no idea why this worked in v0.6, but in v0.7 this fails for me. So we would have to `Pkg.add` it.

schillic:
Closes #545.


schillic:
I just saw that we install this package also in other builds. So apparently we do not need this PR...

PR #568, opened 2018-08-20
#567 - Error tolerance should be compared using <= instead of <

### LINKED ISSUE #567: Error tolerance should be compared using <= instead of <
We define `const TOL(N::Type{Rational{INNER}}) where {INNER} = zero(N)`. During iterative refinement, we compare two vectors with `norm(p1-p2, 2) < TOL(N)`. This will always result in `false`. However, we want to terminate in the case `p1 == p2`.

schillic:
Closes #567.


schillic:
The build fails are due to #569. Locally the tests pass.

Issue #569, opened 2018-08-20
FastRounding's most recent release is incompatible with Julia v0.6


schillic:
See the [releases of `FastRounding.jl`](https://github.com/JeffreySarnoff/FastRounding.jl/releases). In Julia v0.6, we need to stick to version `v1.2` instead of the new version `v2.0`. Locally my version is stuck at `v0.0.5` and does not want to update (not sure why), but Travis uses the newest version (and hence breaks).

schillic:
I reported this issue [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/199).

schillic:
This is now fixed on `METADATA`.

PR #570, opened 2018-08-20
Add v0.6.0 build for Travis

schillic:
I realized that if our `REQUIRE` file says `julia 0.6`, we should actually test if the package runs in `v0.6.0` :face_with_head_bandage:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: julia
 os:
   - linux
 julia:
+  - 0.6.0
   - 0.6
   - 0.7
   - nightly


schillic:
The build fails are due to #569.

mforets:
i suggest to keep the `.travis.yml` build as it is. adding another minor version will make it take even more time for the CI.

the requirement `0.6` is the minimum version for which LazySets works, so 0.6.4 is included.

schillic:
> adding another minor version will make it take even more time for the CI.

They run in parallel.

> the requirement 0.6 is the minimum version for which LazySets works.

Do you have a proof? This is exactly why I want to test it.

mforets:
> They run in parallel.

ok.

> Do you have a proof? This is exactly why I want to test it.

we removed minor julia versions from `.travis.yml` before. shound't we instead update `REQUIRE` with `0.6.4`?

schillic:
> shound't we instead update REQUIRE with 0.6.4?

But why require a new version if there is no reason?

mforets:
i agree. under the assumption that for patch releases in julia base the code remains compatible, there would be no need for extra testing. but i recall there were some issues..

schillic:
Okay, let us keep it as is.

PR #573, opened 2018-08-20
#130 - Add Rational type to unit tests

schillic:
See #130.
This fixes two remaining TODOs in the unit tests.

### DIFF
--- a/test/unit_decompose.jl
+++ b/test/unit_decompose.jl
@@ -3,29 +3,7 @@ import LazySets.Approximations: decompose,
                                 OctDirections,
                                 BoxDiagDirections
 
-for N in [Float64, Float32] # TODO Rational{Int}
-    # =============================
-    # Check that Issue #43 is fixed
-    # =============================
-
-    CPA = CartesianProductArray
-    X = CPA([BallInf(to_N(N, [0.767292, 0.936613]), to_N(N, 0.1)),
-             BallInf(to_N(N, [0.734104, 0.87296]), to_N(N, 0.1))])
-    Y = to_N(N, [1.92664 1.00674 1.0731 -0.995149;
-        -2.05704 3.48059 0.0317863 1.83481;
-        0.990993 -1.97754 0.754192 -0.807085;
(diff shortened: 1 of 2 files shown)


test/unit_decompose.jl
```diff
@@ -3,28 +3,26 @@ import LazySets.Approximations: decompose,
...
-    Ω0 = CH(X, Y)
+    Ω0 = CH(X, A * X)
     dec = decompose(Ω0, set_type=HPolygon)
     dec1 = dec.array[1]
 
-    @test dec1.constraints[1].b ≈ to_N(N, 2.84042586)
-    @test dec1.constraints[2].b ≈ to_N(N, 4.04708832)
-    @test dec1.constraints[3].b ≈ to_N(N, -0.667292)
-    @test dec1.constraints[4].b ≈ to_N(N, -0.836613)
+    @test isapprox(dec1.constraints[1].b, to_N(N, 2.84042586); rtol=1)
```
mforets: why is `rtol=1` 

```julia
For real or complex floating-point values, if an atol > 0 is not specified, rtol defaults to the square root of eps of the type of x or y,
  whichever is bigger (least precise). This corresponds to requiring equality of about half of the significand digits. Otherwise, e.g. for integer
  arguments or if an atol > 0 is supplied, rtol defaults to zero.
```

schillic: It needs to be `> 0` for `Rational`s. `1 > 0` :smiley:

mforets: with `rtol=1`, we have this behavior:

```julia
julia> isapprox(1.0, 0.0, rtol=1)
true

julia> isapprox(2//1, 1//1, rtol=1)
true
```

is this what we want to test?

schillic: Oh, no :disappointed:
Then I will move these tests to the Float tests below, thanks :+1:

schillic:
merge?

mforets:
thanks for working on this!

PR #588, opened 2018-08-24
Use technical term 'centrally symmetric' for 'point symmetric'

schillic:
According to [Wikipedia](https://en.wikipedia.org/wiki/Point_reflection), what I called *point symmetry* is weaker than *central symmetry*:
> An object that is invariant under a point reflection is said to possess *point symmetry*; if it is invariant under point reflection through its center, it is said to possess *central symmetry* or to be *centrally symmetric*.

@mforets:
Should we clarify by merging this PR?
Should we also rename the interfaces?


mforets:
> Should we clarify by merging this PR?

ok

> Should we also rename the interfaces?

👍  , those interfaces have a `center(...)` method, so i think it is good to use `Centrally` in their names.

schillic:
I renamed the interfaces, good to merge from my side!

PR #594, opened 2018-08-28
#534 - Remove redundant 'to_N(N, ...)' from unit tests

### LINKED ISSUE #534: Remove redundant 'to_N(N, ...)' from unit tests
Sometimes they are needed for the `Rational` conversion.

We can probably also simplify integer value conversions like `N(1.)` to `N(1)`.

schillic:
Closes #534.

### DIFF
--- a/test/unit_Polygon.jl
+++ b/test/unit_Polygon.jl
@@ -134,14 +134,14 @@ for N in [Float64, Float32, Rational{Int}]
     @test σ(d, vp) == points[2]
 
     # test that #83 is fixed
-    v = VPolygon(to_N(N, [[2.0, 3.0]]))
+    v = VPolygon([N[2.0, 3.0]])
     @test N[2.0, 3.0] ∈ v
     @test !(N[3.0, 2.0] ∈ v)
-    v = VPolygon(to_N(N, [[2.0, 3.0], [-1.0, -3.4]]))
+    v = VPolygon([N[2.0, 3.0], to_N(N, [-1.0, -3.4])])
     @test to_N(N, [-1.0, -3.4]) ∈ v
 
     # an_element function
-    v = VPolygon(to_N(N, [[2., 3.]]))
+    v = VPolygon([N[2., 3.]])
     @test an_element(v) ∈ v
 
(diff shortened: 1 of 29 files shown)


test/unit_Polygon.jl
```diff
@@ -198,19 +198,19 @@ for N in [Float64, Float32, Rational{Int}]
             @test v1 ∈ h1
         elseif i == 2
             c = h1.constraints[1]
-            @test c.a ≈ to_N(N, [0.4, 0.5])&& c.b ≈ to_N(N, (0.46))
+            @test c.a ≈ to_N(N, [0.4, 0.5]) && c.b ≈ to_N(N, (0.46))
```
mforets: is there sth special about this test that `to_N` is needed ? instead of `N[0.4, 0.5]`

mforets: since

```julia
julia> N = Rational{Int}
Rational{Int64}

julia> N[0.4, 0.5]
2-element Array{Rational{Int64},1}:
 3602879701896397//9007199254740992
                1//2
```

schillic: The numbers are very big. Sometimes this leads to overflow errors.
In this case there is no problem, but if we do computations with those numbers, I think it is better to make them small again to avoid problems if we eentually extend the tests.

schillic: The automatic "good" conversion works for integers (and apparently for `0.5` as well, but I did not consider this case), so I removed the `to_N` for integers.

mforets: OK, in those cases you prefer to use `rationalize`

```julia
julia> rationalize(0.4)
2//5

julia> Rational{Int}(0.4)
3602879701896397//9007199254740992
```

mforets: as an alternative `N(4//10)` is `2//5` if `N = Rational{Int}` or `0.4` if `N` is a float.

schillic: > OK, in those cases you prefer to use rationalize

This is what `to_N` does :+1:

PR #596, opened 2018-08-29
#595 - Polytopes intersect should be intersection

### LINKED ISSUE #595: Polytopes' intersect should be intersection
**Concrete** intersection has been called "intersection", while **lazy** intersection is ∩ , spelled out as "intersect", for the case of rectangles:

```julia
julia> B1 = BallInf(ones(2), 1.);

julia> B2 = BallInf(ones(2), 1.);

julia> intersection(B1, B2)

LazySets.Hyperrectangle{Float64}([1.0, 1.0], [1.0, 1.0])

julia> intersect(B1, B2)
LazySets.Intersection{Float64,LazySets.BallInf{Float64},LazySets.BallInf{Float64}}(LazySets.BallInf{Float64}([1.0, 1.0], 1.0), LazySets.BallInf{Float64}([1.0, 1.0], 1.0))
```

This is to propose adjusting the function's name in 

```julia
    intersect(P1::HPolytope{N}, P2::HPolytope{N};
              [backend]=CDDLib.CDDLibrary(),
              [prunefunc]=removehredundancy!)::HPolytope{N} where {N<:Real}
```

and the corresponging one for `VPolytope`.

mforets:
Closes #595.

### DIFF
--- a/docs/src/lib/representations.md
+++ b/docs/src/lib/representations.md
@@ -349,6 +349,7 @@ dim(::HPolytope)
 addconstraint!(::HPolytope{Real}, ::LinearConstraint{Real})
 constraints_list(::HPolytope)
 tosimplehrep(::HPolytope)
+intersection(::HPolytope, ::HPolytope)
 ```
 Inherited from [`LazySet`](@ref):
 * [`norm`](@ref norm(::LazySet, ::Real))
@@ -366,6 +367,7 @@ VPolytope
 dim(::VPolytope)
 σ(::AbstractVector{Real}, ::VPolytope{Real})
 vertices_list(::VPolytope)
+intersection(::VPolytope, ::VPolytope)
 ```
 Inherited from [`LazySet`](@ref):
 * [`norm`](@ref norm(::LazySet, ::Real))
--- a/docs/src/lib/representations.md
+++ b/docs/src/lib/representations.md
@@ -349,7 +349,6 @@ dim(::HPolytope)
 addconstraint!(::HPolytope{Real}, ::LinearConstraint{Real})
 constraints_list(::HPolytope)
 tosimplehrep(::HPolytope)
-intersection(::HPolytope, ::HPolytope)
 ```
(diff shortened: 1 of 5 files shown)


mforets:
in 0.7 this is not working at all

```julia
julia> using Polyhedra

julia> V1 = VPolytope([rand(2) for i in 1:20]);

julia> V2 = VPolytope([rand(2) for i in 1:20]);

julia> intersection(V1, V2)
ERROR: MethodError: no method matching intersection(::VPolytope{Float64}, ::VPolytope{Float64})
Stacktrace:
 [1] top-level scope at none:0

julia> LazySets.intersection(V1, V2)
ERROR: MethodError: no method matching intersection(::VPolytope{Float64}, ::VPolytope{Float64})
Stacktrace:
 [1] top-level scope at none:0
```

docs/src/lib/representations.md
```diff
@@ -366,6 +367,7 @@ VPolytope
 dim(::VPolytope)
 σ(::AbstractVector{Real}, ::VPolytope{Real})
 vertices_list(::VPolytope)
+intersection(::VPolytope, ::VPolytope)
```
schillic: Currently we cannot create docs for functions inside the `@require` block because we do not load `Polyhedra`, which again is because that package does not build in v0.7 at the moment.
So you have to remove these two lines again.

docs/src/man/concrete_polyhedra.md
```diff
@@ -108,7 +108,7 @@ polyoverapprox(x) = HPolytope(overapproximate(x, 1e-3).constraints)
 
 Epoly = polyoverapprox(E)
 Bpoly = polyoverapprox(B)
-X = intersect(Epoly, Bpoly)
+X = intersecttion(Epoly, Bpoly)
```
schillic: `intersection`

docs/src/lib/representations.md
mforets: ok, got it

PR #602, opened 2018-09-02
Merge recipes for polygons and add polytopes

### LINKED ISSUE #574: Use one plot recipe for polytopic types
We have identical plot recipes for `HPolygon`, `VPolgyon`, and `Zonotope` which could all be unified by using `AbstractPolytope` instead. This would also speed up the plotting for other polytopic set types. And furthermore, this would allow us to plot 2D `VPolytope`s without using the `Polyhedra` library.

mforets:
Closes #574.

### DIFF
--- a/src/plot_recipes.jl
+++ b/src/plot_recipes.jl
@@ -162,22 +162,22 @@ julia> plot([B1, B2], 1e-4);
     end
 end
 
-# =========================
-# Plot recipes for polygons
-# =========================
+# =========================================
+# Plot recipes for polygons or 2D polytopes
+# =========================================
 
 """
-    plot_polygon(P::Union{HPolygon, HPolygonOpt}; ...)
+    plot_polygon(P::Union{AbstractPolygon, HPolytope, VPolytope}; ...)
 
-Plot a polygon in constraint representation.
+Plot a polygon or a 2D polytope.
 
 ### Input
 
-- `P` -- polygon in constraint representation
+- `P` -- polygon or polytope
 
 ### Examples
 
-```jldoctest
+```jldoctest plotting_polygons
 julia> using LazySets, Plots;
 
 julia> P = HPolygon([LinearConstraint([1.0, 0.0], 0.6),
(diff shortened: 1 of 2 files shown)


schillic:
Why not just use `AbstractPolytope`? `vertices_list` should exist for all types.
If we have more efficient implementations, they will be chosen automatically.

schillic:
You also need to adapt the unit tests in [`unit_plot.jl`](https://github.com/JuliaReach/LazySets.jl/blob/master/test/unit_plot.jl).

mforets:
Yeah you're right, i agree to plot every `AbstractPolytope` exactly by default. I think it's more natural to plotting a box overapproximation. If the user wants to plot an overapproximation of the set, `plot_lazyset(S::LazySet, ε::Float64)` can still be used.

mforets:
> You also need to adapt the unit tests in unit_plot.jl.

I had a look, but what is there to change? I wouldn't delete tests even if they are now called from the same recipe.

src/plot_recipes.jl
```diff
@@ -162,22 +162,22 @@ julia> plot([B1, B2], 1e-4);
...
-# =========================================
-# Plot recipes for polygons or 2D polytopes
-# =========================================
+# ==============================
+# Plot recipes for 2D polytopes
+# ==============================
 
 """
-    plot_polygon(P::Union{AbstractPolygon, HPolytope, VPolytope}; ...)
+    plot_polygon(P::Union{AbstractPolytope}; ...)
```
schillic: without `Union` (same below)

schillic:
> Do you agree this closes #574?

Yes.

> >  You also need to adapt the unit tests in unit_plot.jl.
>
> I had a look, but what is there to change? I wouldn't delete tests even if they are now called from the  same recipe. (since for each different type they are testing different code paths)

You may need to replace the `@test_throws X plot(y)` with just `plot(y)`. Just run the tests locally and see if they pass. I am not sure.

PR #604, opened 2018-09-03
defines == generically to work on LazySet pairs

tomerarnon:


mforets:
@tomerarnon Thanks for the contribution. Would you please add a docstring? We've got some [guidelines in this wiki](https://github.com/JuliaReach/LazySets.jl/wiki/Documentation-Guidelines#writing-docstrings-).

tomerarnon:
Actually, here is a definition which I believe works in all cases without the trouble described earlier:

```
function ==(a::LazySet, b::LazySet)
    # If the common supertype of a and b is abstract, they cannot be compared.
    if Base.isabstract(promote_type(typeof(a), typeof(b)))
        return false
    end

    for f in fieldnames(a)
        if getfield(a, f) != getfield(b, f)
            return false
        end
    end
    return true
end

# returns false for:
HPolygon(intconst) == HPolytope(intconst)
HPolygon(intconst) == HPolytope(floatconst)
```

It first checks to see if the common supertype of the two types is abstract, thereby not allowing comparisons between different types. I believe in v1.0 the function is called `isabstracttype`, and an equivalent `isconcretetype` also exists.

mforets:
Hmm I prefer the original option, making it explicit in the new function's signature:

```julia
function ==(a::LazySet{N}, b::LazySet{N}) where {N}
....
end
```
We can add  some approach that covers the case of different `N` if usage arises..

mforets:
> Actually, here is a definition which I believe works in all cases without the trouble described earlier:

Good!

> I believe in v1.0 the function is called isabstracttype,

True. You can use `Compat.isabstracttype`

tomerarnon:
> Good

Does this mean you prefer the newest version?

schillic:
> Hmm I prefer the original option, making it explicit in the new function's signature:
>
>function ==(a::LazySet{N}, b::LazySet{N}) where {N}
>....
>end
>
>We can add some approach at a later stage that covers the case of different N if usage arises..

I agree here.
The new function might return `true` (not checked) for `Ball1([0.], 1.) == Ball2([0.], 1.)` because the field names, field types, and field content are equal.

@tomerarnon: Do you have this use case that you switch between numeric types? Otherwise I would keep your initial function (sorry).

tomerarnon:
> The new function might return true (not checked) for Ball1([0.], 1.) == Ball2([0.], 1.) because the field names, field types, and field content are equal.

It does not. The last version I included returns false if the promotion type of `a` and b` is not a concrete type. This obviated the fieldnames check so I removed it. The docstring I have written for it features this explanation:

> Will not convert `a` or `b` in order to check equivalence of different LazySet types.
> I.e. `a::HPolytope{Int64} == b::HPolytope{Float64}` is a valid comparison, 
> but `a::VPolytope == b::HPolytope` is not, and therefore returns `false` even if `a` and `b`
> represent the same polytope. 

By contrast, the first version (from the issue) would not feature the piece about the L{M} == L{N} being a valid comparison.

> Do you have this use case that you switch between numeric types?

Admittedly, I do not, I just liked the general solution! If you prefer the original one, I can revert to it, and note the restriction in the docstring.

schillic:
>> The new function might return true (not checked) for Ball1([0.], 1.) == Ball2([0.], 1.) because the field names, field types, and field content are equal.
>
>It does not. The last version I included returns false if the promotion type of a and b is not a concrete type.

Indeed, I must have looked at the second version. Then, since the solution is so simple and with no particular overhead, from my side we can go for the new version. A typical Julia user might expect that behavior anyway.
```julia
julia> 1. == 1f0
true

help?> ==
[...] For example, all numeric types are compared by numeric value, ignoring type. [...]
```

schillic:
Actually, we can have both versions. The first version takes precedence if the type parameters are identical (and thus this version is faster), while the second version is the fallback option for the less typical case.

schillic:
> true iff a is equivalent to b. Otherwise false

I would use `equal` instead of `equivalent`.
`Otherwise false` is not needed.

> Will not convert in order to check equivalence of different LazySet types.

This sentence is not clear to me. Maybe:
`The check is purely syntactic and the sets need to have the same base type`.

tomerarnon:
On second thought, one branch or the other compiles away in the == that has the isabstract() check, so actually it should not be slower.
```
julia> @code_warntype Ball1([0.], 1.) == Ball2([0.], 1.)
Variables:
  #temp#@_1 <optimized out>
  #temp#@_2 <optimized out>
  #temp#@_3 <optimized out>

Body:
  begin 
      return $(QuoteNode(false))
  end::Bool
```

```
julia> @code_warntype hs_int == hs_float
Variables:
  #self# <optimized out>
  X::LazySets.HalfSpace{Int64}
  Y::LazySets.HalfSpace{Float64}
  f <optimized out>
  #temp#@_5::Int64
  t::Type{LazySets.HalfSpace{Int64}}
  #temp#@_7::Core.MethodInstance
  #temp#@_8::Bool
  #temp#@_9::Bool

Body:
  begin 
      goto 3 # line 4:
      3:  # line 6:
      $(Expr(:inbounds, false))
      # meta: location REPL[5] equality_helper 2
      # meta: location reflection.jl fieldnames 137
      t::Type{LazySets.HalfSpace{Int64}} = LazySets.HalfSpace{Int64} # line 138:
      goto 12 # line 139:
      12: 
      # meta: pop location
      SSAValue(0) = $(Expr(:invoke, MethodInstance for fieldnames(::DataType), :(Base.fieldnames), :(t)))
      #temp#@_5::Int64 = 1
      16: 
      unless (Base.not_int)((#temp#@_5::Int64 === (Base.add_int)((Base.arraylen)(SSAValue(0))::Int64, 1)::Int64)::Bool)::Bool goto 45
      SSAValue(2) = (Base.arrayref)(SSAValue(0), #temp#@_5::Int64)::Symbol
      SSAValue(3) = (Base.add_int)(#temp#@_5::Int64, 1)::Int64
      #temp#@_5::Int64 = SSAValue(3) # line 3:
      unless ((Main.getfield)(Y::LazySets.HalfSpace{Float64}, SSAValue(2))::Union{AbstractArray{Float64,1}, Float64} isa AbstractArray{Float64,1})::Bool goto 24
      goto 33
      24: 
      unless ((Main.getfield)(Y::LazySets.HalfSpace{Float64}, SSAValue(2))::Union{AbstractArray{Float64,1}, Float64} isa Float64)::Bool goto 31
      unless ((Main.getfield)(X::LazySets.HalfSpace{Int64}, SSAValue(2))::Union{AbstractArray{Int64,1}, Int64} isa Int64)::Bool goto 29
      #temp#@_7::Core.MethodInstance = MethodInstance for !=(::Int64, ::Float64)
      goto 36
      29: 
      goto 33
      31: 
      goto 33
      33: 
      #temp#@_8::Bool = ((Main.getfield)(X::LazySets.HalfSpace{Int64}, SSAValue(2))::Union{AbstractArray{Int64,1}, Int64} != (Main.getfield)(Y::LazySets.HalfSpace{Float64}, SSAValue(2))::Union{AbstractArray{Float64,1}, Float64})::Bool
      goto 38
      36: 
      #temp#@_8::Bool = $(Expr(:invoke, :(#temp#@_7), :(Main.!=), :((Main.getfield)(X, SSAValue(2))::Union{AbstractArray{Int64,1}, Int64}), :((Main.getfield)(Y, SSAValue(2))::Union{AbstractArray{Float64,1}, Float64})))
      38: 
      unless #temp#@_8::Bool goto 43 # line 4:
      #temp#@_9::Bool = false
      goto 48
      43: 
      goto 16
      45:  # line 7:
      #temp#@_9::Bool = true
      48: 
      # meta: pop location
      $(Expr(:inbounds, :pop))
      return #temp#@_9::Bool
  end::Bool
```

schillic:
I would also give a negative example. Also, please use `jldoctest` for the examples.

tomerarnon:
> > true iff a is equivalent to b. Otherwise false
> 
> I would use equal instead of equivalent.
> Otherwise false is not needed.
> 
>> Will not convert in order to check equivalence of different LazySet types.
> 
> This sentence is not clear to me. Maybe:
> The check is purely syntactic and the sets need to have the same base type.

Got it, Made the change.

tomerarnon:
> I would also give a negative example. 

Will do.

> Also, please use jldoctest for the examples.

I am, but it does not show up when I copy-paste to here, interestingly.

You prefer the full version of the docstring though, yes?

schillic:
> You prefer the full version of the docstring though, yes?

Yes, since you already wrote it :wink:

schillic:
> Got it, Made the change.

Did you push the changes already? I do not see them in this PR.
If you have pushed, this might be related to the fact that you work in a fork. The newest comments below the answer [here](https://stackoverflow.com/a/9790530) indicate that this might be a Github bug, but they are quite old already...

tomerarnon:
No, I hadn't yet pushed, I was waiting for your thoughts on the docstring. Now you should see it .

src/LazySet.jl
```diff
@@ -191,12 +191,50 @@ function an_element(S::LazySet{N}) where {N<:Real}
...
+
+julia> HalfSpace([1], 1) == HalfSpace([1.0], 1.0)
+true
+
+julia> Ball1([0.], 1.) == Ball2([0.], 1.)
+false
+```
+"""
+function ==(X::LazySet, Y::LazySet)
+    # if the common supertype of a and b is abstract, they cannot be compared
```
schillic: Sorry, I renamed the variables, but not in the comment :disappointed:

tomerarnon: Ha! I hadn't noticed. If you don't mind I'll change them back to a and b to match the doc

schillic: okay

tomerarnon: Really I was just being lazy 😞 
I prefer X and Y too honestly...
Renamed again.

schillic:
In Julia v0.7:
```julia
Warning: `fieldnames(v)` is deprecated, use `fieldnames(typeof(v))` instead.
```

tomerarnon:
Not a problem. As they say, "if you're going to do something, do it right"!

Just to note again something I alluded to in the original issue, this catch-all definition is not as fast as a function defined on each type. That is primarily because the for-loop creates variables which are Unions over all the field types in `X`, which is a little bit type unstable. So unrolling the loop manually will be faster.

schillic:
@tomerarnon: Feel free to add yourself to [the list of contributors](https://juliareach.github.io/LazySets.jl/latest/about.html#Credits-1).

PR #613, opened 2018-09-08
Add example in Intersection

mforets:


src/Intersection.jl
```diff
@@ -13,6 +13,30 @@ Type that represents the intersection of two convex sets.
 
 - `X` -- convex set
 - `Y` -- convex set
+
+### Examples
+
+Create the intersection of two balls:
```
schillic: circles? balls in the 2 norm?

mforets: thanks, i changed this sentence. 
hmm i'll change again to make sure the difference between lazy and concrete intersection is shown. for `Ball2` the concrete intersection is not available, but i can change to hyperrectanlgles.

PR #627, opened 2018-09-17
#617 - Add function to check for vector permutation

### LINKED ISSUE #617: Add function to check for vector permutation
Something like `ispermutation([1, 2], [2, 1]) == true`.

Reason: Currently we do things like this in the unit tests:
```julia
@test length(vlist) == 4
@test all([vi ∈ vlist for vi in [N[3, 3], N[3, -1], N[-1, -1], N[-1, 3]]])
```
It would be more readable to wrap this in a function.
It suffices to have the function only available in unit tests.

schillic:
Closes #617.

### DIFF
--- a/docs/src/lib/utils.md
+++ b/docs/src/lib/utils.md
@@ -9,6 +9,7 @@ end
 
 ```@docs
 sign_cadlag
+ispermutation
 @neutral
 @absorbing
 @declare_array_version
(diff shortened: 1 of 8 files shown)


schillic:
I added the function to the `LazySets` codebase but did not export it. I am open to change that.
Is it correct to add it to the documentation if it is not exported?

mforets:
>  Is it correct to add it to the documentation if it is not exported?

yes, i think so, even if the function is not in the public API i find it useful to have it in the docs (for searching in the docs or just reading the html docs).

schillic:
Thanks, I fixed it.

PR #636, opened 2018-09-19
#631 - Linear map for AbstractPolygon and AbstractPolytope

### LINKED ISSUE #337: Convert `VPolytope` to `VPolygon`
See also #336.

### LINKED ISSUE #631: Linear map should preserve input set type for polygons
Proposal: the concrete linear map of a `HPolytope` (resp. `VPolytope`), should return a `HPolytope` (resp. `VPolytope`). See [linear_map](https://github.com/JuliaReach/LazySets.jl/blob/master/src/AbstractPolytope.jl#L71), that currently the return type depends on the dimension of the set.

I realized that keeping the input type (concrete polytopes) is convient when writing algorithms because you often want to keep the same family, eg. for intersecting polytopes. Conversion to a polygon can be done outside this function with a simple `convert(HPolygon, X)` (resp `convert(VPolygon, X)`).

The guiding principle is to return objects whose type only depends on the types of the inputs.

### LINKED ISSUE #634: Extend `convert` for AbstractPolygon's
```julia
julia> V = VPolygon([rand(2) for i in 1:5]);

julia> tohrep(V)
LazySets.HPolygon{Float64}(LazySets.HalfSpace{Float64}[LazySets.HalfSpace{Float64}([-0.385579, 0.990618], 0.311256), LazySets.HalfSpace{Float64}([-0.307237, -0.0695505], -0.024366), LazySets.
HalfSpace{Float64}([0.12915, -0.879463], 0.00125324), LazySets.HalfSpace{Float64}([0.563667, -0.0416042], 0.53338)])


julia> convert(HPolygon, V)
ERROR: MethodError: Cannot `convert` an object of type LazySets.VPolygon{Float64} to an object of type LazySets.HPolygon
This may have arisen from a call to the constructor LazySets.HPolygon(...),
```

mforets:
Closes #631.
Closes #634.
Closes #337.


src/AbstractPolygon.jl
```diff
@@ -178,3 +178,9 @@ function <=(u::AbstractVector{N},
             v::AbstractVector{N})::Bool where {N<:AbstractFloat}
     return jump2pi(atan(u[2], u[1])) <= jump2pi(atan(v[2], v[1]))
 end
+
+function linear_map(M::AbstractMatrix, P::T)::T where T<:AbstractPolygon{N} where N
+    @assert dim(P) == size(M, 2)
+    V = VPolygon(broadcast(v -> M * v, vertices_list(P)))
+    return convert(T, V)
+end
```
schillic: new line :smile:

src/AbstractPolytope.jl
```diff
@@ -68,21 +68,9 @@ The polytope in V-representation obtained by applying the linear map ``M`` to
...
-    end
-    vlist = vertices_list(P)
-    new_vlist = Vector{Vector{N}}(undef, length(vlist))
-    @inbounds for (i, vi) in enumerate(vlist)
-        new_vlist[i] =  M * vi
-    end
-    return T(new_vlist)
+function linear_map(M::AbstractMatrix, P::AbstractPolytope{N})::VPolytope{N} where {N<:Real}
+    @assert dim(P) == size(M, 2)
+    return broadcast(v -> M * v, vertices_list(P)) |> VPolytope{N}
```
schillic: Is there a difference between
`broadcast(v -> M * v, vertices_list(P)) |> VPolytope{N}`
and
`VPolytope{N}(broadcast(v -> M * v, vertices_list(P)))` (used for polygons)?
(I do not know `|>`.)
I would stick to one of them for consistency.

schillic: Does the `VPolytope` output type not contradict with the name of this PR? :stuck_out_tongue:

schillic:
Unit tests have to be adapted: The new `AbstractPolytope` method captues all (non-polygon) polytope types, e.g., `Hyperrectangle`s (and always produces a `VPolytope`).

src/AbstractPolytope.jl
mforets: 🔙  How about not changing the given set type by default, but giving the option of return type as a keyword argument?

mforets: > (I do not know |>.)

it is the "pipe" operator that can be used to chaining function calls. i used it here because it seemed more clear in the code to me.

```julia
help?> |>
search: |>

  |>(x, f)

  Applies a function to the preceding argument. This allows for easy function
  chaining.

  julia> [1:5;] |> x->x.^2 |> sum |> inv
  0.01818181818181818
```

schillic: > How about not changing the given set type by default, but giving the option of return type as a keyword argument?

We do not have constructors from vertices for the other types. And some `AbstractPolytope` subtypes are not even closed under linear map (e.g., `BallInf`), so in general we have to change the type.
So I think that we should stick to `VPolytope`. We can still add more specific implementations as usual, like we have for `Zonotope`s.

mforets: > And some AbstractPolytope subtypes are not even closed under linear map (e.g., BallInf), so in general we have to change the type.

good point. i changed the title of this PR 😄 

how about `linear_map` of `AbstractPolygon`? in that case we can add the optional conversion argument, because the subtypes are closed under linear map.

schillic: Okay for me, more work for you :wink:

mforets:
I had to add some new (trivial) conversions between polygons and polytopes. So this also closes #634 and #337.

src/AbstractPolygon.jl
```diff
@@ -179,8 +179,32 @@ function <=(u::AbstractVector{N},
...
+
+### Algorithm
+
+The linear map ``M`` is applied to each vertex of the given set ``P``,
+obtaining a polygon in V-representation. Since polygons are closed under linear
+map, by default ``MP`` is converted to the concrete type of ``P``. If an
+`output_type` is given, the corresponding `convert` method is invoked.
+"""
+function linear_map(M::AbstractMatrix, P::POLYGON;
+                    output_type::Type{<:LazySet}=typeof(P)) where {POLYGON<:AbstractPolygon{N}} where {N}
```
schillic: Either remove the type name `POLYGON` (not used anywhere) or use it for `typeof(P)`.

src/convert.jl
```diff
@@ -28,6 +28,42 @@ function convert(::Type{HPOLYGON1},
...
+### Input
+
+- `VPolygon` -- type used for dispatch
+- `P`        -- polygon in constraint representation
+
+### Output
+
+A polygon in vertex representation.
+"""
+function convert(::Type{VPolygon}, P::HPOLYGON) where {HPOLYGON<:AbstractHPolygon}
```
schillic: Again, `HPOLYGON` is not needed.

src/convert.jl
```diff
@@ -28,6 +28,42 @@ function convert(::Type{HPOLYGON1},
...
+### Input
+
+- `HPOLYGON` -- type used for dispatch
+- `P`        -- polygon in vertex representation
+
+### Output
+
+A polygon in constraint representation.
+"""
+function convert(T::Type{HPOLYGON}, P::VPolygon) where {HPOLYGON<:AbstractHPolygon}
```
schillic: Again, `HPOLYGON` is not needed:
`function convert(T::Type{<:AbstractHPolygon}, P::VPolygon)`

src/convert.jl
```diff
@@ -46,6 +82,110 @@ function convert(::Type{HPolytope}, P::AbstractHPolygon)
...
+### Input
+
+- `type` -- target type
+- `P`    -- source polygon
+
+### Output
+
+The polygon represented as 2D polytope.
+"""
+function convert(::Type{VPolytope}, P::VPolygon)
```
schillic: Should we rather use `P::AbstractPolytope`? Then we can use this function for all polytopic types.

src/convert.jl
```diff
@@ -46,6 +82,110 @@ function convert(::Type{HPolytope}, P::AbstractHPolygon)
...
+### Input
+
+- `type` -- target type
+- `P`    -- source polytope
+
+### Output
+
+The 2D polytope represented as a polygon.
+"""
+function convert(::Type{VPolygon}, P::VPolytope)
```
schillic: Same as above: `P::AbstractPolytope`?

src/convert.jl
```diff
@@ -46,6 +82,110 @@ function convert(::Type{HPolytope}, P::AbstractHPolygon)
...
+### Output
+
+The polygon represented as 2D polytope.
+
+### Algorithm
+
+``P`` is first converted to a polytope in V-representation, then the conversion
+method to a polytope in H-representation is invoked.
+"""
+function convert(::Type{HPolytope}, P::VPolygon)
```
schillic: Same as above?

src/AbstractPolygon.jl
mforets: I didn't understand the comment. The `POLYGON<:AbstractPolygon{N}` is used for dispatch.

src/convert.jl
mforets: Ok, i see what you meant here and in the comment above 👍

src/AbstractPolygon.jl
mforets: The problem is that it should be `P::AbstractPolygon{N}`.

schillic: Yeah, you need ` P::AbstractPolygon{N}` :stuck_out_tongue_winking_eye:

src/convert.jl
mforets: That's a good idea 👍 

I made them to receive `AbstractPolytope` but it started to mess up with unexpected errors because it was trying to use the new functions, instead of using Julia's default no-op. So we have to add the no-op explicitly.

mforets:
Seems like i'm using a wrong function signature for Documenter. @schillic any suggestions?

```julia
!! No docs found for 'linear_map(::AbstractMatrix, P::AbstractPolygon{N}, ::Type{<:LazySet}=typeof(P)) where {N}'. [src/lib/interfaces.md]
```

schillic:
> Seems like i'm using a wrong function signature for Documenter. @schillic any suggestions?

fixed in #665

PR #642, opened 2018-09-20
Fix math in docstring of norm(H::AbstractHyperrectangle, [p]::Real=Inf)::Real

mforets:


mforets:
Note: i can do `α₁` in the REPL, with `\alpha[TAB]\_1[TAB]`, but `\alpha[TAB]\_n[TAB]` does not work; it produces `αₙ` instead of what i would expect `\alpha_n`.

PR #646, opened 2018-09-21
#645 - Concrete intersection for polytopic sets

### LINKED ISSUE #645: Concrete intersection for polytopic sets
```julia
ERROR: MethodError: no method matching
 intersection(::LazySets.HPolytope{Float64}, ::LazySets.VPolygon{Float64})
```

I say we add an inefficient default implementation:
```julia
function intersection(P1::AbstractPolytope{N}, P2::AbstractPolytope{N}) where N
    return intersection(VPolytope(vertices_list(P1)), VPolytope(vertices_list(P2)))
end
```

schillic:
Closes #645.

### DIFF
--- a/src/concrete_intersection.jl
+++ b/src/concrete_intersection.jl
@@ -116,7 +116,7 @@ end
                  P2::Union{HPolytope{N}, VPolytope{N}},
                  [backend]=default_polyhedra_backend(N),
                  [prunefunc]=removehredundancy!,
-                 [output_type]=default_intersection_output_type(P1, P2)) where {N}
+                 [output_type]=default_intersection_output_type(P1, P2)) where N
 
 Compute the intersection of two polytopes in either H-representation or
 V-representation.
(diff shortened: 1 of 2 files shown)


src/concrete_intersection.jl
```diff
@@ -154,5 +154,35 @@ function intersection(P1::Union{HPolytope{N}, VPolytope{N}},
...
+Compute the intersection of two polytopic sets.
+
+### Input
+
+- `P1`          -- polytope
+- `P2`          -- another polytope
+
+### Output
+
+The polytope obtained by the intersection of `P1` and `P2`, in V-representation.
```
mforets: shall we add something in the lines of: "This is a fallback implementation that requires `Polyhedra` to evaluate the concrete intersection. In the general case, the argument sets are converted to `VPolytope` through the `vertices_list` function."

schillic:
@mforets: I rebased (also with `master` for the v0.7 problem). Can you check again?

schillic:
Ah, the problem is that the docs are only available with `Polyhedra`. I will remove them again.

mforets:
> I will remove them again.

it's a pity that we cannot "comment" those lines. so that these undocumented functions are easy to locate for inclusion in the docs afterwards.

PR #651, opened 2018-09-22
#583 - Concrete intersection of two HPolygons

### LINKED ISSUE #583: Concrete intersection of two HPolygons
This is really simple: Just take the constraints of both polygons together.
Since the constraints are already sorted, they can be inserted efficiently such that the result is also sorted.

In #582 we introduce functionality to reduce the number of constraints in a second step.

schillic:
Closes #583.

### DIFF
--- a/src/concrete_intersection.jl
+++ b/src/concrete_intersection.jl
@@ -100,6 +100,110 @@ function intersection(H1::AbstractHyperrectangle{N},
     return Hyperrectangle(high=high, low=low)
 end
 
+"""
+    intersection(P1::AbstractHPolygon{N},
+                 P2::AbstractHPolygon{N}
+                )::Union{HPolygon{N}, EmptySet{N}} where N<:Real
+
+Return the intersection of two polygons in constraint representation.
+
+### Input
+
+- `P1` -- first polygon
+- `P2` -- second polygon
+
+### Output
+
+If the polygons do not intersect, the result is the empty set.
+Otherwise the result is the polygon that describes the intersection.
+
+### Algorithm
+
+We just combine the constraints of both polygons.
+To obtain a linear-time algorithm, we interleave the constraints.
+If there are two constraints with the same normal vector, we choose the tighter
+one.
+"""
+function intersection(P1::AbstractHPolygon{N},
+                      P2::AbstractHPolygon{N}
+                     )::Union{HPolygon{N}, EmptySet{N}} where N<:Real
+    @inline function choose_first_diff_dir(i, i1, i2, c1, c2)
+        c[i] = c1[i1]
+        if i1 == length(c1)
+            c[i+1:length(c)] = c2[i2:length(c2)]
+            return true
+        end
+        return false
+    end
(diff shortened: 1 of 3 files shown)


mforets:
I would expect that `addconstraint!` takes care of adding the constraints properly. And it provides two methods, for search, linear and binary, (eventually can use a specialized `addconstraint!` based on a specific concrete `AbstractHPolygon`, but i don't have an example right now).

```julia
function intersection(...)
    P = HPolygon{N}()
    addconstraint!.(P, constraints_list(P1))
    addconstraint!.(P, constraints_list(P2))
    return P
end
```

But the repeated halfspaces is something to be fixed in `addconstraint!`. Or we add a method to remove the redundant constraints. For example:

```julia
julia> h = convert(HPolygon, BallInf(ones(2), 1.0));

julia> constraints_list(h)
4-element Array{LazySets.HalfSpace{Float64},1}:
 LazySets.HalfSpace{Float64}([1.0, 0.0], 2.0)   
 LazySets.HalfSpace{Float64}([0.0, 1.0], 2.0)   
 LazySets.HalfSpace{Float64}([-1.0, -0.0], -0.0)
 LazySets.HalfSpace{Float64}([-0.0, -1.0], -0.0)

julia> P = HPolygon{Float64}()
LazySets.HPolygon{Float64}(LazySets.HalfSpace{Float64}[])

julia> addconstraint!.(P, [constraints_list(h); constraints_list(h)]);

julia> constraints_list(P)
julia> constraints_list(P)
8-element Array{LazySets.HalfSpace{Float64},1}:
 LazySets.HalfSpace{Float64}([1.0, 0.0], 2.0)   
 LazySets.HalfSpace{Float64}([1.0, 0.0], 2.0)   
 LazySets.HalfSpace{Float64}([0.0, 1.0], 2.0)   
 LazySets.HalfSpace{Float64}([0.0, 1.0], 2.0)   
 LazySets.HalfSpace{Float64}([-1.0, -0.0], -0.0)
 LazySets.HalfSpace{Float64}([-1.0, -0.0], -0.0)
 LazySets.HalfSpace{Float64}([-0.0, -1.0], -0.0)
 LazySets.HalfSpace{Float64}([-0.0, -1.0], -0.0)
```

schillic:
The thing is that adding constraints all at once gives a worst-case quadratic algorithm. Interleaved insertion is worst-case linear.

src/concrete_intersection.jl
```diff
@@ -100,6 +100,109 @@ function intersection(H1::AbstractHyperrectangle{N},
...
+- `P2` -- second polygon
+
+### Output
+
+If the polygons do not intersect, the result is the empty set.
+Otherwise the result is the polygon that describes the intersection.
+
+### Algorithm
+
+We just combine the constraints of both polygons.
```
mforets: Maybe extend a bit: "The thing is that adding constraints all at once gives a worst-case quadratic algorithm. Interleaved insertion is worst-case linear."

PR #655, opened 2018-09-24
Add alias 'isdisjoint' for 'is_intersection_empty'

schillic:


mforets:
I would rather expect `is_disjoint` (or `isdisjoint`), becuse of the boolean output.
But i agree to keep `disjoint`, since we don't have another meaning for it.

schillic:
> I would rather expect is_disjoint (or isdisjoint), becuse of the boolean output.

On second thought, I will change it. Two sets *are* disjoint, but the first set *is* disjoint from the second set :+1:

PR #657, opened 2018-09-24
#656 - Fallback implementation for commutative binary functions

### LINKED ISSUE #656: Fallback implementation for commutative binary functions
For commutative binary functions (currently `intersection` and `is_intersection_empty`), we can add a fallback definition that tries to swap the arguments. If that fails, we can still print a nicer error message.

schillic:
Closes #656.

### DIFF
--- a/src/concrete_intersection.jl
+++ b/src/concrete_intersection.jl
@@ -2,6 +2,64 @@
 
 export intersection
 
+
+# --- fallback implementation that swaps arguments ---
+
+
+global FALLBACK_CACHE_intersection = Vector{Tuple{LazySet, LazySet}}()
+
+
+"""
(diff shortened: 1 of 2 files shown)


schillic:
We could also remove the two symmetric definitions of `is_intersection_empty`, but I am okay with keeping them because the new function adds a little overhead.

schillic:
The downside of having this is that `hasmethod(is_intersection_empty, (X, Y))` always returns `true`.

mforets:
I think it would be nice to have  `@symmetric` macro applicable to functions such as `intersection` or any other function where the arguments can be swapped without changing the meaning.That's more or less the spirit in this PR but it is done for a specific class of functions.

mforets:
This issue is still relevant? Should we close it? cc @schillic 

Swapping the arguments should be handled by the implementation (as it is already the case); if it is missing that's a flaw in the implementation.

PR #659, opened 2018-09-24
#658 - Check for intersection of a compact set and a HalfSpace

### LINKED ISSUE #658: Check for intersection of a compact set and a HalfSpace
This is a special case of #641.

EDIT: The previous version did not work, I fixed it.
The idea is to compute the support function in the opposite direction of the `HalfSpace`'s normal vector and then compare to the `HalfSpace`'s constraint.

schillic:
Closes #658.

### DIFF
--- a/src/is_intersection_empty.jl
+++ b/src/is_intersection_empty.jl
@@ -680,6 +680,55 @@ function is_intersection_empty(hp::Union{Hyperplane{N}, Line{N}},
     return is_intersection_empty(X, hp, witness)
 end
 
+# --- HalfSpace ---
+
+"""
+    is_intersection_empty(X::LazySet{N},
+                          hs::HalfSpace,
+                          [witness]::Bool=false
+                         )::Union{Bool, Tuple{Bool,Vector{N}}} where N<:Real
+
+Check whether a compact set an a half-space do not intersect, and otherwise
+optionally compute a witness.
+
+### Input
+
+- `X`       -- compact set
+- `hs`      -- half-space
+- `witness` -- (optional, default: `false`) compute a witness if activated
+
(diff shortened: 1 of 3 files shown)


src/is_intersection_empty.jl
```diff
@@ -680,6 +680,55 @@ function is_intersection_empty(hp::Union{Hyperplane{N}, Line{N}},
...
+### Notes
+
+We assume that `X` is compact.
+Otherwise, the support vector queries may fail.
+
+### Algorithm
+
+A compact convex set intersects with a half-space iff the support vector in
+the negative direction of the half-space's normal vector ``a`` is contained in
+the half-space: ``σ(-a) ∈ hs``.
```
mforets: I disagree. Take `B = Ball2(zeros(2), 1.0)` and `H = HalfSpace([1.0, 0.0], 1.0) # x <= 0`. Then `is_intersection_empty(B, H)` should be `true` but it is returning `false`.

I think the condition is:  (`is_intersection_empty` is `true`) iff ``\rho_X(n) - b >= 0``.

schillic: You are right, the condition should be *every support vector* in general, so my implementation does not work. Yours does.

> Then is_intersection_empty(B, H) should be true

No, why? The constraint says x <= 1 and the ball is completely contained in that set, so the intersection is not empty.

schillic: > You are right, the condition should be every support vector in general, so my implementation does not work.

I take that back. For half-spaces, if *one* support vector is inside, *all* of them are (because they all lie on a hyperplane, which must be beyond the half-space's hyperplane.

PR #668, opened 2018-09-25
Fix 'eye' export for v1.0

schillic:


mforets:
ah, i see. before we didn't need this export because it was at `Base`. we can also `export exp` directly in `ExponentialMap.jl`, it shouldn't break < 1.0 versions.

mforets:
> Okay, and a v1.0 version?

yes, but i think it goes after we drop the support for 0.6 version.

schillic:
> ah, i see. before we didn't need this export because it was at Base. we can also export exp directly in ExponentialMap.jl, it shouldn't break < 1.0 versions.

Hm, sorry, I thought you were okay with merging.

schillic:
Should we drop the v0.7 build then? If it is just more generous than v1.0...
The reason why I propose this is that we do not get unlimited build servers. Sometimes they only start with two. So the total build time might be shorter if we use fewer versions.

mforets:
i still find the 0.7 build in travis helpful to know how to fix the errors (the deprecation warnings), because i still use v0.6.4 for development .

PR #670, opened 2018-09-25
#667 - vertices_list function for lazy operations

### LINKED ISSUE #667: vertices_list function for lazy operations
For some lazy operations the implementation is rather simple by relying on the implementation of `vertices_list` for the inner set types.

* `CartesianProduct(Array)`: Take the Cartesian product of the vertices.
* `LinearMap`/`ExponentialMap`: Apply the map to each vertex.

For the other operations, this is complicated, though.

* `MinkowskiSumArray`: The vertices are sums of the old vertices. However, this is only sufficient but not necessary (see [this picture](https://upload.wikimedia.org/wikipedia/commons/e/e2/Minkowski-sumex4.svg)). There is [an algorithm based on linear programming](https://arxiv.org/abs/1412.2564) and [one based on dual cones](https://arxiv.org/abs/1412.2562).
* `ConvexHull`: ?
* `Intersection`: ?

schillic:
Closes #667.

### DIFF
--- a/src/CartesianProduct.jl
+++ b/src/CartesianProduct.jl
@@ -308,3 +308,73 @@ function ∈(x::AbstractVector{N}, cpa::CartesianProductArray{N, <:LazySet{N}}
     end
     return true
 end
+
+"""
+    vertices_list(cpa::CartesianProductArray{N})::Vector{Vector{N}} where N<:Real
+
+Return the list of vertices of a (polytopic) Cartesian product.
+
+### Input
+
+- `cpa` -- Cartesian product
+
+### Output
+
+A list of vertices.
+
+### Algorithm
+
+We assume that the underlying sets are polytopic.
+Then the high-dimensional set of vertices is just the Cartesian product of the
+low-dimensional sets of vertices.
+"""
+function vertices_list(cpa::CartesianProductArray{N})::Vector{Vector{N}} where N<:Real
+    # collect low-dimensional vertices lists
+    vlist_low = [vertices_list(X) for X in array(cpa)]
+
+    # create high-dimensional vertices list
+    m = 1
+    for vl in vlist_low
+        m *= length(vl)
+    end
+    vlist = Vector{Vector{N}}(undef, m)
+    indices = ones(Int, length(vlist_low))
+    indices_max = [length(vl) for vl in vlist_low]
+    v = zeros(N, dim(cpa))
+    j = 1
+    for vl in vlist_low
+        v_low = vl[1]
+        v[j:j+length(v_low)-1] = v_low
+        j += 1
+    end
+    i = 1
+    j = 1
+    # iterate through all index combinations
+    while true
+        indices[1] = 0
+        while indices[1] < indices_max[1]
+            indices[1] += 1
+            v_low = vlist_low[1][indices[1]]
+            v[1:length(v_low)] = v_low
+            vlist[i] = copy(v)
+            i += 1
+        end
+        if i > m
+            break
+        end
+        j = 1
+        dim_start_j = 1
+        while indices[j] == indices_max[j]
+            indices[j] = 1
+            v_low = vlist_low[j][1]
+            v[dim_start_j:dim_start_j+length(v_low)-1] = v_low
+            dim_start_j += length(v_low)
+            j += 1
+        end
+        indices[j] += 1
+        v_low = vlist_low[j][indices[j]]
+        v[dim_start_j:dim_start_j+length(v_low)-1] = v_low
+    end
+
+    return vlist
+end
--- a/src/CartesianProduct.jl
+++ b/src/CartesianProduct.jl
@@ -133,6 +133,42 @@ function ∈(x::AbstractVector{<:Real}, cp::CartesianProduct)::Bool
            ∈(view(x, dim(cp.X)+1:length(x)), cp.Y)
 end
 
+"""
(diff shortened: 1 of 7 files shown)


mforets:
> We assume that the underlying sets are polytopic.

This is something that we can guarantee using dispatch on `AbstractPolytope`. Is there a reason not to use it?

schillic:
Well, what about `LinearMap` of a `BallInf`, etc.?

schillic:
Argh, nondeterminism in the builds... I'll fix the error later.

mforets:
`LinearMap(::AbstractMatrix, ::BallInf)` is a polytope  too

mforets:
what if instead of `vertices_list(lm::LinearMap{N})::Vector{Vector{N}} where N<:Real`  we use `vertices_list(lm::LinearMap{N, S<:AbstractPolytope, NM, MAT})::Vector{Vector{N}} where {N, NM, MAT}`

schillic:
I feared that you might say this :smile:
Then what about `LinearMap{::AbstractMatrix, ::LinearMap{::AbstractMatrix, ::AbstractPolytope}}`?

mforets:
hehe this one is automatically merged into a single `LinearMap` !

schillic:
Grrr... `LinearMap{::AbstractMatrix, ::MinkowskiSum{::BallInf, ::BallInf}}`?

mforets:
Consider:

```julia
vertices_list(lm::LinearMap{N}) where {N} = vertices_list(linear_map(L.m, L.X))
```

We rely on the calculation of the concrete linear map, which could be optimized given the set type of `L.X` and `linear_map` returns a `VPolytope` => `vertices_list(..)` is its field.

schillic:
1. We do not support `linear_map` for many types.
2. I do not think that `vertices_list(linear_map(L.M, L.X))` is faster than `L.M * vertices_list(L.X)` for any type for `X`.

src/CartesianProduct.jl
```diff
@@ -272,3 +308,73 @@ function ∈(x::AbstractVector{N}, cpa::CartesianProductArray{N, <:LazySet{N}}
...
+"""
+function vertices_list(cpa::CartesianProductArray{N})::Vector{Vector{N}} where N<:Real
+    # collect low-dimensional vertices lists
+    vlist_low = [vertices_list(X) for X in array(cpa)]
+
+    # create high-dimensional vertices list
+    m = 1
+    for vl in vlist_low
+        m *= length(vl)
+    end
```
mforets: alternate: use `prod(length.(v))` as in 

```julia
julia> prod([1,2,3])
6
```

src/CartesianProduct.jl
```diff
@@ -272,3 +308,73 @@ function ∈(x::AbstractVector{N}, cpa::CartesianProductArray{N, <:LazySet{N}}
...
+    vlist = Vector{Vector{N}}(undef, m)
+    indices = ones(Int, length(vlist_low))
+    indices_max = [length(vl) for vl in vlist_low]
```
mforets: since we need `indices_max` below, we can move its definition upwards and let `m = prod(indices_max)`.

schillic: Actually that code is super ugly, but I could not get `Iterators.product` to work.

src/CartesianProduct.jl
```diff
@@ -272,3 +308,73 @@ function ∈(x::AbstractVector{N}, cpa::CartesianProductArray{N, <:LazySet{N}}
...
+    v = zeros(N, dim(cpa))
+    j = 1
+    for vl in vlist_low
+        v_low = vl[1]
+        v[j:j+length(v_low)-1] = v_low
+        j += 1
```
mforets: i don't get this one. is `j += 1` correct?

schillic: `j` is the block's index. I walk through all blocks until I find a block where `indices[j] < indices_max[j]`.

schillic: Fixed!

src/CartesianProduct.jl
```diff
@@ -272,3 +308,70 @@ function ∈(x::AbstractVector{N}, cpa::CartesianProductArray{N, <:LazySet{N}}
...
+    m = prod(indices_max)
+    vlist = Vector{Vector{N}}(undef, m)
+    indices = ones(Int, length(vlist_low))
+    v = zeros(N, dim(cpa))
+    dim_start_j = 1
+    for vl in vlist_low
+        v_low = vl[1]
+        v[dim_start_j:dim_start_j+length(v_low)-1] = v_low
+        dim_start_j += length(v_low)
+    end
```
mforets: did you forget an assignment to `vlist` here? since below the array `v` is modified before writing in `vlist`.

schillic: No, this loop only writes the first full-dimensional vector `v`. The code below starts by overwriting the first block with the same values, true. I had written to `vlist` in an earlier version and started at `i = 2` below, but then I removed this part to keep the code shorter.

mforets: Ok. Alternative: `v = vcat([vi[1] for vi in vlist_low]...)`.

schillic:
I do not fully understand. Can you do the change? I will not have time the next days.

mforets:
although it is does the same calculation, i expect the `@nloops` version to have less allocations. yes, it is worth an explanation and docstring. we can write the new version on a separate PR.

PR #671, opened 2018-09-25
Add v1.0 build for Travis

schillic:


mforets:
for now i think we should add it to admitted failures in 1.0. 

because we haven't yet completely transitioned to 1.0 (eg. the interaction with Polyhedra, i had some issues building docs.)

schillic:
But if it works?

mforets:
i haven't transitioned yet to a development workflow in 1.0, because some packages that i use and that we use for in LazySets (or Reachability) haven't transitioned.

schillic:
But you do not have to change anything. Travis runs the build for us.

PR #677, opened 2018-09-26
#375 - Intersection of a halfspace with polyhedron: line search method

### LINKED ISSUE #375: Intersection of a halfspace with a polyhedron
http://spaceex.imag.fr/sites/default/files/frehser_adhs2012.pdf
https://sites.google.com/site/frehseg/publications/Frehse_Habil16.pdf

cc: @kostakoida

mforets:
Closes #375.


mforets:
Some examples:

```julia
using LazySets, Optim, Compat
p = Ball2(zeros(2), 1.0)
h = HalfSpace([1.0, 1.0], 0.0) # x + y <= 0

plot(p, aspectratio=1, alpha=0.2, 1e-3)
plot!(p ∩ h, 1e-3, aspectratio=1, alpha=0.2, color="red")
```

screen shot 2018-09-27 at 11 53 02

mforets:
This one seems wrong, but i don't know why:

```julia
p = VPolygon([rand(2) for i in 1:8])
plot(p, aspectratio=1, alpha=0.2, 1e-3)
h = HalfSpace([1.0, 0.0], 0.5) # x <= 0.5
plot!(p ∩ h, 1e-3, aspectratio=1, alpha=0.2, color="red")
```

screen shot 2018-09-27 at 12 05 09

schillic:
How come that `σ` is defined for you? I can only plot the box approximation.

mforets:
sorry, i forgot to push that part. just a second.

mforets:
do you agree about my definition of `σ`?

mforets:
Hmm it is not a problem with `VPolygon`. Here is a similar error (i only get the overapproximation) using a `Ball2`:

```julia
p = Ball2(ones(2), 1.0)
plot(p, aspectratio=1, alpha=0.2, 1e-3)
h = HalfSpace([1.0, 0.0], 1.0) # x <= 1
plot!(p ∩ h, 1e-3, aspectratio=1, alpha=0.2, color="red")
```
screen shot 2018-09-27 at 13 04 01

mforets:
Here i get

```julia
julia> σ([1., 1], p∩h)
2-element Array{Float64,1}:
 1.5
 1.5
```

but i think it should be `[1.0, 2.0]`.

mforets:
ok the problem is that `σ`  does not work like `ρ(ℓvec, cap) * ℓvec` in general, since this condition is not sufficient for the point to belong to the set `p`. ...

for instance to test that the support function is correct, consider the same example but using template directions (which evaluates the supp func) instead of iterative refinement  (which evaluates the supp vector):

```julia
using LazySets.Approximations

p = Ball2(ones(2), 1.0)
h = HalfSpace([1.0, 0.0], 1.0) # x <= 1
pint = overapproximate(p∩h, BoxDiagDirections(2))
pint = convert(HPolygon, pint)

plot(p, aspectratio=1, alpha=0.2, 1e-3)
plot!(pint, aspectratio=1, alpha=0.2, color="red")
```

screen shot 2018-09-27 at 14 07 34

schillic:
> ok the problem is that σ does not work like ρ(ℓvec, cap) * ℓvec in general, since this condition is not sufficient for the point to belong to the set p. ...

Yes, this does not work in general.
```julia
julia> b = BallInf([0., 0.], 1.); d = [1., 0.5];

julia> ρ(d, b)
1.5

julia> ρ(d, b) * d
2-element Array{Float64,1}:
 1.5 
 0.75

julia> ρ(d, b) * d ∈ b
false

julia> σ(d, b)
2-element Array{Float64,1}:
 1.0
 1.0
```

schillic:
> Where do i get the number

I look it up in `.julia/environments/v0.7`.

schillic:
Do we use "ℓ" for directions now? (I am not against that choice.)

I would also add an option to skip the emptiness check. In practice we want to do that in advance before computing ρ. The name of that option is tricky: A possible name `is_known_to_be_nonempty` is a bit overkill. Maybe `skip_emptiness_check`?

mforets:
> Do we use "ℓ" for directions now? (I am not against that choice.)

i was using it because i was following the notation from the paper. but i shall switch back to `d`.

> I would also add an option to skip the emptiness check.

good idea. maybe `emptiness_check`.

src/Intersection.jl
```diff
@@ -287,3 +287,182 @@ function ∈(x::AbstractVector{N}, ia::IntersectionArray{N})::Bool where {N<:Rea
...
+    # Recover results
+    fmin, λmin = sol.minimum, sol.minimizer
+    return (fmin, λmin)
+end # _line_search
+end # quote
+end # load_optim
+
+# Symmetric case
+ρ(ℓ::AbstractVector{N}, cap::Intersection{N, <:HalfSpace, <:LazySet};
+  algorithm::String="line_search", kwargs...) where {N<:AbstractFloat} = ρ(ℓ, cap.Y ∩ cap.X; algorithm=algorithm, kwargs...)
```
schillic: The `check_intersection` argument is missing here.

test/REQUIRE
```diff
@@ -0,0 +1,9 @@
+julia 0.6
+Compat 1.0
+Expokit 0.1
+GLPKMathProgInterface 0.4
+IntervalArithmetic 0.14
+MathProgBase 0.7
+RecipesBase 0.3
+Requires 0.4
+Optim 0.14.1
```
schillic: new line

test/unit_Intersection.jl
```diff
@@ -53,3 +53,16 @@ for N in [Float64, Rational{Int}, Float32]
...
+# ======================
+
+# halfspace - ball1 intersection
+X = Ball1(zeros(2), 1.0);
+H = HalfSpace([-1.0, 0.0], -1.0); # x >= 0 
+d = normalize([1.0, 0.0])
+
+# line search using Optim
+using Optim
+@test ρ(d, X ∩ H) == ρ(d, X ∩ H) == 1.0
```
schillic: Should the second expression be the symmetric case?

mforets:
Thanks for the comments. I put them in https://github.com/JuliaReach/LazySets.jl/pull/681

PR #680, opened 2018-09-30
Fix broken unit tests

### LINKED ISSUE #71: Failing doctests should fail the build

schillic:
EDIT: Does *not* close the issue #71.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ git:
 ## allow failures (tests will run but not make your overall status red)
 matrix:
   allow_failures:
-    - julia: 1.0
     - julia: nightly
 
 ## uncomment and modify the following lines to manually install system packages
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ git:
 ## allow failures (tests will run but not make your overall status red)
 matrix:
   allow_failures:
+    - julia: 1.0
     - julia: nightly
 
 ## uncomment and modify the following lines to manually install system packages
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -45,7 +45,7 @@ Every `LazySet` type must define a function `σ` to compute the support vector.
 
 ```@docs
 support_vector
-ρ(::AbstractVector{Real}, ::LazySet{Real})
+ρ(::AbstractVector{N}, ::LazySet{N}) where {N<:Real}
 support_function
 ```
 
--- a/docs/src/lib/methods_fix.md
+++ b/docs/src/lib/methods_fix.md
@@ -35,6 +35,7 @@ vertices_list
 +
 <=
 σ
+ρ
 ∈
 ⊆
 ```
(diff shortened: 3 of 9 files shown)


schillic:
I vote for activating the other unit tests in the builds. This will take a bit longer, but we will notice when problems are introduced.
`Polyhedra` tests are only available in v0.6 at the moment (that package is also needed for plotting), so I propose the following settings.
```julia
# old, unchanged
global test_suite_basic = true
global test_suite_doctests = VERSION >= v"0.7-" # only run doctests with new Julia version
# new
global test_suite_polyhedra = VERSION < v"0.7-" # only run doctests with old Julia version (for now)
global test_suite_plotting = VERSION < v"0.7-" # only run doctests with old Julia version (for now)
```

mforets:
> I vote for activating the other unit tests in the builds. This will take a bit longer, but we will notice when problems are introduced.

I agree too. Once `Polyhedra` tests pass in 0.7 too we can switch all default tests to `true`.

schillic:
I had to undo the commit for strict doctests because `Optim` has many undocumented functions and for some reason `Documenter` complains about them (although we explicitly say
`modules = Module[LazySets, Approximations]`).

mforets:
Does `strict=true` imply that if we have an undocumented function the build breaks? (If yes, i think that's "too strict".)

schillic:
> Does strict=true imply that if we have an undocumented function the build breaks? (If yes, i think that's "too strict".)

Unfortunately, that is the case (and there is no fine-tuning option ).

mforets:
> Unfortunately, that is the case (and there is no fine-tuning option ).

Alright. Documented [here](https://github.com/JuliaDocs/Documenter.jl/blob/bfa7b6b12581fad302e1521ee82615577646a4a4/src/Documenter.jl#L163), i'd like `strict=true` *but* it doesn't break if `checkdocs` fails. But it seems like if `strict=true` then `checkdocs` option is ignored?

mforets:
i suspect (but i'm not 100% sure) that the failure with the v1.0 is the new restriction of the package manager to refuse installing packages which are known not to work in > 0.7. the error message says

```julia
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package CDDLib [3391f64e]:
 CDDLib [3391f64e] log:
 ├─possible versions are: [0.0.1, 0.1.0-0.1.4, 0.2.0-0.2.3, 0.3.0-0.3.2] or uninstalled
 ├─restricted to versions 0.3.2-* by an explicit requirement, leaving only versions 0.3.2
 └─restricted by julia compatibility requirements to versions: uninstalled — no versions left
```

mforets:
finally

screen shot 2018-09-30 at 23 08 06

let's temporarily allow failure in the 1.0 build until the issue with the dependency is fixed.

schillic:
I would rather remove `CDDLib`/`Polyhedra` from the `test/REQUIRE` file because
1. we do not use these packages in the newer versions, and
2. for the v0.6 build we need to write the `Pkg.add` commands explicitly.

PR #692, opened 2018-10-03
#690 - Fix T parameter and remove deprecated imports

### LINKED ISSUE #690: Adapt `HRep{T, N}`
Polyhedra#master the H- (resp V-) representations are not parametrized anymore in the set dimension. We should adapt to this change in the constructors `HPolytope` and `VPolytope`.

mforets:
Closes #690.

### DIFF
--- a/src/AbstractPolytope.jl
+++ b/src/AbstractPolytope.jl
@@ -117,7 +117,7 @@ function load_polyhedra_abstractpolytope() # function to be loaded by Requires
 return quote
 
 import CDDLib # default backend
-import Polyhedra:polyhedron, SimpleHRepresentation, SimpleHRepresentation,
+import Polyhedra:polyhedron,
                  HRep, VRep,
                  removehredundancy!, removevredundancy!,
                  hreps, vreps,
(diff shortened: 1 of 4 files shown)


mforets:
It fails in v0.6 because this changeset is for the new version of Polyhedra, which does no longer support v0.6. 

What do we want to do now? Shall we keep this change for later when we drop the support for 0.6 in LazySets? I don't like the idea of keep patching with if-else-version for each new udpate in other packages.

schillic:
Okay for waiting. Maybe add a milestone and add this issue/PR to it.

mforets:
As a quick and momentary fix we can special case this function based on the julia version.

mforets:
i keep getting this issue (unrelated to this PR)

```julia
julia> using LazySets

julia> using Polyhedra
┌ Warning: Package LazySets does not have CDDLib in its dependencies:
│   added CDDLib as a dependency but haven't updated your primary
└ Loading CDDLib into LazySets from project dependency, future warnings for LazySets are suppressed.
WARNING: could not import CDDLib.CDDLibrary into LazySets
```

PR #713, opened 2018-10-05
#712 - Add intersection of set and polytope

### LINKED ISSUE #712: Check for empty intersection between lazyset and polytope in constraint representation
In https://github.com/JuliaReach/LazySets.jl/issues/658 we added an efficient intersection check between any lazyset and a halfspace. By extension we can add 

```julia
function is_intersection_empty(X::LazySet{N},
                               P::Union{HPolytope{N}, AbstractHPolygon{N}},
                               witness::Bool=false
                              )::Union{Bool, Tuple{Bool,Vector{N}}} where N<:Real
... # call is_intersection_empty(X, Hi) for each Hi in P
end
```
This algorithm will take `m` evaluations of the support vector of `P`.

Actually this method can be used for any lazysets whose `constraints_list` is known.

mforets:
Closes #712.

### DIFF
--- a/src/is_intersection_empty.jl
+++ b/src/is_intersection_empty.jl
@@ -764,7 +764,7 @@ The algorithm relies on the intersection check between the set ``X`` and each co
 in ``P``. It costs ``m`` support vector evaluations of ``X``, where ``m`` is
 the number of constraints in ``P``.
 
-Note that method can be used with any set ``P`` whose constraints are known.
+Note that this method can be used with any set ``P`` whose constraints are known.
 """
 function is_intersection_empty(X::LazySet{N},
                                P::Union{HPolytope{N}, AbstractHPolygon{N}}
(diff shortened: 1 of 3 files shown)


src/is_intersection_empty.jl
```diff
@@ -737,6 +737,54 @@ function is_intersection_empty(hs::HalfSpace{N},
...
+We assume that `X` is compact. Otherwise, the support vector queries may fail.
+Witness production is not supported.
+
+### Algorithm
+
+The algorithm relies on the intersection check between the set ``X`` and each constraint
+in ``P``. It costs ``m`` support vector evaluations of ``X``, where ``m`` is
+the number of constraints in ``P``.
+
+Note that method can be used with any set ``P`` whose constraints are known.
```
schillic: Note that *this*

PR #714, opened 2018-10-05
Simplify is_intersection_empty for HalfSpace w/o witness production

schillic:


mforets:
i fixed the sign in the comparison (this is equivalent to the supp vec or see part (1) of Lemma 3 [in Frehse&Ray](https://drive.google.com/file/d/0By5Nhm0s_V1cazE2SGZhRXRybDg/view)).

also made a cosmetic deletion of parentheses .

schillic:
I still do not understand the missing step:
```math
  H ∩ X = ∅
⇔ σ(-a, X) ∉ H            | our old def.
⇔ dot(σ(-a, X), a) > b    | def. ∉ H
⇔ -dot(σ(-a, X), -a) > b  | ?
⇔ -ρ(-a, X) > b           | def. ρ
```

PR #723, opened 2018-10-06
#722 - Add upper bound option to intersection

### LINKED ISSUE #722: Add `upper_bound` option to intersection
In https://github.com/JuliaReach/LazySets.jl/pull/721 we added `ρ_upper_bound` in the approximations module.

In other parts of the library eg. `is_intersection_empty` between a set and a halfspace, `ρ` is used. Instead of adding `is_intersection_empty_upper_bound`, we can have an optional keyword argument (eg. `upper_bound`) in the exact `ρ`, for this particular case of intersection, such that 

```julia
function ρ(d::AbstractVector{<:Real}, cap::Intersection, upper_bound=true)::Vector{<:Real} 
   if upper_bound 
        ρ_upper_bound(d, cap)
    else 
        error("the exact support function of an intersection is not implemented yet")
  end 
end
```

We would have to extend the method `is_intersection_empty` to accept keyword arguments that are passed through.

mforets:
Closes #722.

### DIFF
--- a/src/is_intersection_empty.jl
+++ b/src/is_intersection_empty.jl
@@ -717,7 +717,7 @@ the half-space: ``σ(-a) ∈ hs``.
 The support vector is thus also a witness.
 
 Optional keyword arguments can be passed to the `ρ` function. In particular, if
-`X` is an lazy intersection, the option `upper_bound=true` may be useful.
+`X` is a lazy intersection, the option `upper_bound=true` may be useful.
 """
 function is_intersection_empty(X::LazySet{N},
                                hs::HalfSpace{N},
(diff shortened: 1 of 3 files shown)


src/is_intersection_empty.jl
```diff
@@ -715,13 +715,17 @@ A compact convex set intersects with a half-space iff the support vector in
 the negative direction of the half-space's normal vector ``a`` is contained in
 the half-space: ``σ(-a) ∈ hs``.
 The support vector is thus also a witness.
+
+Optional keyword arguments can be passed to the `ρ` function. In particular, if
+`X` is an lazy intersection, the option `upper_bound=true` may be useful.
```
schillic: "a lazy"

mforets:
> Why did you add the kwargs only for one method?

🤔  the other method that could use support function (but now uses support vector so i didn't touch it now) is the intersection with a hyperplane

```julia
function is_intersection_empty(X::LazySet{N},
                               hp::Union{Hyperplane{N}, Line{N}},
                               witness::Bool=false
                              )::Union{Bool, Tuple{Bool,Vector{N}}} where N<:Real
...
```

PR #726, opened 2018-10-06
#725 - Add constraints_list of VPolygon and VPolytope

### LINKED ISSUE #725: Constraints list of `VPolygon` and `VPolytope`
Add `constraints_list(::VPolytope)`` and `VPolygon`.

mforets:
Closes #725.


mforets:
> Move the VPolytope method into the Polyhedra condition?

Hmm i'd like to pull out as much as possible from that condition, keeping inside it only the code that breaks without the `import` statements. The "caller" code will see more more messages of the form "the method X needs the package 'Polyhedra' to be loaded" instead of `MethodError` no method matching `X`.

Issue #416, opened 2018-07-27
Conversion from Zonotope to HPolytope


schillic:
See [this paper](http://www.i6.in.tum.de/Main/Publications/Althoff2009b.pdf) for an algorithm to convert (parallelotopes or) zonotopes to polytopes.

mforets:
A default conversion from `Zonotope` to `HPolytope` is available since #631.

I don't propose to close this one since the conversion proposed here (Theorem 7) is more efficient.

PR #732, opened 2018-10-06
#730 - Add constraints list of Zonotope

### LINKED ISSUE #730: Constraints list of Zonotope
The naive conversion to `HPolytope` from which we can extract the list of constraints was introduced here https://github.com/JuliaReach/LazySets.jl/pull/636.

In https://github.com/JuliaReach/LazySets.jl/issues/416 it is proposed to write the list of constraints using an n-dimensional cross product. We can outsource this more efficient method to a follow-up issue.

mforets:
Closes #730.

### DIFF
--- a/src/Zonotope.jl
+++ b/src/Zonotope.jl
@@ -6,7 +6,8 @@ export Zonotope,
        linear_map,
        scale,
        ngens,
-       reduce_order
+       reduce_order,
+       constraints_list
 
 """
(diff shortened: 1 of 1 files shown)


mforets:
Not sure if this function should be documented right now, since it depends on `tohrep(::VPolytope)`.

schillic:
> Not sure if this function should be documented right now, since it depends on tohrep(::VPolytope).

Let us keep it like that. For #416 we can just replace this implementation.

PR #735, opened 2018-10-07
#731 - Constraints list of Ball1

### LINKED ISSUE #731: Constraints list of Ball1
`B1(c, r)`, in dimension `n`, has `2^n` constraints whose normal directions `di` are

```
julia> Iterators.product([[1, -1] for i = 1:n]...)
Base.Iterators.Prod2{Array{Int64,1},Array{Int64,1}}([1, -1], [1, -1])

julia> collect(ans)
2×2 Array{Tuple{Int64,Int64},2}:
 (1, 1)   (1, -1)
 (-1, 1)  (-1, -1)
```

The constraints can be defined as `di^T (x-c) <= r` for all `di`.

mforets:
Closes #731.

### DIFF
--- a/src/Ball1.jl
+++ b/src/Ball1.jl
@@ -182,3 +182,34 @@ function ∈(x::AbstractVector{N}, B::Ball1{N})::Bool where {N<:Real}
     end
     return sum <= B.radius
 end
+
+"""
+    constraints_list(P::Ball1{N})::Vector{LinearConstraint{N}} where {N<:Real}
+
+Return the list of constraints defining a ball in the 1-norm.
+
+### Input
+
+- `B` -- ball in the 1-norm
+
+### Output
+
+The list of constraints of the ball.
+
+### Algorithm
+
+The constraints can be defined as ``d_i^T (x-c) ≤ r`` for all ``d_i``, where
+``d_i`` is a vector with elements ``1`` or ``-1`` in ``n`` dimensions. To span
+all possible ``d_i``, the function `Iterators.product` is used.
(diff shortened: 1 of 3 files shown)


src/Ball1.jl
```diff
@@ -182,3 +182,34 @@ function ∈(x::AbstractVector{N}, B::Ball1{N})::Bool where {N<:Real}
...
+``d_i`` is a vector with elements ``1`` or ``-1`` in ``n`` dimensions. To span
+all possible ``d_i``, the function `Iterators.product` is used.
+"""
+function constraints_list(B::Ball1{N})::Vector{LinearConstraint{N}} where {N<:Real}
+    n = LazySets.dim(B)
+    c, r = B.center, B.radius
+    clist = Vector{LinearConstraint{N}}(undef)
+    sizehint!(clist, 2^n)
+    for (i, di) in enumerate(Iterators.product([[one(N), -one(N)] for i = 1:n]...))
+        di = collect(di) # tuple -> vector
```
schillic: Generally one should not redefine variables with new types for type stability.

mforets: i added because i was getting `MethodError` no method matching `dot(::Tuple, ::Array)`. i don't know if it can be done in a different way, seems like `Iterators.product...` is always returning tuples.

schillic:
You could have just removed the `undef` :smile:

src/Ball1.jl
schillic: I meant to use a new variable name.

PR #748, opened 2018-10-08
#746 - Wrong support vector of HalfSpace in opposite direction of normal vector

### LINKED ISSUE #746: Wrong support vector of HalfSpace in opposite direction of normal vector
```julia
julia> ρ([-1.], LazySets.HalfSpace([1.0], 1.0))
-1.0
```

schillic:
Closes #746.

### DIFF
--- a/src/Hyperplane.jl
+++ b/src/Hyperplane.jl
@@ -118,16 +118,17 @@ end
 ```
     σ_helper(d::AbstractVector{N},
              hp::Hyperplane{N},
-             [name]::String="hyperplane") where {N<:Real}
+             [halfspace]::Bool=false) where {N<:Real}
 ```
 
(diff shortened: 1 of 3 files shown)


src/Hyperplane.jl
```diff
@@ -167,6 +170,9 @@ Otherwise this function throws an error.
                 elseif first_nonzero_entry_a == -1
                     factor = hp.a[i] / d[i]
                     first_nonzero_entry_a = i
+                    if halfspace && factor < 0
```
mforets: Given `N` why not returning `convert(N, Inf)`?

schillic: I created #750. To keep things consistent, let us do that change in one go.

PR #749, opened 2018-10-08
#747 - Use ρ of lazy HalfSpace etc. intersection also for ρ_upper_bounds

### LINKED ISSUE #747: Use ρ of lazy HalfSpace etc. intersection also for ρ_upper_bounds
Otherwise we get an error that the support function/vector is not defined.

schillic:
Closes #747.


schillic:
One remark: The `is_intersection_empty` check that is performed in the exact `ρ` might fail. I had prepared a version where we check for emptiness in the approximate version first, but then I realized that this does not help because it is only a sufficient test.
What can we do here?

src/Intersection.jl
```diff
@@ -380,17 +385,19 @@ For additional information we refer to:
...
+           kwargs...) where N<:Real
 
     X = cap.X    # compact set
     H = cap.Y    # halfspace or hyperplane
 
     # if the intersection is empty => stop
-    if check_intersection
-        is_intersection_empty(X, H) && return zero(N) # TODO use this convention? error?
+    if check_intersection && is_intersection_empty(X, H)
+        error("the intersection is empty")
```
mforets: 👍

mforets:
problem is that `include("Approximations/Approximations.jl") is done *after* the including `Intersection.jl`

PR #765, opened 2018-10-11
#764 - Add concrete intersection of HalfSpace and HPolytope

### LINKED ISSUE #764: Add concrete intersection of HalfSpace and HPolytope

schillic:
Closes #764.


schillic:
I was not sure whether to move the methods inside the `Polyhedra` block. But we only need that package for redundancy removal, so generally it can be used without loading `Polyhedra`.

Issue #776, opened 2018-10-12
Lazy intersection recomputes supp function too often


mforets:
```julia
julia> X = BallInf(ones(2), 1.0);
julia> H = LazySets.HalfSpace([-1.0, 0.0], 0.0);

julia> @time ρ(d, X∩H, upper_bound=true)
entering rho method
  0.000622 seconds (650 allocations: 51.813 KiB)

julia> @time ρ(d, X∩(X∩H), upper_bound=true)
entering rho method
...  # hundred of times
entering rho method
  1.789980 seconds (1.02 M allocations: 63.388 MiB, 3.47% gc time)
```

mforets:
In current master things got better (#796 , and others):

```julia
julia> @btime ρ($d, $X∩$H)
28.693 μs (465 allocations: 41.97 KiB)

julia> @btime ρ($d, $X∩($X∩$H))
10.791 ms (182097 allocations: 14.92 MiB)
```

Still `ρ(d, X∩(X∩H))` enters many times into `is_intersection_empty` which is kind of unexpected (?).

schillic:
With #803, things will get smooth (but also coarser).

schillic:
#814 fixes a related bug.

PR #777, opened 2018-10-12
improve line_search algorithm

schillic:
See #776.

### DIFF
--- a/src/Approximations/upper_bounds.jl
+++ b/src/Approximations/upper_bounds.jl
@@ -120,7 +120,8 @@ function ρ_upper_bound(d::AbstractVector{N},
 
     X = cap.X    # compact set
     P = cap.Y    # polytope
-    return minimum([ρ(d, X ∩ Hi; kwargs...) for Hi in constraints_list(P)])
+    return minimum([ρ(d, X ∩ Hi; upper_bound=true, kwargs...)
+                    for Hi in constraints_list(P)])
 end
 
 # disambiguation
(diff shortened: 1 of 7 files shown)


schillic:
|feature|performance|works (i.e., correctly implemented)?|
|---|---|---|
|`master`|`0.032610 seconds (185.67 k allocations: 15.251 MiB)`|-|
|depth = 1|`0.031774 seconds (186.05 k allocations: 15.268 MiB)`|no|
|check intersection only 1x|`0.035874 seconds (180.95 k allocations: 14.917 MiB)`|no|
|upper/lower bound|`0.020309 seconds (108.77 k allocations: 8.682 MiB)`|yes|
|all (this branch)|`0.017837 seconds (109.05 k allocations: 8.695 MiB)`|no|

schillic:
I decided to end this mess with passing `upper_bound` around. Now we have `ρ_upper_bound` for all relevant lazy operations. For all other set types we use the default `ρ`. The only remainder is `Intersection`, which we can rectify in another PR.
However, with the current settings we cannot run the bouncing ball model anymore...

schillic:
This global variable stuff does not work. We would need to do a heavy refactoring such that the depth is passed through all `ρ` functions correctly.

mforets:
i seriously  consider removing `upper_bound` altogether and defining the overapprox `ρ` for the intersection (there is no such efficient and exact `ρ` for this operation anyways).

PR #778, opened 2018-10-12
[WIP] Revise ρ_upper_bound

schillic:
I decided to end this mess with passing upper_bound around. Now we have ρ_upper_bound for all relevant lazy operations. For all other set types we use the default ρ. The only remainder is Intersection, which we can rectify in another PR.
However, with the current settings we cannot run the bouncing ball model anymore...


schillic:
Now everything should work again. Note, however, that the execution of ρ is now less efficient, which is not good :disappointed:

`master`:
```
julia> @time ρ(d, X+(X∩H))
  0.000178 seconds (474 allocations: 42.469 KiB)
julia> @time ρ(d, X+(X+X))
  0.000016 seconds (13 allocations: 816 bytes)
```
this branch:
```
julia> @time ρ(d, X+(X∩H))
  0.000183 seconds (582 allocations: 50.906 KiB)
julia> @time ρ(d, X+(X+X))
  0.000009 seconds (22 allocations: 1.375 KiB)
```

I guess this is due to the additional keyword arguments (even if they are not used).

---

There might be a solution by defining additional `ρ`/`ρ_upper_bound` methods without keyword arguments (basically the old implementation). An idea is to define
```julia
ρ_args(kwargs...) = (d, X) -> ρ(d, X, kwargs...)
ρ_upper_bound_args(kwargs...) = (d, X) -> ρ_upper_bound(d, X, kwargs...)
```
globally and then instead of passing `ρ` resp. `ρ_upper_bound` to `ρ_helper`, in the two new implementations we just pass the respective function object and remove the `kwargs...` from `ρ_helper`.
Example:
```julia
function ρ_helper(d, X, ρ_rec)
    ...
    ρ_rec(d2, X2)
    ...
end
function ρ(d, X)
    return ρ_helper(d, X, ρ)
end
function ρ(d, X; kwargs...)
    return ρ_helper(d, X, ρ_args(kwargs...))
end
function ρ_upper_bound(d, X)
    return ρ_helper(d, X, ρ_upper_bound)
end
function ρ_upper_bound(d, X; kwargs...)
    return ρ_helper(d, X, ρ_upper_bound_args(kwargs...))
end
```
Actually, now I realize that we might be able to outsource all these small functions and only need to implement `ρ_helper` for each type :+1:

schillic:
New version in this branch:
```julia
julia> @time ρ(d, X+(X∩H))
  0.000177 seconds (476 allocations: 42.625 KiB)

julia> @time ρ(d, X+(X+X))
  0.000008 seconds (17 allocations: 1008 bytes)
```

schillic:
Closing in favor of #779.

PR #782, opened 2018-10-13
#781 - Intersection emptiness check for two half-spaces

### LINKED ISSUE #781: Intersection emptiness check for two half-spaces

schillic:
Closes #781.


schillic:
> i had some doubts about the type instability in samedir (not using zero(N) etc).

Yes, but if you have `1 / 2` (of type `Int`), then the result will be a `Float64`. The implementation will anyway not work due to numerical issues. We should maybe add a general way to specify a tolerance.

mforets:
i agree; using `N<:AbstractFloat` seems reasonable in those cases too.

PR #784, opened 2018-10-13
#783 - Overapproximation of lazy intersection of polytope and half-space

### LINKED ISSUE #783: Overapproximation of lazy intersection of polytope and half-space

schillic:
Closes #783.


schillic:
I wonder if this should alternatively be implemented like the intersection of a polytope and a half-space. This might be more efficient.

PR #793, opened 2018-10-13
WIP #682 - Sandwich algorithm for lazy intersection

### LINKED ISSUE #682: Lazy intersection of halfspace and polyhedron using sandwich algorithm
See also #375.

mforets:
See #682.


mforets:
The branch contains old code; we should probably experiment in a separate notebook before making this PR, so i think we should close this one.

mforets:
For reference, i stored the code in this branch in https://github.com/mforets/escritoire/blob/master/sets/SandwichIntersection.ipynb

PR #796, opened 2018-10-14
#397 Add cache to Intersection

### LINKED ISSUE #397: Add cache to Intersection
In #396 we implement the concrete intersection of two hyperrectangles. The issue with the lazy operation (`Intersection{N, <:AbstractHyperrectangle, <:AbstractHyperrectangle}`) is this: It needs to know whether the intersection is empty in advance. In the worst case this requires to compute the whole intersection, which to some extent breaks the point of the lazy operation.
Given that we have to compute this information with the first support vector query (or other queries like `isempty`), we should cache this result and only compute it once.
I guess that this observation is general for the intersection (in particular also for #375 and #262), so we should have a general caching mechanism for the lazy `Intersection` type. I cannot quite foresee how this cache would look like. For the moment it would suffice to have a new field `isempty::Tribool`. A `Tribool` would be a (mutable) type that can be `true`, `false`, or `unknown`. The field `isempty` would be initialized to `unknown` and later be set to `true` or `false` accordingly. (Note that `Intersection` would still be immutable.)
Later we may require more specific information for intersections, depending on the set types. Then we could replace the `isempty` field by a general caching field `cache::C` where `C` is a new type parameter of type `IntersectionCache`. The latter would be a new abstract type that we can instantiate for each respective combination of set types. In the instantiation we would store the respective information.

So the question is: Do we want to have the `isempty` field for now, or directly go for the general `IntersectionCache` type?

schillic:
Closes #397.


src/Intersection.jl
```diff
@@ -370,16 +433,30 @@ Return if the intersection is empty or not.
...
 
 `true` iff the intersection is empty.
+
+### Notes
+
+The result will be cached, so a second query will be fast.
 """
 function isempty(cap::Intersection)::Bool
-    return is_intersection_empty(cap.X, cap.Y)
+    if isempty_known(cap.cache)
```
mforets: above there is a method `isempty_konwn(::Intersection)`. should we use it instead? (or delete it?)

Issue #750, opened 2018-10-09
Infinity values for support vector/function of unbounded sets


schillic:
Instead of crashing, we should return infinity values for directions in which a set is unbounded.

schillic:
I also had that thought, but it does not work. The result should be an element of the set.
Take the 2D half-space `1x + 0y >= 0` and direction `d = [-1, 1]`. The point `[-Inf, Inf]` is not in the half-space. The solution should be `[0, Inf]`.

PR #799, opened 2018-10-14
#750 - Allow unbounded ρ for HPoly

schillic:
See #750.


src/HPolyhedron.jl
```diff
@@ -67,6 +67,34 @@ function dim(P::HPoly{N})::Int where {N<:Real}
...
+### Output
+
+The support function of the polyhedron.
+If the polyhedron is unbounded in the given direction, the result is `Inf`.
+
+### Algorithm
+
+This implementation uses `GLPKSolverLP` as linear programming backend.
+"""
+function ρ(d::AbstractVector{N}, P::HPoly{N})::N where {N<:Real}
```
mforets: use `HPolyhedron` instead of `HPoly`?

mforets: on the other hand we dont yet a have a `isbounded` check. . i take back my comment.

schillic: Oh, I forgot to make the result depend on the type :+1:

PR #803, opened 2018-10-14
#802 - Generic depth limit for line search

### LINKED ISSUE #802: Generic depth limit for line search
Idea: Check wether the first argument to the lazy `Intersection` is an `Intersection` again. Based on that information, we either choose the `line_search` algorithm or not.

In general, we want to specify a particular check in the application. So the interface will be a function that is executed on the `Intersection` object. The above will be the default, but in the application we will apply the check also for a `LinearMap` of an `Intersection`.

schillic:
Closes #802.


schillic:
```julia
julia> using LazySets, Optim; import LazySets.line_search_applies

julia> X = BallInf([1.], 1.); H = HalfSpace([1.], 1.); 

julia> ρ([1.], X ∩ H)
1.0000000007394538

julia> function line_search_applies(cap::Intersection{N})::Bool where N<:Real
           return false
       end

julia> ρ([1.], X ∩ H)
2.0

julia> function line_search_applies(cap::Intersection{N})::Bool where N<:Real
           return false
       end

julia> ρ([1.], X ∩ H)
1.0000000007394538
```

mforets:
+1 but i dont quite understand the implementation. i think that `s = min(ρ(d, X; kwargs...), N(Inf))` is a valid approach, alternative to `line_search` and `projection` so it should be on the same level, maybe call it `naive` or `default`.

schillic:
> shall the return value of the 2nd one be true?

Yes, sorry, fixed.

PR #863, opened 2018-10-26
#115 - Avoid array insertion in iterative refinement

### LINKED ISSUE #115: Avoid array insertion in iterative refinement
In the "iterative refinement" of polygons we use array `insert!`, which is not efficient (in practice it might be *okay* because we start at the front and walk to the end).
An efficient alternative is to have two arrays, one for finished refinements and one for unfinished ones:

```julia
initialize with box direction approximations
[] [a b c d]
turn around initial array
[] [d c b a]
# actually just initialize in opposite order, but for clarity this is described in this way
pick a
[] [d c b]
refine a to (d, e)
add them in the opposite order
[] [d c b e d]
pick d
assume it is not refinable, so add it to the left
[d] [d c b e]
pick e
[d] [d c b]
assume it is not refinable, so add it to the left
[d e] [d c b]
...
```

schillic:
Closes #115.

### DIFF
--- a/src/Approximations/iterative_refinement.jl
+++ b/src/Approximations/iterative_refinement.jl
@@ -5,13 +5,13 @@ Type that represents a local approximation in 2D.
 
 ### Fields
 
-- `p1`         -- first inner point
-- `d1`         -- first direction
-- `p2`         -- second inner point
-- `d2`         -- second direction
-- `q`          -- intersection of the lines l1 ⟂ d1 at p1 and l2 ⟂ d2 at p2
-- `refinable`  -- states if this approximation is refinable
-- `err`        -- error upper bound
+- `p1`        -- first inner point
+- `d1`        -- first direction
+- `p2`        -- second inner point
+- `d2`        -- second direction
+- `q`         -- intersection of the lines l1 ⟂ d1 at p1 and l2 ⟂ d2 at p2
+- `refinable` -- states if this approximation is refinable
+- `err`       -- error upper bound
 
 ### Notes
 
@@ -34,14 +34,14 @@ Type that represents the polygonal approximation of a convex set.
 
 ### Fields
 
-- `S`                -- convex set
-- `approx_list_open` -- vector of local approximations that are not finished yet
-- `approx_list_done` -- vector of linear constraints that are already finished
+- `S`           -- convex set
+- `approx_list` -- vector of local approximations that are not finished yet
+- `constraints` -- vector of linear constraints that are already finished
 """
 struct PolygonalOverapproximation{N<:Real}
     S::LazySet{N}
-    approx_list_open::Vector{LocalApproximation{N}}
-    approx_list_done::Vector{LinearConstraint{N}}
+    approx_list::Vector{LocalApproximation{N}}
+    constraints::Vector{LinearConstraint{N}}
 
     PolygonalOverapproximation(S::LazySet{N}) where N<:Real = new{N}(
         S, Vector{LocalApproximation{N}}(), Vector{LinearConstraint{N}}())
@@ -53,11 +53,11 @@ end
 
 ### Input
 
-- `S`          -- convex set
-- `p1`         -- first inner point
-- `d1`         -- first direction
-- `p2`         -- second inner point
-- `d2`         -- second direction
+- `S`  -- convex set
+- `p1` -- first inner point
+- `d1` -- first direction
+- `p2` -- second inner point
+- `d2` -- second direction
 
 ### Output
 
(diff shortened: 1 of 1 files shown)


schillic:
The new implementation is only more efficient for very large numbers of constraints.

```julia
using LazySets, LazySets.Approximations, BenchmarkTools
B = Ball2(zeros(2), 1.);


# master branch:

julia> @btime overapproximate(B, 1e-2);
  104.092 μs (2573 allocations: 132.77 KiB)

julia> @btime overapproximate(B, 1e-5);
  3.959 ms (88850 allocations: 4.51 MiB)

julia> @btime overapproximate(B, 1e-8);
  225.120 ms (2977460 allocations: 147.42 MiB)


# this branch:

julia> @btime overapproximate(B, 1e-2);
  104.749 μs (2631 allocations: 134.19 KiB)

julia> @btime overapproximate(B, 1e-5);
  3.986 ms (87824 allocations: 4.51 MiB)

julia> @btime overapproximate(B, 1e-8);
  221.496 ms (2974378 allocations: 147.37 MiB)

```

src/Approximations/iterative_refinement.jl
```diff
@@ -131,29 +135,26 @@ function refine(Ω::PolygonalOverapproximation, i::Int)::Tuple{LocalApproximatio
...
 
 A polygon in constraint representation.
 """
-function tohrep(Ω::PolygonalOverapproximation{N})::AbstractHPolygon where {N<:Real}
-    p = HPolygon{N}()
-    for ai in Ω.approx_list
-        addconstraint!(p, LinearConstraint(ai.d1, dot(ai.d1, ai.p1)))
-    end
-    return p
+function tohrep(Ω::PolygonalOverapproximation{N}
```
mforets: good! could a `Algorithm` comment that there is no need for using `addconstraint!` because the constraints `Ω.constraints` are known to be already sorted? 

(if the `HPolygon` constructor has at some later point a default sorting of the constraints, we'd like to turn that flag off here)

src/Approximations/iterative_refinement.jl
```diff
@@ -35,60 +35,61 @@ Type that represents the polygonal approximation of a convex set.
 ### Fields
 
 - `S`           -- convex set
-- `approx_list` -- vector of local approximations
+- `approx_list` -- vector of local approximations that are not finished yet
+- `constraints` -- vector of linear constraints that are already finished
```
mforets: Perhaps we should define what it means for an approximation to be "finished" in the docstring (or the docs).

src/Approximations/iterative_refinement.jl
```diff
@@ -97,30 +98,33 @@ the new approximation is returned by this function.
...
-by splitting along the normal direction to the approximation.
+Refine a given local approximation of the polygonal approximation of a convex
+set by splitting along the normal direction to the approximation.
 
 ### Input
 
-- `Ω`   -- polygonal overapproximation of a convex set
-- `i`   -- integer index for the local approximation to be refined
+- `Ω` -- polygonal overapproximation of a convex set
+- `i` -- integer index for the local approximation to be refined
```
mforets: needs update

src/Approximations/iterative_refinement.jl
```diff
@@ -97,30 +98,33 @@ the new approximation is returned by this function.
...
 
 ### Output
 
 The tuple consisting of the refined right and left local approximations.
 """
-function refine(Ω::PolygonalOverapproximation, i::Int)::Tuple{LocalApproximation, LocalApproximation}
-    R = Ω.approx_list[i]
+function refine(Ω::PolygonalOverapproximation,
```
mforets: moreover, `Ω` is not needed anymore, i suggest to remove it from the function's signature.

mforets: false -- i forgot the lines just below.

src/Approximations/iterative_refinement.jl
```diff
@@ -169,46 +170,52 @@ local `Approximation2D`.
...
     pw = σ(DIR_WEST(N), S)
     ps = σ(DIR_SOUTH(N), S)
 
-    Ω = PolygonalOverapproximation{N}(S)
+    Ω = PolygonalOverapproximation(S)
 
-    addapproximation!(Ω, pe, DIR_EAST(N), pn, DIR_NORTH(N))
-    addapproximation!(Ω, pn, DIR_NORTH(N), pw, DIR_WEST(N))
-    addapproximation!(Ω, pw, DIR_WEST(N), ps, DIR_SOUTH(N))
+    # add constraints in reverse (i.e., clockwise) order
```
mforets: Hmm could you plz elaborate on the change? 

Doesn't `# add constraints in reverse (i.e., clockwise) order` conflict with `HPolygon{N}(Ω.constraints)` in `tohrep` constructor, that expects counterclockwise ordering?

src/Approximations/iterative_refinement.jl
schillic: Actually, we only need `Ω.S`. I will pass that instead.

src/Approximations/iterative_refinement.jl
schillic: The `approx_list` became a stack, so you have to add in reverse order. I will rename to `approx_stack` to make that clear.

PR #864, opened 2018-10-26
WIP #202 - Optional algorithms for <=

### LINKED ISSUE #202: Optional algorithms for <=
In #198 we introduced an alternative method to sort directions with the right-hand rule after checking the quadrant of each direction (in particular it doesn't use `atan2`). It could be added as an alternative `algorithm=...` for the `<=`.

A quick test shows that it can be faster too: for `1000` random directions, the  times are approx. `130us` vs. `70us`.

---

The comparison that uses `quadrant` is more general:

```julia
julia> using LazySets

julia> u, v = [1., 0], [0, 1.];

julia> @which u <= v
<=(u::AbstractArray{N,1}, v::AbstractArray{N,1}) where N<:AbstractFloat in LazySets at /Users/forets/.julia/v0.6/LazySets/src/AbstractPolygon.jl:107

julia> u, v = [1//1, 0//1], [0//1, 1//1];

julia> @which u <= v
<=(u::AbstractArray{N,1}, v::AbstractArray{N,1}) where N<:Real in LazySets at /Users/forets/.julia/v0.6/LazySets/src/AbstractPolygon.jl:169
```

Test:

```julia
julia> using Compat, BenchmarkTools

julia> @inline function quadrant(w::AbstractVector{N})::Int where {N<:Real}
           dwx = w[1] >= zero(N) ? 1 : 0
           dwy = w[2] >= zero(N) ? 1 : 0
           return (1 - dwx) + (1 - dwy) + ((dwx & (1 - dwy)) << 1)
       end
quadrant (generic function with 1 method)

julia> function cquadrant(u::AbstractVector{N}, v::AbstractVector{N})::Bool where {N<:Real}
           qu, qv = quadrant(u), quadrant(v)
           if qu == qv
               # same quadrant, check right-turn with center 0
               ans = u[1] * v[2] - v[1] * u[2] >= zero(N)
           else

               # different quadrant
               ans = qu < qv
           end
           return ans
       end
cquadrant (generic function with 1 method)

julia> @inline function jump2pi(x::N)::N where {N<:AbstractFloat}
           x < zero(N) ? 2 * pi + x : x
       end
jump2pi (generic function with 1 method)

julia> function cjump(u::AbstractVector{N},
                   v::AbstractVector{N})::Bool where {N<:AbstractFloat}
           return jump2pi(atan2(u[2], u[1])) <= jump2pi(atan2(v[2], v[1]))
       end
cjump (generic function with 1 method)

julia> u = [rand(2) for i in 1:1000]; v = [rand(2) for i in 1:1000];

julia> @btime cjump.(u, v)
  132.405 μs (23 allocations: 5.13 KiB)

julia> @btime cquadrant.(u, v)
  31.329 μs (23 allocations: 5.13 KiB)

julia> cquadrant.(u, v) == cjump.(u, v)
true
```

schillic:
Closes #202.

I had to find out that Julia is not very efficient for optional arguments, even if the default argument is chosen (I had assumed that this case can be statically optimized).
Hence I defined `<=` without additional arguments (the default implementation), and added a method with an additional `algorithm` argument. If you instead use
`<=(u::AbstractVector{N}, v::AbstractVector{N}; algorithm::String="quadrant")`,
you will get the following timings:
```julia

julia> x = rand(2); y = rand(2);

julia> @btime LazySets.le_quadrant($x, $y);
  5.173 ns (0 allocations: 0 bytes)

julia> @btime <=($x, $y); # should be equivalent to the above, but it is slower
  10.923 ns (0 allocations: 0 bytes)

julia> @btime <=($x, $y, algorithm="quadrant");
  46.827 ns (1 allocation: 96 bytes)
```

### DIFF
--- a/src/AbstractHPolygon.jl
+++ b/src/AbstractHPolygon.jl
@@ -217,7 +217,8 @@ function addconstraint!(P::AbstractHPolygon{N},
         else
             if linear_search
                 # linear search
-                while d <= P.constraints[k].a
+                quadrant_d = quadrant(d)
+                while le_quadrant(d, quadrant_d, P.constraints[k].a)
                     k -= 1
                 end
             else
@@ -267,8 +268,9 @@ function binary_search_constraints(d::AbstractVector{N},
                                   )::Int where {N}
     lower = 1
     upper = n+1
+    quadrant_d = quadrant(d)
     while lower + 1 < upper
-        if constraints[k].a <= d
+        if le_quadrant(constraints[k].a, d, quadrant_d)
             lower = k
         else
             upper = k
@@ -278,7 +280,7 @@ function binary_search_constraints(d::AbstractVector{N},
     if choose_lower
         return lower
     else
-        if lower == 1 && !(constraints[1].a <= d)
+        if lower == 1 && !le_quadrant(constraints[1].a, d, quadrant_d)
             # special case for index 1
             return 1
         end
--- a/src/AbstractPolygon.jl
+++ b/src/AbstractPolygon.jl
@@ -48,7 +48,7 @@ The ambient dimension of the polygon, which is 2.
 end
 
 """
-    jump2pi(x::N)::N where {N<:AbstractFloat}
+    jump2pi(x::N)::N where N<:AbstractFloat
 
 Return ``x + 2π`` if ``x`` is negative, otherwise return ``x``.
 
@@ -75,18 +75,18 @@ julia> jump2pi(0.5)
 0.5
 ```
 """
-@inline function jump2pi(x::N)::N where {N<:AbstractFloat}
+@inline function jump2pi(x::N)::N where N<:AbstractFloat
     x < zero(N) ? 2 * pi + x : x
 end
 
 """
-    quadrant(w::AbstractVector{N})::Int where {N<:Real}
+    quadrant(w::AbstractVector{N})::Int where N<:Real
 
 Compute the quadrant where the direction `w` belongs.
 
 ### Input
 
-- `w` --  direction
+- `w` -- direction
 
 ### Output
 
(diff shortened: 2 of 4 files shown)


schillic:
We probably do not want to take the last commit. Here are the results (random numbers, hence not really comparable; but note the higher allocations):

```julia
using LazySets, BenchmarkTools
P = HPolygon();
for i in 1:100
    addconstraint!(P, LinearConstraint(rand(2), rand(1)[1]));
end
@btime σ($[-0.5, 0.5], $P);
```

`master`:   1.545 μs (25 allocations: 1.08 KiB)
this branch:   1.595 μs (39 allocations: 1.30 KiB)

mforets:
> I had to find out that Julia is not very efficient for optional arguments, even if the default argument is chosen (I had assumed that this case can be statically optimized).

AFAIK this is not further a performance issue in v0.7/1.0.

mforets:
These are the timings that i got (all on v0.7):

```julia
using LazySets, BenchmarkTools

x = rand(2); y = rand(2);

@btime LazySets.le_quadrant($x, $y);
4.172 ns (0 allocations: 0 bytes)

@btime <=($x, $y);
4.177 ns (0 allocations: 0 bytes)

@btime <=($x, $y, "quadrant");
14.584 ns (0 allocations: 0 bytes)
```

---

For `@btime σ($[-0.5, 0.5], $P);` : 

|schillic/202|master|
|-----|------|
|1.037 μs|1.225 μs|
|18 allocations: 944 bytes|18 allocations: 944 bytes|


---

Since keyword argumnts do not cause a problem (🎉) i suggest to use them instead of `le_quadrant`.
`le_quadrant.

schillic:
The code I added in this PR did *not* use keyword arguments, and hence it was fast.
I added a commit that uses them now, but I do not want to merge because it is still slower on v0.7:
```julia
julia> @btime LazySets.le_quadrant($x, $y);
  3.671 ns (0 allocations: 0 bytes)

julia> @btime <=($x, $y);
  8.233 ns (0 allocations: 0 bytes)

julia> @btime <=($x, $y, "quadrant");
  7.763 ns (0 allocations: 0 bytes)
```
And for the polygon constraints, in v0.7 the allocations are fine, but the runtime got worse:
```julia
# without the last commit
julia> @btime σ($[-0.5, 0.5], $P);
  753.192 ns (18 allocations: 944 bytes)

# with the last commit
julia> @btime σ($[-0.5, 0.5], $P);
  823.456 ns (18 allocations: 944 bytes)
```

mforets:
As per experiments, i think its a good idea to change the default for 2d vector comparison, since the "quadrant" is faster. This branch contains deprecated code though.

mforets:
This branch is superseded by https://github.com/JuliaReach/LazySets.jl/pull/1986. @schillic shall we close it?

schillic:
> This branch is superseded by #1986. @schillic shall we close it?

#1986 just removes the `atan` implementation. This branch contains new features.

mforets:
Should we close this as stale?

PR #865, opened 2018-11-01
Fix docs

schillic:
There is one problem remaining: We load `Polyhedra` while we build the doctests for the library (for `vertices_list` in `HPolytope.jl` and `HPolyhedron.jl`). This creates a namespace conflict with `HalfSpace` and thus influences some other doctests. Options:
1. We modify these doctests.
2. We remove the doctests for `vertices_list`.

### DIFF
--- a/docs/src/lib/methods_fix.md
+++ b/docs/src/lib/methods_fix.md
@@ -26,6 +26,7 @@ low
 norm
 is_intersection_empty
 linear_map
+polyhedron
 radius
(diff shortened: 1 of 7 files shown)


mforets:
3. move the examples of `vertices_list(P)` the concrete polyhedra section of the online docs

mforets:
4. just swap the call to `using Polyhedra` and `constraints_list` 😄

PR #869, opened 2018-11-01
#208 - Witness production for empty intersection check of zonotope with hyperplane

### LINKED ISSUE #208: Check for empty intersection of zonotope with hyperplane
* [x] check for intersection
* [ ] produce a witness

schillic:
Closes #208.

### DIFF
--- a/src/Hyperplane.jl
+++ b/src/Hyperplane.jl
@@ -192,7 +192,7 @@ Otherwise this function throws an error.
     end
 
     first_nonzero_entry_a = -1
-    if all(d .== 0)
+    if iszero(d)
         # zero vector
         return an_element(hp)
     else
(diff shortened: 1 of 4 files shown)


src/Hyperplane.jl
```diff
@@ -192,7 +192,7 @@ Otherwise this function throws an error.
     end
 
     first_nonzero_entry_a = -1
-    if all(d .== 0)
+    if iszero(d)
```
mforets: 👍  i didn't know of its existence at the time of https://github.com/JuliaReach/Reachability.jl/pull/373

schillic: Oh, I already forgot about that again :astonished:

PR #870, opened 2018-11-01
#235 - Random set generation functions

### LINKED ISSUE #235: Random set generation functions
Sometimes it is helpful to create a random instance of a set type. What about a function
```julia
random(S::LazySet; N=Float64, dim::Int=2, rng=GLOBAL_RNG, seed=nothing)
```
that we would have to implement for each set type.

I am not sure how rng works in Julia; this needs to be figured out ([starting point](https://docs.julialang.org/en/stable/stdlib/numbers/#Random-Numbers-1)) (reproducibility could be important for experiments/testing).

schillic:
Closes #235.

* [x] The only basic set type missing is `HPolyhedron`. I am not sure how to randomly generate that. One idea is to create an `HPolytope` and randomly remove constraints.


schillic:
Side comment: When we merge this, we will locally get the following warning whenever `LazySets` is loaded in versions >= v0.7:
```julia
┌Warning: Package LazySets does not have Random in its dependencies:
│   added Random as a dependency but haven't updated your primary
└ Loading Random into LazySets from project dependency, future warnings for LazySets are suppressed.
```
This is resolved by
```julia
pkg> resolve
```

src/AbstractHPolygon.jl
```diff
@@ -181,6 +183,47 @@ function ∈(x::AbstractVector{N}, P::AbstractHPolygon{N})::Bool where {N<:Real}
     return true
 end
 
+"""
+    rand(::Type{HPOLYGON}; [N]::Type{<:Real}=Float64, [dim]::Int=2,
+         [rng]::AbstractRNG=GLOBAL_RNG, [seed]::Union{Int, Nothing}=nothing
+        )::HPOLYGON{N} where {HPOLYGON<:AbstractHPolygon}
```
mforets: the option `num_constraints` does not appear here

src/Ellipsoid.jl
```diff
@@ -159,3 +171,46 @@ function ∈(x::AbstractVector{N}, E::Ellipsoid{N})::Bool where {N<:AbstractFloa
...
+A random ellipsoid.
+
+### Algorithm
+
+The center is a normally distributed vector with entries of mean 0 and standard
+deviation 1.
+
+The idea for the shape matrix comes from
+[here](https://math.stackexchange.com/a/358092).
+The matrix is symmetric positive definite, but also diagonally dominant.
```
mforets: Describe with an equation?

```math
Q =  \frac{1}{2}(S + S^T) + nI,
```
where `n = dim` (defaults to 2), and ``S `` is a ``n\\times n`` randomly matrix with coefficients uniformly distributed in the interval `[-1, 1]`.

src/VPolygon.jl
```diff
@@ -276,6 +277,123 @@ function ∈(x::AbstractVector{N}, P::VPolygon{N})::Bool where {N<:Real}
...
+              seed::Union{Int, Nothing}=nothing,
+              num_vertices::Int=-1
+             )::VPolygon{N}
+    @assert dim == 2 "a polygon must have dimension 2"
+    rng = reseed(rng, seed)
+    if num_vertices < 0
+        num_vertices = rand(1:10)
+    end
+
+    # special cases, 0 or 1 vertex
```
mforets: this corner case can be treated in functions that rely on this function, eg. 

```julia
julia> rand(HPolygon, num_constraints=2).constraints
4-element Array{HalfSpace{Float64},1}:
 HalfSpace{Float64}([0.398179, 0.371021], -0.5601983753144615)
 HalfSpace{Float64}([-0.371021, 0.398179], -1.5636591141726837)
 HalfSpace{Float64}([-0.398179, -0.371021], 0.8564018279869186)
 HalfSpace{Float64}([0.371021, -0.398179], 1.5636591141726837)
```

should instead raise an error

src/VPolygon.jl
```diff
@@ -276,6 +277,123 @@ function ∈(x::AbstractVector{N}, P::VPolygon{N})::Bool where {N<:Real}
...
+              N::Type{<:Real}=Float64,
+              dim::Int=2,
+              rng::AbstractRNG=GLOBAL_RNG,
+              seed::Union{Int, Nothing}=nothing,
+              num_vertices::Int=-1
+             )::VPolygon{N}
+    @assert dim == 2 "a polygon must have dimension 2"
+    rng = reseed(rng, seed)
+    if num_vertices < 0
+        num_vertices = rand(1:10)
```
mforets: i suggest to take at least `rand(3:10)`, to avoid some corner cases

src/VPolygon.jl
```diff
@@ -276,6 +277,123 @@ function ∈(x::AbstractVector{N}, P::VPolygon{N})::Bool where {N<:Real}
...
+    end
+
+    # special cases, 0 or 1 vertex
+    if num_vertices <= 1
+        vertices = [randn(rng, N, 2) for i in 1:num_vertices]
+        return VPolygon(vertices; apply_convex_hull=false)
```
mforets: the empty constructor (case `num_vertices=0`) does not work with the kwarg `apply_convex_hull`

```julia
julia> VPolygon()
VPolygon{Float64}(Array{Float64,1}[])

julia> VPolygon([]; apply_convex_hull=false)
ERROR: MethodError: no method matching VPolygon(::Array{Any,1}; apply_convex_hull=false)
Closest candidates are:
  VPolygon() at /Users/forets/.julia/dev/LazySets/src/VPolygon.jl:53 got unsupported keyword argument "apply_convex_hull"
  VPolygon(::Array{Array{N<:Real,1},1}; apply_convex_hull, algorithm) where N<:Real at /Users/forets/.julia/dev/LazySets/src/VPolygon.jl:44
Stacktrace:
 [1] top-level scope at none:0
```

don't know if this is a bug; but we can do here

```
if num_vertices == 0
    return VPolygon()    # convex hull is not applied
elseif num_vertices == 1
    return VPolygon([randn(rng, N, 2)])
end
```

src/VPolygon.jl
```diff
@@ -276,6 +277,123 @@ function ∈(x::AbstractVector{N}, P::VPolygon{N})::Bool where {N<:Real}
...
+    function remove_duplicates_sorted!(v::AbstractVector)
+        for i in length(v)-1:-1:1
+            if v[i] == v[i+1]
+                splice!(v, i+1)
+            end
+        end
+        return v
+    end
+
+    function random_axis_aligned_vectors(rng, N, n)
```
mforets: can one substitute `random_axis_aligned_vectors` to picking n-elements randomly from a normal distribution? or what do these have in common?

```julia
julia> random_axis_aligned_vectors(rng, N, 5)
5-element Array{Float64,1}:
  0.12830892377309455
  2.387169565427545
  0.3232363202950286
 -0.29623066338265214
 -2.542484146113016

julia> random_axis_aligned_vectors(rng, N, 5)
5-element Array{Float64,1}:
  0.9638834005050456
  1.0636721009415822
  0.8336680989665766
 -1.5437334714482494
 -1.317490128964955

julia> random_axis_aligned_vectors(rng, N, 5)
5-element Array{Float64,1}:
  1.8457700493139375
  0.9765805620918393
 -0.40900877044843353
 -0.5103436430987645
 -1.9029981978585788
```

mforets:
A general comment (that could be addressed in another PR): `rand(BallInf, 5)` has the same behavior as `rand(Int, 5)`, ie. to create 5 `BallInf`'s in an array. This is often useful because one can also create arrays of any order easily. However, one is subject to the default options, in particular `rand(BallInf, 5, dim=2)` doesn't work.

src/VPolygon.jl
schillic: I was also thinking about that, but sometimes it is helpful to have these corner cases for testing (my main application for random sets is random testing). Otherwise you will systematically exclude these corner cases. You can control this number with `num_vertices` if you want to, so I think we can keep it like that.

src/VPolygon.jl
schillic: The sum is zero. Have a look at the video [here](http://cglab.ca/~sander/misc/ConvexGeneration/convex.html).

schillic:
Yes, this feature is thanks to
```julia
# hook into random API
import Random.rand
function rand(rng::AbstractRNG, ::SamplerType{T}) where T<:LazySet
    rand(T, rng=rng)
end
```
I am not sure if [this API](https://docs.julialang.org/en/stable/stdlib/Random/#Hooking-into-the-Random-API-1) is made for having keyword arguments.
You can always do `[rand(BallInf; dim=mydim) for i in 1:5]`. Not too bad I think.

src/VPolygon.jl
mforets: Very nice video.

I would suggest to attach a docstring and rename `random_axis_aligned_vectors`, maybe to `random_zero_sum_vector`.

schillic:
> The only basic set type missing is HPolyhedron. I am not sure how to randomly generate that. One idea is to create an HPolytope and randomly remove constraints.

What do you think about that?

src/VPolygon.jl
mforets: would this work?

```julia
function random_zero_sum_vector(rng, N, n)
    list = randn(rng, N, n)
    return [list[1:1:end-1] - list[2:1:end];
                list[end] - list[1]]
end
```

mforets:
Since `HPolyhedron` are `HPolytope`, one can just run the `rand(HPolytope)` and do a `convert(HPolyhedron, ..)`. Doing this and then removing constraints sounds good as well. In general we don't give guarantees on the distribution of the constructed set (that's another story).

src/VPolygon.jl
schillic: Yes, but the numbers are not uniformly distributed. The article I linked to claims that the algorithm guarantees that, and I think it is good to have this property (if you can have it).

mforets: alright, i think this function is interesting in its own right and i think we can put it outside (eventually call it `_random_zero_sum_vector`, we had this discussion 😄 ) and add docstring with reference to the article.

src/VPolygon.jl
schillic: I do not understand exactly, but I now raise an error if the number of constraints is two or less.

src/VPolygon.jl
schillic: Thinking about it, my argument also holds in the other direction. So I will change it.

src/VPolygon.jl
schillic: You have to add a type to the array (as the message suggested):
```julia
julia> VPolygon(Vector{N}[]; apply_convex_hull=false)
VPolygon{Float64}(Array{Float64,1}[])
```

schillic:
I added all the fixes you proposed.

PR #872, opened 2018-11-02
Do not use 'fill' for zeros and ones

schillic:


schillic:
No, just shorter code and (IMO) better readability.

PR #889, opened 2018-11-07
#876 - Convex hull does not sort two vertex input

### LINKED ISSUE #876: convex_hull does not sort two-vertex input
```julia
julia> VPolygon([[1., 2.], [1., 3.]])
VPolygon{Float64}(Array{Float64,1}[[1.0, 2.0], [1.0, 3.0]])

julia> VPolygon([[1., 3.], [1., 2.]])
VPolygon{Float64}(Array{Float64,1}[[1.0, 3.0], [1.0, 2.0]])
```

mforets:
Closes #876.

### DIFF
--- a/src/convex_hull_algorithms.jl
+++ b/src/convex_hull_algorithms.jl
@@ -71,7 +71,21 @@ See the non-modifying version `convex_hull` for more details.
 function convex_hull!(points::Vector{S};
                       algorithm::String="monotone_chain"
                      )::Vector{S} where {N<:Real, S<:AbstractVector{N}}
-    (length(points) == 1 || length(points) == 2) && return points
+
+    # treat 1-vertex and 2-vertex corner cases (see #876)
+    if length(points) == 1
+        return points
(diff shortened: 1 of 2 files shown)


src/convex_hull_algorithms.jl
```diff
@@ -71,7 +71,21 @@ See the non-modifying version `convex_hull` for more details.
...
+
+    # treat 1-vertex and 2-vertex corner cases (see #876)
+    if length(points) == 1
+        return points
+    elseif length(points) == 2
+        p1, p2 = points[1], points[2]
+        if p1 == p2  # check for redundancy
+            pop!(points)
+        elseif p1 <= p2
+            nothing
```
schillic: Skip this case and check for the opposite case.

mforets: It looked a bit more readable this way 📄

schillic: Hm, not to me :smile:

PR #919, opened 2018-11-16
#918 - vertices_list for AbstractHPolygon with infeasible constraints is incorrect

### LINKED ISSUE #918: vertices_list for AbstractHPolygon with infeasible constraints is incorrect
`vertices_list` does not work correctly if the constraints are infeasible.
The result should be an empty list, but it is not.
```julia
julia> P = HPolygon([
    HalfSpace([1.0, 1.0], 0.0),
    HalfSpace([-1.0, 0.0], -1.0),
    HalfSpace([0.0, -1.0], -1.0)]);

julia> vertices_list(P)
 [1.0, -1.0]
 [1.0, 1.0] 
 [-1.0, 1.0]
```
The reason is that we take the intersection of lines without caring for the direction of the normal vector.
This has serious consequences because the emptiness check relies on it.

schillic:
Closes #918.

### DIFF
--- a/src/AbstractHPolygon.jl
+++ b/src/AbstractHPolygon.jl
@@ -80,7 +80,8 @@ end
 
 """
     vertices_list(P::AbstractHPolygon{N},
-                  apply_convex_hull::Bool=false
+                  apply_convex_hull::Bool=false,
+                  check_feasibility::Bool=true
                  )::Vector{Vector{N}} where {N<:Real}
 
 Return the list of vertices of a polygon in constraint representation.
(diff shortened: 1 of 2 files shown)


src/AbstractHPolygon.jl
```diff
@@ -109,6 +123,15 @@ function vertices_list(P::AbstractHPolygon{N},
     end
     points[n] = element(intersection(Line(P.constraints[n]),
                                      Line(P.constraints[1])))
+
+    # check if polygon was empty
```
mforets: did you forget to use the flag `check_feasibility`  below?

src/AbstractHPolygon.jl
```diff
@@ -109,6 +123,15 @@ function vertices_list(P::AbstractHPolygon{N},
...
+    avg = sum(points) / length(points)
+    for c in constraints_list(P)
```
mforets: note that one could ask for (see [this function](https://github.com/JuliaReach/LazySets.jl/blob/51c891e09cebdc24dfe0ac0ff604ca31a5fc1561/src/AbstractHPolygon.jl#L175)):

```julia
if avg ∉ P
    return Vector{Vector{N}}(undef, 0)
end
```

schillic:
Thanks, I rebased.

Issue #824, opened 2018-10-18
Add lazy set Union operation


mforets:
The union of two sets `X` and `Y` is the set of points which are in `X`, in `Y` or in both `X` and `Y`. Convexity is not preserved under set union.

An upper bound on the [support function can be returned using the convex hull approximation](https://juliareach.github.io/LazySets.jl/latest/man/polyhedral_approximations.html#Support-Function-1).

schillic:
> Convexity is not preserved under set union.

This is the reason why we removed it before. I think we do not want to go the way of #245 again.

> An upper bound on the support function can be returned using the convex hull approximation.

Support function and support vector are actually tight. But, for instance, iterative refinement does not give you an epsilon-close approximation (because there might be holes in the set).

PR #266, opened 2018-03-01
remove UnionSet and unit_UnionSet.jl

mforets:
Closes #263.

Issue #759, opened 2018-10-10
Add validation function


schillic:
Add a function `validate(X::LazySet)` that takes a set and validates if the set is correctly constructed according to internal assumptions. For instance, the constraints of a `HPolygon` should be sorted, or an `HPolytope` should be bounded. (For many set types, this would just be the constant function `true`.)

mforets:
Otherwise it could be `validate(X::LazySet, f::Function)`, for example `validate(X, isbounded)` to really check that the set is bounded. Consider for example [this use case](https://github.com/JuliaReach/LazySets.jl/pull/978).

PR #922, opened 2018-11-22
UnionSet

mforets:
See .https://github.com/JuliaReach/LazySets.jl/issues/824.

Note that `UnionSet` was removed previously, see [this PR](https://github.com/JuliaReach/LazySets.jl/pull/266).


src/UnionSet.jl
```diff
@@ -0,0 +1,91 @@
...
+- `array` -- array of convex sets
+"""
+struct UnionSetArray{N<:Real, S<:LazySet{N}}
+    array::Vector{S}
+
+    # default constructor with dimension check
+    function UnionSetArray{N, S}(X::Vector{S}) where {N<:Real, S<:LazySet{N}}
+        @assert !isempty(X) "the array in this `UnionSetArray` is empty"
+        n = dim(X[1])
+        @assert all(x -> dim(x) == n, X) "sets in a union must have the same dimension"
```
schillic: Hm, we do not have such a thing for other array types. I would rather do this in #759.

mforets: Alright, i'm removing that inner constructor.

mforets:
I've declared `UnionSetArray` as the array version of `UnionSet` with the `@declare_array_version` macro. Seems like there is no need to make `UnionSet` inherit from `LazySet` to use it.

src/UnionSet.jl
```diff
@@ -0,0 +1,258 @@
...
+Return the support vector of the union of two convex sets in a given direction.
+
+### Input
+
+- `d`         -- direction
+- `cap`       -- union of two convex sets
+- `algorithm` -- (optional, default: "support_vector"): the algorithm to compute
+                 the support vector; if "support_vector", use the support
+                 vector of each argument; if "support_function" use the support
+                 function of each argument and evalute the support vector of only
```
schillic:
```suggestion
                 function of each argument and evaluate the support vector of only
```

src/UnionSet.jl
```diff
@@ -0,0 +1,258 @@
...
+a given direction.
+
+### Input
+
+- `d`         -- direction
+- `cap`       -- union of a finite number of convex sets
+- `algorithm` -- (optional, default: "support_vector"): the algorithm to compute
+                 the support vector; if "support_vector", use the support
+                 vector of each argument; if "support_function" use the support
+                 function of each argument and evalute the support vector of only
```
schillic:
```suggestion
                 function of each argument and evaluate the support vector of only
```

src/UnionSet.jl
```diff
@@ -0,0 +1,258 @@
...
+### Output
+
+The array that holds the union of a finite number of convex sets.
+"""
+function array(cap::UnionSetArray{N, S})::Vector{S} where {N<:Real, S<:LazySet{N}}
+    return cap.array
+end
+
+"""
+    σ(d::AbstractVector{N}, cap::UnionSetArray{N}; algorithm="support_vector") where {N<:Real}
```
schillic:
```suggestion
    σ(d::AbstractVector{N}, cap::UnionSetArray{N}; [algorithm]="support_vector") where {N<:Real}
```

src/UnionSet.jl
```diff
@@ -0,0 +1,258 @@
...
+### Output
+
+The ambient dimension of the union of two convex sets.
+"""
+function dim(cup::Union)::Int
+    return dim(cup.X)
+end
+
+"""
+    σ(d::AbstractVector{N}, cap::UnionSet{N}; algorithm="support_vector") where {N<:Real}
```
schillic:
```suggestion
    σ(d::AbstractVector{N}, cap::UnionSet{N}; [algorithm]="support_vector") where {N<:Real}
```

src/UnionSet.jl
```diff
@@ -0,0 +1,258 @@
...
+can be obtained as the vector that maximizes the support function, i.e.
+it is sufficient to find the ``\\argmax(ρ(d, X₂), ρ(d, X₂), ...])`` and evaluate
+its support vector.
+
+The default implementation, with option `algorithm="support_vector"`, computes
+the support vector of all ``X₁, X₂, ...`` and then compares the support function using
+a dot product. If it happens that the support function can be more efficiently
+computed (without passing through the support vector), consider using the alternative
+`algorithm="support_function"` implementation, which evaluates the support function
+of each set directly and then calls only the support vector of only one of the ``Xᵢ``.
```
schillic:
```suggestion
of each set directly and then calls only the support vector of one of the ``Xᵢ``.
```

src/UnionSet.jl
schillic: `cap` → `cup` (all over this file)

src/UnionSet.jl
```diff
@@ -0,0 +1,258 @@
...
+"""
+function σ(d::AbstractVector{N}, cap::UnionSetArray{N}; algorithm="support_vector") where {N<:Real}
```
schillic: These algorithms can be implemented in constant memory, but let us keep yours for simplicity.

mforets: Do open an issue if you have an idea for improvement of this one, to modify `"support_function`, `"support_vector"` or add a new one.

schillic: See #993.

PR #923, opened 2018-11-22
#72 - Fix 'where' syntax in doc references

### LINKED ISSUE #72: Fix 'docstrings potentially missing' warnings

schillic:
Closes #72.

* [ ] check if all links to docs still work correctly


src/plot_recipes.jl
```diff
@@ -153,15 +153,15 @@ julia> plot([B1, B2], 1e-4);
 
 ```
 """
-@recipe function plot_lazyset(arr::Vector{<:LazySet}, ε::Float64;
+@recipe function plot_lazyset(arr::Vector{S}, ε::Float64;
```
mforets: while you are at this, `arr` is an ugly name, should we call it `array`, or `Xk`?

src/plot_recipes.jl
```diff
@@ -306,7 +306,7 @@ julia> plot(Singleton([0.5, 1.0]));
 end
 
 """
-    plot_singleton(arr::Vector{<:AbstractSingleton}; ...)
+    plot_singleton(arr::Vector{S}; ...) where {S<:AbstractSingleton}
```
mforets: these changes are motivated by style only? for me both are OK; if i had to choose, the one in the left is a bit shorter but maybe the notation in the right is more clean.. anyway 😄

schillic: No, I did not find a solution for these cases other than introducing a (useless) type parameter. But I rather wanted to close that issue, so I did that change. I kept them in a separate commit if we do not want to go that way (after all, it is weird to change the code just for documentation). But I do not find it that bad.

mforets: so simple cases like `function f(Vector{<:LazySet}) return 1 end` where giving problems?

schillic: Yes, they are not found by `Documenter` (at least I did not find a way).

src/plot_recipes.jl
schillic: done

Issue #127, opened 2018-01-09
More readable string output


schillic:
For nested set types the string output is not very helpful.
* Should we add a shorter output string function?
* Should it be default or optional?
* Do we want to have several levels (e.g., "print only sets up to two nesting levels", "print only the first 10 sets", ...)?

mforets:
i agree that for nested sets there is an ugly output and we can do better with `Base.show`. more readable would be like: 

```julia
julia> b = Ball2(ones(2), 1.)
A Ball2 of center [1.0, 1.0] and radius 1.0.
julia> b ⊕ b
MinkowskiSum of a Ball2 of center [1.0, 1.0] and radius 1.0 and a Ball2 of center [1.0, 1.0] and radius 1.0
```

or even

```julia
julia> b ⊕ b
(Ball2 of center [1.0, 1.0] and radius 1.0) ⊕ (Ball2 of center [1.0, 1.0] and radius 1.0)
```

schillic:
And what if we have deep nesting, e.g., `MinkowskiSum(MinkowskiSum(U, V), MinkowskiSum(X, Y))`?
What about array set types?

PR #925, opened 2018-11-22
WIP #127 - More readable string output

schillic:
See #127.

### DIFF
--- a/src/Ball1.jl
+++ b/src/Ball1.jl
@@ -1,5 +1,6 @@
 import Base: rand,
-             ∈
+             ∈,
+             show
 
 export Ball1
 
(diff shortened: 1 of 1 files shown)


schillic:
I am not sure if this is the expected behavior. In particular, `print` shows the long description only for arrays.
```julia
julia> x = [rand(Ball1), rand(Ball1)]
2-element Array{Ball1{Float64},1}:
 Ball1(c=[0.029414, -0.178756], r=0.351283)
 Ball1(c=[-0.65475, -0.885821], r=1.00673) 

julia> print(x)
Ball1{Float64}[Ball1 of center=[0.029414, -0.178756] and radius=0.351283,
               Ball1 of center=[-0.65475, -0.885821] and radius=1.00673]

julia> y = rand(Ball1)
Ball1(c=[-0.589741, 1.32808], r=0.671615)

julia> print(y)
Ball1(c=[-0.589741, 1.32808], r=0.671615)
```

mforets:
I think that the default printing is just fine.. I think that since we have docs and examples that should be enough for beginners to get the meaning. Another more "beginner-friendly" printing could be added as an optional module..?

PR #929, opened 2018-11-26
Add comparisons in floating point

mforets:


src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+
+This function falls back to `_leq(y, x)`, with type promotion if needed. See the
+documentation of `_leq` for further details.
+"""
+_geq(x::N, y::M) where {N<:Real, M<:Real} = _geq(promote(x, y)...)
+
+"""
+    isapproxzero(x::N; kwargs...) where {N<:Real}
+
+Returns `true` if the real `x` is greater or equal than `y` and false otherwise.
```
schillic:
```suggestion
Determine if `x` is approximately zero.
```

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+whenever `x` is equal to zero in the same numeric type as `x`.
+"""
+function isapproxzero(x::N; kwargs...) where {N<:Real}
+    return x == zero(N)
+end
+
+"""
+    isapproxzero(x::N; ztol::Real=ABSZTOL(N)) where {N<:AbstractFloat}
+
+Returns `true` if `x` is approximately zero `false` otherwise.
```
schillic:
```suggestion
Determine if `x` is approximately zero.
```

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+    return abs(x) < ztol
+end
+
+"""
+    _isapprox(x::N, y::N;
+              rtol::Real=Base.rtoldefault(N),
+              ztol::Real=ABSZTOL(N),
+              atol::Real=zero(N)) where {N<:AbstractFloat}
+
+Returns `true` if `x` is approximately equal than `y` and false otherwise.
```
schillic:
```suggestion
Determine if `x` is approximately equal to `y`.
```

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+
+### Input
+
+- `x`    -- number
+- `y`    -- another number (of the same numeric type than `x`)
```
schillic:
```suggestion
- `y`    -- another number (of the same numeric type as `x`)
```

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+### Algorithm
+
+This comparison is split into both `x` and `y` approximately zero checked using
+`isapproxzero(x, y)`, or `x ≈ y` checked using Julia's `isapprox(x, y)`. In the
+latter function we use zero absolute tolerance and `rtol` relative tolerance.
+"""
+function _isapprox(x::N, y::N;
+                   rtol::Real=Base.rtoldefault(N),
+                   ztol::Real=ABSZTOL(N),
+                   atol::Real=zero(N)) where {N<:AbstractFloat}
```
schillic: `atol` is never used.

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+### Output
+
+A boolean that is `true` iff `x ≈ y`.
+
+### Algorithm
+
+This comparison is split into both `x` and `y` approximately zero checked using
+`isapproxzero(x, y)`, or `x ≈ y` checked using Julia's `isapprox(x, y)`. In the
+latter function we use zero absolute tolerance and `rtol` relative tolerance.
+"""
```
schillic: Suggestion:
```
We first check if `x` and `y` are both approximately zero, using `isapproxzero(x, y)`.
If that fails, we check if `x ≈ y`, using Julia's `isapprox(x, y)`.
In the latter check we use zero absolute tolerance and `rtol` relative tolerance.
```

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+    end
+end
+
+"""
+    _leq(x::N, y::N;
+         rtol::Real=Base.rtoldefault(N),
+         ztol::Real=ABSZTOL(N),
+         atol::Real=zero(N)) where {N<:AbstractFloat}
+
+Returns `true` if the real `x` is smaller or equal than `y` and false otherwise.
```
schillic:
```suggestion
Determine if `x` is smaller than or equal to `y`.
```

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+
+### Input
+
+- `x`    -- number
+- `y`    -- another number (of the same numeric type than `x`)
```
schillic:
```suggestion
- `y`    -- another number (of the same numeric type as `x`)
```

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
+const ABSZTOL(N::Type{<:AbstractFloat}) = sqrt(eps(N))
+const ABSZTOL(N::Type{Rational{INNER}}) where {INNER} = zero(N)
+
+"""
+    _leq(x::N, y::N) where {N<:Real}
+
+Return `true` if the real `x` is smaller or equal than `y` and false otherwise.
```
schillic:
```suggestion
Determine if `x` is smaller than or equal to `y`.
```

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+
+This is a fallback implementation for numbers of type `Real`. If the arguments
+are floating point numbers, see `_leq(x::AbstractFloat, y::AbstractFloat)`.
+"""
+_leq(x::N, y::N) where {N<:Real} = x <= y
+
+"""
+    _leq(x::N, y::M) where {N<:Real, M<:Real}
+
+Return `true` if the real `x` is smaller or equal than `y` and false otherwise.
```
schillic:
```suggestion
Determine if `x` is smaller than or equal to `y`.
```

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+arguments to a common numeric type, returning them as a tuple. The conversion
+is such that the common type to which the values are converted can represent
+them as faithfully as possible.
+"""
+_leq(x::N, y::M) where {N<:Real, M<:Real} = _leq(promote(x, y)...)
+
+"""
+    _geq(x::N, y::N) where {N<:Real}
+
+Returns `true` if the real `x` is greater or equal than `y` and false otherwise.
```
schillic:
```suggestion
Determine if `x` is greater than or equal to `y`.
```

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+
+### Input
+
+- `x` -- number
+- `y` -- another number (of the same numeric type than `x`)
```
schillic:
```suggestion
- `y` -- another number (of the same numeric type as `x`)
```

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+
+This function falls back to `_leq(y, x)`. See the documentation of `_leq` for
+further details. 
+"""
+_geq(x::N, y::N) where {N<:Real} = _leq(y, x)
+
+"""
+    _geq(x::N, y::M) where {N<:Real, M<:Real}
+
+Returns `true` if the real `x` is greater or equal than `y` and false otherwise.
```
schillic:
```suggestion
Determine if `x` is greater than or equal to `y`.
```

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+
+This comparison is split into both `x` and `y` approximately zero checked using
+`isapproxzero(x, y)`, or `x ≈ y` checked using Julia's `isapprox(x, y)`. In the
+latter function we use zero absolute tolerance and `rtol` relative tolerance.
+"""
+function _isapprox(x::N, y::N;
+                   rtol::Real=Base.rtoldefault(N),
+                   ztol::Real=ABSZTOL(N),
+                   atol::Real=zero(N)) where {N<:AbstractFloat}
+    if isapproxzero(x, ztol=ztol) && isapproxzero(y, ztol=ztol)
```
schillic: I do not understand the motivation for this function.
Why should both numbers being close to zero be a special case? If both are close to ten, I would expect the same result.
I had expected something like `if isapproxzero(x-y, ztol=ztol)`.

mforets:
(*Note: This file is inspired from Polyhedra.jl, see [this file](https://github.com/JuliaPolyhedra/Polyhedra.jl/blob/master/src/comp.jl).* I've added this note on the beginning of this file.)

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+
+The `x <= y` comparison is split into `x < y` or `x ≈ y`; the latter is implemented
+by extending Juila's built-in `isapprox(x, y)` with an absolute tolerance that is
+used to compare against zero.
+"""
+function _leq(x::N, y::N;
+              rtol::Real=Base.rtoldefault(N),
+              ztol::Real=ABSZTOL(N),
+              atol::Real=zero(N)) where {N<:AbstractFloat}
+    return x < y || _isapprox(x, y, rtol=rtol, ztol=ztol, atol=atol)
```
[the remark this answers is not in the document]
mforets: No; the idea is that (approximate) equality between `x` and `y` is taken care by `_isapprox`.

src/comparisons.jl
```diff
@@ -0,0 +1,204 @@
...
+
+A boolean that is `true` iff `x ≈ 0`.
+
+### Algorithm
+
+It is considered that `x ≈ 0` whenever `x` (in absolute value) is smaller than the
+tolerance for zero, `ztol`.
+"""
+function isapproxzero(x::N; ztol::Real=ABSZTOL(N)) where {N<:AbstractFloat}
+    return abs(x) < ztol
```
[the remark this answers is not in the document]
mforets: I'm not sure; this function is for FP comparisons (note `N<:AbstractFloat`) where `0` is actually the minimum representable number,

```julia
julia> eps(1.0)
2.220446049250313e-16
```

src/comparisons.jl
```diff
@@ -87,9 +87,7 @@ A boolean that is `true` iff `x ≈ 0`.
 This is a fallback implementation for any real `x` such that `x ≈ 0` is `true`
 whenever `x` is equal to zero in the same numeric type as `x`.
 """
-function isapproxzero(x::N; kwargs...) where {N<:Real}
-    return x == zero(N)
-end
+isapproxzero(x::Real; kwargs...) = x == zero(N)
```
schillic: `N` is not defined here.

src/comparisons.jl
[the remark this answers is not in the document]
schillic: As long as `x < y`, `_isapprox` will not be called. The only case that is added is `x == y`. And no matter what the tolerance settings, if `x == y`, `_isapprox` will return `true` anyway. So IMO using `<=` is just more efficient.

src/comparisons.jl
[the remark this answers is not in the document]
schillic: I do not understand...
I would argue that `isapproxzero(0.0, ztol=0.0)` should return `true`.

src/comparisons.jl
mforets: Ouch, fixed. I should add a comparisons unit tests file.

schillic: definitely :wink:

src/comparisons.jl
mforets: Comparing to zero with default tolerances is a special case in Julia's `isapprox`, see the last paragraph in `?isapprox`. This function tries to combine `isapprox` with its default values and a branch for `x ≈ y ≈ 0` which includes `x == y == 0` but also admits a tolerance `ztol`.

mforets: > the argument atol is not used,

`atol` is passed to `isapprox`. We use an absolutel-like comparison (with `atol=TOL`) for example [here](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Approximations/iterative_refinement.jl#L89). (we can make this check more robust by using a relative comparison or a combination or relative and absolute).

mforets: 😮  below it should read

```julia
return isapprox(x, y, rtol=rtol, atol=atol)
```

src/comparisons.jl
mforets: thanks for pointing it out, see [this comment below](https://github.com/JuliaReach/LazySets.jl/pull/929/files#r247466251)

PR #931, opened 2018-11-28
#930 - Define issubset for EmptySet

### LINKED ISSUE #930: Define issubset for EmptySet
Currently this breaks in some cases such as `issubset(::EmptySet, ::Hyperrectangle)` because the default implementation `issubset(::LazySet, ::Hyperrectangle)` creates the `box_approximation` of the empty set.
We should fix both methods by adding a specific implementation for `EmptySet`.

schillic:
Closes #930.

* [x] ∅ ⊆ X
* [x] X ⊆ ∅
* [x] `box_approximation(∅)`

### DIFF
--- a/src/is_subset.jl
+++ b/src/is_subset.jl
@@ -530,8 +530,60 @@ function ⊆(∅::EmptySet{N}, X::LazySet{N}, witness::Bool=false
     return witness ? (true, N[]) : true
 end
 
-# disambiguity
+# disambiguation
 function ⊆(∅::EmptySet{N}, H::AbstractHyperrectangle{N}, witness::Bool=false
+          )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+    return witness ? (true, N[]) : true
+end
+
+"""
+    ⊆(X::LazySet{N}, ∅::EmptySet{N}, [witness]::Bool=false
+     )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+
+Check whether a set is contained in an empty set.
+
+### Input
+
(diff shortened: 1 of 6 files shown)


src/is_subset.jl
```diff
@@ -499,3 +503,40 @@ Check whether an interval is contained in another interval.
...
+function ⊆(∅::EmptySet{N}, X::LazySet{N}, witness::Bool=false
+          )::Bool where {N<:Real}
+    if witness
+        return (true, N[])
+    end
+    return true
+end
+
+# disambiguity
+function ⊆(∅::EmptySet{N}, H::AbstractHyperrectangle{N}, witness::Bool=false
```
mforets: sorry, why is that `EmptySet` is ambiguous with `AbstractHyperrectangle`?

schillic: The method
`⊆(∅::EmptySet{N}, S::LazySet{N})`
is ambiguous with
`⊆(S::LazySet{N}, H::AbstractHyperrectangle{N})`.

mforets: Got it, thanks. I wonder if Julia has some mechanism to ease this workaround, like allowing to invoke a given method. Maybe with [invoke](https://docs.julialang.org/en/v1/base/base/#Core.invoke).

schillic: I know, but `invoke` is less efficient, which is why I try to avoid it.

mforets: Less efficient in which sense? The call overhead should be negligible, no?

schillic: Not if you call this function very often. In this case I agree, though.

Issue #363, opened 2018-06-06
Add lazy decomposition


schillic:
Add the possibility to just project, not overapproximate. We want to have this in `Reachability`, where for the moment we use a workaround.

Candidate name: `:lazy`?
Should this be the default implementation?

Related: #345.

schillic:
I first thought that the implementation should always do this lazy projection as a first step and then apply overapproximation as a second step. But this might not be good because, e.g., box directions are implemented by support vector queries in high dimensions immediately, which might be a bit more efficient. So instead we should add a new function, which should be called `project`, rename the current `project` methods, and let them use the new function.

mforets:
> Add the possibility to just project, not overapproximate.

Ok, sounds good to me and your implementation [here](https://github.com/JuliaReach/Reachability.jl/pull/211) looks good to me. 

Perhaps i would add an optional keyword argument, `concrete`, with the default value `true`. (Or a default option `lazy`, with the default value `false`. I have no particular preference for one over the other ATM...)

> Should this be the default implementation?

I think we should keep the default non-lazy implementation for a decomposition. This is arguable, sure.

mforets:
> So instead we should add a new function, which should be called project, rename the current project methods, and let them use the new function.

Looks like we've got this project function in `decompositions.jl`, but it is not supposed to be part of the user API, since it is not exported. The current purpose of `project` is to "concretize" the given lazy set into small dimensions. I think it is fair to make it part of the user API.

schillic:
> I think it is fair to make it part of the user API.

Which one, the old one or the new one? I find `project` confusing for `concretize`, which is why I suggested to rename it (but I have no good name yet).

PR #935, opened 2018-12-05
#934 - Add exact projection method

### LINKED ISSUE #934: Add exact projection method
See also #363.

kostakoida:
Close #934 
related #363

### DIFF
--- a/src/Approximations/decompositions.jl
+++ b/src/Approximations/decompositions.jl
@@ -304,8 +304,8 @@ A set of type `set_type` representing an overapproximation of the projection of
                          n::Int=dim(S),
                          ε::Real=Inf
                         )::LazySet{N} where {N<:Real}
-    M = sparse(1:length(block), block, ones(N, length(block)), length(block), n)
-    return overapproximate(M * S, set_type)
+    lm = project(S, block, LinearMap, n)
+    return overapproximate(lm, set_type)
 end
 
 """
@@ -440,8 +440,8 @@ The template direction approximation of the projection of `S`.
                          directions::AbstractDirections{N},
                          n::Int=dim(S)
                         )::HPolytope where {N<:Real}
-    M = sparse(1:length(block), block, ones(N, length(block)), length(block), n)
-    return overapproximate(M * S, directions)
+    lm = project(S, block, LinearMap, n)
+    return overapproximate(lm, directions)
 end
 
 """
--- a/src/Approximations/decompositions.jl
+++ b/src/Approximations/decompositions.jl
@@ -228,7 +228,7 @@ function decompose(S::LazySet{N};
 
     if directions != nothing
         # template directions
-        # potentially defined option set_type is *ignored*       
+        # potentially defined option set_type is *ignored*
         block_start = 1
         @inbounds for bi in blocks
             push!(result, project(S, block_start:(block_start + bi - 1), directions(bi), n))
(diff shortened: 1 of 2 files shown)


src/Approximations/decompositions.jl
```diff
@@ -268,6 +268,42 @@ function decompose(S::LazySet{N};
     return CartesianProductArray(result)
 end
 
+
+"""
+    decompose_explicit(S::LazySet{N};
+                       blocks::AbstractVector{Int}=default_block_structure(S, set_type)
+                      )::CartesianProductArray where {N<:Real}
+
+Decompose a high-dimensional set into a Cartesian product of overapproximations
```
schillic: > overapproximations

No, this method is exact.

src/Approximations/decompositions.jl
```diff
@@ -268,6 +268,42 @@ function decompose(S::LazySet{N};
...
+- `blocks`      -- (optional, default: [2, …, 2] or [1, …, 1])
+                   block structure - a vector with the size of each block
+
+### Output
+
+A `CartesianProductArray` containing lazy low-dimensional exact
+projections.
+"""
+
+function decompose_explicit(S::LazySet{N};
```
schillic: I do not understand the name.
Also, I am not sure what you need this function for.

src/Approximations/decompositions.jl
```diff
@@ -442,3 +478,29 @@ The template direction approximation of the projection of `S`.
     M = sparse(1:length(block), block, ones(N, length(block)), length(block), n)
     return overapproximate(M * S, directions)
 end
+
+"""
+    project(S::LazySet{N},
+            block::AbstractVector{Int},
+            n::Int
+           )::HPolytope where {N<:Real}
```
schillic:
```suggestion
           )::LinearMap where {N<:Real}
```

src/Approximations/decompositions.jl
```diff
@@ -442,3 +478,29 @@ The template direction approximation of the projection of `S`.
...
+### Input
+
+- `S` -- set
+- `block` -- block structure - a vector with the dimensions of interest
+- `n` -- (optional, default: `dim(S)`) ambient dimension of the set `S`
+
+### Output
+
+The lazy linear map of the projection of `S`.
+"""
```
mforets: it doesn't hurt to add an example here.

it also seems that the input parameter `n` can be removed: if `n` is different  from `dim(S)`, then `M * S` will break anyways:

```julia
julia> rand(2, 2) * rand(Interval)
ERROR: AssertionError: a linear map of size (2, 2) cannot be applied to a set of dimension 1
Stacktrace:
 [1] LinearMap{Float64,Interval{Float64,IntervalArithmetic.Interval{Float64}},Float64,Array{Float64,2}}(::Array{Float64,2}, ::Interval{Float64,IntervalArithmetic.Interval{Float64}}) at /Users/forets/.julia/dev/LazySets/src/LinearMap.jl:32
 [2] Type at /Users/forets/.julia/dev/LazySets/src/LinearMap.jl:40 [inlined]
 [3] *(::Array{Float64,2}, ::Interval{Float64,IntervalArithmetic.Interval{Float64}}) at /Users/forets/.julia/dev/LazySets/src/LinearMap.jl:65
 [4] top-level scope at none:0
```

schillic: We added `n` to all `project` methods because the caller might already know `n`. In general, can has to be computed (e.g., for a `CartesianProductArray`).

mforets: Ok. I forgot that we did this as a workaround for caching the dimension of the set. The problem is that passing anything different from `dim(S)` will produce an error. In general computing `dim(..)` shouldn't take more than a few microseconds, unless the set is very nested.

schillic: My point is: That discussion is completely independent from this PR. If you want to remove `n`, we should remove it everywhere.

src/Approximations/decompositions.jl
```diff
@@ -226,7 +226,13 @@ function decompose(S::LazySet{N};
...
     result = Vector{LazySet{N}}()
 
-    if directions != nothing
+    if set_type == LinearMap
+        block_start = 1
+        @inbounds for bi in blocks
+            push!(result, project(S, block_start:(block_start + bi - 1), n))
+            block_start += bi
+    end
+    elseif directions != nothing
```
schillic: I would not add another condition here.
Just use the existing code and pass `set_type` to the `project` method.

schillic:
It would be good to add a small test in `unit_decompose.jl` (I guess just copy lines 21 and 22 and replace the `set_type`).

src/Approximations/decompositions.jl
```diff
@@ -442,3 +443,30 @@ The template direction approximation of the projection of `S`.
...
+
+- `S` -- set
+- `block` -- block structure - a vector with the dimensions of interest
+- `n` -- (optional, default: `dim(S)`) ambient dimension of the set `S`
+
+### Output
+
+The lazy linear map of the projection of `S`.
+"""
+@inline function project(S::LazySet{N},
```
mforets: the signature and docstring do not match (`<ε::Real=Inf`)

src/Approximations/decompositions.jl
```diff
@@ -442,3 +443,30 @@ The template direction approximation of the projection of `S`.
...
     return overapproximate(M * S, directions)
 end
+
+"""
+    project(S::LazySet{N},
+            block::AbstractVector{Int},
+            n::Int
+           )::LinearMap where {N<:Real}
+
+Project a high-dimensional set to a low-dimensional set.
```
schillic: Maybe while you are at it, add a more precise description of the method (but I am not sure if this is better or worse):

    Project a high-dimensional set to a low-dimensional set by a lazy linear map.

PR #940, opened 2018-12-15
#750 - Allow unbounded ρ for Hyperplane/HalfSpace

schillic:
See #750.

This PR brings only the support function.
Intuitively, it is clear what the support vector should look like, but I do not see the general formula.

### DIFF
--- a/docs/src/lib/representations.md
+++ b/docs/src/lib/representations.md
@@ -155,6 +155,7 @@ Inherited from [`LazySet`](@ref):
 HalfSpace
 LinearConstraint
 dim(::HalfSpace)
+ρ(::AbstractVector{N}, ::HalfSpace{N}) where {N<:Real}
 σ(::AbstractVector{N}, ::HalfSpace{N}) where {N<:Real}
 ∈(::AbstractVector{N}, ::HalfSpace{N}) where {N<:Real}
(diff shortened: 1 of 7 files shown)


mforets:
Something that we could add is a note on the algorithm behind `σ_helper`.

It is a consequence [of weak-duality in LPs](https://en.wikipedia.org/wiki/Linear_programming#Duality): if the primal is unbounded then the dual is infeasible. Since there is only 1 constraint, the feasible set of the dual problem is `ay == d`, `y >= 0` with `a` and `d` being the normal vector of the halfspace and `d` the given direction respectively (the objective function is `by`). It is easy to see that this problem is infeasible whenever `a` is not parallel to `d`, or they are parallel with negative combination coefficient (-1, after normalization).

mforets:
> Intuitively, it is clear what the support vector should look like, but I do not see the general formula.

Can't we just let:

```julia
function σ(d::AbstractVector{N}, hp::Hyperplane{N})::N where {N<:Real}
    v, unbounded = σ_helper(d, hp, error_unbounded=true)
    return v
end
```

schillic:
> Something that we could add is a note on the algorithm behind σ_helper.

Done.

Issue #561, opened 2018-08-19
convex_hull of polytopes does not remove redundant vertices


schillic:
Example:
```julia
julia> using LazySets, Polyhedra
julia> v1 = [1., 0.]; v2 = [0., 1.]; v3 = [-1., 0.]; v4 = [0., -1.]; v5 = [0., 0.];
julia> p1 = VPolytope([v1, v2, v5]);
julia> p2 = VPolytope([v3, v4]);
julia> ch = convex_hull(p1, p2);
julia> vl = vertices_list(ch)
5-element Array{Array{Float64,1},1}:
 [1.0, 0.0] 
 [0.0, 1.0] 
 [0.0, 0.0] 
 [-1.0, 0.0]
 [0.0, -1.0]
```
I would say that `v5` should not occur in the hull, at least not in its `vertices_list`.

schillic:
On the other hand, even the constructor of `VPolytope` does not remove redundant vertices.

mforets:
Seems to be fixed now, see related PR #944.

schillic:
No, it is not fixed yet. It either needs to be fixed upstream or we need to merge that PR :wink:

schillic:
There is also this:

> the constructor of VPolytope does not remove redundant vertices.

Here is a proposal for a default constructor, but we have to make sure that we pass `false` internally whenever we know that no redundant vertices exist.

```julia
    function VPolytope{N}(vertices::Vector{Vector{N}};
                          removeredundancy::Bool=true) where {N<:Real}
        if removeredundancy
            @assert isdefined(@__MODULE__, :Polyhedra) "redundancy removal " *
                "needs the package 'Polyhedra' loaded"
            P = polyhedron(Polyhedra.vrep(vertices))
            removevredundancy!(P)
            vertices = P.vrep.points.points
        end
        return new{N}(vertices)
    end
```

mforets:
Ok, got it. I like the idea of making `VPolytope` to remove redundant vertices, which goes in line with the similar constructor for `VPolygon`.

The "drawback" is that `Polyhedra` needs to be loaded (and also `CDDLib` to make `removevredundancy!` fast for high dims) every time one creates `VPolytope`, though-- there are of course other uses of `VPolytope` without `Polyhedra`, such as computing support vectos--.  Anyway those are first-time-to-compile overhead and the kwarg to the constructor is useful to suppress that calls.

---

[...]

PR #944, opened 2018-12-17
#943 - Update convexhull

### LINKED ISSUE #943: Call removevredundancy! in convexhull
See https://github.com/JuliaPolyhedra/Polyhedra.jl/issues/142.

See also #941.

mforets:
Closes #943.

### DIFF
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -499,7 +499,8 @@ The `HPolyhedron` (resp. `HPolytope`) obtained by the concrete convex hull of
 
 ### Notes
 
-It is suggested to use the `CDDLib.Library()` backend.
+For performance reasons, it is suggested to use the `CDDLib.Library()` backend
+for the `convex_hull`.
 
 For further information on the supported backends see
 [Polyhedra's documentation](https://juliapolyhedra.github.io/Polyhedra.jl/).
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -499,6 +499,8 @@ The `HPolyhedron` (resp. `HPolytope`) obtained by the concrete convex hull of
 
 ### Notes
 
+It is suggested to use the `CDDLib.Library()` backend.
+
 For further information on the supported backends see
 [Polyhedra's documentation](https://juliapolyhedra.github.io/Polyhedra.jl/).
 """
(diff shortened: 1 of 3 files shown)


mforets:
It is intended that `removevredundnacy!` is used outside `convexhull`, so i suggest to use this branch after we figure what was is making it to fail.

mforets:
Ok, found the error: the problem explained in the commented part seems to be fixed now:

```julia
        # Note: The redundant vertex v5 is not removed (see #561).
        # This test can be removed (and the length above should be corrected)
        # when that issue is resolved.
        @test length(vl) == 5 && v5 ∈ vl
```

test/unit_Polytope.jl
```diff
@@ -229,11 +229,8 @@ if test_suite_polyhedra
         p2 = VPolytope([v3, v4])
         ch = convex_hull(p1, p2)
         vl = vertices_list(ch)
+        @test length(vl) == 4
         @test v1 ∈ vl && v2 ∈ vl && v3 ∈ vl && v4 ∈ vl
```
schillic: These two lines can now be made more readable with `ispermutation(vl, [v1, v2, v3, v4])`.

src/HPolyhedron.jl
```diff
@@ -499,15 +499,18 @@ The `HPolyhedron` (resp. `HPolytope`) obtained by the concrete convex hull of
 
 ### Notes
 
+It is suggested to use the `CDDLib.Library()` backend.
```
schillic: Why is that?

schillic:
I added above that this PR closes #561, but now I see that there is a second issue there.

> the constructor of `VPolytope` does not remove redundant vertices.

I will hence add an optional argument `prune::Bool=true` to the default constructor.

src/HPolyhedron.jl
mforets: I saw that `CDDLib` is faster than the default backend for `convexhull` by a large margin.
Example:

```julia
using LazySets, Polyhedra

for n in [2, 4, 10]
    @time begin
        V1 = VPolytope([rand(n) for i in 1:100])
        V2 = VPolytope([rand(n) for i in 1:100])
        convex_hull(V1, V2) # uses Polyhedra's default library
    end
end

0.414625 seconds (4.13 M allocations: 230.331 MiB, 12.61% gc time)
# ?
# ?

using CDDLib

for n in [2, 4, 10]
    @time begin
        V1 = VPolytope([rand(n) for i in 1:100])
        V2 = VPolytope([rand(n) for i in 1:100])
        convex_hull(V1, V2, backend=CDDLib.Library())
    end
end

  0.086622 seconds (132.72 k allocations: 6.416 MiB)
  0.033711 seconds (15.75 k allocations: 856.891 KiB)
  0.511388 seconds (17.09 k allocations: 1.092 MiB, 1.62% gc time)
```

(times are shown on the second pass, and when there is a `?` is because i stopped the calculation)

Issue #950, opened 2018-12-18
Revise convex_hull


schillic:
The concrete `convex_hull` methods are currently documented all in one place. I think we should add individual docs to each method.
Note that we have unary and binary versions of this method. The binary versions should go to `binary_functions.md`.
Also, should we reorganize the binary methods into the file `convex_hull_algorithms.jl`, or maybe into a new file `concrete_convex_hull.jl`? This could simplify working on #948 and #949 to catch method ambiguities.
We also need to add that function to our ambiguity tests (`unit_binary_operations.jl`).

mforets:
> Also, should we reorganize the binary methods into the file convex_hull_algorithms.jl, or maybe into a new file concrete_convex_hull.jl?

👍  I have thought of reorganizing the library into subfolders in the past (`convex_hull.jl` is not a problem but `intersection.jl` vs `Intersection.jl` is a problem). I agree to either use subfolders or rename `concrete_...` 

- `concrete_intersection.jl`
- `concrete_convex_hull.jl`
- `concrete_minkowski_sum.jl` (to be added)
...

I'm preparing a PR with the renaming `concrete_convex_hull.jl`.

schillic:
I think the suggested steps have been resolved by now.

Issue #904, opened 2018-11-10
Override copy for lazyset objects


mforets:
See eg. [here](https://github.com/JuliaReach/LazySets.jl/pull/901/files#diff-795b585c4704738125c4e81ad4bff018R446).

PR #953, opened 2018-12-18
#950 - Refactor convex_hull_algorithms

### LINKED ISSUE #949: convex_hull method for VPolygon
Currently a user has to unwrap the vertices of a `VPolygon` to apply the `convex_hull` method.

### LINKED ISSUE #952: Unary convex hull for VPolytope
Proposal: add the same methods as that proposed in https://github.com/JuliaReach/LazySets.jl/issues/949 but for a `VPolytope`.

mforets:
Closes #949.
Closes #952.

See #950.

- [x] refactor `convex_hull_algorithms.jl` into `concrete_convex_hull.jl`
- [x] add binary version of `convex_hull` to `binary_functions.md`
- [x] update `unit_binary_operations.jl`
- [x] add `convex_hull(::VPolygon)`
- [x]  add `convex_hull(::VPolytope)`
- [x] udpate docs
- [x] update / add tests
- [x] requires https://github.com/JuliaReach/LazySets.jl/issues/904 add copy methods for common set reps


schillic:
As I wrote [here](https://github.com/JuliaReach/LazySets.jl/issues/949#issuecomment-448500544):

> Thinking about it, it does not really make sense to apply a convex hull to a single set. The method should rather be called remove_redundancy or something like that. In the background this can call to convex_hull, but a user would not have to know that.

mforets:
> Thinking about it, it does not really make sense to apply a convex hull to a single set.

Agreed.

> The method should rather be called remove_redundancy or something like that.

I'm adding that stuff on this PR too 👍 

> Furthermore, I suggest to extend the convex_hull function to also accept higher-dimensional vertices.

I like this idea. We can put that code in `concrete_convex_hull.jl`; i'll have to change the `@assert length(points[1]) == 2` introduced recently.

src/VPolytope.jl
```diff
@@ -250,6 +249,34 @@ function cartesian_product(P1::VPolytope{N}, P2::VPolytope{N};
...
+    remove_redundant_vertices(P::VPolytope{N};
+                              backend=default_polyhedra_backend(P, N))::VPolytope{N} where {N}
+
+Return the polytope obtained by taking the convex hull of the vertices of the
+given polytope.
+
+### Input
+
+- `P`         -- polytope in vertex representation
+- `backend`   -- (optional, default: `default_polyhedra_backend(P1, N)`) the polyhedral
```
schillic: three spaces

src/VPolytope.jl
```diff
@@ -250,6 +249,34 @@ function cartesian_product(P1::VPolytope{N}, P2::VPolytope{N};
     return VPolytope(Pcp)
 end
 
+"""
+    remove_redundant_vertices(P::VPolytope{N};
+                              backend=default_polyhedra_backend(P, N))::VPolytope{N} where {N}
+
+Return the polytope obtained by taking the convex hull of the vertices of the
+given polytope.
```
schillic: I find this summary confusing. The function should remove redundant vertices. (same for `VPolygon`)

src/VPolygon.jl
```diff
@@ -14,10 +16,10 @@ Type that represents a polygon by its vertices.
 
 ### Notes
 
-The constructor of `VPolygon` runs a convex hull algorithm, and the given
-vertices are sorted in counter-clockwise fashion.
-The constructor flag `apply_convex_hull` can be used to skip the computation of
-the convex hull.
+The constructor of `VPolygon` runs a convex hull algorithm on its vertices by
+default, to remove the possibly redundant constraints. The vertices are sorted in
```
schillic: > to remove the possibly redundant constraints

This is confusing for a `VPolygon`. It sounds as if we compute the constraints.

src/VPolytope.jl
mforets: changing to: "Return the polytope obtained by removing the redundant vertices of the given polytope".

src/VPolygon.jl
mforets: changed to: "The constructor of `VPolygon` runs a convex hull algorithm on its vertices by
default, to remove the possibly redundant vertices."

schillic: Nothing has changed yet :wink:

mforets: could you check again now?

mforets:
Ready from my side. I don't know why does it keep failing, since `remove_redundant_vertices` for a `VPolytope` are documented [here](https://github.com/JuliaReach/LazySets.jl/pull/953/files#diff-835c7b7575bf0754201cebca0a3b5939R489).

PR #956, opened 2018-12-24
#605 - Function for checking boundedness

### LINKED ISSUE #605: Function for checking boundedness
Add a function that checks if an `HPolytope` (and an `AbstractHPolygon`) is bounded.

An efficient but only necessary check is to look at the number of constraints. [In n dimensions you need at least n+1 constraints.](https://en.wikipedia.org/wiki/Feasible_region#Bounded_and_unbounded_feasible_sets)

A naive approach could be to ask for the support vector in the positive and negative unit directions.

schillic:
Closes #605.

This PR adds `isbounded(::Set)`.
There is a default implementation (that uses `isbounded_unit_dims`), but only the implementations of `ExponentialProjectionMap`, `HPolyhedron`, `Intersection(Array)`, and `LinearMap` may fall back to that default if some sufficient/necessary checks fail. For the other set types, the check is trivial (or recursive for operations).

<strike>Some tests do not work before merging #940, so they currently throw an error. Whatever gets merged last has to update those tests.</strike>

@mforets: Can you please check the implementations for `LinearMap`, `ExponentialMap`, and `ExponentialProjectionMap` if they are correct or if they can be improved?

### DIFF
--- a/src/ExponentialMap.jl
+++ b/src/ExponentialMap.jl
@@ -518,7 +518,7 @@ Determine whether an exponential projection map is bounded.
 
 We first check if the wrapped set is bounded or if the left or right projection
 matrix is zero.
-Otherwise, we check boundedness via `isbounded_unit_dims`.
+Otherwise, we check boundedness via [`isbounded_unit_dimensions`](@ref).
 """
 function isbounded(eprojmap::ExponentialProjectionMap)::Bool
     if isbounded(eprojmap.X)
@@ -527,7 +527,7 @@ function isbounded(eprojmap::ExponentialProjectionMap)::Bool
     if iszero(eprojmap.projspmexp.L) || iszero(eprojmap.projspmexp.R)
         return true
     end
-    return isbounded_unit_dims(eprojmap)
+    return isbounded_unit_dimensions(eprojmap)
 end
 
 """
--- a/src/ExponentialMap.jl
+++ b/src/ExponentialMap.jl
@@ -516,8 +516,8 @@ Determine whether an exponential projection map is bounded.
 
 ### Algorithm
 
-We first check if the wrapped set is bounded or if the left or right projection
-matrix is zero.
+We first check if the left or right projection matrix is zero or the wrapped set
+is bounded.
 Otherwise, we check boundedness via [`isbounded_unit_dimensions`](@ref).
 """
 function isbounded(eprojmap::ExponentialProjectionMap)::Bool
--- a/src/ExponentialMap.jl
+++ b/src/ExponentialMap.jl
@@ -276,6 +276,23 @@ function ρ(d::AbstractVector{N}, em::ExponentialMap{N}) where {N<:Real}
 end
 
 """
+    isbounded(em::ExponentialMap)::Bool
+
+Determine whether an exponential map is bounded.
+
+### Input
+
+- `em` -- exponential map
+
+### Output
+
+`true` iff the exponential map is bounded.
+"""
+function isbounded(em::ExponentialMap)::Bool
+    return isbounded(em.X)
+end
+
+"""
     ∈(x::AbstractVector{N}, em::ExponentialMap{N})::Bool where {N<:Real}
 
 Check whether a given point is contained in an exponential map of a convex set.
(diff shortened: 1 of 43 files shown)


src/LazySet.jl
```diff
@@ -103,6 +104,57 @@ Alias for the support vector σ.
...
+
+- `S` -- set
+
+### Output
+
+`true` iff the set is bounded.
+
+### Algorithm
+
+We check boundedness via `isbounded_unit_dims`.
```
```suggestion
 We check boundedness via `[isbounded_unit_dims(::LazySet)](@ref)`.
```

src/ExponentialMap.jl
```diff
@@ -484,6 +501,35 @@ function σ(d::AbstractVector{N},
...
+
+We first check if the wrapped set is bounded or if the left or right projection
+matrix is zero.
+Otherwise, we check boundedness via [`isbounded_unit_dimensions`](@ref).
+"""
+function isbounded(eprojmap::ExponentialProjectionMap)::Bool
+    if isbounded(eprojmap.X)
+        return true
+    end
+    if iszero(eprojmap.projspmexp.L) || iszero(eprojmap.projspmexp.R)
```
mforets: Swap these conditions? These `if` clauses, or write it like this?

```julia
if iszero(eprojmap.projspmexp.L) || iszero(eprojmap.projspmexp.R) || isbounded(eprojmap.X)
```m
Even if it can fail most times, in general it will be easier to check if `L` or `R` are empty than boundedness of `X`.

File: src/ExponentialMap.jl [hunk 3]
```diff
@@ -484,6 +501,35 @@ function σ(d::AbstractVector{N},
...
+
+`true` iff the exponential projection map is bounded.
+
+### Algorithm
+
+We first check if the wrapped set is bounded or if the left or right projection
+matrix is zero.
+Otherwise, we check boundedness via [`isbounded_unit_dimensions`](@ref).
+"""
+function isbounded(eprojmap::ExponentialProjectionMap)::Bool
```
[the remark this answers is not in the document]
schillic: If you want you can add your formulas.
I was more worried that maybe there is a simpler check that does not require the overkill `isbounded_unit_dimensions`.

src/ExponentialMap.jl
schillic: Depends: if `X` is of basic set type different from `HPolyhedron`, the check is constant-time. But okay!

src/ExponentialMap.jl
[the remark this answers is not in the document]
mforets: I think that knowing that only knowing that `X` is unbounded (and the coefficients of `M`) is not enough; one needs to know in which directions `X` is unbounded. For this reason i think that we won't have a formula that works for any `LazySet` `X`.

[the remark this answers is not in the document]
schillic: Okay. If you can think of another sufficient or necessary check, we can add it.
For `ExponentialMap` I am not sure if the conditions are really sufficient *and* necessary.

src/LazySet.jl
```diff
@@ -103,6 +104,57 @@ Alias for the support vector σ.
...
+`true` iff the set is bounded in each unit dimension.
+
+### Algorithm
+
+This function performs ``2n`` support function checks, where ``n`` is the
+ambient dimension of `S`.
+"""
+function isbounded_unit_dimensions(S::LazySet{N})::Bool where {N<:Real}
+    n = dim(S)
+    for i in 1:n
```
```suggestion
     @inbounds for i in 1:n
```

schillic:
Do you agree that for `ExponentialMap` the condition `isbounded(em.X)` is sufficient and necessary?

mforets:
> Do you agree that for ExponentialMap the condition isbounded(em.X) is sufficient and necessary?

Yes, it follows from the invertibility of `M := exp(A)` for any matrix `A`, see the [properties of the matrix exponential](https://en.wikipedia.org/wiki/Matrix_exponential).

Indeed, suppose by contradiction that the set `Y = MX` is bounded but the set `X` is unbounded. Then there exists a sequence of elements of `X`, `x^j` for `j=1,2,...` such that `||x^j|| -> oo`. Let `y^j = Mx^j` for all `j`. Then since `M` is invertible, `x^j = inv(M) * y^j` and thus `||x^j|| <= ||inv(M)|| ||y^j||`, and the right-hand side is bounded by hypothesis, thus the sequence `||x^j||` is bounded, a contradiction. 

We can use this condition in `LinearMap` and `ExponentialProjectionMap` as well.

PR #967, opened 2019-01-02
Pull out remaining VPolytope funcs from load polyhedra

mforets:


mforets:
Thanks for reviewing!

Issue #974, opened 2019-01-06
tohrep and tovrep do not preserve numeric type


schillic:
```julia
julia> N = Float32
Float32

julia> H = BallInf(N[0, 0], N(1))
BallInf{Float32}(Float32[0.0, 0.0], 1.0f0)

julia> P = convert(VPolytope, H)
VPolytope{Float32}(Array{Float32,1}[[1.0, 1.0], [-1.0, 1.0], [1.0, -1.0], [-1.0, -1.0]])

julia> tohrep(P)
HPolytope{Float64}(LazySets.HalfSpace{Float64}[
    HalfSpace{Float64}([-0.0, 0.333333], 0.333333),
    HalfSpace{Float64}([0.181818, -0.0], 0.181818),
    HalfSpace{Float64}([-0.0, -0.15], 0.15),
    HalfSpace{Float64}([-0.139535, -0.0], 0.139535)])
```

schillic:
For the record: We decided to accept this behavior. `CDDLib` only supports `Float64` and `Rational{BigInt}`, and it does not make sense to convert back to a different type on our end. The default `Polyhedra` backend also works for `Rational{Int}`.

PR #958, opened 2018-12-26
#416 - Conversion from Zonotope to HPolytope

schillic:
Closes #416.

PR #975, opened 2019-01-06
#972 - Subset check in polyhedron

### LINKED ISSUE #972: Subset check in HPolytope
Checking inclusion of a set `S` in a polytope `P` with `m` constraints is possible with `m` support function queries.

```julia
function ⊆(S::LazySet{N}, P::HPoly{N}) where {N<:Real}
    for H in constraints_list(P)
        if ρ(H.a, S) > H.b
            return false
        end
    end
    return true
end
```

schillic:
Closes #972.

A unit test currently fails due to #974. This will automatically be resolved by merging #958 first.


src/is_subset.jl
```diff
@@ -195,6 +195,69 @@ function ⊆(P::AbstractPolytope{N}, S::LazySet{N}, witness::Bool=false
...
+each direction of the constraints of ``P``.
+
+For witness generation, we use the support vector in the first direction where
+the above check fails.
+"""
+function ⊆(S::LazySet{N},
+           P::Union{AbstractPolytope{N}, HPolyhedron{N}, HalfSpace{N}},
+           witness::Bool=false
+          )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+    for H in constraints_list(P)
```
```suggestion
    @inbounds for H in constraints_list(P)
```

src/is_subset.jl
```diff
@@ -195,6 +195,69 @@ function ⊆(P::AbstractPolytope{N}, S::LazySet{N}, witness::Bool=false
...
+        if ρ(H.a, S) > H.b
```
```suggestion
        if _geq(ρ(H.a, S), H.b)
```

mforets: actually i'm not sure, if you want strict inequality we can leave it as is

mforets:
Strangely enough, `⊈` works in 1.0 but not in 0.6, if i checkout this branch:

v.1.0

```julia
julia> N = Float64
Float64

julia> P = rand(VPolytope)
VPolytope{Float64}(Array{Float64,1}[[-1.42147, -1.23376], [0.769873, 1.68362], [1.03484, -0.991754], [1.92539, -0.605125], [-1.16005, -0.990606], [0.438257, 0
.598263]])

julia> BallInf(N[0, 0], N(1)) ⊆ P
false

julia> BallInf(N[0, 0], N(1.01)) ⊈ P
true
```

v0.6

```julia
INFO: Recompiling stale cache file /Users/forets/.julia/lib/v0.6/LazySets.ji for module LazySets.
WARNING: Method definition __init__() in module LazySets at /Users/forets/.julia/v0.6/Requires/src/init.jl:9 overwritten at /Users/forets/.julia/v0.6/Requires
/src/init.jl:9.

julia> N = Float64
Float64

julia> P = rand(VPolytope)
LazySets.VPolytope{Float64}(Array{Float64,1}[[0.636787, -0.155192], [-0.116778, -0.27157], [-1.2512, 0.794358]])

julia> BallInf(N[0, 0], N(1)) ⊆ P
false

julia> BallInf(N[0, 0], N(1.01)) ⊈ P
ERROR: MethodError: no method matching ⊈(::LazySets.BallInf{Float64}, ::LazySets.VPolytope{Float64})
```

schillic:
> Strangely enough, ⊈ works in 1.0 but not in 0.6

Yeah, meanwhile we have #1033 for that.

src/is_subset.jl
schillic: Yes, I want to check the negation of `<=`.

src/is_subset.jl
schillic: I incorporated the change in my rebase.

PR #978, opened 2019-01-08
#964 - Check if polytope is bounded on construction

### LINKED ISSUE #964: Check if polytope is bounded on construction
See https://github.com/JuliaReach/LazySets.jl/issues/605.

schillic:
Closes #964.

### DIFF
--- a/src/HPolygon.jl
+++ b/src/HPolygon.jl
@@ -28,7 +28,6 @@ Use `addconstraint!` to iteratively add the edges in a sorted way.
   -- default constructor
 - `HPolygon()`
   -- constructor with no constraints
-- `HPolygon(S::LazySet)` -- constructor from another set
 """
 struct HPolygon{N<:Real} <: AbstractHPolygon{N}
     constraints::Vector{LinearConstraint{N}}
@@ -66,16 +65,6 @@ HPolygon{N}() where {N<:Real} = HPolygon{N}(Vector{LinearConstraint{N}}())
 
 # constructor for an HPolygon with no constraints of type Float64
 HPolygon() = HPolygon{Float64}()
-
-# conversion constructor
-function HPolygon(S::LazySet; validate_boundedness::Bool=false)
-    P = convert(HPolygon, S)
-    if validate_boundedness
-        # trigger boundedness check in constructor
-        HPolygon(P.constraints; validate_boundedness=true)
-    end
-    return P
-end
 
 
 # --- LazySet interface functions ---
--- a/src/HPolygon.jl
+++ b/src/HPolygon.jl
@@ -29,24 +29,31 @@ struct HPolygon{N<:Real} <: AbstractHPolygon{N}
 
     # default constructor that applies sorting of the given constraints
     function HPolygon{N}(constraints::Vector{LinearConstraint{N}};
-                         sort_constraints::Bool=true) where {N<:Real}
+                         sort_constraints::Bool=true,
+                         validate_boundedness::Bool=false) where {N<:Real}
         if sort_constraints
             sorted_constraints = Vector{LinearConstraint{N}}()
             sizehint!(sorted_constraints, length(constraints))
             for ci in constraints
                 addconstraint!(sorted_constraints, ci)
             end
-            return new{N}(sorted_constraints)
+            P = new{N}(sorted_constraints)
         else
-            return new{N}(constraints)
+            P = new{N}(constraints)
         end
+        @assert (!validate_boundedness ||
+                 LazySets.validate_boundedness(P)) "the polygon is not bounded"
+        return P
     end
 end
 
 # convenience constructor without type parameter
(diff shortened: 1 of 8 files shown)


mforets:
Here i have an API / aesthetic concern: i see no need to add a new method `validate_boundedness`, since we already have `isbounded`. We do need a new name for the keyword argument, which following other conventions throughout could be called `check_boundedness`.

schillic:
We need it because `isbounded(::HPolytope)` does *not* perform any checks

mforets:
Ok. What about the keyword in the constructor? I prefer `check_boundedness` because it's shorter 😄

src/HPolygon.jl
```diff
@@ -29,24 +29,31 @@ struct HPolygon{N<:Real} <: AbstractHPolygon{N}
...
                 addconstraint!(sorted_constraints, ci)
             end
-            return new{N}(sorted_constraints)
+            P = new{N}(sorted_constraints)
         else
-            return new{N}(constraints)
+            P = new{N}(constraints)
         end
+        @assert (!validate_boundedness ||
+                 LazySets.validate_boundedness(P)) "the polygon is not bounded"
```
mforets: i think that `LazySets.` is not needed here

mforets: i don't understand the logic behind

```julia
 @assert (!validate_boundedness ||
                 LazySets.validate_boundedness(P)) "the polygon is not bounded"
```

by default `validate_boundedness = false`, meaning that one doesn't want to check if the input polygon is indeed bounded. but then `!validate_boundedness` evaluates to `true` and `LazySets.validate_boundedness(P)` is called.

src/HPolygon.jl
```diff
@@ -29,24 +29,31 @@ struct HPolygon{N<:Real} <: AbstractHPolygon{N}
 
     # default constructor that applies sorting of the given constraints
     function HPolygon{N}(constraints::Vector{LinearConstraint{N}};
-                         sort_constraints::Bool=true) where {N<:Real}
+                         sort_constraints::Bool=true,
```
mforets: i think that we are missing to add `validate_boundedness` to the docstring

src/HPolygon.jl
```diff
@@ -55,7 +62,14 @@ HPolygon{N}() where {N<:Real} = HPolygon{N}(Vector{LinearConstraint{N}}())
 HPolygon() = HPolygon{Float64}()
 
 # conversion constructor
-HPolygon(S::LazySet) = convert(HPolygon, S)
+function HPolygon(S::LazySet; validate_boundedness::Bool=false)
```
mforets: this function is strange. i think that we should eliminate it.

src/HPolygon.jl
schillic: > i think that LazySets. is not needed here

true

src/HPolygon.jl
schillic: The reason we have them is probably this warning in v0.7:
```julia
julia> H = Hyperrectangle(low=N[-1, -1], high=N[1, 1]);

julia> HPolygon(H)
┌ Warning: Constructors no longer fall back to `convert`.
A constructor `HPolygon(::Hyperrectangle{Float64})` should be defined instead.
```
If you still find it necessary, let us do this change somewhere else.

src/HPolygon.jl
schillic: > but then `!validate_boundedness` evaluates to `true`

correct

> and `LazySets.validate_boundedness(P)` is called

incorrect: The `||` is already satisfied by `true` above, so short-circuiting will terminate the evaluation.

src/HPolygon.jl
mforets: these constructors are old. i wrote that we should remove them in the past but i don't find the issue.

schillic: We lose a feature if we remove them. Why do we want that?

src/HPolygon.jl
mforets: > fewer names sounds better

yes. but since one is a kwarg and another a function, we can have them distinct, and we could use:

- `check_boundedness` in the kwarg
-  `validate(S, isbounded)` for validation

for the function i really like: `isbounded(S, use_type_assumption::Bool=true)`, so no new function name needs to be added and one can control the behavior if `isbounded` in a program with one external flag.

src/HPolygon.jl
mforets: `convert(HPolygon, S)` should be used in favor of `HPolygon(S)` IMO. we have these `convert` aliases only for the `HPolygon`s.

src/HPolygon.jl
schillic: > for the function i really like: `isbounded(S, use_type_assumption::Bool=true)`

okay

schillic:
The last two commits are new. Can you check again?

PR #980, opened 2019-01-09
#962 #769 - Invertible linear map

### LINKED ISSUE #769: Concrete linear map of polytope in H-rep should not change to V-rep
The linear map of a polytope in H-rep can be defined without going to the V-rep, provided the map is invertible (see [here](https://github.com/JuliaReach/LazySets.jl/pull/373#issuecomment-406605362)).

### LINKED ISSUE #962: Faster boundedness check for invertible LinearMap
In #956 we implemented `isbounded(::LinearMap)` as follows:
```julia
function isbounded(lm::LinearMap)::Bool
    if iszero(lm.M) || isbounded(lm.X)
        return true
    end
    return isbounded_unit_dimensions(lm)
end
```
Another sufficient criterion is that `lm.M` is invertible. Since Julia does not provide such a check, we discussed to add a sufficient check of a matrix being invertible/not singular as follows (based on [this reference](https://stackoverflow.com/questions/13145948/how-to-find-out-if-a-matrix-is-singular)):
```julia
function isinvertible_sufficient(M::AbstractMatrix; cond_tol::Number=1e6)
    return cond(M) < cond_tol
end
```
> i think that we can use `cond`
it correspond to the ratio between the largest and smallest singular value of the matrix
if it is bigger than some given tolerance, then the matrix is singular
there will always be some tolerance because we make the computations in floating point

Can we also use this idea for `ExponentialProjectionMap`?

schillic:
Closes #962.
Closes #769.

### DIFF
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -522,7 +522,7 @@ then ``y = M x`` implies ``x = \\text{inv}(M) y`` and we transform the
 constraint system ``A x ≤ b`` to ``A \\text{inv}(M) y ≤ b``.
 """
 function linear_map(M::AbstractMatrix{N}, P::PT) where {N<:Real, PT<:HPoly{N}}
-    if !isinvertible_sufficient(M)
+    if !isinvertible(M)
         if P isa HPolyhedron
             error("linear maps for polyhedra need to be invertible")
         end
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -16,6 +16,7 @@ export HPolyhedron,
        vertices_list,
        singleton_list,
        isempty,
+       linear_map,
        remove_redundant_constraints,
        remove_redundant_constraints!,
        constrained_dimensions
@@ -497,6 +498,60 @@ function remove_redundant_constraints!(P::PT;
 
     deleteat!(P.constraints, setdiff(1:m, non_redundant_indices))
     return P
+end
+
+"""
+    linear_map(M::AbstractMatrix{N}, P::PT) where {N<:Real, PT<:HPoly{N}}
+
+Concrete linear map of a polyhedron in constraint representation.
+
+### Input
+
+- `M` -- matrix
+- `P` -- polyhedron in constraint representation
+
+### Output
+
+A polyhedron of the same type as the input (`PT`).
+
+### Algorithm
+
+If the matrix ``M`` is invertible (which we check with a sufficient condition),
+then ``y = M x`` implies ``x = \\text{inv}(M) y`` and we transform the
+constraint system ``A x ≤ b`` to ``A \\text{inv}(M) y ≤ b``.
+"""
+function linear_map(M::AbstractMatrix{N}, P::PT) where {N<:Real, PT<:HPoly{N}}
+    if !isinvertible_sufficient(M)
+        if P isa HPolyhedron
+            error("linear maps for polyhedra need to be invertible")
+        end
+        # use the implementation for general polytopes
+        return invoke(linear_map, Tuple{typeof(M), AbstractPolytope{N}}, M, P)
+    end
+    # matrix is invertible
+    invM = inv(M)
+    constraints = Vector{LinearConstraint{N}}(undef, length(constraints_list(P)))
+    for c in constraints_list(P)
+        push!(constraints, LinearConstraint(vec(c.a' * invM), c.b))
+    end
+    return PT(constraints)
+end
+
+"""
+    copy(P::PT) where {N, PT<:HPoly{N}}
+
+Create a copy of a polyhedron.
+
+### Input
+
+- `P` -- polyhedron
+
+### Output
+
+The polyhedron obtained by copying the constraints in `P` using `Base.copy`.
+"""
+function copy(P::PT) where {N, PT<:HPoly{N}}
+    return PT(copy(P.constraints))
 end
 
 # ========================================================
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -2,8 +2,7 @@ using MathProgBase, GLPKMathProgInterface
 
 import Base: isempty,
              rand,
-             convert,
-             copy
+             convert
 
 export HPolyhedron,
        dim, σ, ∈,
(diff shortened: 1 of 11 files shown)


src/HPolyhedron.jl
```diff
@@ -499,6 +500,60 @@ function remove_redundant_constraints!(P::PT;
...
+
+### Input
+
+- `P` -- polyhedron
+
+### Output
+
+The polyhedron obtained by copying the constraints in `P` using `Base.copy`.
+"""
+function copy(P::PT) where {N, PT<:HPoly{N}}
```
mforets: now that we have `copy(::LazySet)`, shall we delete this one?

src/HPolyhedron.jl
```diff
@@ -499,6 +501,43 @@ function remove_redundant_constraints!(P::PT;
...
+
+A polyhedron of the same type as the input (`PT`).
+
+### Algorithm
+
+If the matrix ``M`` is invertible (which we check with a sufficient condition),
+then ``y = M x`` implies ``x = \\text{inv}(M) y`` and we transform the
+constraint system ``A x ≤ b`` to ``A \\text{inv}(M) y ≤ b``.
+"""
+function linear_map(M::AbstractMatrix{N}, P::PT) where {N<:Real, PT<:HPoly{N}}
```
mforets: since `isinvertible_sufficient` receives a tolerance, we could allow it in `linear_map` as well, either explicitly or with a `kwargs..` (at least mentioning on the docs)

src/HPolyhedron.jl
```diff
@@ -499,6 +501,43 @@ function remove_redundant_constraints!(P::PT;
...
+            error("linear maps for polyhedra need to be invertible")
+        end
+        # use the implementation for general polytopes
+        return invoke(linear_map, Tuple{typeof(M), AbstractPolytope{N}}, M, P)
+    end
+    # matrix is invertible
+    invM = inv(M)
+    constraints = Vector{LinearConstraint{N}}(undef, length(constraints_list(P)))
+    for c in constraints_list(P)
+        push!(constraints, LinearConstraint(vec(c.a' * invM), c.b))
```
mforets: probably more or less the same efficiency but shorter:

```julia
A, b = tosimplehrep(P)
return PT(A * invM, b)
```

src/HPolyhedron.jl
```diff
@@ -499,6 +501,43 @@ function remove_redundant_constraints!(P::PT;
...
+### Algorithm
+
+If the matrix ``M`` is invertible (which we check with a sufficient condition),
+then ``y = M x`` implies ``x = \\text{inv}(M) y`` and we transform the
+constraint system ``A x ≤ b`` to ``A \\text{inv}(M) y ≤ b``.
+"""
+function linear_map(M::AbstractMatrix{N}, P::PT) where {N<:Real, PT<:HPoly{N}}
+    if !isinvertible_sufficient(M)
+        if P isa HPolyhedron
+            error("linear maps for polyhedra need to be invertible")
```
mforets: this is a library requirement but i think that we should change and revisit this (take for example a bounded polyhedron); probably after https://github.com/JuliaReach/LazySets.jl/issues/998

schillic:
Before we merge, we should investigate the following warnings during the tests:
```
┌ Warning: Deprecated syntax `implicit assignment to global variable `N``.
│ Use `global N` instead.
```

src/HPolyhedron.jl
```diff
@@ -499,6 +501,43 @@ function remove_redundant_constraints!(P::PT;
...
+        end
+        # use the implementation for general polytopes
+        return invoke(linear_map, Tuple{typeof(M), AbstractPolytope{N}}, M, P)
+    end
+    # matrix is invertible
+    invM = inv(M)
```
mforets: `inv` has complexity O(n^3) (see [here](https://en.wikipedia.org/wiki/Computational_complexity_of_mathematical_operations#Matrix_algebra)), see also [this report](http://www.diva-portal.org/smash/get/diva2:1219121/FULLTEXT01.pdf); we could probably be better off with a lazy matrix inverse (related [LazyArrays.jl](https://github.com/JuliaArrays/LazyArrays.jl)). i propose to move this comment to another issue.

src/LinearMap.jl
```diff
@@ -216,12 +216,18 @@ Determine whether a linear map is bounded.
...
+If not, we perform a sufficient check whether the matrix is invertible.
+If the matrix is invertible, then the map being bounded is equivalent to the
+wrapped set being bounded, and hence the map is unbounded.
 Otherwise, we check boundedness via [`isbounded_unit_dimensions`](@ref).
 """
 function isbounded(lm::LinearMap)::Bool
     if iszero(lm.M) || isbounded(lm.X)
         return true
     end
+    if isinvertible_sufficient(lm.M)
```
mforets: we could pass around the `cond_tol` argument to `isbounded` too

src/HPolyhedron.jl
schillic: Cool, but it needs to create two additional matrices. So I keep mine :man_shrugging:

src/helper_functions.jl
```diff
@@ -86,6 +86,31 @@ function ispermutation(u::AbstractVector{T}, v::AbstractVector{T})::Bool where T
...
+If the result is `true`, `M` is invertible.
+If the result is `false`, this function could not conclude.
+
+### Algorithm
+
+We check whether the
+[matrix condition](https://en.wikipedia.org/wiki/Condition_number#Matrices)
+`cond(M)` is below some prescribed tolerance.
+"""
+function isinvertible_sufficient(M::AbstractMatrix; cond_tol::Number=1e6)
```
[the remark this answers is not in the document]
schillic: alright

schillic:
The last two commits are new.

src/helper_functions.jl
```diff
@@ -1,3 +1,6 @@
+# default tolerance for matrix condition number (see 'isinvertible')
+DEFAULT_COND_TOL = 1e6
```
mforets:
```suggestion
const DEFAULT_COND_TOL = 1e6
```

PR #985, opened 2019-01-10
#974 - tohrep and tovrep do not preserve numeric type

### LINKED ISSUE #974: tohrep and tovrep do not preserve numeric type

### LINKED ISSUE #994: Review test with empty polytope
See [this line](https://github.com/JuliaReach/LazySets.jl/blob/master/test/unit_Polytope.jl#L189).

Maybe change the current one for this triangle:

```julia
julia> A = [N(0) N(1); N(1) N(0); N(-1) N(-1)];
julia> b = N[0.25, 0.25, 0];
julia> P = HPolytope(A, b);

julia> vertices_list(P)
 [0.25, 0.25]
 [-0.25, 0.25]
 [0.25, -0.25]
```

schillic:
Closes #974.
Closes #994.

### DIFF
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -593,6 +593,8 @@ in constraint representation.
 
 ### Notes
 
+The conversion may not preserve the numeric type (e.g., with `N == Float32`)
+depending on the backend.
 For further information on the supported backends see
 [Polyhedra's documentation](https://juliapolyhedra.github.io/Polyhedra.jl/latest/installation.html#Getting-Libraries-1).
 """
@@ -601,10 +603,7 @@ function tovrep(P::HPoly{N};
     @assert isdefined(@__MODULE__, :Polyhedra) "the function `tovrep` needs " *
                                         "the package 'Polyhedra' to be loaded"
     P = polyhedron(P; backend=backend)
-    Q = VPolytope(P)
-    @assert Q isa VPolytope{N} "'polyhedron' did not preserve the numeric " *
-        "type; consider using a different polyhedra backend"
-    return Q
+    return VPolytope(P)
 end
 
 """
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -582,9 +582,9 @@ Transform a polyhedron in H-representation to a polytope in V-representation.
 
 ### Input
 
-- `P`          -- polyhedron in constraint representation
-- `backend`    -- (optional, default: `default_polyhedra_backend(P, N)`)
-                  the polyhedral computations backend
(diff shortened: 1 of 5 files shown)


src/HPolyhedron.jl
```diff
@@ -601,7 +601,11 @@ function tovrep(P::HPoly{N};
     @assert isdefined(@__MODULE__, :Polyhedra) "the function `tovrep` needs " *
                                         "the package 'Polyhedra' to be loaded"
     P = polyhedron(P; backend=backend)
-    return VPolytope(P)
+    Q = VPolytope(P)
+    if Q isa VPolytope{N}
+        return Q
+    end
+    return VPolytope{N}(vertices_list(Q))
```
mforets: The function that preserves the numeric type should be `polyhedron` with some backend, otherwise we should raise an error. Perhaps make it an assertion:

```julia
    @assert Q isa VPolytope{N} "the numeric type is not preserved by `polyhedron`"
    return Q
```

schillic: At least `CDDLib` only supports `Float64` and `Rational` (so for instance no `Float32`). Then should we just remove those tests that work in `Float32`?

mforets: I agree that we should only use `Float64` and `Rational`  with `CDDLib`.

schillic: Okay. Should I remove the first commit as well (not needed anymore) or should we keep it?

mforets:
The error is because `polyhedron` makes an implicit conversion `Rational{Int} -> Rational{BigInt}`.

I'm doubting about my comment [here](https://github.com/JuliaReach/LazySets.jl/pull/985#discussion_r246960334). Actually, if solver needs more space to produce a correct result, then why wouldn't it use it? For sure i wouldn't convert back from `Float64 -> Float32`, or `Rational{BigInt} -> Rational{Int}`, say.

PR #999, opened 2019-01-13
#599 - Set membership for VPolytope

### LINKED ISSUE #599: Set membership for VPolytope
```julia
julia> [1.0, 2.0] ∈ VPolytope([[1.0, 2.0]])
ERROR: MethodError: no method matching start(::LazySets.VPolytope{Float64})
Closest candidates are:
  start(::SimpleVector) at essentials.jl:258
  start(::Base.MethodList) at reflection.jl:560
  start(::ExponentialBackOff) at error.jl:107
  ...
Stacktrace:
 [1] any(::Base.##138#139{Array{Float64,1}}, ::LazySets.VPolytope{Float64}) at ./reduce.jl:573
 [2] in(::Array{Float64,1}, ::LazySets.VPolytope{Float64}) at ./reduce.jl:631
```

See:

https://mathoverflow.net/questions/142418/point-in-polytope

https://www.sciencedirect.com/science/article/pii/0097849390900667

schillic:
Closes #599.

### DIFF
--- a/src/VPolytope.jl
+++ b/src/VPolytope.jl
@@ -121,7 +121,7 @@ Check whether a given point is contained in a polytope in vertex representation.
 We check, using linear programming, the definition of a convex polytope that a
 point is in the set if and only if it is a convex combination of the vertices.
 
-Let ``v_j`` be the ``m`` vertices of `P`.
+Let ``\\{v_j\\}`` be the ``m`` vertices of `P`.
 Then we solve the following ``m``-dimensional linear program.
 
 ```math
--- a/src/VPolytope.jl
+++ b/src/VPolytope.jl
@@ -134,20 +134,20 @@ Then we solve the following ``m``-dimensional linear program.
 function ∈(x::AbstractVector{N}, P::VPolytope{N};
            solver=GLPKSolverLP(method=:Simplex))::Bool where {N<:Real}
     vertices = P.vertices
+    m = length(vertices)
 
     # special cases: 0 or 1 vertex
-    if length(vertices) == 0
+    if m == 0
         return false
-    elseif length(vertices) == 1
+    elseif m == 1
         return x == vertices[1]
     end
 
-    @assert length(x) == dim(P) "a vector of length $(length(x)) cannot be contained in a polytope of dimension $(dim(P))"
+    n = length(x)
+    @assert n == dim(P) "a vector of length $(length(x)) cannot be " *
+        "contained in a polytope of dimension $(dim(P))"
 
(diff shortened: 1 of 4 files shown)


src/VPolytope.jl
```diff
@@ -101,6 +102,77 @@ function σ(d::AbstractVector{N},
...
+"""
+function ∈(x::AbstractVector{N}, P::VPolytope{N};
+           solver=GLPKSolverLP(method=:Simplex))::Bool where {N<:Real}
+    vertices = P.vertices
+
+    # special cases: 0 or 1 vertex
+    if length(vertices) == 0
+        return false
+    elseif length(vertices) == 1
+        return x == vertices[1]
```
mforets:
```suggestion
        return _isapprox(x, vertices[1])
```

src/VPolytope.jl
```diff
@@ -101,6 +102,77 @@ function σ(d::AbstractVector{N},
...
+    end
+
+    @assert length(x) == dim(P)
```
mforets:
```suggestion
    @assert length(x) == dim(P) "a vector of length $(length(x)) cannot be contained in a polytope of dimension $(dim(P))"
```

src/VPolytope.jl
```diff
@@ -101,6 +102,77 @@ function σ(d::AbstractVector{N},
...
+
+    m, n = length(vertices), length(x)
```
mforets: simplification: `length(x)` is used a couple of times above, you could move it on top 

```julia
isempty(P) && return false
vertices = P.vertices
m, n = length(vertices), length(first(vertices))
...
```

src/VPolytope.jl
```diff
@@ -101,6 +102,77 @@ function σ(d::AbstractVector{N},
...
+    A = Matrix{N}(undef, n+1, m)
```
mforets: equivalent to the loops below, but for the sake of mentioning it:

```julia
A = [hcat(vertices...); ones(1, m)]
```

src/VPolytope.jl
schillic: This would be inconsistent with [`VPolygon`](https://github.com/JuliaReach/LazySets.jl/blob/master/src/VPolygon.jl#L321) and other types. There is #126 which could be implemented in such a way, but we also need #1028. I suggest we keep this version for now.

src/VPolytope.jl
schillic: Thanks. I keep mine now that it is already writte.

schillic:
Happy with the new commit?

PR #1002, opened 2019-01-14
#904 - Override copy for LazySets

### LINKED ISSUE #904: Override copy for lazyset objects

mforets:
Closes #904.

Related to this PR:

- https://discourse.julialang.org/t/what-is-the-difference-between-copy-and-deepcopy/3918/2
- https://stackoverflow.com/questions/51956958/how-to-copy-a-struct-in-julia

### DIFF
--- a/src/LazySet.jl
+++ b/src/LazySet.jl
@@ -301,9 +301,9 @@ end
 end
 
 """
-    copy(S::T)::T where {T<:LazySet}
+    copy(S::LazySet)
 
-Return a new set independent of the given one by copying its values recursively.
+Return a deep copy of the given set by copying its values recursively.
 
 ### Input
 
@@ -316,8 +316,7 @@ A copy of `S`.
 ### Notes
 
 This function performs a `deepcopy` of each field in `S`, resulting in a
-completely independent object.
+completely independent object. See the documentation of `?deepcopy` for further
+details.
 """
-function copy(S::T)::T where {T<:LazySet}
-    return T([copy(getfield(S, fi)) for fi in fieldnames(T)]...)
-end
+copy(S::LazySet) = deepcopy(S)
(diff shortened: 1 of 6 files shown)


src/LazySet.jl
```diff
@@ -299,3 +299,26 @@ end
...
+
+A copy of `S`.
+
+### Notes
+
+This function performs a `deepcopy` of each field in `S`, resulting in a
+completely independent object.
+"""
+function copy(S::LazySet)
+    T = typeof(S)
```
schillic: Just define `copy(S::T) where {T<:LazySet}`?

src/LazySet.jl
```diff
@@ -299,3 +299,26 @@ end
...
+    return T([deepcopy(getfield(S, fi)) for fi in fieldnames(T)]...)
```
schillic: `copy` should not use `deepcopy`. Rather add a `deepcopy` method as well.

test/unit_Polytope.jl
```diff
@@ -132,6 +132,14 @@ for N in [Float64, Rational{Int}, Float32]
...
         @test ispermutation(constraints_list(V), constraints_list(tohrep(V)))
     end
+
+    # copy (see #1002)
+    p, q = [1], [2]
+    P = VPolytope([p, q])
+    Pcopy = copy(P)
+    p[1] = 5
+    # test that Pcopy is independent of P ( = deepcopy)
+    @test Pcopy.vertices[1] == [1]
```
schillic: That test does not use the numeric type `N`. So we execute the same test three times.

Issue #1011, opened 2019-01-15
Support vector of HPolyhedron requires `Vector` direction


schillic:
The implementation of the support vector for `HPolytope`/`HPolyhedron` uses a linear program, where it uses the direction as objective function. Since the direction can be an `AbstractVector`, it can be a `SparseVector`; however, the LP backend requires a `Vector`.

Trigger: The default implementation of `an_element` creates a sparse unit vector and queries the support vector.
```julia
julia> an_element(HPolytope([HalfSpace([1.0, 0.0], 1.0), HalfSpace([0.0, 1.0], 1.0), HalfSpace([-1.0, -1.0], -1.0)]))
ERROR: MethodError: no method matching linprog(::SparseArrays.SparseVector{Float64,Int64}, ::Array{Float64,2}, ::Char, ::Array{Float64,1}, ::Float64, ::Float64, ::GLPKMathProgInterface.GLPKInterfaceLP.GLPKSolverLP)
Closest candidates are:
  linprog(::Union{Char, Array{T,1}, Real} where T<:Union{Char, Real}, ::AbstractArray{T,2} where T, ::Union{Char, Array{T,1}, Real} where T<:Union{Char, Real}, ::Union{Char, Array{T,1}, Real} where T<:Union{Char, Real}, ::Union{Char, Array{T,1}, Real} where T<:Union{Char, Real}, ::Union{Char, Array{T,1}, Real} where T<:Union{Char, Real}, ::MathProgBase.SolverInterface.AbstractMathProgSolver) at MathProgBase/rVlFR/src/HighLevelInterface/linprog.jl:106
  linprog(::Union{Char, Array{T,1}, Real} where T<:Union{Char, Real}, ::AbstractArray{T,2} where T, ::Union{Char, Array{T,1}, Real} where T<:Union{Char, Real}, ::Union{Char, Array{T,1}, Real} where T<:Union{Char, Real}, ::Union{Char, Array{T,1}, Real} where T<:Union{Char, Real}, ::Union{Char, Array{T,1}, Real} where T<:Union{Char, Real}) at MathProgBase/rVlFR/src/HighLevelInterface/linprog.jl:115
  linprog(::Any, ::Any, ::Any, ::Any) at MathProgBase/rVlFR/src/HighLevelInterface/linprog.jl:116
  ...
Stacktrace:
[...]

PR #1012, opened 2019-01-15
#1003 - Add more methods for UnionSet

### LINKED ISSUE #1003: Add more methods for UnionSet
Basically all common `LazySet` methods are trivial to implement:
* unary
  * `∈`
  * `an_element`
  * `isempty`
  * `isbounded`
* binary (careful: The trivial implementations only work with `LazySet`s.)
  * `⊆(::UnionSet, ::LazySet)`
  * `isdisjoint(::UnionSet, X)` and its symmetric version (note that `X` does not need to be convex)
  * `∩(::LazySet, ::UnionSet)` and its symmetric version (returns a `UnionSet` of the pairwise intersections)

(Also add these methods for `UnionSetArray.)

schillic:
Closes #1003.

Currently some tests fail due to #1011. I would rather like to keep the tests and fix that other one first.


src/is_subset.jl
```diff
@@ -587,3 +587,68 @@ function ⊆(X::EmptySet{N}, ∅::EmptySet{N}, witness::Bool=false
...
+
+* If `witness` option is deactivated: `true` iff ``\\text{cup} ⊆ X``
+* If `witness` option is activated:
+  * `(true, [])` iff ``\\text{cup} ⊆ X``
+  * `(false, v)` iff ``\\text{cup} \\not\\subseteq X`` and
+    ``v ∈ \\text{cup} \\setminus X``
+"""
+function ⊆(cup::UnionSetArray{N}, X::LazySet{N}, witness::Bool=false
+          )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+    for Y in array(cup)
```
mforets: add `local result` above the `for Y in array(cup)` to make `result` be available in the `return` statement

schillic: I initialized `result` and `w` instead (this has the benefit that it even works for the empty array).

schillic:
I will work on the seeming nontermination in #1026.

schillic:
Unfortunately, I had to undo the usage of `⊈` because of incompatibility with v0.6.

src/concrete_intersection.jl
```diff
@@ -536,8 +536,75 @@ function intersection(S::AbstractSingleton{N},
...
+- `cup` -- union of two convex sets
+- `X`   -- convex set
+
+### Output
+
+The union of the pairwise intersections, expressed as a `UnionSet`.
+If one of those sets is empty, only the other set is returned.
+"""
+function intersection(cup::UnionSet{N}, X::LazySet{N}) where {N<:Real}
+    return intersection(cup.X, X) ∪ intersection(cup.Y, X)
```
mforets: i would expect `intersection` to return a concrete set, not a `UnionSet`, but ok it can be used as a first step.

mforets:
> Unfortunately, I had to undo the usage of ⊈ because of incompatibility with v0.6.

should we create an issue for this?

src/concrete_intersection.jl
schillic: In general we do not have a concrete set type for that.
The only other option I see is to remove this method again.

schillic:
> >  Unfortunately, I had to undo the usage of ⊈ because of incompatibility with v0.6.
>
> should we create an issue for this?

Done. I will wait with the merge until you answered [this question](https://github.com/JuliaReach/LazySets.jl/pull/1012#pullrequestreview-194062941).

src/concrete_intersection.jl
mforets: true. unless we we make further assumptions in the sets (eg. polytopes). ok for keeping it.

PR #1037, opened 2019-01-20
Performance regression in LinearMap

schillic:


mforets:
@schillic what is "old" and what is "new" above? what do you get for the same benchmark if you use the branch `mforets/linearmap`? i get the following times:

```julia
328.716 ns (15 allocations: 1.00 KiB)  # mforets/linearmap using option 1 (with parameter MAT)
514.663 ns (15 allocations: 1.00 KiB)  # mforets/linearmap using option 2 (without parameter MAT)

333.828 ns (15 allocations: 1.00 KiB) # master
534.875 ns (15 allocations: 1.00 KiB) # schillic/linearmap
```
this indicates that the parameter is beneficial, doesn't explain https://github.com/JuliaReach/Reachability.jl/issues/423 though :/

schillic:
True, I mixed them up, thanks for double-checking! Still, https://github.com/JuliaReach/Reachability.jl/issues/423 gets fixed by this change.

mforets:
We should close this PR; it is fixed by using concrete type containers. See eg. escritoire/reachability/BFFPSV18.

mforets:
Should we close this issue? cc @schillic

schillic:
> Should we close this issue?

No, we have never investigated whether this was the reason for the regression.

mforets:
I have investigated :) it's working with vectors of lazysets which are not concretely typed which causes this regression.

schillic:
What I mean is: Did you go back to the old version, apply the fix that you suggest, and see if you get the old runtime?

mforets:
> Did you go back to the old version, apply the fix that you suggest, and see if you get the old runtime?

No, i didn't need to do that in this case. I wrote a version which stores all the type parameters and checked that it is much faster.

mforets:
In any case, we won't remove type parameters for the lazy linear map, so we can close this issue and store a branch.. I don't think there is much more to be learned from this example.

Issue #843, opened 2018-10-21
Empty polyhedra intersection is broken


schillic:
```julia
julia> using LazySets, Polyhedra

julia> H = LazySets.HalfSpace([-1.0], -21.0); # x >= 21

julia> P = HPolytope([LazySets.HalfSpace([1.0], 18.1), # x <= 18.1
                      LazySets.HalfSpace([-1.0], -18.0)]); # x >= 18
```
In v0.6 this results in:
```julia
julia> cap = intersection(H, P)
LazySets.HPolytope{Float64}(LazySets.HalfSpace{Float64}[
 LazySets.HalfSpace{Float64}([1.0], 18.1),
 LazySets.HalfSpace{Float64}([-1.0], -18.0),
 LazySets.HalfSpace{Float64}([-1.0], -18.1),
 LazySets.HalfSpace{Float64}([1.0], 18.0)])

julia> isempty(cap)
true
```
In v0.7 this results in:
```julia
julia> cap = intersection(H, P)
HPolytope{Float64}(LazySets.HalfSpace{Float64}[])

julia> isempty(cap)
false
```
Problems:
1. Why does v0.6 not simplify?
It might be because the simplification with `CDDLib` is hard to interpret (contains two hyperplanes):
```julia
julia> Q = polyhedron(cap, CDDLib.CDDLibrary())
Polyhedron CDDLib.CDDPolyhedron{1,Float64}:
4-element iterator of Polyhedra.HalfSpace{1,Float64,Array{Float64,1}}:
 HalfSpace([1.0], 18.1)
 HalfSpace([-1.0], -18.0)
 HalfSpace([-1.0], -18.1)
 HalfSpace([1.0], 18.0)

julia> removehredundancy!(Q)

julia> Q
Polyhedron CDDLib.CDDPolyhedron{1,Float64}:
2-element iterator of Polyhedra.HyperPlane{1,Float64,Array{Float64,1}}:
 HyperPlane([1.0], 18.1)
 HyperPlane([-1.0], -18.0)
```
2. The v0.7 result (polyhedron without constraints) is due to #842. By hard-coding a `1` in the code, the result is as follows:
```julia
julia> cap = intersection(H, P)
[...]

Issue #1038, opened 2019-01-20
Intersection of polytope and polyhedron results in unbounded polyhedron


schillic:
```julia
julia> P = HPolytope([
 HalfSpace([0.0, 0.0, 0.0, 0.0, -1.0], -1.0),
 HalfSpace([-0.0, -0.0, -0.0, -0.0, 1.0], 1.0),
 HalfSpace([1.0, 0.0, 0.0, 0.0, 0.0], 0.308261),
 HalfSpace([-1.0, 0.0, 0.0, 0.0, 0.0], -0.2),
 HalfSpace([0.0, 1.0, 0.0, 0.0, 0.0], 0.1),
 HalfSpace([0.0, -1.0, 0.0, 0.0, 0.0], 0.106045),
 HalfSpace([0.0, 0.0, 1.0, 0.0, 0.0], 0.0153778),
 HalfSpace([0.0, 0.0, -1.0, 0.0, 0.0], 0.0),
 HalfSpace([0.0, 0.0, 0.0, 1.0, 0.0], 0.00075663),
 HalfSpace([0.0, 0.0, 0.0, -1.0, 0.0], 0.000156122)
]);

julia> Q = HPolyhedron([
 HalfSpace([-1.0, 0.0, 0.0, 0.0, 0.0], 0.0),
 HalfSpace([-0.714286, -1.0, 0.0, 0.0, 0.0], 0.0),
 HalfSpace([0.714286, 1.0, 0.0, 0.0, 0.0], 0.0)
]);

julia> isdisjoint(P, Q)
false

julia> isdisjoint(Q, P)
true

# the above is already worrying

# the intersection should be bounded, as P is already bounded
julia> R = intersection(P, Q)
HPolytope([
 HalfSpace([0.714286, 1.0, 0.0, 0.0, 0.0], 0.0),
 HalfSpace([-0.0, -0.0, -0.0, -0.0, 1.0], 1.0),
 HalfSpace([-0.714286, -1.0, -0.0, -0.0, -0.0], -0.0),
 HalfSpace([0.0, 0.0, 0.0, 0.0, -1.0], -1.0)
])
```

schillic:
The `isdisjoint` calls [the same method](https://github.com/JuliaReach/LazySets.jl/blob/master/src/is_intersection_empty.jl#L865) in both cases.

schillic:
Here is a 2D projection with the same problems:
```julia
[...]

Issue #754, opened 2018-10-09
HPolytope's removehredundancy! removes non-redundant constraints


schillic:
```julia
julia> using LazySets, LazySets.Approximations, Polyhedra, Optim

julia> P1 = LazySets.HPolytope{Float64}(
              LazySets.HalfSpace{Float64}[
              LazySets.HalfSpace{Float64}([1.0, -1.0], 5.465),
              LazySets.HalfSpace{Float64}([-1.0, 1.0], -4.4),
              LazySets.HalfSpace{Float64}([-1.0, -1.0], 4.5),
              LazySets.HalfSpace{Float64}([1.0, 1.0], -3.435),
              LazySets.HalfSpace{Float64}([1.0, 0.0], 0.965),
              LazySets.HalfSpace{Float64}([0.0, 1.0], -4.4),
              LazySets.HalfSpace{Float64}([0.0, -1.0], 4.5),
              LazySets.HalfSpace{Float64}([-1.0, 0.0], 1.57415e-9)
              ]);

julia> P2 = LazySets.HPolytope{Float64}(LazySets.HalfSpace{Float64}[
              LazySets.HalfSpace{Float64}([0.0, 1.0], 0.0),
              LazySets.HalfSpace{Float64}([-1.0, 0.0], 0.0),
              LazySets.HalfSpace{Float64}([1.0, 0.0], 0.0)
              ]);

julia> C = P1 ∩ P2;

julia> isempty(C)
false

julia> using Plots; plot(P1)
```
![p1](https://user-images.githubusercontent.com/9656686/46697473-fbc1e280-cc14-11e8-932c-226191801b46.png)
```julia
julia> p = polyhedron(P1)
Polyhedron CDDLib.CDDPolyhedron{2,Float64}:
8-element iterator of Polyhedra.HalfSpace{2,Float64,Array{Float64,1}}:
 HalfSpace([1.0, -1.0], 5.465)
 HalfSpace([-1.0, 1.0], -4.4)
 HalfSpace([-1.0, -1.0], 4.5)
 HalfSpace([1.0, 1.0], -3.435)
 HalfSpace([1.0, 0.0], 0.965)
 HalfSpace([0.0, 1.0], -4.4)
 HalfSpace([0.0, -1.0], 4.5)
 HalfSpace([-1.0, 0.0], 1.57415e-9)

julia> Polyhedra.removehredundancy!(p); p
Polyhedron CDDLib.CDDPolyhedron{2,Float64}:
[...]

PR #1040, opened 2019-01-21
polyhedron redundant constraints

mforets:
See #1038.

- [ ] check if fixes https://github.com/JuliaReach/LazySets.jl/issues/843

### DIFF
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -418,7 +418,7 @@ end
 
 """
     remove_redundant_constraints(P::PT;
-                                 backend=GLPKSolverLP())::Union{PT, EmptySet} where {N, PT<:HPoly{N}}
+                                 backend=GLPKSolverLP())::Union{PT, EmptySet{N}} where {N, PT<:HPoly{N}}
 
 Remove the redundant constraints in a polyhedron in H-representation.
 
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -437,7 +437,7 @@ if `P` is detected to be empty (which happens if the constraints are infeasible)
 See [`remove_redundant_constraints!`](@ref) for details.
 """
 function remove_redundant_constraints(P::PT;
-                                      backend=GLPKSolverLP())::Union{PT, EmptySet} where {N, PT<:HPoly{N}}
+                                      backend=GLPKSolverLP())::Union{PT, EmptySet{N}} where {N, PT<:HPoly{N}}
     Pred = copy(P)
     if remove_redundant_constraints!(Pred, backend=backend)
         return Pred
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -490,7 +490,7 @@ function remove_redundant_constraints!(P::PT;
             error("LP is not optimal; the status of the LP is $(lp.status)")
         end
         objval = -lp.objval
-        if objval <= b[j]
+        if _leq(objval, b[j])
             # the constraint is redundant
             non_redundant_indices = setdiff(non_redundant_indices, j)
         else
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -484,8 +484,10 @@ function remove_redundant_constraints!(P::PT;
         br = b[non_redundant_indices]
         br[i] = b[j] + one(N)
         lp = linprog(-α, Ar, '<', br, -Inf, Inf, backend)
-        if lp.status != :Optimal
-            error("LP is not optimal")
+        if lp.status == :Infeasible
+            throw(DomainError(P, "the polyhedron is empty"))
+        else
+            error("LP is not optimal; the status of the LP is $(lp.status)")
         end
         objval = -lp.objval
         if objval <= b[j]
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -494,7 +494,7 @@ function remove_redundant_constraints!(P::PT;
             return false
         elseif lp.status == :Optimal
             objval = -lp.objval
-            if _leq(objval, b[j])
+            if objval <= b[j]
                 # the constraint is redundant
                 non_redundant_indices = setdiff(non_redundant_indices, j)
             else
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -418,10 +418,9 @@ end
 
 """
     remove_redundant_constraints(P::PT;
-                                 backend=GLPKSolverLP()) where {N, PT<:HPoly{N}}
+                                 backend=GLPKSolverLP())::Union{PT, EmptySet} where {N, PT<:HPoly{N}}
 
-Given a polyhedron in H-representation, return a new polyhedron with no reundant
-constraints.
+Remove the redundant constraints in a polyhedron in H-representation.
 
 ### Input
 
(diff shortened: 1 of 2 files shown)


mforets:
`intersection` could return an instance of `∅` whenever the intersection is empty; we already have this union return type for other set intersections.

src/HPolyhedron.jl
```diff
@@ -484,11 +484,13 @@ function remove_redundant_constraints!(P::PT;
         br = b[non_redundant_indices]
         br[i] = b[j] + one(N)
         lp = linprog(-α, Ar, '<', br, -Inf, Inf, backend)
-        if lp.status != :Optimal
-            error("LP is not optimal")
+        if lp.status == :Infeasible
+            throw(DomainError(P, "the polyhedron is empty"))
```
mforets: perhaps this should instead throw an assertion error, to be checked from the outside (e.g. from `intersection`)

src/HPolyhedron.jl
```diff
@@ -484,11 +484,13 @@ function remove_redundant_constraints!(P::PT;
...
+        else
+            error("LP is not optimal; the status of the LP is $(lp.status)")
```
mforets: what type of error should be thrown here?

schillic: Why does it matter?

src/HPolyhedron.jl
schillic: What is "the polyhedron"? And why should an empty intersection not be allowed?
I think this case should just return an `EmptySet`.

src/HPolyhedron.jl
```diff
@@ -484,11 +484,13 @@ function remove_redundant_constraints!(P::PT;
...
         end
         objval = -lp.objval
-        if objval <= b[j]
+        if _leq(objval, b[j])
```
schillic: I think we need to clarify the cases when we want to use `_leq` etc. Do we want to overapproximate or underapproximate the result? I think without an answer there is no point in using approximations.
The documentation should clearly mention the behavior for floating point, too.
In this case, I would say that we should not remove more constraints but rather less to be on the safe side, so here I would check for the opposite condition and use `!_geq(...)`.

src/HPolyhedron.jl
mforets: i threw an error because `remove_redundant_constraints!` is supposed to modify an `HPolyhedron` in place. changing `P` to an emptyset breaks type stability.

otherwise we could change the signature to something like:

```julia
function remove_redundant_constraints!(P::Union{PT<:HPoly{N}, EmptySet{N}},
                                       backend=GLPKSolverLP()) where {N}
....
```

but this could break type stability in other places, no? like functions that use `remove_redundant_constraints!` should be aware that it can return empty sets.

so just throwing an error didn't seems ok to me.

schillic: Oh, I thought this was in `intersection`. But why would you have to throw an error? Can we not just print a warning and return the original set?

src/HPolyhedron.jl
mforets: thinking of a beter practice, because it helps to debug for example:

``` try
              remove_redundant_constraints!(P)
       catch y
           if isa(y, DomainError)
                  return EmptySet()
           elseif isa(y, BoundsError)
                   . . . .
           else  
                   ...
            end
       end
```

schillic: Then you have to study all cases when this can happen. I would say that this is not worth the effort.

src/HPolyhedron.jl
mforets: ok. for this change, i was having in mind https://github.com/JuliaReach/LazySets.jl/issues/754  (but i haven't tried it). then, i'll remove this commit which can be done in a separate PR.

i think that *once the floating point tolerance has been specified*, like a `ztol`, it doesn't matter if the result is under or overapproximated, one just wants that `objval` is either strictly smaller than `b[j]` or above but close to it. this is the purpose of `_leq` that is a relaxed version of `<=` for floating point. for rationals this is exact, and for FP we can pass the tolerances through an argument to  `remove_redundant_constraints!`.

src/concrete_intersection.jl
```diff
@@ -364,23 +364,33 @@ A new same polytope in H-representation with just one more constraint.
...
+
     # concatenate the linear constraints
     Q = HPOLY([constraints_list(P1);
                constraints_list(P2)])
 
     # remove redundant constraints
-    if prunefunc == removehredundancy!
+    if prunefunc == remove_redundant_constraints!
+        if backend == nothing
+            backend = GLPKSolverLP()
```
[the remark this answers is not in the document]
mforets: i thought about this but i didn't come up with a simpler way.

basically, the `prunefunc` specifies the algorithm to be used, and the backend is an option to it, but each algorithm needs a different default backend.

src/HPolyhedron.jl
schillic: > i was having in mind #754

You relaxed the condition for removing, so you removed *more* constraints.

mforets: ok. the examples there use `Polyhedra.removehredundancy!`, i should try in the first place how does it behave with `remove_redundant_constriants!`

src/concrete_intersection.jl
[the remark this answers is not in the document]
schillic: Okay, I see that it is not so simple. But since you do a case split based on the function name, should we instead use clearer names, like `"polyhedra"` and `"default"`? You anyway do not use `prunefunc` inside the cases.

src/concrete_intersection.jl
```diff
@@ -341,49 +341,69 @@ end
...
+### Notes
+
+The default value of the backend is set internally and depends on whether the
+Polyhedra backend is used or not. The default backends are `GLPKSolverLP()`
+and `default_polyhedra_backend(P1, N)` respectively.
+
+Note that if `use_polyhedra_interface` is set to `true`, there is no guarantee
+that the removal of constraints keep the set empty (see #1038 and Polyhedra#146),
+so it is better to check for emptiness of intersection before using this function
+if that case.
```
schillic:
```suggestion
in that case.
```

src/HPolyhedron.jl
```diff
@@ -430,25 +429,29 @@ constraints.
...
 
-See `remove_redundant_constraints!`. 
+See [`remove_redundant_constraints!`](@ref) for details.
 """
 function remove_redundant_constraints(P::PT;
-                                      backend=GLPKSolverLP()
-                                     ) where {N, PT<:HPoly{N}}
-    return remove_redundant_constraints!(PT(copy(constraints_list(P))),
-                                         backend=backend)
+                                      backend=GLPKSolverLP())::Union{PT, EmptySet} where {N, PT<:HPoly{N}}
```
schillic:
```suggestion
                                      backend=GLPKSolverLP())::Union{PT, EmptySet{N}} where {N, PT<:HPoly{N}}
```

src/HPolyhedron.jl
```diff
@@ -418,10 +418,9 @@ end
 
 """
     remove_redundant_constraints(P::PT;
-                                 backend=GLPKSolverLP()) where {N, PT<:HPoly{N}}
+                                 backend=GLPKSolverLP())::Union{PT, EmptySet} where {N, PT<:HPoly{N}}
```
schillic:
```suggestion
                                 backend=GLPKSolverLP())::Union{PT, EmptySet{N}} where {N, PT<:HPoly{N}}
```

PR #1050, opened 2019-01-22
fix intersection

### LINKED ISSUE #843: Empty polyhedra intersection is broken

mforets:
Closes #843.

### DIFF
--- a/src/concrete_intersection.jl
+++ b/src/concrete_intersection.jl
@@ -394,7 +394,7 @@ function intersection(P1::HPoly{N},
         # convert to an hrep, remove the redundancies and convert back to HPOLY
         ph = polyhedron(Q; backend=backend)
         removehredundancy!(ph)
-        Q = convert(HPOLY, ph)
+        return convert(HPOLY, ph)
     else
         if backend == nothing
             backend = GLPKSolverLP()
@@ -406,7 +406,6 @@ function intersection(P1::HPoly{N},
             return EmptySet{N}()
         end
     end
-    return Q
 end
 
 """
--- a/src/concrete_intersection.jl
+++ b/src/concrete_intersection.jl
@@ -400,7 +400,11 @@ function intersection(P1::HPoly{N},
             backend = GLPKSolverLP()
         end
         # here, detection of empty intersection may be reported as an infeasible LP
-        remove_redundant_constraints!(Q, backend=backend)
+        if remove_redundant_constraints!(Q, backend=backend)
+            return Q
+        else
+            return EmptySet{N}()
+        end
     end
     return Q
 end


src/concrete_intersection.jl
```diff
@@ -400,7 +400,11 @@ function intersection(P1::HPoly{N},
...
         end
         # here, detection of empty intersection may be reported as an infeasible LP
-        remove_redundant_constraints!(Q, backend=backend)
+        if remove_redundant_constraints!(Q, backend=backend)
+            return Q
+        else
+            return EmptySet{N}()
+        end
     end
     return Q
```
schillic: The `return Q` at the end of the function could be pulled inside the `if` case above to make the flow clearer.

PR #1075, opened 2019-01-26
#1074 - Intersection with bounded set computes an unbounded support value

### LINKED ISSUE #1074: Intersection with bounded set computes an unbounded support value
```julia
julia> H = Hyperrectangle([0., 0.], [1., 1.]);

julia> P = HPolyhedron([HalfSpace([1., 0.], 0.)]);

julia> C = Intersection(H, P);

julia> ρ(ones(2), C)
Inf
```

schillic:
Closes #1074.


mforets:
Alternatively, now that we have `isbounded` we could use it. It provides a generalization of the idea behind this `ρ` because it could handle for example unbounded types that are not `HPolyhedron`; for example an operation over an unbounded polyhedron that also returns an unbounded set.

PR #1080, opened 2019-01-29
Use left division in linear map

mforets:
I haven't benchmarked this, but i would expect it to be more efficient to calculate the application of the linear map using the left division than calculating the full matrix inverse.

### DIFF
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -400,7 +400,7 @@ end
 """
     linear_map(M::AbstractMatrix{N}, P::PT;
               [cond_tol=DEFAULT_COND_TOL]::Number,
-              [use_inv]::Bool) where {N<:Real, PT<:HPoly{N}}
+              [use_inv]::Bool=!issparse(M)) where {N<:Real, PT<:HPoly{N}}
 
 Concrete linear map of a polyhedron in constraint representation.
 
@@ -443,6 +443,7 @@ function linear_map(M::AbstractMatrix{N},
     end
 
     constraints = similar(constraints_list(P))
+
     # matrix M is invertible => the normal vectors are vec(c.a' * inv(M))
     if use_inv
         invM = inv(M)
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -433,7 +433,7 @@ function linear_map(M::AbstractMatrix{N},
     end
     # matrix M is invertible: the normal vectors are vec(c.a' * inv(M)), or taking
     # the left division for each constraint c, transpose(M) \ c.a
-    Mt = tranpose(M)
+    Mt = transpose(M)
     constraints = Vector{LinearConstraint{N}}(undef,
                                               length(constraints_list(P)))
     @inbounds for (i, c) in enumerate(constraints_list(P))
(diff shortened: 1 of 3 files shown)


src/HPolyhedron.jl
```diff
@@ -431,12 +431,13 @@ function linear_map(M::AbstractMatrix{N},
         # use the implementation for general polytopes
         return invoke(linear_map, Tuple{typeof(M), AbstractPolytope{N}}, M, P)
     end
-    # matrix is invertible
-    invM = inv(M)
+    # matrix M is invertible: the normal vectors are vec(c.a' * inv(M)), or taking
+    # the left division for each constraint c, transpose(M) \ c.a
+    Mt = tranpose(M)
```
mforets:
```suggestion
    Mt = transpose(M)
```

mforets:
(note: this code is thought using v1.0 in mind, for earlier versions we should rather use the corresponding BLAS function name)

schillic:
> (note: this code is thought using v1.0 in mind, for earlier versions we should rather use the corresponding BLAS function name)

Are you going to change that?

mforets:
yes. i will do a trick similar to [this one](https://github.com/JuliaReach/LazySets.jl/blob/master/src/compat.jl#L17).

i plan to add a "linear map" benchmark to `benchmark/benchmarks.jl` to evaluate this change before merging.

mforets:
```julia
 using this branch
 julia> @btime linear_map(rand(50, 50), convert(HPolytope, rand(Hyperrectangle, dim=50)));
  50.610 ms (1122 allocations: 2.13 MiB)

 julia> @time linear_map(rand(50, 50), convert(HPolytope, rand(Hyperrectangle, dim=50)));
  0.001384 seconds (1.04 k allocations: 213.891 KiB)
```

src/HPolyhedron.jl
```diff
@@ -398,8 +398,9 @@ function remove_redundant_constraints!(P::HPoly{N};
 end
 
 """
-    linear_map(M::AbstractMatrix{N}, P::PT; [cond_tol=DEFAULT_COND_TOL]::Number)
-        where {N<:Real, PT<:HPoly{N}}
+    linear_map(M::AbstractMatrix{N}, P::PT;
+              [cond_tol=DEFAULT_COND_TOL]::Number,
+              [use_inv]::Bool) where {N<:Real, PT<:HPoly{N}}
```
schillic: Add the default value for `use_inv`?

test/unit_Polytope.jl
```diff
@@ -112,12 +112,21 @@ for N in [Float64, Rational{Int}, Float32]
...
+    # =====================
+    # Concrete linear map
+    # =====================
+    linear_map(N[2 3; 1 2], P) # invertible matrix
     if test_suite_polyhedra
         linear_map(N[2 3; 0 0], P)  # noninvertible matrix
     end
 
+    M = N[2 1; 0 1]
+    linear_map(M, P)
```
schillic: Use `H` instead of `P` for the tests to also work with `N = Rational{Int}`.

mforets: It didn't help, because the new keyword argument is defined for polytopes, not for `BallInf` 😢 . So i'll just put this inside a `N <: AbstractFloat` check..

mforets: The problem is that `an_element` is not type-stable: https://github.com/JuliaReach/LazySets.jl/issues/1105 The `linear_map`s for Rational are built properly.

Issue #1041, opened 2019-01-21
Decomposed intersection for CartesianProductArray


kostakoida:
<img width="492" alt="screen shot 2019-01-25 at 12 29 07 am" src="https://user-images.githubusercontent.com/7996008/51681299-45a1cd80-2038-11e9-8b7a-02ac6e54df22.png">

 We preprocess our system in such way,  we know what blocks are necessary to make an intersection with guard or invariant. Then we apply new intersection method(LazySets.jl) of two CartesianProductArray`s(which are represent our decomposed RSets and decomposed constraints of guard or invariant). It is important to note, that our RSets and constraints should be decomposed in the same number of blocks for CartesianProductArray.

mforets:
@kostakoida Thanks for the write-up.

Can you add a minimal example to imagine how we plan to actually implement this?

Is it going to be a concrete intersection algorithm (see [here](https://juliareach.github.io/LazySets.jl/latest/lib/binary_functions.html#LazySets.intersection-Union{Tuple{N},%20Tuple{Interval{N,IN}%20where%20IN%3C:AbstractInterval{N},Interval{N,IN}%20where%20IN%3C:AbstractInterval{N}}}%20where%20N%3C:Real)), or an algorithm to calculate the support function of a lazy intersection more efficiently (see [here](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Intersection.jl#L245))?

kostakoida:
> What function(s) need to be provided by this package that are of independent interest of this reachability algorithm?
We should implement new intersection algorithm to compute intersection between two CartesianProductArrays smartly(only between required blocks)
 
> Are the intersections appearing in lines 20, 21, 23 all lazy? (`intersect` in LazySets)
Yes

PR #1081, opened 2019-01-29
[WIP]Concrete intersection of CartesianProductArray and polyhedron

kostakoida:
See #1041.


src/concrete_intersection.jl
```diff
@@ -656,3 +656,27 @@ end
...
+
+ - `ca1` -- Cartesian Product Array of convex sets
+ - `ca2` -- Cartesian Product Array of LazySets (it might by unbounded sets)
+ - `nonzero_blocks` -- Required blocks to intersect
+
+### Output
+
+The Cartesian Product Array obtained by the intersection nonzero blocks of `ca1` and `ca2`.
+"""
+function intersection(ca1::CartesianProductArray{N, S}, L2::CartesianProductArray{N, S}, nonzero_blocks::Vector{Int})
```
schillic: `nonzero_blocks` should be an *optional* argument. The code should run without having to specify this. You can write a function to compute the default value.

`L2` should be `ca2` I guess.

src/concrete_intersection.jl
```diff
@@ -656,3 +656,27 @@ end
 function intersection(L1::LinearMap{N}, L2::LinearMap{N}) where {N}
     return intersection(linear_map(L1.M, L1.X), linear_map(L2.M, L2.X))
 end
+
+"""
+intersection(ca1::CartesianProductArray{N, S}, L2::CartesianProductArray{N, S}, nonzero_blocks::Vector{Int})
+              ) where {N<:Real, S<:LazySet{N}}
+
+Return the intersection of two CartesianProductArray's only for necessary blocks.
```
schillic: "necessary" is not self-explanatory.
Also, since this argument should be optional (see my other comment), maybe better do not talk about this in the short description but add a `### Notes` section below.

src/concrete_intersection.jl
```diff
@@ -656,3 +656,27 @@ end
...
+ - `nonzero_blocks` -- Required blocks to intersect
+
+### Output
+
+The Cartesian Product Array obtained by the intersection nonzero blocks of `ca1` and `ca2`.
+"""
+function intersection(ca1::CartesianProductArray{N, S}, L2::CartesianProductArray{N, S}, nonzero_blocks::Vector{Int})
+              ) where {N<:Real, S<:LazySet{N}}
+    for i in nonzero_blocks
+        ca1.array[i] = intersection(ca1.array[i], ca2.array[i])
```
schillic: Do not modify `ca1`. Create a new set instead.

src/concrete_intersection.jl
```diff
@@ -656,3 +656,27 @@ end
...
+ - `ca1` -- Cartesian Product Array of convex sets
+ - `ca2` -- Cartesian Product Array of LazySets (it might by unbounded sets)
+ - `nonzero_blocks` -- Required blocks to intersect
+
+### Output
+
+The Cartesian Product Array obtained by the intersection nonzero blocks of `ca1` and `ca2`.
+"""
+function intersection(ca1::CartesianProductArray{N, S}, L2::CartesianProductArray{N, S}, nonzero_blocks::Vector{Int})
+              ) where {N<:Real, S<:LazySet{N}}
```
```suggestion
              where {N<:Real, S<:LazySet{N}}
```

schillic: This is a problem. You cannot have the `where` alone on the next line. Rather remove the closing parenthesis in the previous line.

mforets:
This function asks that both sets are cartesian product arrays, and then use `nonzero_blocks`. A problem is that having this argument does not make sense in general; i would at least add a suitable assertion or clarify the docs. But the idea that is behind this code is to simplify the intersection of a set (we can think that it is given as a CPA, although it is no necessarily the case), with an *unbounded set*, that is, a set that has free variables. Then in [this gist](https://gist.github.com/mforets/f2c5b3aff985ac32f3cb75f9f40f50a0) i propose an alternative implementation. Thoughts?

src/concrete_intersection.jl
```diff
@@ -654,3 +655,48 @@ end
 function intersection(L1::LinearMap{N}, L2::LinearMap{N}) where {N}
     return intersection(linear_map(L1.M, L1.X), linear_map(L2.M, L2.X))
 end
+
+"""
+function intersection(X::CartesianProductArray{N},
+                      Y::AbstractPolyhedron{N}) where {N}
+
+Return the intersection of CartesianProductArray and AbstractPolyhedron.
```
```suggestion
Return the intersection of the cartesian product of a finite number of convex sets and a polyhedron.
```

src/concrete_intersection.jl
```diff
@@ -654,3 +655,48 @@ end
...
+
+### Input
+
+ - `X` -- Cartesian Product Array of convex sets
```
```suggestion
 - `X` -- cartesian product of a finite number of convex sets
```

src/concrete_intersection.jl
```diff
@@ -654,3 +655,48 @@ end
...
+ - `Y` -- AbstractPolyhedron
```
```suggestion
 - `Y` -- polyhedron
```

src/concrete_intersection.jl
```diff
@@ -654,3 +655,48 @@ end
...
+
+### Output
+
+The Cartesian Product Array obtained by the intersection of the nonzero blocks of `X` and `Y`.
```
```suggestion
The concrete intersection betewen `X` and `Y`.
```

src/concrete_intersection.jl
```diff
@@ -654,3 +655,48 @@ end
...
+
+### Notes
+
+This function computes an intersection of CartesianProductArray and AbstractPolyhedron only for blocks, which are not free in AbstractPolyhedron object.
```
```suggestion
This function takes into account that if the polyhedron is unbounded, the intersection is only needed to be taken in the elements of the cartesian product array (subsets of variables, or "blocks") which are constrained; those which are not constrained do not need to be intersected. 
```

src/concrete_intersection.jl
```diff
@@ -654,3 +655,48 @@ end
...
+"""
+function intersection(X::CartesianProductArray{N},
```
mforets: missing: symmetric function `intersection(::AbstractPolyhedron, ::CartesianProductArray)`

src/concrete_intersection.jl
```diff
@@ -654,3 +655,48 @@ end
...
+### Input
+
+ - `X` -- Cartesian Product Array of convex sets
+ - `Y` -- AbstractPolyhedron
+
+### Output
+
+The Cartesian Product Array obtained by the intersection of the nonzero blocks of `X` and `Y`.
+
+### Notes
```
```suggestion
### Algorithm
```

test/unit_CartesianProduct.jl
```diff
@@ -198,4 +200,14 @@ for N in [Float64, Float32, Rational{Int}]
...
           EmptySet
     @test b × e == e × b == cpa × e == e × cpa == e × e == e
+
+    # intersection
+    set1 = Hyperrectangle(low=[0.;0.], high=[1.; 1.]);
+    set2 = Hyperrectangle(low=[0.5;0.5], high=[1.5; 1.5]);
+
+    ca1 = decompose(set1, blocks=[1, dim(set1)-1])
+
+    intersect1 = intersection(ca1,set2)
```
mforets: have you timed this, in master and on this branch?

would be nice to add an artificial example of more dimensionality. for example consider a high dim halfspace that is only constrained in 2 variables, and the cartesian product of hyperrectangles in 2-dim blocks.

test/unit_CartesianProduct.jl
```diff
@@ -198,4 +200,14 @@ for N in [Float64, Float32, Rational{Int}]
     @test absorbing(CartesianProduct) == absorbing(CartesianProductArray) ==
           EmptySet
     @test b × e == e × b == cpa × e == e × cpa == e × e == e
+
+    # intersection
+    set1 = Hyperrectangle(low=[0.;0.], high=[1.; 1.]);
+    set2 = Hyperrectangle(low=[0.5;0.5], high=[1.5; 1.5]);
+
+    ca1 = decompose(set1, blocks=[1, dim(set1)-1])
```
mforets:
```suggestion
    ca1 = decompose(set1, blocks=[1, 1])
```

test/unit_CartesianProduct.jl
mforets: i don't think that this test corresponds to the function of above...

mforets: i don't think that this tests the interesting part in the new function, since `set2` is bounded.

test/unit_CartesianProduct.jl
```diff
@@ -198,4 +200,14 @@ for N in [Float64, Float32, Rational{Int}]
...
+
+    intersect1 = intersection(ca1,set2)
+
+    @test Hyperrectangle(low=[0.5;0.5], high=[1.; 1.]) == project(intersect1,[1,2], Hyperrectangle)
```
mforets: i don't understand this test. maybe you can test set membership `intersect1 in ca1` and also in `set2`.

src/concrete_intersection.jl
```diff
@@ -654,3 +655,48 @@ end
...
+"""
+function intersection(X::CartesianProductArray{N},
+                      Y::AbstractPolyhedron{N}) where {N}
+
+    # preallocate the resulting CartesianProductArray with size hint
+    result = CartesianProductArray(length(X.array), N)
+
+    if isbounded(Y)
+        # no free variables
+        vars = Vector{Int}(1:dim(Y))
```
mforets: why convert to vector? this allocates. just let `vars = 1:dim(Y)` should work?

also could you please change the `vars` to something more readable, like `constrained_vars`?

src/concrete_intersection.jl
```diff
@@ -654,3 +655,48 @@ end
...
+    else
+        vars = constrained_dimensions(Y)
+    end
+    blocks = block_indices(X, vars)
+
+    for bi in 1:length(X.array)
+        if haskey(blocks,bi)
+            # otherwise, make the intersection with the projection of the halfspace
+            push!(result.array, intersection(X.array[bi], LazySets.Approximations.project(Y,variable_indices(X, bi, blocks[bi]), Hyperrectangle)))
```
mforets: (add `import LazySets.Approximations: project` on top of the file and use `project` here)

kostakoida: It causes compilation error for LazySets library

src/concrete_intersection.jl
kostakoida: Because `1:dim(Y)` return `Range`, while `constrained_dimensions` returns an array.

src/concrete_intersection.jl
```suggestion
            push!(result.array, intersection(X.array[bi], Approximations.project(Y,variable_indices(X, bi, blocks[bi]), Hyperrectangle)))
```

src/CartesianProduct.jl
```diff
@@ -549,3 +551,68 @@ function vertices_list(cpa::CartesianProductArray{N}
...
+number of sets.
+
+### Input
+
+- `ca` -- Cartesian product array
+- `vars` -- List of non free variables
+
+### Output
+
+A dictionary of block indices and indices of their first variables in the whole system.
```
mforets: Can you add an `Example` block with one or two illustrative examples?

src/CartesianProduct.jl
```diff
@@ -549,3 +551,68 @@ function vertices_list(cpa::CartesianProductArray{N}
...
+
+"""
+function block_indices(ca::CartesianProductArray{N}, vars::Vector{Int}) where {N}
```
mforets: make `vars` optional, defaulting to the empty set? also, i don't understand why you restrict it to `Vector{Int}`, why not letting it not annotated, so other iterable structs can be used?

mforets: on a second read, i would split this function on two smaller functions, one with one argument, and another one with no arguments. the behavior for `vars = empty` is generic (no need to dispatch on `CartesianProductArray`).

the behavior on the two argument function can also be used for other set types, so again i would not annotate the first argument to a `CartesianProductArray` (note that you can pass the input as `dim.(ca)`), instead use a vector (or abstractvector) of `Int`'s.

src/CartesianProduct.jl
```diff
@@ -549,3 +551,68 @@ function vertices_list(cpa::CartesianProductArray{N}
...
+    #key is the index of block, value is the index of the starting variable
+    result = Dict{Int,Int}()
+    start_index = 1
+    if isempty(vars)
+        for i in 1:length(ca.array)
+            result[i] = start_index
+            start_index += dim(ca.array[i])
+        end
+    else
+        for var in vars
```
mforets: i think that this logic is.. wrong? let the block partition be `[1,2], [3], [4,5,6]`. and suppose that `vars = [4]`. 

- at the beginning `start_index_1` and the loop in `i` is from `1` to `3`. 
- first `start_index <= var < start_index_dim(..) => 1 <= 4 < 3 => false`; 
- then `1 <= 4 < 1+1=2 => false`, and `1 <= 4 < 4 => false`.

so the `result` is empty. but i think that it should be the dictionary with keys `[3]` and values `[4]`?

probably the `start_index` variable has to be updated if the check in parentheses fails.

mforets:
(The function [compute_blocks](https://github.com/JuliaReach/Reachability.jl/blob/master/src/ReachSets/ContinuousPost/BFFPSV18/BFFPSV18.jl#L298) in Reachability may be useful to this PR)

mforets:
This branch seems to have conflict with master. @kostakoida could you check and rebase? thanks

PR #1090, opened 2019-02-07
[WIP]Concrete linear map of a cartesian product array

### LINKED ISSUE #1091: Decomposed linear map of CartesianProductArray
Outsourced from [this comment](https://github.com/JuliaReach/LazySets.jl/issues/1085#issuecomment-459389779).

kostakoida:
Closes #1091.


src/LinearMap.jl
```diff
@@ -389,8 +389,26 @@ Return the linear map of a lazy linear map.
...
+        block_dim = dim(ls)
+        proj = view(M, start_block_variable:(start_block_variable+block_dim-1), start_block_variable:(start_block_variable+block_dim-1))
+
+        push!(result.array, LinearMap(Array(proj), ls))
+        start_block_variable+=block_dim
+    end
+    return result
+end
+
+function linear_map(cp::CartesianProductArray{N}, M::AbstractMatrix{N}) where {N}
```
mforets: there is no need for the symmetric method, since matrix acts from the left to the set.

mforets:
We should try to exploit the existing `linear_map(::AbstractMatrix, ::AbstractPolyhedron)`  and perhaps deifne a new method that covers your case, with this signature:

```julia
linear_map(::AbstractMatrix, ::CartesianProductArray{N, <:AbstractPolyhedron})
```

src/CartesianProduct.jl
```diff
@@ -549,3 +550,36 @@ function vertices_list(cpa::CartesianProductArray{N}
...
+    if !isinvertible(M; cond_tol=cond_tol)
+        error("linear maps for Cartesian product arrays need to be invertible")
+    end
+
+    result = CartesianProductArray(length(cp.array), N)
+    col_st_index, col_end_ind = 0, 0
+    for bi in 1:length(cp.array)
+        col_st_index = col_end_ind + 1
+        col_end_ind += dim(cp.array[bi])
+        n = dim(cp.array[bi])
```
mforets: you can swap lines 567 and 568 and reuse n

src/CartesianProduct.jl
```diff
@@ -549,3 +550,36 @@ function vertices_list(cpa::CartesianProductArray{N}
...
+            push!(matrices, M[col_st_index:col_end_ind,start_ind:end_ind])
+            start_ind = end_ind + 1
+            end_ind += dim(cp.array[bi])
+        end
+        v_block = MinkowskiSumArray()
+        for m in matrices
+            push!(v_block.array,LinearMap(m, cp.array[bi]))
+        end
+
+        push!(result.array, overapproximate(v_block))
```
schillic: This is a serious problem: `overapproximate(v_block)` returns a hyperrectangle. Why is this needed?

src/CartesianProduct.jl
```diff
@@ -549,3 +550,36 @@ function vertices_list(cpa::CartesianProductArray{N}
...
+
+function linear_map(M::AbstractMatrix{N},
+    cp::CartesianProductArray{N};
+    cond_tol::Number=DEFAULT_COND_TOL,
+    use_inv::Bool=!issparse(M)) where {N<:Real}
+    if !isinvertible(M; cond_tol=cond_tol)
+        error("linear maps for Cartesian product arrays need to be invertible")
+    end
+
+    result = CartesianProductArray(length(cp.array), N)
```
mforets: Here it would expect it is preferable to preallocate a `CartesianProductArray` of a given concrete type if it is known, for example, if they are `LinearMap` or `Hyperrectangle` then use those. You can run the code with `@code_warntype` to check and fix those type instabilities.

src/CartesianProduct.jl
```diff
@@ -549,3 +550,36 @@ function vertices_list(cpa::CartesianProductArray{N}
 
     return vlist
 end
+
+
+function linear_map(M::AbstractMatrix{N},
+    cp::CartesianProductArray{N};
+    cond_tol::Number=DEFAULT_COND_TOL,
+    use_inv::Bool=!issparse(M)) where {N<:Real}
```
schillic: Align these arguments with `M::` above.

src/CartesianProduct.jl
```diff
@@ -549,3 +550,36 @@ function vertices_list(cpa::CartesianProductArray{N}
 
     return vlist
 end
+
+
+function linear_map(M::AbstractMatrix{N},
```
schillic: Add a docstring, including an explanation of the implementation.

src/CartesianProduct.jl
```diff
@@ -549,3 +550,36 @@ function vertices_list(cpa::CartesianProductArray{N}
...
+    cp::CartesianProductArray{N};
+    cond_tol::Number=DEFAULT_COND_TOL,
+    use_inv::Bool=!issparse(M)) where {N<:Real}
+    if !isinvertible(M; cond_tol=cond_tol)
+        error("linear maps for Cartesian product arrays need to be invertible")
+    end
+
+    result = CartesianProductArray(length(cp.array), N)
+    col_st_index, col_end_ind = 0, 0
```
mforets: I suggest to outsource this preparation step that identifies the starting and ending positions of blocks of a given Cartesian product to its own separate function. It is easier to reason about small functions with a concrete purpose, and this strategy is good for code reuse in general.

src/CartesianProduct.jl
```diff
@@ -549,3 +550,36 @@ function vertices_list(cpa::CartesianProductArray{N}
...
+    for bi in 1:length(cp.array)
+        col_st_index = col_end_ind + 1
+        col_end_ind += dim(cp.array[bi])
+        n = dim(cp.array[bi])
+        start_ind, end_ind = 1, dim(cp.array[bi])
+        matrices = Vector{Matrix{N}}()
+        while end_ind <= size(M)[1]
```
mforets: This seems like a bug to me -- shouldn't we have `size(M, 2)` here?

src/CartesianProduct.jl
```diff
@@ -549,3 +550,36 @@ function vertices_list(cpa::CartesianProductArray{N}
...
+
+
+function linear_map(M::AbstractMatrix{N},
+    cp::CartesianProductArray{N};
+    cond_tol::Number=DEFAULT_COND_TOL,
+    use_inv::Bool=!issparse(M)) where {N<:Real}
+    if !isinvertible(M; cond_tol=cond_tol)
+        error("linear maps for Cartesian product arrays need to be invertible")
+    end
+
```
mforets: It is reasonable to assert here that the linear map is commensurate with the cartesian product: `size(M, 2) = dim(cp)`.

src/CartesianProduct.jl
schillic: agreed

src/CartesianProduct.jl
```diff
@@ -549,3 +550,36 @@ function vertices_list(cpa::CartesianProductArray{N}
...
+    result = CartesianProductArray(length(cp.array), N)
+    col_st_index, col_end_ind = 0, 0
+    for bi in 1:length(cp.array)
+        col_st_index = col_end_ind + 1
+        col_end_ind += dim(cp.array[bi])
+        n = dim(cp.array[bi])
+        start_ind, end_ind = 1, dim(cp.array[bi])
+        matrices = Vector{Matrix{N}}()
+        while end_ind <= size(M)[1]
+            push!(matrices, M[col_st_index:col_end_ind,start_ind:end_ind])
```
mforets: This `matrices` vector just creates extra allocations that are not needed. Use a view array for the particular block of `M` where it needs to be applied instead (line 578).

schillic:
I think this implementation by taking the lazy `LinearMap` and then overapproximating is not good. Instead I suggest to apply the concrete `linear_map` to each block.

PR #1093, opened 2019-02-07
Add invertibility check for concrete linear maps

### LINKED ISSUE #1085: 2-norm condition number is not implemented for sparse matrices
To reproduce this issue run the following script:
```
julia> using SparseArrays

julia> using LazySets

julia> set1 = HalfSpace{Float64}([-1.0, 0.0], 0.0)
HalfSpace{Float64}([-1.0, 0.0], 0.0)

julia> lmap1 = sparse([1, 2],[1, 2],[1., 1.])
2×2 SparseMatrixCSC{Float64,Int64} with 2 stored entries:
  [1, 1]  =  1.0
  [2, 2]  =  1.0

julia> L1 = LinearMap(lmap1, set1)
LinearMap{Float64,HalfSpace{Float64},Float64,SparseMatrixCSC{Float64,Int64}}(
  [1, 1]  =  1.0
  [2, 2]  =  1.0, HalfSpace{Float64}([-1.0, 0.0], 0.0))

julia> set21 = Interval(9.99499, 10.2051)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([9.99498, 10.2051])

julia> set22 = Interval(-0.100001, 0.)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([-0.100002, 0])

julia> ca2 = CartesianProductArray([set21, set22])
CartesianProductArray{Float64,Interval{Float64,IntervalArithmetic.Interval{Float64}}}(Interval{Float64,IntervalArithmetic.Interval{Float64}}[Interval{Float64,Interval{Float64}}([9.99498, 10.2051]), Interval{Float64,Interval{Float64}}([-0.100002, 0])])

julia> L2 = LinearMap(lmap1, ca2)
LinearMap{Float64,CartesianProductArray{Float64,Interval{Float64,IntervalArithmetic.Interval{Float64}}},Float64,SparseMatrixCSC{Float64,Int64}}(
  [1, 1]  =  1.0
  [2, 2]  =  1.0, CartesianProductArray{Float64,Interval{Float64,IntervalArithmetic.Interval{Float64}}}(Interval{Float64,IntervalArithmetic.Interval{Float64}}[Interval{Float64,Interval{Float64}}([9.99498, 10.2051]), Interval{Float64,Interval{Float64}}([-0.100002, 0])]))

julia> intersection(L1, L2)
ERROR: ArgumentError: 2-norm condition number is not implemented for sparse matrices, try cond(Array(A), 2) instead
Stacktrace:
 [1] cond(::SparseMatrixCSC{Float64,Int64}, ::Int64) at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/SparseArrays/src/linalg.jl:950
 [2] cond at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/SparseArrays/src/linalg.jl:941 [inlined]
 [3] #isinvertible#1(::Float64, ::Function, ::SparseMatrixCSC{Float64,Int64}) at /Users/forets/.julia/dev/LazySets/src/helper_functions.jl:115
 [4] #isinvertible at ./none:0 [inlined]
 [5] #linear_map#54(::Float64, ::Function, ::SparseMatrixCSC{Float64,Int64}, ::HPolyhedron{Float64}) at /Users/forets/.julia/dev/LazySets/src/HPolyhedron.jl:427
 [6] linear_map(::SparseMatrixCSC{Float64,Int64}, ::HPolyhedron{Float64}) at /Users/forets/.julia/dev/LazySets/src/HPolyhedron.jl:427
 [7] linear_map(::SparseMatrixCSC{Float64,Int64}, ::HalfSpace{Float64}) at /Users/forets/.julia/dev/LazySets/src/HalfSpace.jl:376
 [8] intersection(::LinearMap{Float64,HalfSpace{Float64},Float64,SparseMatrixCSC{Float64,Int64}}, ::LinearMap{Float64,CartesianProductArray{Float64,Interval{Float64,IntervalArithmetic.Interval{Float64}}},Float64,SparseMatrixCSC{Float64,Int64}}) at /Users/forets/.julia/dev/LazySets/src/concrete_intersection.jl:657
 [9] top-level scope at none:0
```

### LINKED ISSUE #1092: Add option to linear map of abstract polytope
Proposal: add an option `use_constraints::Bool` such that `linear_map(M, X)` uses the current implementation (vertices) if it is false, and otherwise it uses the implementation for `HPolytope` (check for invertibility criterion in which case the linear map is applied using its inverse).

This is useful for sets such that their constraints list is available or can be computed more easily than its vertices:

```julia
julia> n = 20; M = rand(n, n);

julia> H = rand(Hyperrectangle, dim=n);

julia> @time linear_map(M, H);  # current master
  1.652406 seconds (4.19 M allocations: 920.006 MiB, 12.87% gc time)

julia> @time linear_map(M, convert(HPolytope, H)); # using the hpolytope implementation
  0.000282 seconds (438 allocations: 57.891 KiB)
```

### LINKED ISSUE #1132: Remove conversion from concrete linear map of VPolytope

### LINKED ISSUE #1133: Revise linear map of AbstractPolygon
- [Concrete linear maps of an HPolygon](https://github.com/JuliaReach/LazySets.jl/blob/f9cfbed30507c44e78ff3aa1def2372e7166fff0/src/AbstractPolygon.jl#L225) should use the method from `AbstractPolyhedron`. 

- Write a `linear_map` method that dispatches on `VPolygon`.

mforets:
Closes #1085.
Closes #1092.
Closes #1132.
Closes #1133.

~~Requires: #1080.~~


mforets:
I don't know if `check_invertibility` is good, maybe we prefer to call it `use_vertex_representation` or `use_vertices`. We should use the same convention for https://github.com/JuliaReach/LazySets.jl/issues/1092 .

src/HPolyhedron.jl
```diff
@@ -422,23 +425,29 @@ constraint system ``A x ≤ b`` to ``A \\text{inv}(M) y ≤ b``.
...
-    if !isinvertible(M; cond_tol=cond_tol)
-        if P isa HPolyhedron
-            error("linear maps for polyhedra need to be invertible")
+    if check_invertibility
+        if !isinvertible(M; cond_tol=cond_tol)
+            if P isa HPolyhedron
+                error("linear maps for polyhedra need to be invertible")
+            end
+            # use the implementation for general polytopes
+            return invoke(linear_map, Tuple{typeof(M), AbstractPolytope{N}}, M, P)
```
schillic: I do not like the flow of this method, in particular the duplication of the `invoke`.

mforets: Suppose that we name it `use_vertex_representation`, and that it is set to `false`. Then it may happen that it tries to use the invertibility criterion but it fails. Hence the name of the flag is misleading in this situation.

But `check_invertibility` does not suffer from this because if it is `true` the check is made, and it computes without passing to v-rep if possible, and if `false` the check is not made and it always computes in the v-rep.

mforets: > I do not like the flow of this method, in particular the duplication of the `invoke`.

I have uploaded a new version, which uses `invoke` only once. The idea is to use short-circuited OR for the cases that will make the function use the v-rep. I removed the check for P being a HPolyhedron and linear map being not invertible, because i think it is not needed. But let me note if there is some use case that i'm missing.

schillic: > `use_vertex_representation`, and that it is set to `false`. Then it may happen that it tries to use the invertibility criterion

But it should not, that is exactly my point. If that option is used, we have to go to vertices. Otherwise, we may do whatever we want.

schillic: > I have uploaded a new version, which uses `invoke` only once.

Yes, it is :+1: now, but read on :wink:

> I removed the check for P being a HPolyhedron and linear map being not invertible, because i think it is not needed.

It is needed because `vertices_list` is not working as this method would it expect to. I am worried that [this test](https://github.com/JuliaReach/LazySets.jl/blob/master/test/unit_Polyhedron.jl#L84) does not fail now.

mforets: To be clear, i expect for an option `use_vertex_representation` that is set to `false` => it never uses the vertices. Otherwise we need to make it a kwarg: "always", "if_not_invertible".

schillic: I would say:
* `use_vertex_representation == true` => use vertices
* `use_vertex_representation == false` => do whatever you want

src/HPolyhedron.jl
```diff
@@ -426,20 +430,17 @@ sparse matrices. In that case, either assure that `use_inv=false`, or use
 """
 function linear_map(M::AbstractMatrix{N},
                     P::PT;
+                    check_invertibility::Bool=true,
                     cond_tol::Number=DEFAULT_COND_TOL,
                     use_inv::Bool=!issparse(M)
                    ) where {N<:Real, PT<:HPoly{N}}
-    if !isinvertible(M; cond_tol=cond_tol)
-        if P isa HPolyhedron
-            error("linear maps for polyhedra need to be invertible")
```
schillic: I think keeping this error message is helpful.

src/HPolyhedron.jl
```diff
@@ -426,20 +430,17 @@ sparse matrices. In that case, either assure that `use_inv=false`, or use
 """
 function linear_map(M::AbstractMatrix{N},
                     P::PT;
```
schillic: Let us generalize this method to `P::AbstractPolyhedron{N}`.

src/HPolyhedron.jl
```diff
@@ -415,26 +419,42 @@ Concrete linear map of a polyhedron in constraint representation.
 
 ### Output
 
-A polyhedron of the same type as the input (`PT`).
+A polyhedron of the same type as the input if `M` is invertible; otherwise,
```
schillic: This is not true in general. Imagine `P` is a `Singleton.

src/HPolyhedron.jl
```diff
@@ -411,35 +419,48 @@ Concrete linear map of a polyhedron in constraint representation.
...
+
+To switch off the check for invertibility, set the option
+`check_invertibility=false`. If `M` is not invertible and the polyhedron
+is unbounded, this function returns an exception.
 
-The option `use_inv` is a workaround to allow using the invertibility condition
-when `M` is a sparse matrix, since the `inv` function is not available for
-sparse matrices. In that case, either assure that `use_inv=false`, or use
-`linear_map(Matrix(M), P)`.
+The option `use_inv` lets the user to control - in case `M` is invertible - if
```
schillic:
```suggestion
The option `use_inv` lets the user control - in case `M` is invertible - if
```

src/HPolyhedron.jl
```diff
@@ -411,35 +419,48 @@ Concrete linear map of a polyhedron in constraint representation.
...
+the full matrix inverse is computed, or only the left division on the normal
+vectors. Note that this helps as a workaround when `M` is a sparse matrix, since
+the `inv` function is not available for sparse matrices. In this case, either
+let that `use_inv=false`, or use `linear_map(Matrix(M), P)`.
```
schillic:
```suggestion
use the option `use_inv=false` or convert `M` as in `linear_map(Matrix(M), P)`.
```

test/unit_Polyhedron.jl
```diff
@@ -187,5 +189,46 @@ if test_suite_polyhedra
...
+
+        Pbdd = HPolyhedron([HalfSpace([0.68, 1.22], -0.76),
+                            HalfSpace{Float64}([-0.75, -0.46], 0.68),
+                            HalfSpace{Float64}([1.72, 0.33], 0.37),
+                            HalfSpace{Float64}([-1.60, -0.41], 0.67),
+                            HalfSpace{Float64}([-0.44, 0.06], 0.78)])
+        @assert isbounded(Pbdd)
+
+        Mnotinv = [1.0 0.0; 2.0 0.0]
+        @assert !LazySets.isinvertible(Mnotinv)
```
schillic: Maybe add `using LazySets: isinvertible` to this file.

test/unit_Polyhedron.jl
```diff
@@ -187,5 +189,46 @@ if test_suite_polyhedra
...
+
+        # invertible matrix times an unbounded polyhedron 
+        L = linear_map(Minv, Punbdd)
+        @test L isa HPolyhedron
+
+        # not invertible matrix times a bounded polyhedron 
+        L = linear_map(Mnotinv, Pbdd) # Requires Polyhedra because it works on vertices
+        @test L isa VPolytope
+
+        # not invertible matrix times a unbbounded polyhedron
```
schillic:
```suggestion
        # not invertible matrix times an unbounded polyhedron
```

src/HPolyhedron.jl
```diff
@@ -411,47 +419,88 @@ Concrete linear map of a polyhedron in constraint representation.
...
+end
+
+function _linear_map_hrep(M::AbstractMatrix{N}, P::AbstractPolytope{N},
+                          use_inv::Bool) where {N<:Real}
+    constraints = _linear_map_hrep_helper(M, P, use_inv)
+    return dim(P) == 2 ? HPolygon(constraints) : HPolytope(constraints)
+end
+
+function _linear_map_hrep(M::AbstractMatrix{N}, P::HalfSpace{N},
+                          use_inv::Bool) where {N<:Real}
```
schillic: Do we want to have a specific implementation for the other polyhedral types (`Hyperplane`, `Line`) as well?

src/VPolytope.jl
```diff
@@ -218,9 +219,37 @@ function rand(::Type{VPolytope};
...
+The linear map ``M`` is applied to each vertex of the given set ``P``, obtaining
+a polytope in V-representation.
+
+The default output is a `VPolytope`. If an `output_type` is given, the
+corresponding `convert` method is invoked.
+"""
+function linear_map(M::AbstractMatrix{N}, P::VPolytope{N};
+                    output_type::Type{<:LazySet}=VPolytope{N}) where {N<:Real}
+    @assert dim(P) == size(M, 2) "a linear map of size $(size(M)) cannot be applied to a set of dimension $(dim(P))"
+    return convert(output_type, _linear_map_vrep(M, P)) # TODO: remove conversion?
```
schillic: Yes, let us remove this conversion.

src/HPolyhedron.jl
```diff
@@ -411,47 +419,88 @@ Concrete linear map of a polyhedron in constraint representation.
...
+function _linear_map_hrep(M::AbstractMatrix{N}, P::AbstractPolyhedron{N},
+                          use_inv::Bool) where {N<:Real}
+    constraints = _linear_map_hrep_helper(M, P, use_inv)
+    return HPolyhedron(constraints)
+end
+
+function _linear_map_hrep(M::AbstractMatrix{N}, P::AbstractPolytope{N},
+                          use_inv::Bool) where {N<:Real}
+    constraints = _linear_map_hrep_helper(M, P, use_inv)
+    return dim(P) == 2 ? HPolygon(constraints) : HPolytope(constraints)
```
schillic: I begin to not like this line. If the user works with `HPolytope`s, it would be surprising if the result becomes an `HPolygon`. Let us add another method for `AbstractHPolygon` :blush:

src/VPolytope.jl
```diff
@@ -218,9 +219,37 @@ function rand(::Type{VPolytope};
...
 
+### Algorithm
+
+The linear map ``M`` is applied to each vertex of the given set ``P``, obtaining
+a polytope in V-representation.
+
+The default output is a `VPolytope`. If an `output_type` is given, the
+corresponding `convert` method is invoked.
+"""
+function linear_map(M::AbstractMatrix{N}, P::VPolytope{N};
```
schillic: Can we have such a method for `VPolygon` as well?

test/unit_HalfSpace.jl
```diff
@@ -97,10 +97,10 @@ for N in [Float64, Rational{Int}, Float32]
...
           LazySets.is_tighter_same_dir_2D(c3, c2, strict=true)
 
-    # test linear map of a half-space
+    # test concrete linear map of a half-space
     H = HalfSpace(N[1.0, -1.0], N(0.0)) # x <= y
-    M = Matrix(-N(1.0)*I, 2, 2)
-    MH = linear_map(M, H)
-    @test constraints_list(MH)[1] == HalfSpace(N[-1.0, 1.0], N(0.0)) # x >= y
-
+    M = N[1. 0; 0 0] # non-invertible matrix
```
schillic:
```suggestion
    M = N[1 0; 0 0] # non-invertible matrix
```

test/unit_HalfSpace.jl
```diff
@@ -97,10 +97,10 @@ for N in [Float64, Rational{Int}, Float32]
...
+    @test_throws ArgumentError linear_map(M, H)
+    M = N[2. 2; 0 1.] # invertible matrix
```
schillic:
```suggestion
    M = N[2 2; 0 1] # invertible matrix
```

test/unit_HalfSpace.jl
```diff
@@ -97,10 +97,10 @@ for N in [Float64, Rational{Int}, Float32]
     @test !LazySets.is_tighter_same_dir_2D(c1, c2, strict=true) &&
           LazySets.is_tighter_same_dir_2D(c3, c2, strict=true)
 
-    # test linear map of a half-space
+    # test concrete linear map of a half-space
     H = HalfSpace(N[1.0, -1.0], N(0.0)) # x <= y
```
schillic:
```suggestion
    H = HalfSpace(N[1, -1], N(0)) # x <= y
```

test/unit_Polyhedron.jl
```diff
@@ -187,5 +190,46 @@ if test_suite_polyhedra
...
+        L = linear_map(Minv, Pbdd)
+        @test L isa HPolyhedron
+
+        # invertible matrix times an unbounded polyhedron 
+        L = linear_map(Minv, Punbdd)
+        @test L isa HPolyhedron
+
+        # not invertible matrix times a bounded polyhedron 
+        L = linear_map(Mnotinv, Pbdd) # Requires Polyhedra because it works on vertices
+        @test L isa VPolytope
```
schillic: This test should probably be moved to `unit_Polytope`, and we need a respective test (for the output type) for the `HPolytope` case as well.

test/unit_Polyhedron.jl
```diff
@@ -1,3 +1,6 @@
+using SparseArrays
```
schillic: `@static if VERSION >= v"0.7-" using SparseArrays end`

src/HPolyhedron.jl
mforets: True, this line should be improved. Since the input types can be quite different, i was thinking in something along these lines

```julia
A set whose type is as "close" as possible to `P`. A polygon (or polytope) in vertex representation if this algorithm has to resort to the vertex representation of `P`; otherwise a set whose type is that same as that of `P`.
```

schillic: I think you do not understand my point.
If the input is a `Hyperrectangle`, we will create an `HPolytope`. And `Hyperrectangle != HPolytope` :wink:

Here is a suggestion which already incorporates my suggested change for `HPolygon`.

> The result type depends on the input and the approach that was used.
If the input is a `HalfSpace`, then the output is a `HalfSpace`.
If the vertex-based approach was used, then the output is an `VPolytope`.
If the input set is an `AbstractPolygon` , then the output is an `HPolygon`.
If the input set is an `AbstractPolytope` , then the output is an `HPolytope`.
Otherwise, the output is an `HPolyhedron`.

docs/src/lib/operations.md
```diff
@@ -298,7 +298,6 @@ Inherited from [`LazySet`](@ref):
 Inherited from [`AbstractPolytope`](@ref):
 * [`isbounded`](@ref isbounded(::AbstractPolytope))
 * [`singleton_list`](@ref singleton_list(::AbstractPolytope{N}) where {N<:Real})
-* [`linear_map`](@ref linear_map(::AbstractMatrix{N}, ::AbstractPolytope{N}) where {N<:Real})
```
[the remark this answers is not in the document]
mforets: Yes, you're right. Let me replace it and put back this method there.

src/HPolyhedron.jl
```diff
@@ -411,47 +419,88 @@ Concrete linear map of a polyhedron in constraint representation.
...
     # matrix M is invertible => the normal vectors are vec(c.a' * inv(M))
+    return _linear_map_hrep(M, P, use_inv)
+end
+
+function _linear_map_vrep(M::AbstractMatrix{N}, P::AbstractPolyhedron{N}) where {N<:Real}
+    if !isbounded(P)
+        throw(ArgumentError("the linear map in vertex representation for an unbounded set is not defined"))
+    else
+        # since P is bounded, we pass an HPolytope and then convert it to vertex representation
+        P = tovrep(HPolytope(constraints_list(P), check_boundedness=false))
```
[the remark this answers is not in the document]
mforets: The purpose of asking for the `constraints_list` is that it is guaranteed by the type of the agrument (`P::AbstractPolyhedron{N}`). So i'm not sure to use change to `vertices_list` here for any `P` (at least not until https://github.com/JuliaReach/LazySets.jl/issues/1129 of course).

> Usually the `vertices_list` method is defined and more efficient than going to H-rep first.

For zonotope and hyperrectangle this is what i get:

```julia
julia> Z = rand(Zonotope, dim=5);

julia> @btime vertices_list($Z);
  53.660 μs (587 allocations: 57.38 KiB)

julia> @btime constraints_list($Z);
  128.345 μs (1322 allocations: 392.06 KiB)

julia> H = rand(Hyperrectangle, dim=5);

julia> @btime vertices_list($H);
  2.982 μs (46 allocations: 5.53 KiB)

julia> @btime constraints_list($H);
  476.206 ns (24 allocations: 1.63 KiB)
```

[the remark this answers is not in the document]
schillic: Your comparison is not fair. You have to see the whole context. In this code, you take a polyhedron, compute its constraints, and then compute its vertices from the polytope with those constraints.
But okay, I just saw that you have a similar function for `AbstractPolytope`. So we can keep it as is.

mforets:
> Should we maybe check if the input is a `VPolytope` or `VPolygon`? In these cases we should probably use the vertices.

Right, we should do something about this, but it can be done in a subsequent PR. At the moment, `VPolygon` and `HPolygon` use the `linear_map` function defined in the `AbstractPolygon.jl` interface. I've created the issue #1133 for this.

On the other hand, `VPolytope` uses the specific method in `VPolytope.jl` which i think it's good. But i think that we should remove the option `output_type`, since `convert` can be done from the outside more clearly. I've opened #1132.

Finally, `HPolytope` uses the new method in this PR so it should be ok.

docs/src/lib/operations.md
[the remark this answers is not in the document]
mforets: 👍  I've added back the reference to `AbstractPolyhedron` [here](https://github.com/JuliaReach/LazySets.jl/pull/1093/files#diff-b32ad525d8110d7cd7862f2fa2758ba5R322).

src/HPolyhedron.jl
mforets: Thanks for the good remark; for those types the linear map can be closed too. Initially i outsourced this to https://github.com/JuliaReach/LazySets.jl/issues/1131 but it turns out to be so simple that we don't need that anymore.

src/VPolytope.jl
mforets: See #1132.

src/VPolytope.jl
mforets: Yes, having a `linear_map` to dispatch on `VPolygon` should be cleaner. This is the purpose of #1133.

src/HPolyhedron.jl
mforets: Again thanks for the good remark. Let's keep this method as fallback (`_linear_map_hrep` to dispatch on the general `AbstractPolytope`) but as you propose add two new methods to guarantee that the types are stable, one for `AbstractHPolygon` and another one for `HPolytope`.

src/HPolyhedron.jl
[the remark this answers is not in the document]
mforets: I was thinking about this and if we add a fallback implementation `tovrep` method and some specific methods it's better. I've opened this https://github.com/JuliaReach/LazySets.jl/issues/1134 .

The argument is the following: suppose that you begin with a `HPolygon` and that the matrix is not invertible, so one passes to the vertex representation. But then the `tovrep` of HPolygon is largely more efficient than the one for general `HPolytope`. If we use `vertices_list(P)` it also works as you proposed so these ideas are very close. But to the `tovrep` allows to control the return type using dispatch (`VPolygon` / `VPolytope`).

mforets:
Thanks a lot for the review!

Do you want to double check that i didn't miss any comment?

Tests pass locally but i only tested in v1.1.

src/AbstractPolyhedron_functions.jl
```diff
@@ -211,3 +212,113 @@ function remove_redundant_constraints(constraints::AbstractVector{LinearConstrai
...
+
+### Output
+
+The result type depends on the input and the algorithm that was used:
+
+- If the vertex-based approach was used and the dimension of `P` is `2`, the output
+  is a `VPolygon`, otherwise a `VPolytope`.
+- If the input set is an `AbstractPolytope` , then the output is an `HPolytope`.
+- If the input set is an `AbstractPolygon` , then the output is an `HPolygon`.
+- The types of `HalfSpace`, `Hyperplane` and `Line` are preserved.
```
schillic: Merge the last two lines:
``The types of `HalfSpace`, `Hyperplane`, `Line`, and `AbstractHPolygon` are preserved.``

src/AbstractPolyhedron_functions.jl
```diff
@@ -211,3 +212,113 @@ function remove_redundant_constraints(constraints::AbstractVector{LinearConstrai
...
+- `use_inv`  -- (optional, default: `false` if `M` is sparse and `true`
+                otherwise) whether to compute the full left division through
+                `inv(M)`, or to use the left division for each vector; see below 
+
+### Output
+
+The result type depends on the input and the algorithm that was used:
+
+- If the vertex-based approach was used and the dimension of `P` is `2`, the output
+  is a `VPolygon`, otherwise a `VPolytope`.
```
schillic: The "otherwise" in this sentence is ambiguous.
``If the vertex-based approach was used, the output is a `VPolytope` (resp. a `VPolygon` in 2D).``

src/AbstractPolyhedron_functions.jl
```diff
@@ -211,3 +212,113 @@ function remove_redundant_constraints(constraints::AbstractVector{LinearConstrai
...
+- Otherwise, the output is an `HPolyhedron`.
+
+### Algorithm
+
+This function implements two algorithms for the linear map:
+
+- If the matrix ``M`` is invertible (which we check with a sufficient condition),
+  then ``y = M x`` implies ``x = \\text{inv}(M) y`` and we transform the
+  constraint system ``A x ≤ b`` to ``A \\text{inv}(M) y ≤ b``.
+- Otherwise, transform the polyhedron to vertex representation and apply the map
```
schillic:
```suggestion
- Otherwise, we transform the polyhedron to vertex representation and apply the map
```

src/AbstractPolyhedron_functions.jl
```diff
@@ -211,3 +212,113 @@ function remove_redundant_constraints(constraints::AbstractVector{LinearConstrai
...
+is unbounded, this function returns an exception.
+
+The option `use_inv` lets the user control - in case `M` is invertible - if
+the full matrix inverse is computed, or only the left division on the normal
+vectors. Note that this helps as a workaround when `M` is a sparse matrix, since
+the `inv` function is not available for sparse matrices. In this case, either
+use the option `use_inv=false` or convert the type of `M` as in
+`linear_map(Matrix(M), P)`.
+"""
+function linear_map(M::AbstractMatrix{N},
```
schillic: Can you add a comment on `_linear_map_hrep` and `_linear_map_vrep`? This is helpful when adding new subtypes to the interface.

src/AbstractPolyhedron_functions.jl
```diff
@@ -211,3 +212,113 @@ function remove_redundant_constraints(constraints::AbstractVector{LinearConstrai
...
+end
+
+function _linear_map_vrep(M::AbstractMatrix{N}, P::AbstractPolyhedron{N}) where {N<:Real}
+    if !isbounded(P)
+        throw(ArgumentError("the linear map in vertex representation for an unbounded set is not defined"))
+    else
+        # since P is bounded, we pass an HPolytope and then convert it to vertex representation
+        P = tovrep(HPolytope(constraints_list(P), check_boundedness=false))
+    end
+    return _linear_map_vrep(M, P)
```
schillic: I would pull out the `else` case for clarity.

src/AbstractPolytope.jl
```diff
@@ -128,6 +97,22 @@ function default_polyhedra_backend(P, N)
...
+# given a polytope P, apply the linear map P to each vertex of P
+# it is assumed that the interface function `vertices_list(P)` is available 
+@inline function _linear_map_vrep(M::AbstractMatrix{N}, P::AbstractPolytope{N}) where {N<:Real}
+    return broadcast(v -> M * v, vertices_list(P)) |> VPolytope{N}
+end
+
+function _linear_map_hrep(M::AbstractMatrix{N}, P::AbstractPolytope{N},
+                          use_inv::Bool) where {N<:Real}
+    constraints = _linear_map_hrep_helper(M, P, use_inv)
+    return dim(P) == 2 ? HPolygon(constraints) : HPolytope(constraints)
```
schillic: A few lines above the `_linear_map_vrep` always returns a `VPolytope` even in 2D. I find this inconsistent.

src/HalfSpace.jl
```diff
@@ -419,20 +417,7 @@ function is_tighter_same_dir_2D(c1::LinearConstraint{N},
...
-
-### Output
-
-The half-space obtained by applying the given linear map to the half-space.
-"""
-function linear_map(M::AbstractMatrix{N}, hs::HalfSpace{N}) where {N}
-    return linear_map(M, HPolyhedron([hs]))
+function _linear_map_hrep(M::AbstractMatrix{N}, P::HalfSpace{N}, use_inv::Bool) where {N<:Real}
+    constraint = _linear_map_hrep_helper(M, P, use_inv)
+    return HalfSpace(constraint[1].a, constraint[1].b)
```
schillic:
```julia
    constraint = _linear_map_hrep_helper(M, P, use_inv)[1]
    return HalfSpace(constraint.a, constraint.b)
```
(same for `Hyperplane` and `Line`)

src/Hyperplane.jl
```diff
@@ -395,3 +395,8 @@ function _constraints_list_hyperplane(a::AbstractVector{N}, b::N
                                      )::Vector{LinearConstraint{N}} where {N<:Real}
     return [HalfSpace(a, b), HalfSpace(-a, -b)]
 end
+
+function _linear_map_hrep(M::AbstractMatrix{N}, P::Hyperplane{N}, use_inv::Bool) where {N<:Real}
+    constraint = _linear_map_hrep_helper(M, P, use_inv)
+    return Hyperplane(constraint[1].a, constraint[1].b)
```
schillic: Hm, so for the linear map with the identity matrix we might change the `b` value to `-b`. I think that is fine, but I wanted to mention it.

src/AbstractPolytope.jl
mforets: Yes, to fix the case of polygons and polytopes by writing specific methods for these is the purpose of #1132 and #1133.

src/AbstractPolyhedron_functions.jl
mforets: Done, and i also splitted the paragraphs into two separate ones, depending on the algorithm that was used.

src/AbstractPolyhedron_functions.jl
mforets: I've re-written this, could you please review again?

src/AbstractPolyhedron_functions.jl
```diff
@@ -211,3 +212,126 @@ function remove_redundant_constraints(constraints::AbstractVector{LinearConstrai
...
+consider post-processing the result of this function with a call to a suitable
+`convert` method.
+
+In particular, the output depends on the type of `P` and the algorithm that was used:
+
+- If the vertex-based approach was used:
+    
+    - If `P` is a `VPolygon` then the output is a `VPolygon`.
+    - If `P` is a `VPolytope` then the output is a `VPolytope`.
+    - Otherwise, the output is a `VPolygon` if the dimension of `P` is `2`, or a
```
schillic:
```suggestion
    - Otherwise, the output is a `VPolygon` if the dimension of `P` is two and a
```
(to make it consistent with the other paragraph)

src/Hyperplane.jl
mforets: True. At first i thought about adding some "normaization" but since the two forms `ax = b`  and `-ax = -b` are mathematically equivalent, I think that it is acceptable.

schillic: Yes, the sets are the same. It may just be unexpected.

src/AbstractPolytope.jl
mforets: They are so related to this PR that let's do them on this issue as well.

src/HPolyhedron.jl
mforets: Alright, i have added two new methods to guarantee that the types are stable. In particular:

- the linear map of an `HPolytope` passes first through the `AbstractPolyhedron` interface `linear_map`; then goes to `_linear_map_vrep` and returns a `VPolygon` or a `VPolytope` depending on the dimension, and if h-rep is used returns `HPolytope`

src/AbstractPolyhedron_functions.jl
```diff
@@ -211,3 +212,139 @@ function remove_redundant_constraints(constraints::AbstractVector{LinearConstrai
...
+               cond_tol::Number=DEFAULT_COND_TOL,
+               use_inv::Bool=!issparse(M)
+               ) where {N<:Real}
+
+Concrete linear map of a polyhedron in constraint representation.
+
+### Input
+
+- `M` -- matrix
+- `P` -- polyhedron in constraint representation
```
mforets:
```suggestion
- `P` -- abstract polyhedron
```

PR #1108, opened 2019-02-12
WIP #1105 - Add an_element specific for rational input

### LINKED ISSUE #1105: An element for HPolytope is not type stable
```julia
julia> P
HPolytope{Rational{Int64}}(HalfSpace{Rational{Int64}}[HalfSpace{Rational{Int64}}(Rational{Int64}[1//1, 0//1], 1//1), HalfSpace{Rational
{Int64}}(Rational{Int64}[0//1, 1//1], 1//1), HalfSpace{Rational{Int64}}(Rational{Int64}[-1//1, 0//1], 1//1), HalfSpace{Rational{Int64}}
(Rational{Int64}[0//1, -1//1], 1//1)])

julia> an_element(P)
2-element Array{Float64,1}:
 1.0
 0.0
```

mforets:
Closes #1105.

With respect to the comment [here](https://github.com/JuliaReach/LazySets.jl/issues/1105#issuecomment-462601735), i decided to go for using the method for floating point, then checking membership, if it works good, otherwise use the Chebyshev method (which should always work, i believe).

### DIFF
--- a/src/HPolyhedron.jl
+++ b/src/HPolyhedron.jl
@@ -454,6 +454,29 @@ function linear_map(M::AbstractMatrix{N},
     return PT(constraints)
 end
 
+# specific an_element function for rational inputs
+function an_element(P::HPolytope{N};
+                    backend=nothing) where {N<:Rational{Int}}
+    # use the floating point method, then convert to rational and check membership
+    p = invoke(an_element, Tuple{HPolytope}, P)
+    p_rat = convert(Vector{N}, p)
+    p_rat ∈ P && return p_rat
+
+    # if the above went wrong, use the Chebyshev center
+    @assert isdefined(@__MODULE__, :Polyhedra) "this function needs the package " *
+                                               "'Polyhedra' to be loaded"
+    if backend == nothing
+        backend = default_polyhedra_backend(P, N)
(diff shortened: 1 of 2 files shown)


src/HPolyhedron.jl
```diff
@@ -454,6 +454,29 @@ function linear_map(M::AbstractMatrix{N},
     return PT(constraints)
 end
 
+# specific an_element function for rational inputs
+function an_element(P::HPolytope{N};
+                    backend=nothing) where {N<:Rational{Int}}
+    # use the floating point method, then convert to rational and check membership
+    p = invoke(an_element, Tuple{HPolytope}, P)
```
schillic: I am quite convinced that this would be an infinite loop.

mforets: Haha i tried it! `Tuple{HPolytope}` uses the `Float64` constructor.

schillic: Okay, then I do not like that exploit :smile:

schillic:
Even if we decide to not use this implementation, we could offer `chebyshev_center` as a new method.

schillic:
Superseded by #1697.

Issue #1113, opened 2019-02-12
Add AbstractAffineMap interface


schillic:
Subtypes: `AffineMap` (#1111), `LinearMap`, `ExponentialMap`, `ResetMap` (#1112), `Translation` (#882); (`ExponentialProjectionMap` ?).

To make use of the interface, all subtypes should provide a method to obtain the `A` matrix and the `b` vector. Then the functionality for affine maps can be written generically and automatically be used by all subtypes (as usual).

A useful function is `apply(M::AbstractAffineMap, x::AbstractVector)` (and the `!` version).

mforets:
> A useful function is apply(M::AbstractAffineMap, x::AbstractVector) (and the ! version).

Ok, i opened a new issue, https://github.com/JuliaReach/LazySets.jl/issues/1997

PR #1114, opened 2019-02-12
#1112 - Add lazy Resetmap operation

### LINKED ISSUE #1112: Add lazy Resetmap operation
A reset map wraps a list of reset variables.
A reset map is a special case of an affine map (#1111) where the linear map is the identity in all non-reset dimensions and zero in all reset dimensions, and the reset is then applied as a translation to this linear map.

### Example

```julia
X = BallInf([2.0, 2.0, 2.0], 1.0)
rm = [x1 => 4.0, x3 => 0.0]
M = ResetMap(rm, X)
```
Here `M` modifies the set `X` such that `x1` is reset to 1 and `x3` is reset to 0, while `x2` is kept as it is.
The affine map `A x + b` would be
```
[0.0 0.0 0.0]     [4.0]
[0.0 1.0 0.0] x + [0.0]
[0.0 0.0 0.0]     [0.0]
```

(Note that, in the vector, the two zero entries cannot be distinguished. Hence we need the input `rm` as a dictionary and not as a vector.)

The resulting set is equivalent to the set `Hyperrectangle([4.0, 2.0, 0.0], [0.0, 1.0, 0.0])`. (In this case, this even holds if `X` is a `Ball2`.)

A lazy implementation has the benefit that the support vector can be computed in the original set and the reset is applied only afterward. For instance, for direction `[1.0, 1.0, 1.0]`, the support vector of `X` is `[1.0, 1.0, 1.0]` and the support vector of `M` is `[4.0, 3.0, 0.0]`.

### Non-lazy extension (future work)

For later, the non-lazy implementation will be easy to do for `AbstractHyperrectangle`s.
The affine map can be taken as a fallback implementation, which should be optimized for `CartesianProductArray`s where we only have to do the affine maps in low dimensions, or even replace the set completely by a `Singleton` if all sets in a block are reset.

schillic:
Closes #1112.

i have found an unrelated problem with the support function for unbounded sets:
```julia
julia> dot([1., 0.], [Inf, -Inf])
NaN
```
This was not what I would expect, but the reason for this is:
```julia
julia> Inf * 0.
NaN
```
So I defined a new dot product. I am not too happy with the current implementation, so I created #1115 as a follow-up. We can outsource these changes (but I need them for the tests to work, or modify the tests).


mforets:
I agree that it is strange to define a new dot product -- these rules follow some standards and are there for a reason. I wonder what is the interest that we support such things as `dot([1., 0.], [Inf, -Inf])`? If don't expect that library reasons about the `Inf`, unless in simple cases such as taking `max(a, b)` or `min(a, b)` where `a` and ` b` are results of taking support functions.

src/ResetMap.jl
```diff
@@ -0,0 +1,169 @@
...
+```jldoctest resetmap
+julia> X = BallInf([2.0, 2.0, 2.0], 1.0);
+
+julia> r = Dict(1 => 4.0, 3 => 0.0);
+
+julia> rm = ResetMap(X, r);
+
+```
+
+Here `rm` modifies the set `X` such that `x1` is reset to 1 and `x3` is reset to
```
mforets:
```suggestion
Here `rm` modifies the set `X` such that `x1` is reset to 4 and `x3` is reset to
```

src/ResetMap.jl
```diff
@@ -0,0 +1,169 @@
+import Base: isempty
+
+export ResetMap,
+       apply,
+       apply!
+
+"""
+    ResetMap{N<:Real} <: LazySet{N}
+
+Type that represents a lazy reset map.
```
mforets: What is the role of the wrapped type `X`? In the example it is not used. Compare to `MathematicalSystems.ResetMap`.

mforets: OK, perhaps write explicitly: ``ResetMap(X) = \\{x : Ax + b, x \in X\}}`` and add an example of support vector / function evaluation below.

src/ResetMap.jl
```diff
@@ -0,0 +1,169 @@
...
+
+The same (but see the Notes below) vector `x` but after `rm` was applied.
+
+### Notes
+
+The vector `x` is modified in-place if it has type `Vector` or `SparseVector`.
+Otherwise, we first create a new `Vector` from it.
+"""
+function apply!(rm::ResetMap{N}, x::AbstractVector{N}) where {N<:Real}
+    return substitute!(rm.resets, x)
```
mforets: I think that this function should check for set membership `x \in rm.X`.

mforets: Also, why call it `apply` / `apply!`? In this library we would call this the concrete reset map, `reset_map(...)`. It may turn out to be interesting to add an `apply` function of a lazy set, why not, but that's another discussion.

src/ResetMap.jl
```diff
@@ -0,0 +1,169 @@
...
+- `d`  -- direction
+- `rm` -- reset map
+
+### Output
+
+The support vector in the given direction.
+If the direction has norm zero, the result depends on the wrapped set.
+"""
+function σ(d::AbstractVector{N}, rm::ResetMap{N}) where {N<:Real}
+    return apply!(rm, σ(d, rm.X))
```
mforets: Is this correct? This formula coded is `A * σ(d, X) + b`. But if i start from the set, `A*X + b`, then: `σ(d, A*X + b) = Aσ(A^T*d) + b`, see [Prop. 2.2 here](https://juliareach.github.io/LazySets.jl/latest/man/polyhedral_approximations.html#Support-Vector-1).

schillic: The outer `A` just makes some entries 0, namely those that are reset in `b`. The inner `A^T = A`  makes the same entries in `d` 0. I thought that the 0-entries do not hurt, but now I am not sure anymore.
My new formula derived from yours would be `apply!(rm, σ(d*, rm.X))`, where `d*` has all entries reset to zero that are reset in `rm`.

schillic: :100: Your hunch was right: If the support vector of `X` is unique, it should work, but if there are choices, then the result can be wrong. Consider a `Ball1` in 2D, the direction `d = [1.0, 1.0]`, and the reset `2 => 0.0`. If `σ(d, X)` returns `[1.0, 0.0]`, then the result will be `[1.0, 0.0]` (correct), but if `σ(d, X)` returns `[0.0, 1.0]`, then the result will be `[0.0, 0.0]`.
Great that you found this. It would have been a painful debugging at some point.

schillic:
> I agree that it is strange to define a new dot product

Alright, I reverted the change. I still use it for the `ResetMap`, though.

src/ResetMap.jl
mforets: > This is not the semantics of `ResetMap`.

Hmm but why? Otherwise, should we clarify in the docstring that it is not checked if `x` belongs to `X`?

schillic: It is not a `ConstrainedResetMap` as in `MathematicalSystems`. The represented set is
`A x + b, x ∈ X`.

src/ResetMap.jl
mforets: Saw the new function, looks good 👍

src/ResetMap.jl
schillic: Okay, I see now the confusion...

src/helper_functions.jl
```diff
@@ -1,6 +1,30 @@
...
+Dot product with preference for zero value in the presence of infinity values.
+
+### Input
+
+- `x` -- first vector
+- `y` -- second vector
+
+### Output
+
+The dot product of `x` and `y`, but with the rule that `0 * Inf == 0`.
```
mforets: Perhaps add a short note about where is this function practically used, to not scare mathematicians ;p  (see some arguments about why this operation should be undefined [here](https://math.stackexchange.com/questions/28940/why-is-infty-cdot-0-not-clearly-equal-to-0))

mforets:
> Alright, I reverted the change. I still use it for the `ResetMap`, though.

Saw it, looks good! 👍 

> The idea was for [a new general affine-map interface](https://github.com/JuliaReach/LazySets.jl/issues/1113#issue-409447984) to define this method. Note that it is only applied to a vector, not to a set. I do not really need this method, it is just for convenience. I can remove it, but why?

I had a thought on the design, we could compose the `LazySets.ResetMap` with the corresponding `MathematicalSystem.ResetMap`, thus leveraging on MathematicalSystems for the methods and traits. I'll propose this in a new issue, which is not particular to the `ResetMap` but applies to any of `<:AbstractMap`.

src/ResetMap.jl
```diff
@@ -0,0 +1,340 @@
...
+### Output
+
+The (sparse) vector for the affine map ``A x + b, x ∈ X`` represented by the
+reset map.
+The vector contains the reset value for all reset dimensions, and is zero for
+all other dimensions.
+"""
+function get_b(rm::ResetMap{N}) where {N<:Real}
+    n = dim(rm)
+    b = sparsevec(N[], N[], n)
```
mforets:
```suggestion
    b = sparsevec(Int[], N[], n)
```

test/unit_ResetMap.jl
```diff
@@ -0,0 +1,45 @@
...
+    # an_element function
+    an_element(rm)
+    elem = an_element(ResetMap(s, r_1))
+    @test elem == N[1, 0] && element(s) == N[0, 0]  # does not modify s
+
+    # isempty
+    @test !isempty(rm)
+
+    # getter for affine map
+    @test get_A(R) == sparse([2], [2], N[1], 3, 3)
```
mforets:
```suggestion
    @test get_A(rm) == sparse([2], [2], N[1], 3, 3)
```

test/unit_ResetMap.jl
```diff
@@ -0,0 +1,45 @@
...
+    @test get_b(R) == sparsevec([1], N[4], 3)
```
mforets:
```suggestion
    @test get_b(rm) == sparsevec([1], N[4], 3)
```

src/ResetMap.jl
```diff
@@ -0,0 +1,287 @@
...
+`Hyperrectangle([4.0, 2.0, 0.0], [0.0, 1.0, 0.0])`, i.e., an axis-aligned line
+segment embedded in 3D.
+
+The corresponding affine map ``A x + b`` would be:
+
+```math
+    \begin{pmatrix} 0 & 0 & 0 \\ 0 & 1 & 0 \\ 0 & 0 & 0 \end{pmatrix} x +
+    \begin{pmatrix} 4 & 0 & 0 \end{pmatrix}
+```
+
```
mforets: "Use the function `get_A` (resp. `get_b`) to create the matrix `A` (resp. vector `b`) corresponding to a given reset map."

src/ResetMap.jl
```diff
@@ -0,0 +1,287 @@
...
+### Output
+
+The (sparse) vector for the affine map ``A x + b, x ∈ X`` represented by the
+reset map.
+The vector contains the reset value for all reset dimensions, and is zero for
+all other dimensions.
+"""
+function get_b(rm::ResetMap{N}) where {N<:Real}
+    n = dim(rm)
+    b = sparsevec(Int[], N[], n)
```
mforets: Minimalist version:

```julia
function get_b(rm::ResetMap{N}) where {N<:Real}
    return sparsevec(pairs(rm.resets), dim(rm))
end
```

schillic: Not sure if that always works. The order in a `sparsevec` could matter.

mforets: I think that it works because `pairs` returns `key_i => val_i`, even if it is not ordered:

```julia
julia> sparsevec([1,2], [2.0, 4.0], 3)
3-element SparseVector{Float64,Int64} with 2 stored entries:
  [1]  =  2.0
  [2]  =  4.0

julia> sparsevec([2, 1], [4.0, 2.0], 3)
3-element SparseVector{Float64,Int64} with 2 stored entries:
  [1]  =  2.0
  [2]  =  4.0
```
That said, i'm not asking to change the implementation; it was a comment after i found that it can be put in one-line using `pairs` :)

schillic: Cool, I was not sure if the indices for `sparsevec` can be unsorted. In that case your version could even be more efficient, so I chose that.

src/ResetMap.jl
```diff
@@ -0,0 +1,290 @@
...
+
+### Algorithm
+
+We fall back to `constraints_list` of a `LinearMap`, which results in
+zero-constraints for the dimensions that are reset.
+Then, for these dimensions, we modify the constraints by pushing them to their
+new reset value.
+"""
+function constraints_list(rm::ResetMap{N}) where {N<:Real}
+    function find_unique_nonzero_entry(vector::AbstractVector{N})
```
mforets: Say what this function does:

```julia
# if `vector` has exactly one non-zero entry, return its index; in all other cases return 0 
function find_unique_nonzero_entry(vector::AbstractVector{N})
...
end
```
That said, note that it can be written as:

```julia
    return count(!iszero, vector) != 1 ? findfirst(!iszero, vector) : 0
```
But this function is worse than yours because it needs two passes on `vector`  😄

src/ResetMap.jl
```diff
@@ -0,0 +1,285 @@
...
+                else
+                    # first non-zero entry so far
+                    res = i
+                end
+            end
+        end
+        return res
+    end
+
+    constraints = copy(constraints_list(LinearMap(get_A(rm), rm.X)))
```
mforets: The outer `copy` is not needed if we agree on https://github.com/JuliaReach/LazySets.jl/issues/1118. The line just below this one is a `println` that can be removed.

src/ResetMap.jl
schillic: `pairs` is not defined in v0.6...I'll revert...

mforets: Ouch, we still live in the past...

PR #1121, opened 2019-02-13
#1100 - More functionality for Universe

### LINKED ISSUE #1100: More functionality for Universe
In #1098 we added the `Universe`. Useful operations in `Reachability` will be the `is_subset` test (easy to do with the `Universe` on the left and on the right), the (concrete) `intersection` (lazy `Intersection` already defines `Universe` as its neutral element) (again easy to do), and the `isdisjoint` test.

(I also forgot to update the "set operations" page in the manual, the plot tests, and the ambiguity tests. This could be done here as well.)

schillic:
Closes #1100.

* [x] add to manual ("set operations" section)
* [x] fix method ambiguities


schillic:
I will update the manual later after rebasing with other open PRs modifying that file. So from my side this PR is ready for review.

src/LazySet.jl
```diff
@@ -341,3 +342,50 @@ list of linear constraints.
...
+* If `witness` option is deactivated: `true` iff ``X`` is universal
+* If `witness` option is activated:
+  * `(true, [])` iff ``X`` is universal
+  * `(false, v)` iff ``X`` is not universal and ``v ∉ X``
+
+### Notes
+
+This is a naive fallback implementation.
+"""
+function isuniversal(X::LazySet{N}, witness::Bool=false
```
mforets: I think that it's better to use dispatch for `isuniversal` and do the appropriate tests on each type (or abstract type).

schillic: Yes, I wanted to create a follow-up issue for this, but forgot about it. I added #1125.

PR #1139, opened 2019-02-21
#1123 - Concrete intersection between half-space and polytopic reset map

### LINKED ISSUE #1123: Concrete intersection between half-space and polytopic reset map
Wanted: `intersection(::HalfSpace{N}, ::LazySets.ResetMap{N, <:AbstractPolytope})`

This would work:
```julia
function intersection(hs::HalfSpace{N}, rm::ResetMap{N, <:AbstractPolytope{N}}) where {N<:Real}
    return HPolytope([rs; constraints_list(rm)])
end
```

schillic:
Closes #1123.

### DIFF
--- a/src/concrete_intersection.jl
+++ b/src/concrete_intersection.jl
@@ -550,3 +550,43 @@ end
 function intersection(S::AbstractSingleton{N}, U::Universe{N}) where {N<:Real}
     return S
 end
+
+"""
+    intersection(P::AbstractPolyhedron{N}, rm::ResetMap{N}) where {N<:Real}
+
+Return the intersection of a polyhedron and a polyhedral reset map.
+
+### Input
+
+- `P`  -- polyhedron
+- `rm` -- polyhedral reset map
+
+### Output
+
+A polyhedron.
+
+### Notes
+
+We assume that `rm` is polyhedral, i.e., has a `constraints_list` method
+defined.
+"""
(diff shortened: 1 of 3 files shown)


src/concrete_intersection.jl
```diff
@@ -550,3 +550,31 @@ end
...
+
+An `HPolytope`.
+
+### Notes
+
+We assume that `rm` is polytopic, i.e., is bounded and has a `constraints_list`
+method defined.
+"""
+function intersection(P::AbstractPolyhedron{N}, rm::ResetMap{N}) where {N<:Real}
+    return HPolytope([constraints_list(P); constraints_list(rm)])
```
mforets: Have you considered removing redundant constraints? This is done by default for polytope intersections [here](https://github.com/JuliaReach/LazySets.jl/blob/master/src/concrete_intersection.jl#L340) and i would suggest to use it here as well.

mforets: Perhaps we can just invoke that function but doing `HPolytope(constraints_list(rm))`?

schillic: Yes, I considered it, but decided against it.
I changed the code (it was actually unsound) and now it uses the method you linked to.

mforets: Great, looks good to me. The argument behind removing redundant constraints by deafult is that is simplifies a lot of cases where the intersection is actually empty. And, well if the intersection is not empty you get a simpler polyhedron. But we can add a flag to turn this on or off, currently there is no such flag.

PR #1148, opened 2019-02-23
WIP - Revise msa

mforets:
This branch makes the support function of msa a bit faster:

```julia
julia> msa = MinkowskiSumArray([rand(Hyperrectangle) for _ in 1:10]);
julia>  @btime ρ(rand(2), $msa)
  638.899 ns (11 allocations: 1.03 KiB)
10.184097859097406
```

vs (master)

```julia
julia> @btime ρ(rand(2), $msa)
  722.187 ns (14 allocations: 1.23 KiB)
10.731613830299803
```

---

There is also a change in the output annotation of sigma_helper : i don't see why the output should always be a `Vector`? I made it similar to `d`.

### DIFF
--- a/src/MinkowskiSum.jl
+++ b/src/MinkowskiSum.jl
@@ -640,7 +640,7 @@ end
 
 @inline function σ_helper(d::AbstractVector{N},
                           array::AbstractVector{<:LazySet}) where {N<:Real}
-    svec = zeros(N, length(d))
+    svec = similar(d)
     @inbounds for sj in array
         svec += σ(d, sj)
     end
--- a/src/MinkowskiSum.jl
+++ b/src/MinkowskiSum.jl
@@ -298,7 +298,11 @@ The support function of the Minkowski sum of sets is the sum of the support
 functions of each set. 
 """
 function ρ(d::AbstractVector{N}, msa::MinkowskiSumArray{N}) where {N<:Real}
-    return sum([ρ(d, Xi) for Xi in msa.array])
+    res = zero(N)
+    @inbounds for Xi in msa.array
+        res = res + ρ(d, Xi)::N
+    end
+    return res
 end
 
 """
@@ -635,10 +639,9 @@ end
 # ================
 
 @inline function σ_helper(d::AbstractVector{N},
-                          array::AbstractVector{<:LazySet}
-                         )::Vector{N} where {N<:Real}
+                          array::AbstractVector{<:LazySet}) where {N<:Real}
     svec = zeros(N, length(d))
-    for sj in array
+    @inbounds for sj in array
         svec += σ(d, sj)
     end
     return svec


src/MinkowskiSum.jl
```diff
@@ -298,7 +298,11 @@ The support function of the Minkowski sum of sets is the sum of the support
 functions of each set. 
 """
 function ρ(d::AbstractVector{N}, msa::MinkowskiSumArray{N}) where {N<:Real}
-    return sum([ρ(d, Xi) for Xi in msa.array])
+    res = zero(N)
+    @inbounds for Xi in msa.array
+        res = res + ρ(d, Xi)::N
```
schillic: I do not mind, but since when do you add type annotations? Note that this might actually be slower, as it has to check for the type.

src/MinkowskiSum.jl
```diff
@@ -635,10 +639,9 @@ end
 # ================
 
 @inline function σ_helper(d::AbstractVector{N},
-                          array::AbstractVector{<:LazySet}
-                         )::Vector{N} where {N<:Real}
-    svec = zeros(N, length(d))
-    for sj in array
+                          array::AbstractVector{<:LazySet}) where {N<:Real}
+    svec = similar(d)
```
schillic: You have to start from a zero vector. So I would just keep it as before.

mforets: You are right, `similar` is wrong here. There is no generic `zeros` for an abstract vector. I wanted to avoid converting to a dense vector, because `d` is of abstract type. I'll try with [mapreduce](https://docs.julialang.org/en/v1/base/collections/#Base.mapreduce-Tuple{Any,Any,Any}), like `mapreduce(sj -> σ(d, sj), +, array)`.

schillic: I think you are overengineering this :smile:

mforets: `reach_blocks!` generate a big MSA "horizontally", e.g. [reach_blocks.jl#L186](https://github.com/JuliaReach/Reachability.jl/blob/master/src/ReachSets/ContinuousPost/BFFPSV18/reach_blocks.jl#L186).

i found that the profiler was giving type instabilities there and began to investigate and put up the results so far in this branch. no need to merge it (i changed the title to WIP to make this clear).

mforets:
> i found that the profiler was giving type instabilities there and began to investigate and put up the results so far in this branch.

The instability probably comes from [this line](https://github.com/JuliaReach/Reachability.jl/blob/master/src/ReachSets/ContinuousPost/BFFPSV18/reach_blocks.jl#L188) since `arr` does not have a concrete type, from [this line](https://github.com/JuliaReach/Reachability.jl/blob/master/src/ReachSets/ContinuousPost/BFFPSV18/reach_blocks.jl#L175). So i'll close this issue.

Issue #965, opened 2018-12-31
linear_map(M{R}, S{T})  error


tomerarnon:
Defining the promotion:
```julia
LazySets.linear_map(M::AbstractMatrix{R}, P::LazySet{T}) where {R,T} = linear_map(T.(M), P)
```
Fixes the following error:
```julia
julia> H = HPolytope(rand(5, 2), rand(5))
HPolytope{Float64}(HalfSpace{Float64}[HalfSpace{Float64}([0.228874, 0.484057], 0.502584), HalfSpace{Float64}([0.424035, 0.628926], 0.488933), HalfSpace{Float64}([0.81033, 0.409093], 0.527271), HalfSpace{Float64}([0.915492, 0.310442], 0.173226), HalfSpace{Float64}([0.126785, 0.134684], 0.941033)])

julia> M = Matrix(I, 2, 2);

julia> linear_map(M, H)
ERROR: MethodError: no method matching linear_map(::Array{Bool}, ::HPolytope{Float64})
```

I find that this sort of "autopromotion" is very convenient. It is likely missing from other methods as well, but this is the one I just encountered 😄

schillic:
We have these "same type" assumptions in the library quite often.
Autopromotions are useful on the one hand, but on the other hand they may hide errors in the program logic (in this case it is fine, though).
From the code I did not see that `M` is a matrix of `Bool`s.
```julia
julia> M = Matrix(I, 2, 2)
2×2 Array{Bool,2}:
  true  false
 false   true
```
I would rather create the "correct" matrix in the first place. This also avoids the performance penalty that you get if you first create a `Bool` matrix and then convert it to `Float64` again.
```julia
M = Matrix{Float64}(I, 2, 2)
2×2 Array{Float64,2}:
 1.0  0.0
 0.0  1.0
```
It is probably a matter of taste whether the programmer or the library should take care of those things. I personally prefer the former paradigm, but I guess the Julian way is indeed to implement the promotion you described. So I am fine with adding the method.
@mforets: Any opinion?

mforets:
I also don't expect the library to make some conversions automatically for me, especially when conversions could have different meanings.
  
But automatic promotions that involve numbers such as:  

```julia
julia> rand(Float64, 2, 2) * rand(Float32, 2, 1)
2×1 Array{Float64,2}:
 1.220159659966665
 1.389455069686564
```

[...]

PR #1156, opened 2019-02-24
Use promotion for linear map of polyhedra

mforets:
This covers a particular case from https://github.com/JuliaReach/LazySets.jl/issues/965.

### DIFF
--- a/src/AbstractPolyhedron_functions.jl
+++ b/src/AbstractPolyhedron_functions.jl
@@ -321,16 +321,12 @@ end
 # handle different numeric types
 function linear_map(M::AbstractMatrix{NM},
                     P::AbstractPolyhedron{NP};
-                    check_invertibility::Bool=true,
-                    cond_tol::Number=DEFAULT_COND_TOL,
-                    use_inv::Bool=!issparse(M)
-                   ) where {NM<:Real, NP<:Real}
+                    kwargs...) where {NM<:Real, NP<:Real}
     N = promote_type(NM, NP)
     if N != NP
         error("conversion between numeric types of polyhedra not implemented yet (see #1181)")
     else
-        return linear_map(N.(M), P, check_invertibility=check_invertibility,
-                cond_tol=cond_tol, use_inv=use_inv)
+        return linear_map(N.(M), P; kwargs...)
     end
 end
 
--- a/src/AbstractPolyhedron_functions.jl
+++ b/src/AbstractPolyhedron_functions.jl
@@ -318,6 +318,22 @@ function linear_map(M::AbstractMatrix{N},
     end
 end
 
+# handle different numeric types
+function linear_map(M::AbstractMatrix{NM},
+                    P::AbstractPolyhedron{NP};
+                    check_invertibility::Bool=true,
+                    cond_tol::Number=DEFAULT_COND_TOL,
+                    use_inv::Bool=!issparse(M)
+                   ) where {NM<:Real, NP<:Real}
+    N = promote_type(NM, NP)
+    if N != NP
+        error("conversion between numeric types of polyhedra not implemented yet (see #1181)")
+    else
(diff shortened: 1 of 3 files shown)


test/unit_Polytope.jl
```diff
@@ -135,6 +135,11 @@ for N in [Float64, Rational{Int}, Float32]
     L4 = linear_map(M, P, check_invertibility=false)
     @test L4 isa VPolygon
 
+    # linear map for mixed types
+    M = [2 1; 0 1] # Int's
+    LM = linear_map(M, P)
+    @test LM isa HPolytope
```
schillic: I would also test for the numeric type (wasn't that the whole point? :wink:)!

src/HPolytope.jl
```diff
@@ -236,3 +238,10 @@ function vertices_list(P::HPolytope{N};
     prunefunc(P)
     return collect(points(P))
 end
+
+function linear_map(M::Matrix{MN}, P::HPolytope{SN}; kwargs...) where {MN<:Real, SN<:Real}
```
schillic: Do we always require a `Matrix` (in constrast to `AbstractMatrix`)?

mforets:
> IMO adding features for single set types is not consistent, but this seems to be the most important type

Certainly, but i don't know how to solve this problem in general :(

schillic:
What is "this problem"?

test/unit_Polytope.jl
mforets: The point was to make `linear_map` work; in master this gives an error. Yes, i should add a test `LM isa HPolytope{N}` because `Int` is promoted to the other `N`'s in the outer loop.

src/HPolytope.jl
mforets: I don't know if it's possible to make the numeric type conversion for *any* `AbstractMatrix` work, but i can try: if `promote_rule` has been defined by the type of `M` perhaps this just works..

mforets:
To use Julia's [conversion and promotion](https://docs.julialang.org/en/v1.1/manual/conversion-and-promotion/#conversion-and-promotion-1) system with LazySets. This PR only adds the `HPolytope` case because it already exposes some convert methods.

src/HPolytope.jl
mforets: Broadcasting on the numeric type works without having to assume a specific matrix type, so ill do it that way...

src/AbstractPolyhedron_functions.jl
```diff
@@ -318,6 +318,22 @@ function linear_map(M::AbstractMatrix{N},
...
 
+# handle different numeric types
+function linear_map(M::AbstractMatrix{NM},
+                    P::AbstractPolyhedron{NP};
+                    check_invertibility::Bool=true,
+                    cond_tol::Number=DEFAULT_COND_TOL,
+                    use_inv::Bool=!issparse(M)
+                   ) where {NM<:Real, NP<:Real}
+    N = promote_type(NM, NP)
+    if N != NP
```
schillic: I first read "`P != NP`" :smile:

src/AbstractPolyhedron_functions.jl
```diff
@@ -318,6 +318,22 @@ function linear_map(M::AbstractMatrix{N},
     end
 end
 
+# handle different numeric types
+function linear_map(M::AbstractMatrix{NM},
+                    P::AbstractPolyhedron{NP};
+                    check_invertibility::Bool=true,
+                    cond_tol::Number=DEFAULT_COND_TOL,
+                    use_inv::Bool=!issparse(M)
```
schillic: Can we use `kwargs...` instead? I prefer spelling out these arguments and their default values in one place only. This makes changing things later much easier.

mforets: Agreed. I changed to use `kwargs...`

PR #1157, opened 2019-02-25
#1140 - Add conversion to interval methods

### LINKED ISSUE #1140: Convert hyperrectangle to interval
```julia
julia> H = rand(Hyperrectangle, dim=1)
Hyperrectangle{Float64}([0.433192], [0.0685728])

julia> convert(Interval, H)
ERROR: MethodError: Cannot `convert` an object of type Hyperrectangle{Float64} to an object of type Interval
Closest candidates are:
  convert(::Type{#s16} where #s16<:Union{Number, T}, ::MultivariatePolynomials.AbstractPolynomialLike{T}) where T at /Users/forets/.julia/dev/MultivariatePolynomials/src/conversion.jl
:10
  convert(::Type{T}, ::T) where T at essentials.jl:154
Stacktrace:
 [1] top-level scope at none:0
```

mforets:
Closes #1140.

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -343,6 +343,51 @@ function convert(::Type{Hyperrectangle}, x::Interval)
 end
 
 """
+    convert(::Type{Interval}, H::AbstractHyperrectangle)
+
+Converts a hyperrectangular set to a unidimensional interval.
+
+### Input
+
+- `H`        -- hyperrectangular set 
+- `Interval` -- interval type, used for dispatch
+
+### Output
+
+An interval.
+
+### Examples
+
+```jldoctest convert_hyperrectangle_interval
+julia> convert(Interval, Hyperrectangle{Float64}([0.5], [0.5]))
+Interval{Float64,IntervalArithmetic.Interval{Float64}}([0, 1])
+```
+"""
+function convert(::Type{Interval}, H::AbstractHyperrectangle)
+    @assert dim(H) == 1 "can only convert a one-dimensional $(typeof(H)) to `Interval`"
+    return Interval([low(H); high(H)])
+end
+
+"""
+    convert(::Type{Interval}, S::LazySet{N}) where {N<:Real}
+
+Converts a convex set to a unidimensional interval.
+
+### Input
+
+- `S`        -- set 
+- `Interval` -- interval type, used for dispatch
+
+### Output
+
+An interval.
+"""
+function convert(::Type{Interval}, S::LazySet{N}) where {N<:Real}
+    @assert dim(S) == 1 "can only convert a one-dimensional $(typeof(S)) to `Interval`"
+    return Interval(-ρ(N[-1], S), ρ(N[1], S))
+end
+
+"""
     convert(::Type{HPolytope}, H::AbstractHyperrectangle)
 
 Converts a hyperrectangular set to a polytope in constraint representation.
--- a/test/unit_Interval.jl
+++ b/test/unit_Interval.jl
@@ -97,7 +97,7 @@ for N in [Float64, Float32, Rational{Int}]
     A = convert(Interval, H)
     @test A isa Interval && low(A) == [N(-1/2)] && high(A) == [N(1/2)]
     # conversion from a lazyset to an interval
-    M = reshape([N(2)], 1, 1)
+    M = hcat(N[2])
     B = convert(Interval, M*H)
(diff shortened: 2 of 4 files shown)


test/unit_Interval.jl
```diff
@@ -91,4 +91,13 @@ for N in [Float64, Float32, Rational{Int}]
     # check empty intersection
     E = intersection(A, Interval(N(0), N(1)))
     @test isempty(E)
+
+    # conversion from a hyperrectangular set to an interval
+    H = Hyperrectangle(N[0], N[1/2])
+    A = convert(Interval, H)
+    @test A isa Interval && low(A) == [N(-1/2)] && high(A) == [N(1/2)]
+    # conversion from a lazyset to an interval
+    M = reshape([N(2)], 1, 1)
```
schillic: Shorter: `hcat(N[2])`

src/convert.jl
```diff
@@ -342,6 +342,51 @@ function convert(::Type{Hyperrectangle}, x::Interval)
...
+- `S`        -- set 
+- `Interval` -- interval type, used for dispatch
+
+### Output
+
+An interval.
+"""
+function convert(::Type{Interval}, S::LazySet{N}) where {N<:Real}
+    @assert dim(S) == 1 "can only convert a one-dimensional $(typeof(S)) to `Interval`"
+    return Interval(-ρ(N[-1], S), ρ(N[1], S))
```
schillic: You could reuse [this method](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Approximations/overapproximate.jl#L208) here.

mforets: That's correct, but `overapproximate` is allowed to produce overapproximations and `convert` is exact so i prefer to not use that method here. The other way round (let overapproximate use convert) is better imo.

test/unit_Interval.jl
mforets: Thanks!

src/convert.jl
schillic: Then do you want to change this here?

mforets: The methods in convert run a bit faster than the support vector used in `overapproximate` if hyperrectangle are used:

```julia
julia> using LazySets, LazySets.Approximations, BenchmarkTools

julia> N = Float64; H = Hyperrectangle(N[0], N[1/2])
Hyperrectangle{Float64}([0.0], [0.5])

julia> @btime convert(Interval, $H)
  129.107 ns (3 allocations: 288 bytes)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([-0.5, 0.5])

julia> @btime overapproximate($H, Interval)
  152.951 ns (4 allocations: 384 bytes)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([-0.5, 0.5])
```

```julia
julia> M = hcat(N[2])
1×1 Array{Float64,2}:
 2.0

julia> S=M*H;

julia> @btime convert(Interval, $S)
  364.971 ns (6 allocations: 576 bytes)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([-1, 1])

julia> @btime overapproximate($S, Interval)
  520.937 ns (8 allocations: 768 bytes)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([-1, 1])
```

schillic: This is probably because you use the support function and the other method uses the support vector :+1:

mforets: Yes. And in the first case (overapproximate a Hyperrectangle), doing `Interval([low(H); high(H)])` is a bit faster than `Interval(-ρ(N[-1], S), ρ(N[1], S))`.

schillic: I assumed that `convert(Interval, Interval(1., 2.))` would return the same interval, but it does not. `convert(::Type{T}, x::T) where {T} = x` should do the trick.
If we define that method, Julia can inline the identity:
```julia
julia> @btime convert(Interval, $A);
  101.935 ns (3 allocations: 288 bytes)

julia> @btime overapproximate($A, Interval);
  102.302 ns (3 allocations: 288 bytes)

julia> convert(::Type{Interval}, i::Interval) = i

julia> @btime convert(Interval, $A)
  0.023 ns (0 allocations: 0 bytes)

julia> @btime overapproximate($A, Interval)
  0.023 ns (0 allocations: 0 bytes)
```
I think we should (in a new approach) define the identity for `convert`.

PR #1164, opened 2019-02-26
#1163 - Check for square matrix in isinvertible

### LINKED ISSUE #1163: "Error: matrix is not square" in inv(M) in AbstractPolyhedron linear_map
The problem:
```julia
M = rand(5,4)
LazySets.isinvertible(M) #true
inv(M) #error
```

Should `linear_map` with `check_invertibility = true` use `pinv` instead of `inv` to deal with non square matrices?

schillic:
Closes #1163.

### DIFF
--- a/test/unit_util.jl
+++ b/test/unit_util.jl
@@ -21,9 +21,19 @@ for _dummy_ in 1:1 # avoid global variable warnings
     @test !LazySets.issquare(hcat([2 3]))
     @test !LazySets.issquare(sparse([1], [1], [1], 3, 4))
 
-    # invertible matrix
+    # invertible matrix: (1) invertible, (2) non-invertible, (3) non-square
+    # dense matrix
     @test LazySets.isinvertible([2 3; 1 2])
     @test !LazySets.isinvertible([2 3; 0 0])
+    @test !LazySets.isinvertible(hcat([2 3]))
+    # sparse matrix
+    @test LazySets.isinvertible(sparse([2 3; 1 2]))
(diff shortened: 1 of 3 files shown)


mforets:
I suggest to change the default value of `use_inv`:

```julia
use_inv::Bool=!issparse(M) && issquare(M)
```

I think that we should let `\` be used even when the matrix is rectangular (`use_inv=false`). Moreover, we can add `use_pinv` as an option too.

schillic:
> I suggest to change the default value of `use_inv`

Okay, can you do the change?

mforets:
Yes. But on a second thought, this PR is an independent fix to `isinvertible` and we should anyways make this change 👍  

I'll change the default of `use_inv` in another PR.

test/unit_util.jl
```diff
@@ -15,9 +15,25 @@ for _dummy_ in 1:1 # avoid global variable warnings
...
+    # dense matrix
     @test LazySets.isinvertible([2 3; 1 2])
     @test !LazySets.isinvertible([2 3; 0 0])
+    @test !LazySets.isinvertible(hcat([2 3]))
+    # sparse matrix
+    @test LazySets.isinvertible(sparse([2 3; 1 2]))
+    @test !LazySets.isinvertible(sparse([2 3; 0 0]))
+    @test !LazySets.isinvertible(sparse(hcat([2 3])))
+    # diagonal matrix
+    @test LazySets.isinvertible(Diagonal([2 0; 0 2]))
```
mforets: (Diagonal matrices also accept vectors: `Diagonal([2, 2])`)

PR #1172, opened 2019-02-27
#1170 - Faster hyperrectangular overapproximation of the linear map of a hyperrectangular set

### LINKED ISSUE #1170: Faster hyperrectangular overapproximation of the linear map of a hyperrectangular set
Using only center and radius one can take the overapprox of the lazy linear map of a hyperrectangle:

```julia
function overapproximate(lm::LinearMap{N, <:AbstractHyperrectangle{N}}, ::Type{Hyperrectangle}) where {N}
    center_MX = lm.M * LazySets.center(lm.X)
    radius_MX = abs.(lm.M) * radius_hyperrectangle(lm.X)
    return Hyperrectangle(center_MX, radius_MX)
end
```

@changliuliu claimed [here](https://github.com/sisl/NeuralVerification.jl/issues/49) that this idea is more efficient than the fallback implementation using support functions, and i have checked that it is the case for different dimensions. Thanks a lot!

The benchmarks are in [this notebook HyperrectangleApproxOfLinearMap](https://github.com/mforets/escritoire/blob/master/sets/HyperrectangleApproxOfLinearMap.ipynb).

This issue proposes to add the method above to `overapproximate`, for any hyperrectangular set.

mforets:
Closes #1170.

---

For a comparison with the fallback implementation using support functions see [this notebook](https://github.com/mforets/escritoire/blob/master/sets/HyperrectangleApproxOfLinearMap.ipynb).

For a randomly generated `M*H` of dimension 10, the speedup is around 10x.

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -116,6 +116,37 @@ function overapproximate(S::CartesianProductArray{N, <:AbstractHyperrectangle{N}
 end
 
 """
+    overapproximate(lm::LinearMap{N, <:AbstractHyperrectangle{N}},
+                    ::Type{Hyperrectangle}) where {N}
+
+Return a tight overapproximation of the linear map of a hyperrectangular set
+using a hyperrectangle.
+
+### Input
+
+- `S`              -- linear map of a hyperrectangular set
+- `Hyperrectangle` -- type for dispatch
+
+### Output
+
+A hyperrectangle.
+
+### Algorithm
+
+If `c` and `r` denote the center and vector radius of a hyperrectangle `H`,
+a tight hyperrectangular overapproximation of `M * H` is obtained by transforming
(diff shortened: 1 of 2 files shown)


src/Approximations/overapproximate.jl
```diff
@@ -115,6 +115,37 @@ function overapproximate(S::CartesianProductArray{N, <:AbstractHyperrectangle{N}
...
+
+### Output
+
+A hyperrectangle.
+
+### Algorithm
+
+If `c` and `r` denote the center and vector radius of a hyperrectangle `H`,
+a tight hyperrectangular overapproximation of `M * H` is obtained by transforming
+`c ↦ M*c` and `r ↦ abs.(⋅) * c`, where `abs.(⋅)` denotes the element-wise absolute
```
```suggestion
`c ↦ M*c` and `r ↦ abs.(M) * c`, where `abs.(⋅)` denotes the element-wise absolute
```

PR #1173, opened 2019-02-27
#882 - Add lazy Translation

### LINKED ISSUE #882: Translation operation
Example: Given a polygon ``P``, I want to translate it by some distance ``d`` (a vector), written ``P + d`` with the semantics ``P ⊕ {d}``.
Proposal: Override the `+` function for a set and a vector and create a new (concrete) set of the same type where each number is translated accordingly (this can be done recursively for lazy operations). Alternatively we can let ``+`` be a new lazy operation ``Translation`` and have a concrete function ``translation``.

mforets:
Closes #882.

### DIFF
--- a/src/Translation.jl
+++ b/src/Translation.jl
@@ -336,13 +336,10 @@ function constraints_list(tr::Translation{N}, ::Val{true}) where {N<:Real}
 end
 
 function constraints_list(tr::Translation{N}) where {N<:Real}
-    if applicable(constraints_list, tr.X)
-        return constraints_list(tr, Val(true))
-    else
-        return constraints_list(tr, Val(false))
-    end
+    has_constraints = applicable(constraints_list, tr.X)
+    return constraints_list(tr, Val(has_constraints))
 end
 
 function constraints_list(tr::Translation{N}, ::Val{false}) where {N<:Real}
-    error("this function requires that the `constraints_list` method is appliable")
+    error("this function requires that the `constraints_list` method is applicable")
 end
--- a/src/Translation.jl
+++ b/src/Translation.jl
@@ -1,0 +1,338 @@
+import Base: isempty
+
+export Translation,
+       an_element,
+       constraints_list
+
+"""
+    Translation{N<:Real, VN<:AbstractVector{N}, S<:LazySet{N}} <: LazySet{N}
+
+Type that represents a lazy translation.
+
+The translation of set `X` along vector `v` is the map: 
+
+```math
+x ↦ x + v,\\qquad x ∈ X
+```
+
+A translation is a special case of an affine map ``A x + b, x ∈ X`` where the
+linear map ``A`` is the identity matrix and the translation vector ``b = v``.
+
+### Fields
+
+- `X` -- convex set
+- `v` -- vector that defines the translation
+
+### Example
+
+```jldoctest translation
+julia> X = BallInf([2.0, 2.0, 2.0], 1.0);
+
+julia> v = [1.0, 0.0, 0.0] # translation along dimension 1
+
+julia> tr = Translation(X, v);
+
+julia> typeof(tr)
+Translation{Float64,Array{Float64,1},BallInf{Float64}}
+
+julia> tr.X
+BallInf{Float64}([2.0, 2.0, 2.0], 1.0)
+
+julia> tr.v
+3-element Array{Float64,1}:
+ 1.0
+ 0.0
+ 0.0
+```
+The sum operator `+` is overloaded to create translations:
+
+```jldoctest translation
+julia> X + v == Translation(X, v)
+true
+```
+And so does the Minkowski sum operator, `⊕`:
+
+```jldoctest
+julia> X ⊕ v == Translation(X, v)
+true
+```
+
+The translation of a translation is performed immediately:
+
+```jldoctest translation
+julia> tr = (X+v)+v
+Translation{Float64,Array{Float64,1},BallInf{Float64}}(BallInf{Float64}([2.0, 2.0, 2.0], 1.0), [2.0, 0.0, 0.0])
+
+julia> tr.v
+3-element Array{Float64,1}:
+ 2.0
+ 0.0
+ 0.0
+```
+
+The dimension of a translation is obtained with the `dim` function:
+
+```jldoctest translation
+julia> dim(tr)
+3
+```
+
+For the support vector (resp. support function) along vector `d`, use `σ` and
+`ρ` respectively:
+
+```jldoctest translation
+julia> σ([1.0, 0.0, 0.0], tr)
+3-element Array{Float64,1}:
+ 5.0
+ 3.0
+ 3.0
+
+julia> ρ([1.0, 0.0, 0.0], tr)
+5.0
+```
+See the docstring of each of these functions for details.
+
+The `an_element` function is useful to obtain an element of a translation:
+
+```jldoctest translation
+julia> e = an_element(tr)
+3-element Array{Float64,1}:
+ 4.0
+ 2.0
+ 2.0
+ ```
+
+The lazy linear map of a translation is again a translation, since the following
+simplification rule applies: ``M * (X⊕v) = (M*X) ⊕ (M*v)``:
+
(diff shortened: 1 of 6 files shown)


test/unit_Translation.jl
```diff
@@ -0,0 +1,51 @@
+for N in [Float64, Rational{Int}, Float32]
+
+    # ==================================
+    # Constructor and interface methods
+    # ==================================
+ 
+    B = BallInf(N[2, 2, 2], N(1))
+    v = N[0, 0, 0]
```
schillic: Please use some other vector to actually test the correct behavior.

test/unit_Translation.jl
```diff
@@ -0,0 +1,51 @@
...
+    # support function
+    @test ρ(N[1, 1, 1], tr) == N(9)
+
+    # boundedness
+    @test isbounded(tr)
+    @test !isbounded(Translation(Universe{N}(3), v))
+    
+    if N ∈ [Float64]
+        @test !isbounded(Translation(HPolyhedron([HalfSpace(N[1, 1], N(1))]), N[0, 0]))
+        @test !isbounded(Translation(HPolyhedron([HalfSpace(N[1, 1], N(1))]), N[0, 0]))
```
schillic: Those lines are identical :confused:

src/Translation.jl
```diff
@@ -0,0 +1,338 @@
...
+```
+"""
+struct Translation{N<:Real, VN<:AbstractVector{N}, S<:LazySet{N}} <: LazySet{N}
+    X::S
+    v::VN
+
+    # default constructor with dimension check
+    function Translation(X::S, v::VN) where {N, VN<:AbstractVector{N}, S<:LazySet{N}}
+        @assert dim(X) == length(v) "cannot create a translation of a set of dimension $(dim(X)) " *
+                                    " along a vector of length $(length(v))"
```
schillic:
```suggestion
                                    "along a vector of length $(length(v))" 
```

src/Translation.jl
```diff
@@ -0,0 +1,338 @@
...
+
+@static if VERSION < v"0.7-"
+    @eval begin
+        # convenience constructor without type parameter
+        Translation(X::S, v::VN) where {N<:Real, VN<:AbstractVector{N}, S<:LazySet{N}} =
+            Translation{N, VN, S}(X, v)
+    end
+end
+
+# constructor from a Translatio:: perform the translation immediately
```
schillic:
```suggestion
# constructor from a Translation: perform the translation immediately
```

src/Translation.jl
```diff
@@ -0,0 +1,338 @@
...
+### Input
+
+- `X` -- convex set
+- `v` -- vector
+
+### Output
+
+The symbolic translation of ``X`` along vector ``v``.
+"""
++(X::LazySet, v::AbstractVector) = Translation(X, v)
```
schillic: I think you should/have to add `+` and `⊕` to the docs.

src/Translation.jl
```diff
@@ -0,0 +1,338 @@
...
+
+- `d`  -- direction
+- `tr` -- translation
+
+### Output
+
+The support function in the given direction.
+"""
+function ρ(d::AbstractVector{N}, tr::Translation{N}) where {N<:Real}
+    return ρ(d, Singleton(tr.v)) + ρ(d, tr.X)
```
schillic: While I like the readability, we could have inline the definition of `ρ(d, Singleton(tr.v))` to avoid the creation of the object.

src/Translation.jl
```diff
@@ -0,0 +1,338 @@
...
+- `tr` -- lazy translation of a polyhedron
+
+### Output
+
+The list of constraints of the translation.
+
+### Notes
+
+We assume that the set wrapped by the lazy translation `X` is a polyhedron, i.e.,
+it offers a method `constraints_list(⋅)`.
```
schillic: Hm, this sounds like you wanted to omit the set type constraint `<:AbstractPolyhedron{N}`. This could actually be helpful for our favorite application `Reachability` :+1:

src/Translation.jl
mforets: 👍  which is `dot(d, tr.v)`

src/Translation.jl
schillic: For the record, there are several problems with the docs (see [here](https://travis-ci.org/JuliaReach/LazySets.jl/jobs/499554056#L804)).

src/Translation.jl
mforets: How about a trait-based solution like:

```julia
function constraints_list(tr::Translation{N}, ::Val{true}) where {N<:Real}
    constraints_X = constraints_list(tr.X)
    constraints_TX = similar(constraints_X)
    @inbounds for (i, ci) in enumerate(constraints_X)
        constraints_TX[i] = HalfSpace(ci.a, ci.b + dot(ci.a, tr.v))
    end
    return constraints_TX
end

function constraints_list(tr::Translation{N}) where {N<:Real}
    if applicable(constraints_list, tr.X)
        return constraints_list(tr, Val(true))
    else
        return constraints_list(tr, Val(false))
    end
end

function constraints_list(tr::Translation{N}, ::Val{false}) where {N<:Real}
    error("this function requires that the `constraints_list` method is appliable")
end
```

Example:

```julia
julia> constraints_list(rand(BallInf)+rand(2))
4-element Array{HalfSpace{Float64},1}:
 HalfSpace{Float64}([1.0, 0.0], 1.9328477832060151)
 HalfSpace{Float64}([0.0, 1.0], 1.1484644378760902)
 HalfSpace{Float64}([-1.0, -0.0], 0.7971425949785014)
 HalfSpace{Float64}([-0.0, -1.0], 1.5815259403084265)

julia> constraints_list(rand(Ball2)+rand(2))
ERROR: this function requires that the `constraints_list` method is appliable
Stacktrace:

 [2] constraints_list(::Translation{Float64,Array{Float64,1},Ball2{Float64}}, ::Val{false}) at /Users/forets/.julia/dev/LazySets/src/Translation.jl:348
 [3] constraints_list(::Translation{Float64,Array{Float64,1},Ball2{Float64}}) at /Users/forets/.julia/dev/LazySets/src/Translation.jl:343
 [4] top-level scope at none:0
```

schillic: Generally yes, but maybe we should do this in a second step and for all lazy operations.
For now, you can add the first method (without the second argument) I would say.

And by the way:
```julia
if x
    return f(y, true)
else
    return f(y, false)
end
```
↓
```julia
return f(y, x)
```

src/helper_functions.jl
```diff
@@ -525,3 +525,30 @@ end
...
 end # @eval
 end # if
+
+#=
+"""
+    concrete_subtypes(interface)
+
+
+"""
+function concrete_subtypes(interface)
```
schillic: Does this belong here?

src/Translation.jl
schillic: Okay, now that you already added it, we can keep it :+1:

test/unit_Translation.jl
mforets: Good. I added a translation along dimension 1 of a 3d set.

src/Translation.jl
```diff
@@ -0,0 +1,345 @@
...
+### Input
+
+- `tr` -- translation
+
+### Output
+
+`true` iff the wrapped set is empty.
+"""
+function isempty(tr::Translation)::Bool
+    return isempty(tr.X)
```
mforets: Wait, is it correct? If `tr.X` is empty but the translation is not the zero vector, we get a singleton. Then:

```julia
function isempty(tr::Translation)::Bool
     return isempty(tr.X) && isempty(tr.v)
end
```
But i don't think that we can create a translation with an empty vector, so `isempty` should be return `false` in all cases?

schillic: It is correct. The empty set is absorbing for the Minkowski sum:
https://github.com/JuliaReach/LazySets.jl/blob/30422364694ac4610ff3ed0ad80980f4742cd149/src/MinkowskiSum.jl#L46

PR #1185, opened 2019-03-01
#1159 - Add subtypes function with concrete argument

### LINKED ISSUE #1159: Add method to get all concrete subtypes of a given set interface
`subtypes` returns concrete and abstract types:

```julia
julia> first(subtypes(LazySet))
AbstractCentrallySymmetric

julia> last(subtypes(LazySet))
ResetMap
```
In some cases it is useful to be able to ask for all concrete subtypes of a given set interface.

mforets:
Closes #1159.

### DIFF
--- a/docs/src/lib/utils.md
+++ b/docs/src/lib/utils.md
@@ -46,3 +46,9 @@ CachedPair
 Approximations.UnitVector
 StrictlyIncreasingIndices
 ```
+
+### Inspection of set interfaces
+
+```@docs
+subtypes(::Type, ::Bool=false)
+```
--- a/docs/src/lib/utils.md
+++ b/docs/src/lib/utils.md
@@ -50,5 +50,5 @@ StrictlyIncreasingIndices
 ### Inspection of set interfaces
 
 ```@docs
-subtypes(::Type, ::Bool=false)
+subtypes(::Any, ::Bool)
 ```
--- a/docs/src/lib/utils.md
+++ b/docs/src/lib/utils.md
@@ -50,5 +50,5 @@ StrictlyIncreasingIndices
 ### Inspection of set interfaces
 
 ```@docs
-subtypes(::Any, ::Bool)
+LazySets.subtypes(::Any, ::Bool)
 ```
--- a/src/LazySet.jl
+++ b/src/LazySet.jl
@@ -62,43 +62,43 @@ If we only consider *concrete* subtypes, then:
 ```jldoctest
 julia> subtypes(LazySet, true)
 37-element Array{Type,1}:
+ Ball1
+ Ball2
+ BallInf
+ Ballp
  CacheMinkowskiSum
  CartesianProduct
  CartesianProductArray
  ConvexHull
  ConvexHullArray
+ Ellipsoid
  EmptySet
  ExponentialMap
  ExponentialProjectionMap
+ HPolygon
+ HPolygonOpt
+ HPolyhedron
+ HPolytope
+ HalfSpace
+ Hyperplane
+ Hyperrectangle
  Intersection
  IntersectionArray
+ Interval
+ Line
+ LineSegment
  LinearMap
  MinkowskiSum
  MinkowskiSumArray
  ResetMap
- Translation
- Ball2
- Ballp
- Ellipsoid
- HPolyhedron
- HalfSpace
- Hyperplane
- Line
- Universe
- HPolytope
- VPolytope
- Ball1
- LineSegment
- Zonotope
- VPolygon
- BallInf
- Hyperrectangle
- Interval
- SymmetricIntervalHull
- HPolygon
- HPolygonOpt
  Singleton
+ SymmetricIntervalHull
+ Translation
+ Universe
+ VPolygon
+ VPolytope
  ZeroSet
+ Zonotope
  ```
 """
 abstract type LazySet{N} end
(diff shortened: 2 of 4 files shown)


src/helper_functions.jl
```diff
@@ -525,3 +527,117 @@ end
...
+ UInt16
+ UInt32
+ UInt64
+ UInt8
+```
+"""
+function subtypes(interface, concrete::Bool=true)
+
+    subtypes_to_test = subtypes(interface)
+    @assert !isempty(subtypes_to_test) "an interface must have a subtype"
```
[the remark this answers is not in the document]
mforets: i agree to remove it. in case it doesn't have subtypes, an empty list is returned:

```julia
julia> subtypes(MyType)
0-element Array{Any,1}

julia> subtypes(MyType, false)
ERROR: AssertionError: an interface must have a subtype
Stacktrace:
 [1] subtypes(::Type, ::Bool) at /Users/forets/.julia/dev/LazySets/src/helper_functions.jl:621
 [2] top-level scope at none:0

julia> subtypes(MyType, true)
ERROR: AssertionError: an interface must have a subtype
Stacktrace:
 [1] subtypes(::Type, ::Bool) at /Users/forets/.julia/dev/LazySets/src/helper_functions.jl:621

 [2] top-level scope at none:0
```

src/LazySet.jl
```diff
@@ -33,6 +33,8 @@ Every concrete `LazySet` must define the following functions:
     `Real`
 - `dim(S::LazySet)::Int` -- the ambient dimension of `S`
 
+The subytpes of `LazySet`, including abstract interfaces:
```
schillic: I am not sure if this is interesting, because it does not contain all interfaces. But okay...

src/helper_functions.jl
```diff
@@ -529,15 +529,16 @@ end # @eval
...
 Return the concrete subtypes of a given interface.
 
 ### Input
 
 - `interface` -- an abstract type, usually a set interface
-- `concrete` -- (optional, default: `true`) if `true`, seek further the inner
-                subtypes of the given interface
+- `concrete`  -- (optional, default: `false`) if `true`, seek further the inner
+                 abstract subtypes of the given interface, otherwise return abstract
+                 subtypes of `interface`
```
schillic: otherwise return abstract subtypes of `interface`
↓
otherwise return only the direct subtypes of `interface`

src/LazySet.jl
mforets: Ok, i see. Would be nice to have an option to list all abstract interfaces as well...

docs/src/lib/utils.md
```diff
@@ -46,3 +46,9 @@ CachedPair
 Approximations.UnitVector
 StrictlyIncreasingIndices
 ```
+
+### Inspection of set interfaces
+
+```@docs
+subtypes(::Any, ::Bool)
```
mforets:
```suggestion
LazySets.subtypes(::Any, ::Bool)
```

src/LazySet.jl
```diff
@@ -54,6 +56,50 @@ julia> subtypes(LazySet)
  ResetMap
  Translation
 ```
+
+If we only consider *concrete* subtypes, then:
+
+```jldoctest
+julia> subtypes(LazySet, true)
```
```suggestion
julia> LazySets.subtypes(LazySet, true)
```

src/helper_functions.jl
```diff
@@ -525,3 +527,116 @@ end
...
+
+A list with the subtypes of the abstract type `interface`, sorted alphabetically.
+
+### Examples
+
+Consider the `AbstractPolytope` interface. If we include the abstract subtypes
+of this interface,
+
+```jldoctest
+julia> subtypes(AbstractPolytope, false)
```
mforets:
```suggestion
julia> LazySets.subtypes(AbstractPolytope, false)
```

src/helper_functions.jl
```diff
@@ -525,3 +527,116 @@ end
...
+ HPolytope
+ VPolytope
+```
+
+We can use this function to obtain the concrete subtypes of
+`AbstractCentrallySymmetricPolytope` and `AbstractPolygon` (further until all
+concrete types are obtained), using the `concrete` flag:
+
+```jldoctest
+julia> subtypes(AbstractPolytope, true)
```
mforets:
```suggestion
julia> LazySets.subtypes(AbstractPolytope, true)
```

src/helper_functions.jl
```diff
@@ -1,3 +1,5 @@
+import InteractiveUtils: subtypes
```
schillic: This must be done in a `VERSION` check for v0.6. I guess for the old version you need to do `import Base: subtypes`.

mforets: ooops i forgot about that. thanks

src/helper_functions.jl
```diff
@@ -525,3 +527,118 @@ end
...
+ VPolygon
+ VPolytope
+ ZeroSet
+ Zonotope
+```
+
+This function can be applied to other abstract types as well:
+
+```jldoctest subtypes
+julia> subtypes(Real, false)
```
schillic: I would drop this part because it depends too much on external packages. After all, we do not export this function.

mforets:
Thanks for the review. There is still this travis CI timeout problem that affect v1.0 tests. Not related to this PR.

PR #1188, opened 2019-03-02
Add doc and examples to uniform partition

mforets:


docs/src/lib/approximations.md
```diff
@@ -26,6 +26,12 @@ decompose
 project
 ```
 
+### Convenience functions
+
+``@docs
```
schillic:
```suggestion
```@docs
```
This fixes the build.

src/Approximations/decompositions.jl
```diff
@@ -1,4 +1,55 @@
...
+2-element Array{UnitRange{Int64},1}:
+ 1:1
+ 2:2
+
+julia> LazySets.Approximations.uniform_partition(2, 2)
+1-element Array{UnitRange{Int64},1}:
+ 1:2
+```
+
+If the block size argument is not compatible (i.e. does not divide) `n`, the
```
schillic:
```suggestion
If the block size argument is not compatible with (i.e. does not divide) `n`, the
```

PR #1192, opened 2019-03-03
#1184 - Concrete translation

### LINKED ISSUE #1184: Concrete translation of basic set types
In https://github.com/JuliaReach/LazySets.jl/pull/1173 we added a lazy `Translation` operator.
We should add the concrete operation.

It can preserve the type in many cases, for example, the translation of an abstracthyperrectangle amounts to translating the center. For a general polytopic set, one translates the vertices. For a set represented by constraints, one can translate the constraints using `constraints_list(::Translation)`.

(About the name `translate` or `translation` -- I prefer `translate` ... but for the concrete intersection we kept `Intersection` for the type and `intersection`, i wonder we we didn't choose `intersect`..)

schillic:
Closes #1184.

### DIFF
--- a/src/Ellipsoid.jl
+++ b/src/Ellipsoid.jl
@@ -251,7 +251,7 @@ A translated ellipsoid.
 
 ### Notes
 
-The shape matrix is shared with the original ellipsoid.
+The shape matrix is shared with the original ellipsoid if `share == true`.
 
 ### Algorithm
 
--- a/src/Ellipsoid.jl
+++ b/src/Ellipsoid.jl
@@ -231,3 +231,32 @@ function rand(::Type{Ellipsoid};
                    Matrix{N}(dim*I, dim, dim)
     return Ellipsoid(center, shape_matrix)
 end
+
+"""
+    translate(E::Ellipsoid{N}, v::AbstractVector{N}) where {N<:AbstractFloat}
+
+Translate (i.e., shift) an ellipsoid by a given vector.
+
+### Input
+
+- `E` -- ellipsoid
+- `v` -- translation vector
+
+### Output
+
+A translated ellipsoid.
+
+### Notes
+
+The shape matrix is shared with the original ellipsoid.
+
+### Algorithm
+
+We add the vector to the center of the ellipsoid.
+"""
+function translate(E::Ellipsoid{N},
+                   v::AbstractVector{N}) where {N<:AbstractFloat}
+    @assert length(v) == dim(E) "cannot translate a $(dim(E))-dimensional " *
+                                "set by a $(length(v))-dimensional vector"
+    return Ellipsoid(center(E) + v, E.shape_matrix)
+end
--- a/src/Ellipsoid.jl
+++ b/src/Ellipsoid.jl
@@ -233,14 +233,17 @@ function rand(::Type{Ellipsoid};
 end
 
 """
-    translate(E::Ellipsoid{N}, v::AbstractVector{N}) where {N<:AbstractFloat}
+    translate(E::Ellipsoid{N}, v::AbstractVector{N}; share::Bool=false
+             ) where {N<:AbstractFloat}
 
 Translate (i.e., shift) an ellipsoid by a given vector.
 
 ### Input
 
-- `E` -- ellipsoid
-- `v` -- translation vector
+- `E`     -- ellipsoid
+- `v`     -- translation vector
+- `share` -- (optional, default: `false`) flag for sharing unmodified parts of
+             the original set representation
 
 ### Output
 
@@ -254,9 +257,11 @@ The shape matrix is shared with the original ellipsoid.
 
 We add the vector to the center of the ellipsoid.
 """
-function translate(E::Ellipsoid{N},
-                   v::AbstractVector{N}) where {N<:AbstractFloat}
+function translate(E::Ellipsoid{N}, v::AbstractVector{N}; share::Bool=false
+                  ) where {N<:AbstractFloat}
     @assert length(v) == dim(E) "cannot translate a $(dim(E))-dimensional " *
                                 "set by a $(length(v))-dimensional vector"
-    return Ellipsoid(center(E) + v, E.shape_matrix)
+    c = center(E) + v
+    shape_matrix = share ? E.shape_matrix : copy(E.shape_matrix)
+    return Ellipsoid(c, shape_matrix)
 end
--- a/src/HPolygon.jl
+++ b/src/HPolygon.jl
@@ -167,7 +167,8 @@ A translated polygon in constraint representation.
 
 ### Notes
 
-The `a` vectors of the constraints are shared with the original constraints.
+The normal vectors of the constraints (vector `a` in `a⋅x ≤ b`) are shared with
+the original constraints if `share == true`.
 
 ### Algorithm
 
--- a/src/HPolygon.jl
+++ b/src/HPolygon.jl
@@ -148,15 +148,18 @@ function σ(d::AbstractVector{N}, P::HPolygon{N};
 end
 
 """
-    translate(v::AbstractVector{N}, P::HPolygon{N}) where {N<:Real}
+    translate(v::AbstractVector{N}, P::HPolygon{N}; share::Bool=false
+             ) where {N<:Real}
 
-Translate (i.e., shift) a polygon in constraint representation by a
-given vector.
+Translate (i.e., shift) a polygon in constraint representation by a given
+vector.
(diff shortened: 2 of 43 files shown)


src/Ellipsoid.jl
```diff
@@ -231,3 +231,32 @@ function rand(::Type{Ellipsoid};
...
+
+### Algorithm
+
+We add the vector to the center of the ellipsoid.
+"""
+function translate(E::Ellipsoid{N},
+                   v::AbstractVector{N}) where {N<:AbstractFloat}
+    @assert length(v) == dim(E) "cannot translate a $(dim(E))-dimensional " *
+                                "set by a $(length(v))-dimensional vector"
+    return Ellipsoid(center(E) + v, E.shape_matrix)
```
mforets: This choice doesn't convince me -- why share the shape matrix and not the center? 

Have you considered adding a `translate!` method (in addition to `translate`, which returns a new object), which shares both center and shape matrix?

schillic: > why share the shape matrix

The sharing is for efficiency. Normally you do not modify a set in-place, so it is safe to share structure.

> and not the center

Because the center would be modified.

> Have you considered adding a `translate!` method

I thought about it, but the object is not actually modified, so the `!` is not correct.
Maybe we want to have an additional optional argument (something like be `share` or `copy`)?

mforets: My proposal was to either (i) make `translate` to return a new independent object or (ii) make translate return a new independent object *and* also offer `translate!` for performance, which translates the object in-place. The implementation is as follows:

```julia
function translate(E::Ellipsoid{N},
                   v::AbstractVector{N}) where {N<:AbstractFloat}
    @assert length(v) == dim(E) "cannot translate a $(dim(E))-dimensional " *
    return translate!(copy(E), v)
end

function translate!(E::Ellipsoid{N},
                   v::AbstractVector{N}) where {N<:AbstractFloat}
    E.center .+= v
    return E
end
```

schillic: And I do not like that because this does not give you the possibility to share things.

mforets: To share things? Can you elaborate?

The behavior of `translate / translate!` of above is:

```julia
julia> E = Ellipsoid(zeros(2), Diagonal([1.0, 2.0]))
Ellipsoid{Float64}([0.0, 0.0], [1.0 0.0; 0.0 2.0])

julia> E_right = translate(E, [1.0, 0.0])
Ellipsoid{Float64}([1.0, 0.0], [1.0 0.0; 0.0 2.0])

julia> E_right.shape_matrix === E.shape_matrix
false

julia> E_right.center === E.center
false

julia> E_right_inplace = translate!(E, [1.0, 0.0])

julia> E_right_inplace.shape_matrix === E.shape_matrix
true

julia> E_right_inplace.center === E.center
true
```

schillic: Applying `translate!` modifies `E.center`.
I want a way to share `E.shape_matrix` but to copy `E.center`. This is the minimal amount of copying (in order to not modify `E`) and the maximal amount of sharing (i.e., the unmodified part is shared), which is not possible with your proposal.

src/Ellipsoid.jl
```diff
@@ -231,3 +231,37 @@ function rand(::Type{Ellipsoid};
...
+- `share` -- (optional, default: `false`) flag for sharing unmodified parts of
+             the original set representation
+
+### Output
+
+A translated ellipsoid.
+
+### Notes
+
+The shape matrix is shared with the original ellipsoid.
```
mforets:
```suggestion
The shape matrix is shared with the original ellipsoid if `share=true`.
```

src/HPolygon.jl
```diff
@@ -146,3 +146,39 @@ function σ(d::AbstractVector{N}, P::HPolygon{N};
...
+- `share` -- (optional, default: `false`) flag for sharing unmodified parts of
+             the original set representation
+
+### Output
+
+A translated polygon in constraint representation.
+
+### Notes
+
+The `a` vectors of the constraints are shared with the original constraints.
```
mforets:
```suggestion
The normal vectors of the constraints (vector `a` in `a⋅x ≤ b`) are shared with the original constraints if `share=true`.
```

src/HPolygonOpt.jl
```diff
@@ -182,3 +182,39 @@ function σ(d::AbstractVector{N}, P::HPolygonOpt{N};
...
+- `share` -- (optional, default: `false`) flag for sharing unmodified parts of
+             the original set representation
+
+### Output
+
+A translated optimized polygon in constraint representation.
+
+### Notes
+
+The `a` vectors of the constraints are shared with the original constraints.
```
mforets: The normal vectors of the constraints (vector `a` in `a⋅x ≤ b`) are shared with the original constraints if `share=true`.

mforets:
```suggestion
The normal vectors of the constraints (vector `a` in `a⋅x ≤ b`) are shared with the original constraints if `share=true`.
```

src/HPolyhedron.jl
```diff
@@ -392,6 +392,40 @@ function remove_redundant_constraints!(P::HPoly{N};
...
+- `share` -- (optional, default: `false`) flag for sharing unmodified parts of
+             the original set representation
+
+### Output
+
+A translated polyhedron in constraint representation.
+
+### Notes
+
+The `a` vectors of the constraints are shared with the original constraints.
```
mforets:
```suggestion
The normal vectors of the constraints (vector `a` in `a⋅x ≤ b`) are shared with the original constraints if `share=true`.
```

src/HalfSpace.jl
```diff
@@ -421,3 +421,38 @@ function _linear_map_hrep(M::AbstractMatrix{N}, P::HalfSpace{N}, use_inv::Bool)
...
+- `share` -- (optional, default: `false`) flag for sharing unmodified parts of
+             the original set representation
+
+### Output
+
+A translated half-space.
+
+### Notes
+
+The `a` vector is shared with the original half-space.
```
mforets:
```suggestion
The normal vectors of the halfspace (vector `a` in `a⋅x ≤ b`) is shared with the original halfspace if `share=true`.
```

src/Hyperplane.jl
```diff
@@ -400,3 +400,38 @@ function _linear_map_hrep(M::AbstractMatrix{N}, P::Hyperplane{N}, use_inv::Bool)
...
+- `share` -- (optional, default: `false`) flag for sharing unmodified parts of
+             the original set representation
+
+### Output
+
+A translated hyperplane.
+
+### Notes
+
+The `a` vector is shared with the original hyperplane.
```
mforets:
```suggestion
The normal vectors of the hyperplane (vector `a` in `a⋅x = b`) is shared with the original hyperplane if `share=true`.
```

src/Hyperrectangle.jl
```diff
@@ -186,3 +186,37 @@ function rand(::Type{Hyperrectangle};
...
+- `share` -- (optional, default: `false`) flag for sharing unmodified parts of
+             the original set representation
+
+### Output
+
+A translated hyperrectangle.
+
+### Notes
+
+The radius vector is shared with the original hyperrectangle.
```
mforets:
```suggestion
The radius vector is shared with the original hyperrectangle if `share=true`.
```

src/Line.jl
```diff
@@ -301,3 +301,38 @@ function _linear_map_hrep(M::AbstractMatrix{N}, P::Line{N}, use_inv::Bool) where
...
+- `share` -- (optional, default: `false`) flag for sharing unmodified parts of
+             the original set representation
+
+### Output
+
+A translated line.
+
+### Notes
+
+The `a` vector is shared with the original line.
```
mforets:
```suggestion
The normal vector of the line (vector `a` in `a⋅x = b`) is shared with the original line if `share=true`.
```

src/Zonotope.jl
```diff
@@ -590,3 +590,37 @@ function constraints_list(Z::Zonotope{N}
...
+- `share` -- (optional, default: `false`) flag for sharing unmodified parts of
+             the original set representation
+
+### Output
+
+A translated zonotope.
+
+### Notes
+
+The generator matrix is shared with the original zonotope.
```
mforets:
```suggestion
The generator matrix is shared with the original zonotope if `share=true`.
```

schillic:
Thanks! I added the changes in the latest commit.

PR #1197, opened 2019-03-05
#1196 - Add overapproximate with raw template-directions type

### LINKED ISSUE #1196: Add overapproximate with raw template-directions type
I want a method like the following but which accepts `Type{<:AbstractDirections}` (instead of an instance thereof).
https://github.com/JuliaReach/LazySets.jl/blob/0751b040531e09bc091e6bd19e5125d06eb8e390/src/Approximations/overapproximate.jl#L236-L246

We offer the same in `decompose`:
https://github.com/JuliaReach/LazySets.jl/blob/0751b040531e09bc091e6bd19e5125d06eb8e390/src/Approximations/decompositions.jl#L434-L441

schillic:
Closes #1196.

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -226,12 +226,12 @@ Overapproximating a set with template directions.
 
 ### Input
 
-- `X`           -- set
-- `dir`         -- direction representation
+- `X`   -- set
+- `dir` -- (concrete) direction representation
 
 ### Output
 
-A `HPolytope` overapproximating the set `X` with the directions from `dir`.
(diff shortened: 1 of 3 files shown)


src/Approximations/overapproximate.jl
```diff
@@ -245,6 +245,27 @@ function overapproximate(X::LazySet{N},
...
+Overapproximating a set with template directions.
+
+### Input
+
+- `X`   -- set
+- `dir` -- type of direction representation
+
+### Output
+
+A `HPolytope` overapproximating the set `X` with the directions from `dir`.
```
```suggestion
An `HPolytope` overapproximating the set `X` with the directions from `dir`.
```

PR #1204, opened 2019-03-06
#1152 - Add Complement set type

### LINKED ISSUE #1152: Type that represents the Complement
_Not an issue, just a note._

In `NeuralVerification`, we have found it useful to define a `PolytopeComplement` type ([see here](https://github.com/sisl/NeuralVerification.jl/blob/complements/src/utils/problem.jl#L24)). For the most part, all it does is serve as a wrapper for a set that reverses operations that act on it. I.e.`issubset(comp)` becomes `is_intersection_empty(comp.set)` etc.

I don't see any reason why it cannot be used to represent also more complex types like set unions, lazy convex hulls, etc., and maybe it is of some use to you as well, so I though I should mention it.

schillic:
Closes #1152.


docs/src/lib/operations.md
```diff
@@ -380,3 +380,17 @@ an_element(::UnionSetArray{N}) where {N<:Real}
 isempty(::UnionSetArray)
 isbounded(::UnionSetArray)
 ```
+
+## Complement
+
+Note that a complement of a convex set is generally not convex.
+Hence thee set type is not part of the convex-set family `LazySet`.
```
mforets:
```suggestion
Hence this set type is not part of the convex-set family `LazySet`.
```

src/is_subset.jl
```diff
@@ -842,3 +846,41 @@ function ⊆(::Universe{N}, ::EmptySet{N}, witness::Bool=false
...
+
+* If `witness` option is deactivated: `true` iff ``X ⊆ C``
+* If `witness` option is activated:
+  * `(true, [])` iff ``X ⊆ C``
+  * `(false, v)` iff ``X \\not\\subseteq C`` and
+    ``v ∈ X \\setminus C``
+
+### Algorithm
+
+Hence we fall back to `isdisjoint(X, C.X)`, which can be justified as follows.
```
```suggestion
We fall back to `isdisjoint(X, C.X)`, which can be justified as follows:
```

PR #1205, opened 2019-03-06
#1199 - Add order reduction in overapproximation

### LINKED ISSUE #1199: Overapproximation of the convex hull of zonotopes
In the method `function overapproximate(S::ConvexHull{N, Zonotope{N}, Zonotope{N}},
                         ::Type{<:Zonotope})::Zonotope where {N<:Real}` it is required that the order of the zonotopes is the same. 

Proposal: if the order is not the same, use order reduction and proceed.

mforets:
Closes #1199.


mforets:
(The build failure is unrelated to this PR.)

PR #1206, opened 2019-03-06
Add isdisjoint for Complement

schillic:
Follow-up of #1204.


src/is_intersection_empty.jl
```diff
@@ -1225,6 +1229,56 @@ function is_intersection_empty(U::Universe{N},
...
+# --- Complement ---
+
+
+"""
+    is_intersection_empty(C::Complement{N},
+                          X::LazySet{N},
+                          [witness]::Bool=false
+                         )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+
+Check whether a complement of a convex set and another set do not intersect.
```
mforets:
```suggestion
Check whether the complement of a convex set and another set do not intersect.
```

src/is_intersection_empty.jl
```diff
@@ -1225,6 +1229,56 @@ function is_intersection_empty(U::Universe{N},
...
+### Output
+
+* If `witness` option is deactivated: `true` iff ``X ∩ C = ∅``
+* If `witness` option is activated:
+  * `(true, [])` iff ``X ∩ C = ∅``
+  * `(false, v)` iff ``X ∩ C ≠ ∅`` and ``v ∈ X ∩ C``
+
+### Algorithm
+
+We fall back to `X ⊆ C.X`, which can be justified as follows.
```
mforets:
```suggestion
We fall back to `X ⊆ C.X`, which can be justified as follows:
```

PR #1215, opened 2019-03-08
Add conversion to and from IntervalBox

mforets:


test/unit_Hyperrectangle.jl
```diff
@@ -1,3 +1,5 @@
+using IntervalArithmetic: IntervalBox
```
mforets:
```suggestion
using IntervalArithmetic
```

src/convert.jl
```diff
@@ -497,3 +499,41 @@ function convert(X::Type{HPOLYGON},
...
+An `IntervalBox`.
+"""
+function convert(::Type{IntervalBox}, H::AbstractHyperrectangle)
+    return IntervalBox(low(H), high(H))
+end
+
+"""
+    convert(::Type{Hyperrectangle}, IB::IntervalBox)
+
+Converts an `IntervalBox` from `IntervalArithmetic` to an hyperrectangular set.
```
```suggestion
Converts an `IntervalBox` from `IntervalArithmetic` to a hyperrectangular set.
```

test/unit_Hyperrectangle.jl
```diff
@@ -157,4 +159,9 @@ for N in [Float64, Rational{Int}, Float32]
     @test ispermutation(clist,
                         [HalfSpace(N[1, 0], N(3)), HalfSpace(N[0, 1], N(3)),
                          HalfSpace(N[-1, 0], N(1)), HalfSpace(N[0, -1], N(1))])
+
+    # conversion to and from IntervalArithmetic's IntervalBox type
+    B = IntervalBox(0..1, 0..1)
+    H = convert(Hyperrectangle, B)
+    @test convert(IntervalBox, H) == B
```
schillic: Can you add a test with non-`Vector` vectors for the TODO part?

test/unit_Hyperrectangle.jl
```diff
@@ -1,3 +1,6 @@
+import IntervalArithmetic
```
schillic:
```suggestion
using IntervalArithmetic
```

mforets: Well this is the commit i made previously, but Travis complained..

mforets: https://travis-ci.org/JuliaReach/LazySets.jl/jobs/503687286

schillic: Then I suggest to remove this line completely and only get what you need.

mforets: another solution is probably to do `using IntervalArithmetic` *and* also `using LazySets: radius` to avoid the conflict.

schillic: You already have `using IntervalArithmetic: IntervalBox` one line below. So why do you need `IntervalArithmetic` at all?

mforets: >  Then I suggest to remove this line completely and only get what you need.

How?

In the REPL one does `using LazySets, IntervalArithmetic` (without warnings) and can construct the `IntervalBox` using the common notation `IntervalBox(0..1, 0..1)` which requires `IntervalArithmetic`.

Otherwise, use the more verbose `IntervalBox(IntervalArithmetic.Interval(0, 1), IntevalArithmetic.Inteval(0, 1)`, which requires that the name `IntervalArithmetic` is available (this is the purpose of the `import IntervalArithmetic`).

mforets: > So why do you need IntervalArithmetic at all?

Because `IntervalArithmetic.Interval` is undefined.

schillic: Then let's do `using IntervalArithmetic: Interval`?

mforets: Hmm this conflicts with LazySets's `Interval` type, and will cause problems if another test file uses lazysets' interval after this one. What is the argument against using `import`?

mforets: ... or otherwise, keep `using IntervalArithmetic` and also `using LazySets: radius` to avoid the namespace conflict. Both options look good to me.

schillic: I don't understand this. The only difference between `import X` and `using X` [should be](https://docs.julialang.org/en/v1/manual/faq/#What-is-the-difference-between-%22using%22-and-%22import%22?-1) that in the first case one can write `y` instead of `X.y`. Why is there now a difference?

schillic: Alright, the correct source is [here](https://docs.julialang.org/en/v1/manual/modules/#Summary-of-module-usage-1). `import` is correct here.

mforets:
(the failure on v1.1. is a travis CI timeout, unrelated to this PR)

PR #1217, opened 2019-03-10
#1119 - constraints_list for ResetMap of hyperrectangle

### LINKED ISSUE #1119: constraints_list for ResetMap of hyperrectangle
The implementation of `constraints_list` for a `ResetMap` wrapping an `AbstractHyperrectangle` can be much more efficient. To keep things simple, we should just implement the concrete `reset_map` operation for this case.

schillic:
Closes #1119.


mforets:
#1119 proposes to add a new function `reset_map` that is not added here. Is it still on the plans?

schillic:
Good point, I opened #1223 for the `reset_map` operation.

PR #1225, opened 2019-03-11
#1219 - Subset check between CartesianProductArrays

### LINKED ISSUE #1219: Subset check between CartesianProductArrays
A subset check between two `CartesianProductArray`s can be efficiently performed if they have the same block structure: simply perform the check block-wise in lower dimensions. Benefits:
1. Low-dimensional checks are more efficient.
2. Julia automatically chooses a suitable algorithm for the sub-blocks, which again is more efficient.

We should initially verify if the block structure is identical. There can be an option to bypass this test.

schillic:
Closes #1219.

### DIFF
--- a/src/is_subset.jl
+++ b/src/is_subset.jl
@@ -898,8 +898,8 @@ end
      )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
 
 Check whether a Cartesian product of finitely many convex sets is contained in
-the another Cartesian product of finitely many convex sets, and otherwise
-optionally compute a witness.
+another Cartesian product of finitely many convex sets, and otherwise optionally
+compute a witness.
 
 ### Input
 
@@ -917,7 +917,6 @@ optionally compute a witness.
   * `(false, v)` iff ``X \\not\\subseteq Y`` and
     ``v ∈ X \\setminus Y``
 
-
 ### Notes
 
 This algorithm requires that the two Cartesian products share the same block
@@ -927,6 +926,10 @@ Depending on the value of `check_block_equality`, we check this property.
 ### Algorithm
 
 We check for inclusion for each block of the Cartesian products.
+
+For witness production, we obtain a witness in one of the blocks.
+We then construct a high-dimensional witness by obtaining any point in the other
+blocks (using `an_element`) and concatenating these points.
 """
 function ⊆(X::CartesianProductArray{N},
            Y::CartesianProductArray{N},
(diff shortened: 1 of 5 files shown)


src/is_subset.jl
```diff
@@ -887,3 +887,75 @@ function ⊆(X::LazySet{N}, C::Complement{N}, witness::Bool=false
...
+# --- CartesianProduct ---
+
+
+"""
+    ⊆(X::CartesianProductArray{N}, Y::CartesianProductArray{N},
+      witness::Bool=false; check_block_equality::Bool=true
+     )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+
+Check whether a Cartesian product of finitely many convex sets is contained in
+the another Cartesian product of finitely many convex sets, and otherwise
```
```suggestion
another Cartesian product of finitely many convex sets, and otherwise
```

mforets:
These ideas can also be brought to the binary `CartesianProduct` type; I've opened https://github.com/JuliaReach/LazySets.jl/issues/1227.

src/is_subset.jl
```diff
@@ -887,3 +887,75 @@ function ⊆(X::LazySet{N}, C::Complement{N}, witness::Bool=false
...
+    if check_block_equality && !same_block_structure(aX, aY)
+        throw(ArgumentError("this inclusion check requires Cartesian products" *
+                            "with the same block structure"))
+    end
+
+    for i in 1:length(aX)
+        result = ⊆(aX[i], aY[i], witness)
+        if !witness && !result
+            return false
+        elseif witness && !result[1]
```
mforets: (side note: this reads a bit odd, that in one `if` one uses `!result` and in the second indexes the reuslt, `!result[1]`, but it's a consequence of the return type of inclusion checks which is either a `Bool` or a tuple `(::Bool, ::Vector)`)

src/is_subset.jl
```diff
@@ -887,3 +887,75 @@ function ⊆(X::LazySet{N}, C::Complement{N}, witness::Bool=false
...
+            # construct a witness
+            w = Vector{N}(undef, dim(X))
```
mforets: witness construction consists of concatenating the first witness found with any element of `X`, for the resto of the blocks in `X` => add this explanation to the docstring under `Algorithm`?

src/is_subset.jl
schillic: Yes, but what can I do?

mforets: i think that it is good like that!

> Yes, but what can I do?

(writing an outer case `if witness . . .` and repeating the line `result = ⊆(aX[i], aY[i], witness)`, of course (not a big deal) is one way, otherwise with dispatch..).

PR #1246, opened 2019-03-23
#1245 - Zonotope conversion from flat hyperrectangle contains zero generators

### LINKED ISSUE #1245: Zonotope conversion from flat hyperrectangle contains zero generators

schillic:
Closes #1245.

### DIFF
--- a/src/helper_functions.jl
+++ b/src/helper_functions.jl
@@ -605,3 +605,37 @@ function subtypes(interface, concrete::Bool)
     end
     return sort(result, by=string)
 end
+
+"""
+    delete_zero_columns(A::AbstractMatrix)
+
+Remove all columns that only contain zeros from a given matrix.
+
+### Input
+
+- `A`           -- matrix
+- `copy_always` -- (optional, default: `false`) flag to copy the matrix even if
+                   it does not contain any zero column
+
+### Output
+
+A matrix.
+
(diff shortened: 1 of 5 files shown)


mforets:
Let's remove the zero columns in creation of the zonotope? With an option `remove_zero_gens` (default, true).

src/helper_functions.jl
```diff
@@ -605,3 +605,37 @@ function subtypes(interface, concrete::Bool)
...
+
+### Output
+
+A matrix.
+
+If the input matrix `A` does not contain any zero column, we return `A` unless
+the option `copy_always` is set.
+If the input matrix contains zero columns, we always return a copy.
+"""
+function delete_zero_columns(A::AbstractMatrix, copy_always::Bool=false)
```
mforets: The argument `copy_always` doesn't seem to be used in the body of the function.

src/helper_functions.jl
```diff
@@ -605,3 +605,37 @@ function subtypes(interface, concrete::Bool)
...
+
+- `A`           -- matrix
+- `copy_always` -- (optional, default: `false`) flag to copy the matrix even if
+                   it does not contain any zero column
+
+### Output
+
+A matrix.
+
+If the input matrix `A` does not contain any zero column, we return `A` unless
```
mforets: To avoid returning a copy in all cases, we can return a `@view` instead.

PR #1253, opened 2019-03-26
Remove use_polyhedra_library arg in intersection

mforets:
I think this simplifies the typical use of this function, by removing the second keyword argument. For example, with this commit one can use `intersection(P, Q, backend=CDDLib.Library())`.

### DIFF
--- a/src/concrete_intersection.jl
+++ b/src/concrete_intersection.jl
@@ -297,11 +297,12 @@ function intersection(P1::AbstractHPolygon{N},
     return P
 end
 
+using MathProgBase.SolverInterface: AbstractMathProgSolver
+
 """
     intersection(P1::AbstractPolyhedron{N},
                  P2::AbstractPolyhedron{N};
-                 backend=nothing,
-                 use_polyhedra_interface=false) where {N<:Real}
+                 backend=GLPKSolverLP()) where {N<:Real}
 
 Compute the intersection of two polyhedra.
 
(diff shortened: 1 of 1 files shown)


src/concrete_intersection.jl
```diff
@@ -339,33 +339,32 @@ This implementation unifies the constraints of the two sets obtained from the
...
-        # convert to an hrep, remove the redundancies and convert back to HPOLY
-        ph = polyhedron(Q; backend=backend)
-        removehredundancy!(ph)
-        return convert(HPOLY, ph)
-    else
-        if backend == nothing
-            backend = GLPKSolverLP()
-        end
-        # here, detection of empty intersection may be reported as an infeasible LP
+    if backend isa AbstractMathProgSolver
```
schillic: I am not sure that this is the correct criterion to distinguish between `Polyhedra`, but let's go for it.
In the `else` case we could check for [`isa Polyhedra.Library`](https://github.com/JuliaPolyhedra/Polyhedra.jl/blob/fc51bb5efdb4270ef8ff71561bf07bedc63ece0d/src/Polyhedra.jl#L10) and throw an error otherwise.

mforets: > In the else case we could check for isa Polyhedra.Library and throw an error otherwise.

But i would need `using Polyhedra: Library` which requires Polyhedra. Is there some workaround around this?

schillic: Oh, I see. You could outsource to another method that you then override with `Requires`, but that is much too complicated for a nicer error message. Let's keep it as is!

mforets: In the last commit i added a comment about this issue.

PR #1255, opened 2019-03-27
Add spherical directions

mforets:


mforets:
Test pass locally on v1.1, i haven't tested in v0.6 though..

src/Approximations/template_directions.jl
```diff
@@ -330,3 +337,127 @@ end
...
+
+### Fields
+
+- `Nθ`    -- length of the partition of the acimutal angle
+- `Nφ`    -- length of the partition of the polar angle
+- `stack` -- list of computed directions
+
+### Notes
+
+The `SphericalDirections` constructor provides a sample of the unit sphere
```
schillic: Add a link to https://en.wikipedia.org/wiki/Spherical_coordinate_system ?

src/Approximations/template_directions.jl
```diff
@@ -330,3 +337,127 @@ end
...
+in ``\\mathbb{R}^3``, which is parameterized by the acimutal and polar angles
```
schillic: > acimutal

In English this seems to be called [azimuthal](https://en.wikipedia.org/wiki/Azimuth) (occurs at least once more below).

src/Approximations/template_directions.jl
```diff
@@ -330,3 +337,127 @@ end
...
+and ``Dφ`` are discretized in ``Nθ`` and ``Nφ`` respectively. Then the Cartesian
+componentes of each direction are obtained with
+
+```math
+[sin(θᵢ)*cos(φᵢ), sin(θᵢ)*sin(φᵢ), cos(θᵢ)]
+```
+Note that by discretization, the polar point may be computed more than once, in
+general. Removing duplicates takes time; you can choose to remove duplicates or
+not by setting the keyword argument `remove_duplicates` in the constructor
+(default, `true`).
```
schillic:
```suggestion
(default `true`).
```

src/Approximations/template_directions.jl
```diff
@@ -330,3 +337,127 @@ end
...
+Repeated directions are removed by default. Use the keyword argument `remove_duplicates`
+to control this behavior:
+
+```jldoctest spherical_directions
+julia> sd_with_duplicates = SphericalDirections(3, remove_duplicates=true);
+
+julia> length(sd_with_duplicates)
+2
+
+julia> sd_without_duplicates = SphericalDirections(3, remove_duplicates=false);
```
schillic: The names of these variables should be swapped.

src/Approximations/template_directions.jl
```diff
@@ -330,3 +337,127 @@ end
 
 end # @eval
 end # if
+
+# ==================================================
+# Spherical directions
+# ==================================================
+
+@inline _isapprox_included(d, S) = any(isapproxzero.([norm(d - si, Inf) for si in S]))
```
schillic: I have the impression that this first creates the whole vector and only then starts evaluating. Is my impression wrong?

src/Approximations/template_directions.jl
```diff
@@ -330,3 +337,127 @@ end
...
+julia> length(sd_4_8)
+15
+```
+"""
+struct SphericalDirections <: AbstractDirections{Float64}
+    Nθ::Int
+    Nφ::Int
+    stack::Vector{Vector{Float64}} # stores the spherical directions
+
+    function SphericalDirections(Nθ::Int, Nφ::Int, stack::Vector{Vector{Float64}}; remove_duplicates=true)
```
schillic: What is the reason to make `stack` an input argument? Should this not always be initialized to an empty vector?

src/Approximations/template_directions.jl
```diff
@@ -330,3 +337,127 @@ end
...
+julia> length(sd_4_5)
+5
+
+julia> sd_4_8 = SphericalDirections(4, 8);
+
+julia> length(sd_4_8)
+15
+```
+"""
+struct SphericalDirections <: AbstractDirections{Float64}
```
schillic: Why not make this parametric in the numeric type (here: `N<:AbstractFloat`)?

src/Approximations/template_directions.jl
mforets: That's correct. I've revised the implementation, please reopen if you have further comments.

src/Approximations/template_directions.jl
mforets: True, it doesn't need to be an input.

src/Approximations/template_directions.jl
```diff
@@ -330,3 +337,139 @@ end
...
+# ==================================================
+
+# returns `true` if and only if there is a vector in the list of vectors S
+# such that it is approximately d
+function _isapprox_included(d::Vector{N}, S::Vector{Vector{N}}) where {N}
+    result = false
+    for si in S
+        if isapproxzero(norm(d-si))
+            result = true
+            break
```
schillic: Slightly shorter: `return true` here and `return `false` below (saves the `result` variable).

src/Approximations/template_directions.jl
```diff
@@ -330,3 +337,139 @@ end
...
+julia> length(sd_4_8)
+15
+```
+"""
+struct SphericalDirections{N} <: AbstractDirections{N}
+    Nθ::Int
+    Nφ::Int
+    stack::Vector{Vector{N}} # stores the spherical directions
+
+    function SphericalDirections(Nθ::Int, Nφ::Int; N::Type{<:AbstractFloat}=Float64, remove_duplicates=true)
```
schillic: Either you restrict the type of `N` in the struct or you should omit the restriction here.

test/unit_template_directions.jl
```diff
@@ -40,10 +40,19 @@ for N in [Float64, Float32, Rational{Int}]
         boxdiag = overapproximate(X, BoxDiagDirections)
         @test length(dir) == length(boxdiag.constraints) ==
               (n == 1 ? 2 : 2^n + 2*n)
+
+        # spherical directions approximation
+        if n == 3 && N <: Float64
```
schillic: `Float64` is a base type and cannot be subtyped.

src/Approximations/template_directions.jl
mforets: By omit the restriction you mean to use `N::Type{<:Number}=Float64` here?

schillic: I mean to have the same restriction as in the struct. There are no restrictions, so there should be no restrictions here as well.
Also, it is better to use `function SphericalDirections{N}(...) where {N}` instead of adding the type as a normal argument.

mforets: Ok, got it. I've used `AbstractFloat` consistently now. I noted that using exact arithmetic and the trigonometric functions don't play well, for instance `Rational{Int}` easily overflows already `Nθ=5`. Since we don't test with `Rational{BigInt}` i would just leave it like that.. If usage arises we can remove the annotation..

src/Approximations/template_directions.jl
```diff
@@ -330,3 +337,116 @@ end
...
+``θ ∈ Dθ := [0, π]`` and ``φ ∈ Dφ := [0, 2π]`` respectively, see the wikipedia
+entry [Spherical coordinate system](https://en.wikipedia.org/wiki/Spherical_coordinate_system).
+The domains ``Dθ`` and ``Dφ`` are discretized in ``Nθ`` and ``Nφ`` respectively.
+Then the Cartesian componentes of each direction are obtained with
+
+```math
+[sin(θᵢ)*cos(φᵢ), sin(θᵢ)*sin(φᵢ), cos(θᵢ)].
+```
+
+The north and south poles are treated separately, so that no those points
```
schillic:
```suggestion
The north and south poles are treated separately so that those points
```

schillic:
This changeset was [breaking](https://travis-ci.org/JuliaReach/LazySets.jl/jobs/512629571#L1420) in Julia v0.6!

PR #1260, opened 2019-03-28
#1252 - Add mesh.jl file

### LINKED ISSUE #1252: Add 3D plots using Makie's mesh
We can exploit `Polyhedra.Mesh` constructor and use it for plotting via Makie:

```julia
using Makie, LazySets, Polyhedra

Gp = LazySets.HalfSpace([2.0, 1.0, -1.0], 20.0)
H = Hyperrectangle(low=[0.0, 0.0, 0.0], high=[10.0, 10.0, 10.0])
Q = HPolytope([constraints_list(H); constraints_list(Gp)])
Q_Polyhedra = LazySets.polyhedron(Q, backend=CDDLib.Library())
m = Polyhedra.Mesh(Q_Polyhedra)
mesh(m, color=:blue)
```

Screen Shot 2019-03-26 at 10 28 15

I'm not sure if we can just find a way around using the same function name `plot` (used for the plot recipes) to also work with Makie for 3d plots through this `Mesh` -- note that we can basically plot any set whose `constraints_list` function is computable!

On the other hand we can add a new function `plot3d` instead of trying to extend `plot`.

Other examples found on the web:

- https://makie.juliaplots.org/stable/examples-mesh.html
- https://github.com/JuliaPolyhedra/Polyhedra.jl/blob/21f7f0f376aef4d12f23879e72076c6c08c2b173/examples/3D%20Plotting%20a%20projection%20of%20the%204D%20permutahedron.ipynb
- https://github.com/JuliaPolyhedra/Polyhedra.jl/issues/153#issue-418447905

mforets:
Closes #1252.

This file adds functionality to plot any abstract polytope or any lazy set whose constraints list can be computed. It relies on converting an `HPolytpe` into a `Polyhedra.Mesh`, which can be plotted with `Makie`. 

It remains to see if we can refactor `plot3d` and `plot3d!`, and to load this file using a `Requires` block.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -28,6 +28,7 @@ GLPKMathProgInterface = "≥ 0.4.0"
 GR = "≥ 0.39.1"
 IntervalArithmetic = "≥ 0.14.0"
 JuMP = "≥ 0.19.2"
+Makie = "≥ 0.9.4"
 MathProgBase = "≥ 0.7.0"
 Optim = "≥ 0.14.1"
 Plots = "≥ 0.25.1"
@@ -44,6 +45,7 @@ Expokit = "a1e7a1ef-7a5d-5822-a38c-be74e1bb89f4"
 GLPK = "60bf3e95-4087-53dc-ae20-288a0d20c6a6"
 GR = "28b8d3ca-fb5f-59d9-8090-bfdbd6d07a71"
 JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
+Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 Polyhedra = "67491407-f73d-577b-9b50-8179a7c68029"
@@ -51,5 +53,5 @@ TaylorModels = "314ce334-5f6e-57ae-acf6-00b6e903104a"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["CDDLib", "Distributions", "Documenter", "Expokit", "GLPK", "GR", "JuMP", "Optim", "Plots", "Polyhedra",
-        "TaylorModels", "Test"]
+test = ["CDDLib", "Distributions", "Documenter", "Expokit", "GLPK", "GR", "JuMP",
+        "Makie", "Optim", "Plots", "Polyhedra", "TaylorModels", "Test"]
--- a/src/mesh.jl
+++ b/src/mesh.jl
@@ -1,0 +1,78 @@
+#using .LazySets
+#using .LazySets.Approximations
+#using LazySets: default_polyhedra_backend, dim # needed?
+
+using Makie: mesh, mesh!
+using AbstractPlotting: Automatic
+using Polyhedra: Mesh
+
+export plot3d
+#import Plots: plot3d
+
+function plot3d(S::LazySet{N}, num_dirs=10; backend=default_polyhedra_backend(S, N),
+                alpha=1.0, color=:blue, colormap=:viridis, colorrange=Automatic(), interpolate=false,
+                linewidth=1, overdraw=false, shading=true, transparency=true, visible=true) where {N}
+
+    if dim(S) != 3
+        throw(ArgumentError("plot3d can only be used to plot 3D sets, but the given set is $(dim(S))-dimensional"))
+    end
+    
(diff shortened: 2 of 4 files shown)


src/mesh.jl
```diff
@@ -0,0 +1,78 @@
...
+
+    if dim(S) != 3
+        throw(ArgumentError("plot3d can only be used to plot 3D sets, but the given set is $(dim(S))-dimensional"))
+    end
+    
+    if !applicable(constraints_list, S)
+        throw(ArgumentError("the `constraints_list` function is not applicable to a set of type $S, hence plot3d cannot be used"))
+    end
+
+    P = overapproximate(S, SphericalDirections(num_dirs))
```
schillic: I do not understand why you require the `constraints_list` if you do not use it.
Warning: `constraints_list` does not imply boundedness (for instances `constraints_list(::Universe) == []`).

src/mesh.jl
```diff
@@ -0,0 +1,78 @@
...
+    remove_redundant_constraints!(P)
+    return plot3d!(P, backend=backend, alpha=alpha, color=color, colormap=colormap, colorrange=colorrange,
```
schillic: Can we avoid these duplications, e.g., by calling a helper function with the respective function symbol (`mesh`/`mesh!`) as additional argument?

mforets:
Example:

```julia
julia> using LazySets, Polyhedra, Makie
[ Info: Recompiling stale cache file /Users/forets/.julia/compiled/v1.1/LazySets/NjrGc.ji for LazySets [b4f0291d-fe17-52bc-9479-3d1a343d9043]

julia> plot3d(10. * rand(Hyperrectangle, dim=3)); plot3d!(10. * rand(Hyperrectangle, dim=3), color=:red)
```

Screen Shot 2019-07-28 at 14 05 26

src/mesh.jl
mforets: True; i've dropped the overapproximation and kept the call to `constraints_list`. Some cases may be problematic (eg. lazy intersection) but we can address it later.

src/mesh.jl
mforets: I've outsourced the common parts to `plot3d_helper`. As it turns out, with Makie one has to define the pairs `plot3d / plot3d!` and export both.

Project.toml
```diff
@@ -21,6 +21,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 julia = "≥ 1.0.0"
 CDDLib = "≥ 0.5.2"
 Distributions = "≥ 0.20.0"
+Makie = "≥ 0.9.4"
```
schillic: Can we keep packages sorted alphabetically here as well, please? :heart:

src/mesh.jl
```diff
@@ -0,0 +1,43 @@
+function load_mesh()
+return quote
+
+using .Polyhedra: Mesh
```
schillic: If `Polyhedra` is not loaded, this would crash. This requires a two-level check in both the `Polyhedra` and `Makie` loading. Whichever is loaded later should execute this `load_mesh` function. See [here](https://github.com/JuliaReach/Reachability.jl/issues/449#issuecomment-462451906) for a proposal.

src/mesh.jl
```diff
@@ -0,0 +1,165 @@
...
+
+```julia
+julia> using CDDLib
+
+julia> plot3d(Sapprox, backend=CDDLib.Library())
+```
+
+### Examples
+
+The functionality in this files requires *both* `Polyhedra` and `Makie`; so after
```
schillic:
```suggestion
The functionality requires *both* `Polyhedra` and `Makie`; so after
```

src/mesh.jl
```diff
@@ -0,0 +1,165 @@
...
+                    the colormap, e.g. `color = [-2, 4]` with `colorrange = (-2, 4)`
+                    will map to the lowest and highest color value of the colormap
+- `interpolate`  -- (optional, default: `false`) a bool for heatmap and images,
+                    it toggles color interpolation between nearby pixels
+- `linewidth`    -- (optional, default: `1`) a number that specifies the width of
+                    the line in `line` and `linesegments` plots
+- `overdraw`     -- (optional, default: `false`)
+- `shading`      -- (optional, default: `false`) a boolean that specifies if shading
+                    should be on or not (for meshes)
+- `transparency` -- (optional, default: `false`)
```
schillic:
```suggestion
```

src/mesh.jl
```diff
@@ -0,0 +1,165 @@
...
```
schillic:
```suggestion
```

Issue #1281, opened 2019-04-12
In convex hull?


kmundnic:
I have gone through the docs, but cannot seem to find a way to check whether a vector or Singleton is in a convex hull. Is this an implemented operation?

This is how I imagine it:
```julia
> points = N -> [randn(2) for i in 1:N]
> v = points(30)
> hull = convex_hull(v)
> Singleton(v[1]) ∈ VPolygon(hull)
true
```
However, this throws the following error:
```julia
ERROR: MethodError: no method matching iterate(::VPolygon{Float64})
Closest candidates are:
  iterate(::Core.SimpleVector) at essentials.jl:589
  iterate(::Core.SimpleVector, ::Any) at essentials.jl:589
  iterate(::ExponentialBackOff) at error.jl:171
  ...
Stacktrace:
 [1] in(::Singleton{Float64,Array{Float64,1}}, ::VPolygon{Float64}) at ./operators.jl:949
 [2] top-level scope at none:0
```
which unfortunately is not very informative for a newcomer.

I also thought about creating a set from a Singleton, and try using `⊆`, but that always throws a `false` for convex hulls.

If this is supported, it would be a good addition in the convex hull section of the docs.

Thanks for your help!

schillic:
> `> Singleton(v[1]) ∈ VPolygon(hull)`

`Singleton(x)` gives you a set containing the point `x`. `VPolygon(y)` also gives you a set of points. So the l.h.s. and the r.h.s. of `∈` are sets of points in your example. What you want to do is either of the following (they are equivalent):

```julia
julia> v[1] ∈ VPolygon(hull)
true

julia> Singleton(v[1]) ⊆ VPolygon(hull)
true
```

mforets:
> If this is supported, it would be a good addition in the convex hull section of the docs.

See https://github.com/JuliaReach/LazySets.jl/pull/1282, comments welcome!

kmundnic:
I thought `Singleton` wasn't a set, that's why I was using `∈` instead of `⊆`. Thanks for the clarification!

mforets:
No problem! Thanks for pushing us to improve the docs ;)

PR #1282, opened 2019-04-12
More documentation for singleton set membership 

mforets:
See https://github.com/JuliaReach/LazySets.jl/issues/1281, this PR adds a better error message and a comment in the convex hull docs.

### DIFF
--- a/docs/src/man/convex_hulls.md
+++ b/docs/src/man/convex_hulls.md
@@ -89,12 +89,20 @@ plot!(p, VPolygon(hull), alpha=0.2)
 ```
 
 !!! note
-    To check if a point `p::AbstractVector` is in a polygon in vertex representation `V`,
-    use `p ∈ V`. However, if you are working with a `Singleton` (it's a *set* with one
-    element), use *set inclusion* `⊆`. The following example illustrates this.
+    To check if a point `p::AbstractVector` is in another set, e.g. a polygon in
+    vertex representation `V`, use `p ∈ V`. However, if you are working with a
+    `Singleton`, which is a *set* with one element, use *set inclusion* `⊆`.
+    The following example illustrates this difference.
 
-```@example example_ch
-Singleton(v[1]) ∈ VPolygon(hull)
+```jldoctest
+julia> points = N -> [randn(2) for i in 1:N];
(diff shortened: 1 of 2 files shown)


mforets:
Just checked building the docs locally and it doesn't behave as i expected: the error message is not displayed in the generated docs (although it is shown in the docs build). I'll just not run the command in the example block.

Screen Shot 2019-04-12 at 08 11 30

Screen Shot 2019-04-12 at 08 12 53

schillic:
Maybe use a [`jldoctest` block](https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Exceptions-1) instead of an `example` block?

kmundnic:
@mforets the comments you're proposing for the docs look good to me! I would include a section or comment for explicitly stating that you can check whether a point is in a convex hull. If you Google for this in Julia, there's mostly a very old comment (circa 2015) using JuMP, since you can pose the problem as a LP. However, in many cases (and in lower dimensions), this solution seems cleaner to me than posing the problem as a feasibility problem using an optimization package (JuMP, Convex.jl).

mforets:
Ok, i see your use case. Indeed, in 2D it's much more efficient to exploit that the points in the plane can be ordered (cf. the implementation in [this function](https://github.com/JuliaReach/LazySets.jl/blob/master/src/VPolygon.jl#L324), in particular it doesn't allocate, see below).

The feasibility test is still good in higher dims and is the fallback implementation for a general `VPolytope` (cf. [this function](https://github.com/JuliaReach/LazySets.jl/blob/master/src/VPolytope.jl#L135)).

We can compare both:

```julia
julia> points = N -> [randn(2) for i in 1:N];
julia> p = points(100);
julia> hull = convex_hull(p);
julia> V_polygon = VPolygon(hull, apply_convex_hull=false);
julia> V_polytope = convert(VPolytope, V_polygon);
julia> x = sum(hull)/length(hull);
```

```julia
julia> @btime x ∈ $V_polygon
  47.688 ns (0 allocations: 0 bytes)
true

julia> @btime $x ∈ $V_polytope
  33.998 μs (106 allocations: 7.56 KiB)
true
```

PR #1289, opened 2019-04-17
Add assertion and message in _linear_map_vrep

mforets:


test/unit_Polyhedron.jl
```diff
@@ -89,6 +89,14 @@ for N in [Float64, Rational{Int}, Float32]
         # concrete linear map with noninvertible matrix throws an error
         @test_throws ArgumentError linear_map(N[2 3; 0 0], P)
     end
+
+    if !test_suite_polyhedra
+        # concrete linear map of a bounded polyhedron by a non-invertible matrix
+        # throws an assertion error, since tovrep(HPolytope(...)) is required
```
schillic: :+1:, but note that we always test with `Polyhedra` loaded on Travis (so this is a "dead test"). Should we have one build where we don't load `Polyhedra`? I don't know if we can still pass an argument to the tests when we switch to the new `Project.toml`-based scheme.

PR #1293, opened 2019-04-18
#1292 - Custom template directions

### LINKED ISSUE #1292: Custom template directions
Add a type `CustomDirections <: AbstractDirections` which just wraps a list of user-defined directions.

Note that there is no guarantee that an overapproximation resulting from those directions is bounded, so we have to check that (best in the constructor of that new type). How do we check that?

schillic:
Closes #1292.

@mforets: Please in particular check whether in the first commit the `isbounded` methods for polar and spherical directions are correct. I am not very familiar with them.

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -290,8 +290,7 @@ function overapproximate(Z::Zonotope, ::Type{<:Hyperrectangle})::Hyperrectangle
 end
 
 """
-    overapproximate(X::LazySet{N}, dir::AbstractDirections{N})::HPolytope{N}
-        where {N}
+    overapproximate(X::LazySet{N}, dir::AbstractDirections{N}) where {N}
 
 Overapproximating a set with template directions.
 
@@ -302,14 +301,15 @@ Overapproximating a set with template directions.
 
 ### Output
 
-An `HPolytope` overapproximating the set `X` with the directions from `dir`.
+A polyhedron overapproximating the set `X` with the directions from `dir`.
+If the directions are known to be bounded, the result is an `HPolytope`,
+otherwise the result is an `HPolyhedron`.
 """
-function overapproximate(X::LazySet{N},
-                         dir::AbstractDirections{N}
-                        )::HPolytope{N} where {N}
+function overapproximate(X::LazySet{N}, dir::AbstractDirections{N}) where {N}
     halfspaces = Vector{LinearConstraint{N}}()
     sizehint!(halfspaces, length(dir))
-    H = HPolytope(halfspaces)
+    T = isbounded(dir) ? HPolytope : HPolyhedron
+    H = T(halfspaces)
     for d in dir
         addconstraint!(H, LinearConstraint(d, ρ(d, X)))
     end
@@ -317,8 +317,7 @@ function overapproximate(X::LazySet{N},
 end
 
 """
-    overapproximate(X::LazySet{N},
-                    dir::Type{<:AbstractDirections})::HPolytope{N} where {N}
+    overapproximate(X::LazySet{N}, dir::Type{<:AbstractDirections}) where {N}
 
 Overapproximating a set with template directions.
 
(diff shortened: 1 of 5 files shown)


mforets:
To summarize what i've suggested in https://github.com/JuliaReach/LazySets.jl/issues/1292#issuecomment-485106888 is to not check for boundedness, that is to not add `isbounded` here. The argument is that directions are *always* unbounded.

OTOH, everything in `CustomDirections{N}` looks good to me (i would suggest to remove the `isbounded::Bool=determine_boundedness(directions)` check though).

schillic:
`isbounded` is essential for custom directions.

schillic:
Any comments here? I would like to merge it.

src/Approximations/overapproximate.jl
```diff
@@ -300,23 +299,23 @@ Overapproximating a set with template directions.
...
+otherwise the result is an `HPolyhedron`.
 """
-function overapproximate(X::LazySet{N},
-                         dir::AbstractDirections{N}
-                        )::HPolytope{N} where {N}
+function overapproximate(X::LazySet{N}, dir::AbstractDirections{N}) where {N}
     halfspaces = Vector{LinearConstraint{N}}()
     sizehint!(halfspaces, length(dir))
-    H = HPolytope(halfspaces)
+    T = isbounding(dir) ? HPolytope : HPolyhedron
```
mforets: I thought if this check should rather be done through dispatch on the type of`dir`, but i don't know if this choice impacts performance in a considerable way.

src/Approximations/template_directions.jl
```diff
@@ -30,6 +30,35 @@ function dim(ad::AbstractDirections)::Int
     return ad.n
 end
 
+"""
+    isbounding(::AbstractDirections)::Bool
```
mforets:
```suggestion
    isbounding(ad::AbstractDirections)::Bool
```

src/Approximations/template_directions.jl
```diff
@@ -374,8 +407,80 @@ SphericalDirections(Nθ::Int, Nφ::Int) = SphericalDirections{Float64}(Nθ, Nφ)
...
+                                               "need a specified dimension"))
+    return length(directions[1])
+end
+
+function determine_boundedness(directions::AbstractVector{<:AbstractVector{N}}
+                              ) where {N}
+    isempty(directions) && return false
+
+    # determine boundedness from a trial with a BallInf
+    set = BallInf(zeros(N, length(directions[1])), N(1))
```
mforets: What if we check boundedness of the polyhedron whose defining halfspaces have normal directions `directions` and the displacement from the origin is 1:

```julia
n = length(first(directions))
isbounded(HPolyhedron(vcat(directions'...), ones(n)))
```

src/Approximations/template_directions.jl
```diff
@@ -374,8 +407,80 @@ SphericalDirections(Nθ::Int, Nφ::Int) = SphericalDirections{Float64}(Nθ, Nφ)
...
+Since custom directions may not be bounded, boundedness can either be asserted
+in the constructor or it will be determined automatically by performing an
+experiment (more precisely, we overapproximate the unit ball in the infinity
+norm using the given directions).
+
+The dimension will also be determined automatically, unless the empty vector is
+passed (in which case the optional argument `n` needs to be specified).
+"""
+struct CustomDirections{N} <: AbstractDirections{N}
+    directions::AbstractVector{<:AbstractVector{N}}
```
mforets:
```suggestion
    directions::Vector{AV}
```
where `AV<:AbstractVector{N}` is a new type parameter

src/Approximations/template_directions.jl
```diff
@@ -374,8 +407,80 @@ SphericalDirections(Nθ::Int, Nφ::Int) = SphericalDirections{Float64}(Nθ, Nφ)
...
+                              ) where {N}
+    isempty(directions) && return false
+
+    # determine boundedness from a trial with a BallInf
+    set = BallInf(zeros(N, length(directions[1])), N(1))
+    dirs = CustomDirections(directions; isbounding=false)
+    return isbounded(overapproximate(set, dirs))
+end
+
+Base.eltype(::Type{CustomDirections{N}}) where {N} = AbstractVector{N}
```
mforets: return `AV` from the type

src/Approximations/template_directions.jl
schillic: There may be applications where the directions are created dynamically, so a restriction to `Vector` may not be good.

src/Approximations/overapproximate.jl
schillic: This should be in the order of milliseconds. Maybe the compiler can even add this information statically.

src/Approximations/template_directions.jl
schillic: We decided to do the change. It can be changed again later if the need arises.

src/Approximations/overapproximate.jl
schillic: FYI, this line has been removed with the new changes.

PR #1299, opened 2019-04-22
Better error msg for inclusion and membership

### LINKED ISSUE #1296: Better error message for set inclusion usage
Similar in spirit to https://github.com/JuliaReach/LazySets.jl/pull/1282, we can add a better error message to the (incorrect) usage:

```julia
julia> Z = rand(Zonotope);

julia> center(Z) ⊆ Z  # explain saying why this is wrong and show the correct usage
ERROR: MethodError: no method matching length(::Zonotope{Float64})
Closest candidates are:
  length(::Core.SimpleVector) at essentials.jl:561
  length(::Base.MethodList) at reflection.jl:801
  length(::Core.MethodTable) at reflection.jl:875
  ...
Stacktrace:
 [1] issubset(::Array{Float64,1}, ::Zonotope{Float64}) at ./abstractset.jl:234
 [2] top-level scope at none:0
```

mforets:
Closes #1296.

### DIFF
--- a/src/AbstractSingleton.jl
+++ b/src/AbstractSingleton.jl
@@ -259,5 +259,5 @@ end
 function ∈(S::AbstractSingleton{N}, X::LazySet{N})::Bool where {N<:Real}
     error("cannot make a point-in-set check if the left-hand side is a set; either " *
           "check for set inclusion, as in `S ⊆ X`, or check for membership, " *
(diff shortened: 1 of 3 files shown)


mforets:
Thanks for the review! I was planning to add a unit test for the new function but i wasn't sure where to add it.

PR #1303, opened 2019-04-22
#1272 - Use eps by default in plots

### LINKED ISSUE #575: Add optional ε parameter to certain plot recipes
Currently `plot(Interval(1., 2.), 1e-2)` fails because it falls back to the iterative refinement recipe, which requires a 2D set.
I suggest to add an optional ε parameter which is just ignored.

We should probably also do this for other plot recipes, such as the polytopic sets (see #574), if they do not get more precise with smaller ε.

### LINKED ISSUE #578: Generalize numeric type of ε in plot recipes
Currently we have `ε::Float64`. I would change it to `ε::AbstractFloat` or even `ε::Real`.
This currently prevents plotting of sets with numeric type different from `Float64` (because `overapproximate` expects the same numeric type).

### LINKED ISSUE #1272: Let plot use epsilon-close approximation by default
`plot(X)` should become `plot(X, 1e-2)` by default.

mforets:
Closes #1272.
Closes #578.
Closes #575.

With this change, the manual has to be updated accordingly (in several places passing the epsilon explicitly is no longer needed). We can do this in a follow-up PR -- this is now https://github.com/JuliaReach/LazySets.jl/issues/1321.

### DIFF
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -59,10 +59,9 @@ diameter(::LazySet, ::Real=Inf)
 isbounded(::LazySet)
 isbounded_unit_dimensions(::LazySet{N}) where {N<:Real}
 an_element(::LazySet{N}) where {N<:Real}
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::LazySet)
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::Vector{S}) where {S<:LazySet}
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::LazySet, ::Float64)
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::Vector{S}, ::Float64) where {S<:LazySet}
+RecipesBase.apply_recipe(∅::EmptySet{Float64}, ε::Float64=0.0)
+RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::LazySet{Float64}, ε::Float64=1e-3)
+RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::Vector{LazySet{Float64}}, ε::Float64=1e-3)
 tosimplehrep(::LazySet)
 isuniversal(::LazySet{N}, ::Bool=false) where {N<:Real}
 ```
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -59,11 +59,15 @@ diameter(::LazySet, ::Real=Inf)
 isbounded(::LazySet)
 isbounded_unit_dimensions(::LazySet{N}) where {N<:Real}
 an_element(::LazySet{N}) where {N<:Real}
-RecipesBase.apply_recipe(∅::EmptySet{Float64}, ε::Float64=0.0)
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::LazySet{Float64}, ε::Float64=1e-3)
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::Vector{LazySet{Float64}}, ε::Float64=1e-3)
 tosimplehrep(::LazySet)
 isuniversal(::LazySet{N}, ::Bool=false) where {N<:Real}
+```
+
+The following functions dispatch in a general `LazySet` and work provided that the overapproximation using iterative refinement in two dimensions is available:
+
+```@docs
+RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::LazySet{N}, ::N=N(1e-3)) where {N<:Real}
+RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::Vector{XN}, ::N=N(1e-3)) where {N<:Real, XN<:LazySet{N}}
 ```
 
 ### Set functions that override Base functions
@@ -134,8 +138,13 @@ This interface defines the following functions:
 isbounded(::AbstractPolytope)
 singleton_list(::AbstractPolytope{N}) where {N<:Real}
 isempty(::AbstractPolytope)
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::AbstractPolytope)
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::Vector{S}) where {S<:AbstractPolytope}
+```
+
+Plotting abstract polytopes is available too:
+
+```@docs
+RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::AbstractPolytope{N}, ::N=zero(N)) where {N<:Real}
+RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::Vector{PN}, ::N=zero(N)) where {N<:Real, PN<:AbstractPolytope{N}}
 ```
 
 #### Polygon
@@ -243,6 +252,6 @@ high(::AbstractSingleton{N}, ::Int) where {N<:Real}
 low(::AbstractSingleton{N}) where {N<:Real}
 low(::AbstractSingleton{N}, ::Int) where {N<:Real}
 linear_map(::AbstractMatrix{N}, ::AbstractSingleton{N}) where {N<:Real}
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::AbstractSingleton)
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::Vector{S}) where {S<:AbstractSingleton}
+RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::AbstractSingleton{N}, ::N=zero(N)) where {N<:Real}
+RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::Vector{SN}, ::N=zero(N)) where {N<:Real, SN<:AbstractSingleton{N}}
 ```
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -63,7 +63,7 @@ tosimplehrep(::LazySet)
 isuniversal(::LazySet{N}, ::Bool=false) where {N<:Real}
 ```
 
-The following functions dispatch in a general `LazySet` and work provided that the overapproximation using iterative refinement in two dimensions is available:
+The following functions work with general two-dimensional `LazySet`s, provided that the overapproximation using iterative refinement is available:
 
 ```@docs
 RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::LazySet{N}, ::N=N(1e-3)) where {N<:Real}
--- a/src/plot_recipes.jl
+++ b/src/plot_recipes.jl
@@ -617,7 +617,7 @@ the iterative refinement is not available for lazy intersections, since it uses
 the support vector (but see #1187).
 
 Also note that if the set is a *nested* intersection, e.g., the lazy linear map,
-a nested intersection you may have to manually overapproximate this set before
+you may have to manually overapproximate this set before
 plotting, see `LazySets.Approximations.overapproximate` for details.
 
 ```julia
--- a/src/plot_recipes.jl
+++ b/src/plot_recipes.jl
@@ -515,14 +515,14 @@ end
 # ==============================
 
 """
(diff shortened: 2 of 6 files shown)


mforets:
> Why remove the tests for Rational{Int} and Float32?

I was planning to put `Float32` back after generalizing the epsilon threshold to `N`. Note, however that the tests don't pass now, because of missing docstrings (how to fix this?).

About `Rational{Int}` i'm not sure that it even makes sense to apply iterative refinement, since 2-norms etc are used in the algorithm etc (one would have to check carefully that the operations make sense and take overapproximations, like overapproximation of those norms using rational, etc.). As a consequence, even for "simple" lazy sets the refinement fails if you pass rationals:

```julia
# the test below also fails if you take X = Ball1([0//1, 0//1], 1//1), but if you take vertices_list(X) works 
# so it can be plotted

julia> X = Diagonal([1//1, 1//1]) * Ball1([0//1, 0//1], 1//1)
LinearMap{Rational{Int64},Ball1{Rational{Int64}},Rational{Int64},Diagonal{Rational{Int64},Array{Rational{Int64},1}}}(Rational{Int64}[1//1 0//1; 0//1 1//1], Ball1{Rational{Int64}}(Rational{Int64}[0//1, 0//1], 1//1))

julia> overapproximate(X, 1//10)
ERROR: MethodError: no method matching σ(::Array{Float64,1}, ::LinearMap{Rational{Int64},Ball1{Rational{Int64}},Rational{Int64},Diagonal{Rational{Int64},Array{Rational{Int64},1}}})
Closest candidates are:
  σ(::AbstractArray{N<:Real,1}, ::HalfSpace{N<:Real}) where N<:Real at /Users/forets/.julia/dev/LazySets/src/HalfSpace.jl:110
  σ(::AbstractArray{N<:Real,1}, ::ZeroSet{N<:Real}) where N<:Real at /Users/forets/.julia/dev/LazySets/src/ZeroSet.jl:98
  σ(::AbstractArray{N<:Real,1}, ::AbstractSingleton{N<:Real}) where N<:Real at /Users/forets/.julia/dev/LazySets/src/AbstractSingleton.jl:231
  ...
Stacktrace:
 [1] new_approx(::LinearMap{Rational{Int64},Ball1{Rational{Int64}},Rational{Int64},Diagonal{Rational{Int64},Array{Rational{Int64},1}}}, ::Array{Rational{Int64},1}, ::Array{Rational{Int64},1}, ::Array{Rational{Int64},1}, ::Array{Rational{Int64},1}) at /Users/forets/.julia/dev/LazySets/src/Approximations/iterative_refinement.jl:95
 [2] addapproximation!(::LazySets.Approximations.PolygonalOverapproximation{Rational{Int64}}, ::Array{Rational{Int64},1}, ::Array{Rational{Int64},1}, ::Array{Rational{Int64},1}, ::Array{Rational{Int64},1}) at /Users/forets/.julia/dev/LazySets/src/Approximations/iterative_refinement.jl:124
 [3] approximate(::LinearMap{Rational{Int64},Ball1{Rational{Int64}},Rational{Int64},Diagonal{Rational{Int64},Array{Rational{Int64},1}}}, ::Rational{Int64}) at /Users/forets/.julia/dev/LazySets/src/Approximations/iterative_refinement.jl:219
 [4] overapproximate(::LinearMap{Rational{Int64},Ball1{Rational{Int64}},Rational{Int64},Diagonal{Rational{Int64},Array{Rational{Int64},1}}}, ::Type{HPolygon}, ::Rational{Int64}) at /Users/forets/.julia/dev/LazySets/src/Approximations/overapproximate.jl:51
 [5] overapproximate(::LinearMap{Rational{Int64},Ball1{Rational{Int64}},Rational{Int64},Diagonal{Rational{Int64},Array{Rational{Int64},1}}}, ::Rational{Int64}) at /Users/forets/.julia/dev/LazySets/src/Approximations/overapproximate.jl:63
 [6] top-level scope at none:0
```

schillic:
> About Rational{Int} i'm not sure that it even makes sense to apply iterative refinement

No, but we can just move these tests to the other loop without `Rational{Int}`. We can still plot `AbstractPolytope`s as before.

src/plot_recipes.jl
```diff
@@ -595,7 +535,52 @@ julia> plot(∅, 1e-2);
...
+    push!(y, vlist[1, 2])
+
+    seriestype := norm(vlist[1, :] - vlist[2, :]) ≈ 0 ? :scatter : :shape
+
+    x, y
+end
+
+#@recipe function plot_intersection(::Intersection{N}, ε::N=N(0.0)) where {N<:Real}
+#    error("cannot plot a lazy intersection using iterative refinement " *
+#         "(the exact support vector of an intersection is not implemented)")
```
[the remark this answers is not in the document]
mforets: I was planning to keep this error message to give a better explanation why `plot(::Intersection, ::eps)` fails and say that `plot(::Intersection, ::Int)` works.

mforets:
Some of those tests in `Rational{Int}` of the outer-most loop worked just because the previous code was creating box overapproximations by default, since `overapproximate(X, Inf)` in the previous example *does* work if the coeffs of `X` are rationals, but it breaks whenever you want to refine the box.

schillic:
That is why I said we can move those tests to the other loop.

mforets:
In `unit_plot.jl`, we can now remove the tests with a custom epsilon, since iterative refinement is used anyways if you `plot(X)`.

schillic:
We could still test whether passing the value works, but I agree.

docs/src/lib/interfaces.md
```diff
@@ -59,14 +59,17 @@ diameter(::LazySet, ::Real=Inf)
...
 an_element(::LazySet{N}) where {N<:Real}
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::LazySet)
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::Vector{S}) where {S<:LazySet}
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::LazySet, ::Float64)
-RecipesBase.apply_recipe(::Dict{Symbol,Any}, ::Vector{S}, ::Float64) where {S<:LazySet}
 tosimplehrep(::LazySet)
 isuniversal(::LazySet{N}, ::Bool=false) where {N<:Real}
 ```
 
+The following functions dispatch in a general `LazySet` and work provided that the overapproximation using iterative refinement in two dimensions is available:
```
schillic:
```suggestion
The following functions work with general two-dimensional `LazySet`s, provided that the overapproximation using iterative refinement is available:
```

test/unit_plot.jl
```diff
@@ -88,108 +114,41 @@ for N in [Float64, Rational{Int}, Float32]
...
+    @test_throws ErrorException plot(itsa) # TODO not implemented yet
 end
 
-# set types that do not work with Rational{Int}
-for N in [Float64, Float32]
-    v0 = zeros(N, 2)
+# -----------------------------------------------------------------
+# tests for rationals; iterative refinement is not available
+# -----------------------------------------------------------------
+for N in [Rational{Int}]
```
schillic: I would rather merge this loop with the one above. Simply add all `Rational{Int}` tests to the top and then `continue` the loop if `N != Rational{Int}`.

src/plot_recipes.jl
```diff
@@ -68,144 +89,59 @@ plotting singletons.
...
-- `ε`  -- approximation error bound
-
-### Examples
-
-```jldoctest
-julia> using Plots, LazySets;
+An iterative refinement method is applied to obtain an overapproximation of each
+set in `X` in constraint representation, which is then plotted. To change the
+default tolerance for the iterative refinement, use the second argument; it applies
+all sets in the array.
```
schillic:
```suggestion
to all sets in the array.
```

src/plot_recipes.jl
```diff
@@ -227,39 +163,43 @@ end
...
 """
-    plot_polygon(P::AbstractPolytope; ...)
+    plot_polytope(P::AbstractPolytope, [ε]::N=zero(N); ...) where {N<:Real}
 
 Plot a 2D polytope as the convex hull of its vertices.
 
 ### Input
 
 - `P` -- polygon or polytope
+- `ε` -- (optional, default: `0.0`) ignored, used for dispatch
```
schillic:
```suggestion
- `ε` -- (optional, default: `0`) ignored, used for dispatch
```
several times in this file

src/plot_recipes.jl
```diff
@@ -575,27 +495,182 @@ end
...
+### Input
 
-julia> plot(∅);
+- `U` -- universal set
+- `ε` -- (optional, default: `0`) ignored, used for dispatch
 
-julia> plot(∅, 1e-2);
+### Output
+
+An error; plotting the universal set is not implemented.
```
schillic:
```suggestion
An error, since plotting the universal set is not implemented yet (see #576).
```

src/plot_recipes.jl
```diff
@@ -575,27 +495,182 @@ end
...
+- `Nφ` -- (optional, default: `40`) number of template directions used in the
+          template overapproximation
+
+### Output
+
+A plot with the overapproximation of the given lazy intersection.
+
+### Notes
+
+This function is separated from the main one `LazySet` plot recipe because
```
schillic:
```suggestion
This function is separated from the main `LazySet` plot recipe because
```

src/plot_recipes.jl
```diff
@@ -575,27 +495,182 @@ end
...
+the iterative refinement is not available for lazy intersections, since it uses
+the support vector (but see #1187).
+
+Also note that if the set is a *nested* intersection eg. the lazy linear map
```
schillic:
```suggestion
Also note that if the set is a *nested* intersection, e.g., the lazy linear map,
```

src/plot_recipes.jl
```diff
@@ -575,27 +495,182 @@ end
...
+a nested intersection you may have to manually overapproximate this set before
```
schillic:
```suggestion
you may have to manually overapproximate this set before
```

src/plot_recipes.jl
```diff
@@ -575,27 +495,182 @@ end
...
+plotting, see `LazySets.Approximations.overapproximate` for details.
```
schillic:
```suggestion
plotting (see `LazySets.Approximations.overapproximate` for details).
```

src/plot_recipes.jl
```diff
@@ -575,27 +495,182 @@ end
...
+julia> using Polyhedra, LazySets.Approximations
+
+julia> X = rand(Ball2) ∩ rand(Ball2); # lazy intersection
+
+julia> plot(X)
+```
+
+The vertices list of an `HPolygon` has known issues (until #1306 is fixed).
+Consider using the `Polyhedra` backend to compute the dual representation.
+One can specify the accuaracy of the overapproximation of the lazy intersection
```
schillic:
```suggestion
One can specify the accuracy of the overapproximation of the lazy intersection
```

src/plot_recipes.jl
```diff
@@ -575,27 +495,182 @@ end
...
+passing a higher value in `Nφ`, which stands for amount of polar directions chosen.
```
schillic:
```suggestion
passing a higher value in `Nφ`, which stands for the number of polar directions chosen.
```

src/plot_recipes.jl
```diff
@@ -575,27 +495,182 @@ end
...
-    return []
+@recipe function plot_intersection(X::Intersection{N}, ε::N=-one(N); Nφ=40,
+                                   color="blue", label="", grid=true, alpha=0.5) where {N<:Real}
+
+    @assert dim(X) == 2 "this recipe only plots two-dimensional sets"
+
+    if ε != -one(N)
+        error("cannot plot a lazy intersection using iterative refinement with " *
+              "error threshold `ε = $ε`, because the exact support vector of an " *
+              "intersection is not available; using instead a set of `Nφ` " *
```
schillic:
```suggestion
              "intersection is currently not available; using instead a set of `Nφ` " *
```

test/unit_plot.jl
mforets: Good, i have merged them, it does simplify the code! Tests pass locally.

Thanks a lot for the review.

PR #1304, opened 2019-04-22
#1300 - Set plot defaults properly

### LINKED ISSUE #703: Revise defaults in plot recipes
In 2D plots i use the options `aspectratio=1, alpha=.5` so often that i wonder if they should be defaults? 

`alpha=.5` is an example but it could be any number `<1`, this option is particularly useful for plotting sets on top of each other.

### LINKED ISSUE #1300: Set plot defaults properly
Inside plot recipes, one has to use the full names (no aliases). And in `seriestype := :shape` some options are ignored (see Plots#1639).

As a workaround such that we can properly set defaults (eg. color, alpha) we can use `-->` with the full names and use the aliases as kwargs, as in:

```julia
 mutable struct UnitSquare end

 @recipe function f(::UnitSquare, color=:lightblue, alpha=.5)
    
     seriescolor --> color   #  color --> :lightblue   would be ignored
     label --> ""
     seriesalpha --> alpha  # same here
     seriestype := :shape
 
     vlist = transpose(hcat([[0.0, 0.0], [1.0, 0.0], [1.0, 1.0], [0.0, 1.0]]...))
     (x, y) = vlist[:, 1], vlist[:, 2]
 
     # add first vertex to "close" the polygon
     push!(x, vlist[1, 1])
     push!(y, vlist[1, 2])
 
     x, y 
end
```

mforets:
Closes #1300.
Closes #703.

Requires #1272.

### DIFF
--- a/src/plot_recipes.jl
+++ b/src/plot_recipes.jl
@@ -65,9 +65,15 @@ returned by `vertices_list` are the same. In that case, a scatter plot is used
 maps of singletons.
 """
 @recipe function plot_lazyset(X::LazySet{N}, ε::N=N(1e-3);
-                              color="blue", label="", grid=true, alpha=0.5) where {N<:Real}
+                              color=:auto, alpha=0.5) where {N<:Real}
 
     @assert dim(X) == 2 "cannot plot a $(dim(X))-dimensional set using iterative refinement"
+
+    seriescolor --> color
+    seriesalpha --> alpha
+    label --> ""
+    grid --> true
+    aspect_ratio --> 1.0
 
     P = overapproximate(X, ε)
     vlist = transpose(hcat(convex_hull(vertices_list(P))...))
@@ -129,10 +135,14 @@ Refer to the documentation of `plot_lazyset(S::LazySet, [ε]::Float64; ...)` for
 further details.
 """
 @recipe function plot_lazysets(X::Vector{XN}, ε::N=N(1e-3);
-                               seriescolor="blue", label="", grid=true,
-                               alpha=0.5) where {N<:Real, XN<:LazySet{N}}
+                               color=:auto, alpha=0.5) where {N<:Real, XN<:LazySet{N}}
 
     seriestype := :shape
+    seriescolor --> color
+    seriesalpha --> alpha
+    label --> ""
+    grid --> true
+    aspect_ratio --> 1.0
 
     for Xi in X
         if Xi isa EmptySet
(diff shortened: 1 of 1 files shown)


src/plot_recipes.jl
```diff
@@ -10,45 +12,72 @@ end
...
+
+julia> plot(B, 1e-2); # faster than the previous algorithm, but less accurate
 ```
 
 ### Algorithm
 
-For any 2D lazy set we compute its box overapproximation, followed by the list of
-vertices. A post-processing `convex_hull` is applied to the vertices list;
-this ensures that the shaded area inside the convex hull of the vertices is covered
+In first stage, an overapproximation of the given set to a polygon in constraint
```
schillic:
```suggestion
In the first stage, an overapproximation of the given set to a polygon in constraint
```

src/plot_recipes.jl
```diff
@@ -10,45 +12,72 @@ end
...
+that the shaded area inside the convex hull of the vertices is covered
 correctly.
 
 ### Notes
 
-This recipe detects if the axis-aligned approximation is such that the first two
-vertices returned by `vertices_list` are the same. In that case, a scatter plot
-is used (instead of a shape plot). This use case arises, for example, when
-plotting singletons.
+This recipe detects if overapproximation is such that the first two vertices
```
schillic:
```suggestion
This recipe detects if the overapproximation is such that the first two vertices
```

src/plot_recipes.jl
```diff
@@ -10,45 +12,72 @@ end
...
+returned by `vertices_list` are the same. In that case, a scatter plot is used
+(instead of a shape plot). This corner case arises, for example, when lazy linear
+maps of singletons.
```
schillic:
```suggestion
maps of singletons are plotted.
```

src/plot_recipes.jl
```diff
@@ -445,23 +418,43 @@ Plot a line segment.
...
+control the color of the endpoints, use the `markercolor` keyword argument.
+To control the width, use `linewidth`.
+
+```jldoctest plot_linesegment
+julia> plot(L, markercolor="green", linecolor="red", linewidth=2.);
+
+```
+
+The option `add_marker` (optional, default: `true`) can be used to specify if
+endpoints markers should be plotted or not.
```
schillic:
```suggestion
endpoint markers should be plotted or not.
```

mforets:
> I am not sure if doctests are useful here. We already have tests for plots.

Good point. I will remove `jldoctests` keeping the examples. 

This branch still needs to be rebased after we merge https://github.com/JuliaReach/LazySets.jl/pull/1303, which turned out to raise many problems (fixes on the way).

PR #1323, opened 2019-05-01
#1315 - Faster disjointess check for intervals

### LINKED ISSUE #1315: Faster disjointness check for intervals
The disjointness check for 1D intervals uses the `AbstractHyperrectangle` implementation (check if the difference of center is offset wrt the sum of radius). But this is suboptimal for intervals, since it's enough to make at most two comparison operations using the endpoints.
Using this approach gives ~1000x speedup.

```julia
julia> I1 = rand(Interval)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([-0.200441, 1.10675])

julia> I2  = rand(Interval)
 Interval{Float64,IntervalArithmetic.Interval{Float64}}([-1.08952, -0.122091])
```

In `master`:

```julia
@btime is_intersection_empty($I1, $I2)
  119.789 ns (3 allocations: 288 bytes)

false

@btime is_intersection_empty($I1, $I2, false)
  119.087 ns (3 allocations: 288 bytes)

false

@btime is_intersection_empty($I1, $I2, true)
  221.764 ns (7 allocations: 608 bytes)

(false, [-0.605804])
```

Using endpoints comparison:

```julia
function is_intersection_empty_v2(I1::Interval{N}, I2::Interval{N}, witness::Bool=false) where {N<:Real}
    check = min(I2) > max(I1) || min(I1) > max(I2)
    if !witness
        return check
    else
        return (check, N[max(min(I1), min(I2))])
    end
end
```

Then:

```julia
@btime is_intersection_empty_v2($I1, $I2)
  0.028 ns (0 allocations: 0 bytes)

false

@btime is_intersection_empty_v2($I1, $I2, false)
  0.021 ns (0 allocations: 0 bytes)

false

@btime is_intersection_empty_v2($I1, $I2, true)
  39.970 ns (2 allocations: 128 bytes)

(false, [-0.20044])
```

mforets:
Closes #1315.

### DIFF
--- a/src/is_intersection_empty.jl
+++ b/src/is_intersection_empty.jl
@@ -1,6 +1,8 @@
 export is_intersection_empty,
        isdisjoint
 
+# --- disjointness check for lazy sets ---
+
 """
     is_intersection_empty(X::LazySet{N},
                           Y::LazySet{N},
@@ -46,9 +48,16 @@ function is_intersection_empty(X::LazySet{N},
     return empty_intersection
 end
 
+# --- alias ---
+
+"""
+    isdisjoint(X, Y)
+
+An alternative name for `is_intersection_empty(X, Y)`.
+"""
+const isdisjoint = is_intersection_empty
 
 # --- AbstractHyperrectangle ---
-
 
 """
     is_intersection_empty(H1::AbstractHyperrectangle{N},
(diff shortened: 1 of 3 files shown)


test/unit_Interval.jl
```diff
@@ -93,6 +93,12 @@ for N in [Float64, Float32, Rational{Int}]
     E = intersection(A, Interval(N(0), N(1)))
     @test isempty(E)
 
+    # disjointness check
+    @test !isdisjoint(A, B)
+    @test isdisjoint(A, Interval(N(-1), N(1)), false)
+    s, w = isdisjoint(A, A, true)
+    @test s == false && w == [min(A)]
```
schillic: Can we have a nontrivial test with a new interval `C` instead of `A` and `A`?
The concrete return value is an implementation detail and can change, so I would change the test to
```suggestion
    @test s == false && w ∈ A && w ∈ C
```

src/is_intersection_empty.jl
```diff
@@ -117,9 +126,47 @@ function is_intersection_empty(H1::AbstractHyperrectangle{N},
...
+
+* If `witness` option is deactivated: `true` iff ``I1 ∩ I2 = ∅``
+* If `witness` option is activated:
+  * `(true, [])` iff ``I1 ∩ I2 = ∅``
+  * `(false, v)` iff ``I1 ∩ I2 ≠ ∅`` and ``v ∈ I1 ∩ I2``
 
+### Algorithm
+
+``I1 ∩ I2 ≠ ∅`` iff there is a gap between the left-most point of the second
+interval and the left-most point of the first interval, or vice-versa.
```
schillic:
```suggestion
interval and the right-most point of the first interval, or vice-versa.
```

src/is_intersection_empty.jl
```diff
@@ -117,9 +126,47 @@ function is_intersection_empty(H1::AbstractHyperrectangle{N},
...
+A witness is computed by taking the maximum over the left-most points of each
+interval, which is guaranteed to belong to the intersection.
+"""
+function is_intersection_empty(I1::Interval{N}, I2::Interval{N},
+                               witness::Bool=false) where {N<:Real}
+    check = min(I2) > max(I1) || min(I1) > max(I2)
+    if !witness
+        return check
+    else
+        return (check, N[max(min(I1), min(I2))])
```
schillic: The `N` here is not needed. I am not sure if it makes the code more or less efficient.

mforets: True; the elements of this vector are already of type `N`. I ran it and it doesn't seem to affect runtime. I'll remove it though, since it is not needed anyways.

test/unit_Interval.jl
```diff
@@ -93,6 +93,12 @@ for N in [Float64, Float32, Rational{Int}]
     E = intersection(A, Interval(N(0), N(1)))
     @test isempty(E)
 
+    # disjointness check
+    @test !isdisjoint(A, B)
+    @test isdisjoint(A, Interval(N(-1), N(1)), false)
+    s, w = isdisjoint(A, B, true)
+    @test s == false && w ∈ A && w ∈ C
```
```suggestion
    @test s == false && w ∈ A && w ∈ B
```

mforets: Oh i had forgotten to fix that typo, thanks 🕵  :

PR #1330, opened 2019-05-01
#1316 - Set difference between intervals

### LINKED ISSUE #1316: Set difference between intervals
In the 1D case, the set difference `I1 \ I2` between the intervals `I1` and `I2` can be either:

a- the empty set
b- one interval (either corresponding to the left side or to the right side)
c- two intervals (one subinterval corresponding to the left side and another one to the right side)

I thought about this and i'm inclined to make the set difference return a `UnionSetArray` of intervals *both* in b and c. This may simplify extending to two or higher dimensions.
 
See also https://github.com/JuliaReach/LazySets.jl/issues/1305

mforets:
Closes #1316.


src/concrete_difference.jl
```diff
@@ -0,0 +1,69 @@
...
+
+"""
+    difference(I1::Interval{N}, I2::Interval{N}) where {N}
+
+Return the set difference between the given intervals.
+
+The set difference is defined as:
+
+```math
+    I1 \\ I2 = \\{x: x ∈ I₁~\\textrm{and} x ∉ I₂ \\}
```
schillic:
```suggestion
    I1 \\setminus I2 = \\{x: x ∈ I₁ \\text{ and } x ∉ I₂ \\}
```
I am not sure that `\\` displays `\`. I suspect that you need `\\\\` in LaTeX. But `\\setminus` is the better way.

src/concrete_difference.jl
```diff
@@ -0,0 +1,69 @@
...
+
+### Input
+
+- `I1` -- first interval
+- `I2` -- second interval
+
+### Output
+
+- An `EmptySet` if the set difference is empty.
+- A `UnionSetArray` of `Interval` sets otherwise.
```
schillic: I would rather expext one of the following two options:
1. a `UnionSetArray` in all cases
2. an `EmptySet` or an `Interval` or a `Union` of two `Interval`s

test/unit_Interval.jl
```diff
@@ -37,7 +37,7 @@ for N in [Float64, Float32, Rational{Int}]
     v = vertices_list(m)
     @test N[1.5] in v && N[-2] in v
 
-    # difference
+    # substraction
```
```suggestion
    # subtraction
```

src/concrete_difference.jl
mforets: Ok, i thought about this. Maybe the second choice is nicer in the sense that the outputs are somehow "canonical". The drawback is that it can make the code that consumes the new `difference` function a bit trickier, namely the set difference between hyperrectangles. I was planning to write it by considering the set differences along each dimension then taking the cartesian product with the other dimensions (without considering the removed parts).

In sum i vote to keep option 2️⃣

src/concrete_difference.jl
mforets: I haven't checked how does it display (cannot build docs locally, i hit the `Expokit` dependency problem).

Note that the alias that i used is the usual backslash, `\`, not `\setminus[TAB]`. I tried to use `setminus` at first but julia doesn't like it; it says `ERROR: syntax: invalid character "∖" `.

schillic: It is fine to not use `\setminus` in the code, but I prefer it in formulas in the docs.

src/concrete_difference.jl
schillic: I do not think that your CP idea works in general.

schillic:
I am thinking about using `-` instead of `\`. The reason is that `\` is already used for left division.

src/difference.jl
```diff
@@ -0,0 +1,71 @@
...
+        Ileft = Interval(min(I1), min(I12))
+        Iright = Interval(max(I12), max(I1))
+
+        Ileft_iszero = isapproxzero(IntervalArithmetic.diam(Ileft.dat))
+        Iright_iszero = isapproxzero(IntervalArithmetic.diam(Iright.dat))
+
+        if Ileft_iszero && Iright_iszero
+            return EmptySet{N}()
+        elseif Ileft_iszero && !Iright_iszero
+            return Iright
```
schillic: You should never return `Iright`. This is the only set of points that is *always* missing in the difference.

mforets: This case corresponds to the case where the set difference excludes the left part of `I1`, so returning the right part of `I1` when you cut them by their intersection, example:

```julia
julia> difference(Interval(1, 4), Interval(0, 2))
Interval{Float64,IntervalArithmetic.Interval{Float64}}([2, 4])
```

src/difference.jl
```diff
@@ -0,0 +1,71 @@
...
+        return I1
+    else
+        Ileft = Interval(min(I1), min(I12))
+        Iright = Interval(max(I12), max(I1))
+
+        Ileft_iszero = isapproxzero(IntervalArithmetic.diam(Ileft.dat))
+        Iright_iszero = isapproxzero(IntervalArithmetic.diam(Iright.dat))
+
+        if Ileft_iszero && Iright_iszero
+            return EmptySet{N}()
```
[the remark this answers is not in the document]
mforets: The code in this branch gives the correct result here too:

```julia
julia> difference(Interval(1, 1), Interval(2, 2))
Interval{Float64,IntervalArithmetic.Interval{Float64}}([1, 1])
```

src/difference.jl
```diff
@@ -0,0 +1,71 @@
...
+function difference(I1::IN, I2::IN)::Union{EmptySet{N}, IN, UnionSet{N, IN, IN}} where {N, IN<:Interval{N}}
+    I12 = intersection(I1, I2)
+    if isempty(I12)
+        return I1
+    else
+        Ileft = Interval(min(I1), min(I12))
+        Iright = Interval(max(I12), max(I1))
+
+        Ileft_iszero = isapproxzero(IntervalArithmetic.diam(Ileft.dat))
+        Iright_iszero = isapproxzero(IntervalArithmetic.diam(Iright.dat))
```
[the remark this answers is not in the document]
mforets: 👍  we can even outsource `isflat` to a separate function.

mforets:
> I am thinking about using - instead of \. The reason is that \ is already used for left division.

Thanks for raising the issue. Note that `-` is the difference between intervals in the arithmetic sense, so it would be a source of confusion if we use it in LazySets as alias for set diff ([and it is already taken](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Interval.jl#L196)). Example:

```julia
julia> I1 = Interval(0, 1)
julia> I2 = Interval(2, 4)

julia> I1 - I2
Interval{Float64,IntervalArithmetic.Interval{Float64}}([-4, -1])

julia> I1.dat - I2.dat # the above implemented this rule
[-4, -1]
```

On the other hand, it is also true that backslash is used for division (although we don't offer this method in the `Interval` wrapper), example:

```julia
julia> I1.dat \ I2.dat
[2, ∞]
```
 
So i would suggest that we just keep `difference` for now. Unfortunately, set minus `\` doesn't seem to be available as function name. Maybe there is another unicode alias?

src/difference.jl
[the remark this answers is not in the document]
schillic: Okay! I do not know what I was thinking here...

schillic:
> Note that - is the difference between intervals in the arithmetic sense, so it would be a source of confusion if we use it in LazySets as alias for set diff (and it is already taken).

True! So let us keep `\`.

mforets:
> I do not see how your function computes the difference. At least give an explanation in the docs, but I have the feeling that it is just wrong.

I've added an `Algorithm` section. What do you think?

src/difference.jl
```diff
@@ -0,0 +1,100 @@
...
+
+Let ``I₁ = [a, b]`` and ``I₂ = [c, d]`` be intervals. Their set difference is
+``I₁ \\setminus I₂ = \\{x: x ∈ I₁ \\text{ and } x ∉ I₂ \\}`` and depending on their
+position three different results may occur:
+
+- If ``I₁`` and ``I₂`` do not overlap, i.e. if their intersection is empty, then the
+  set difference is just ``I₁``.
+- Otherwise, let `I₁₂ = I₁ ∩ I₂` and assume that it is not empty, then either
+  ``I₁₂`` splits `I₁` into one interval or into two intervals. The latter case
+   happens when the inclusion is strict.
```
schillic:
```suggestion
  happens when the inclusion is strict on both ends of ``I₂``.
```

src/difference.jl
```diff
@@ -0,0 +1,100 @@
...
+
+To check for strict inclusion, we asssume that the inclusion is strict and then
```
schillic:
```suggestion
To check for strict inclusion, we assume that the inclusion is strict and then
```

mforets:
Alright, thanks for your review. I'm adding some docs and moving `isflat` to `Interval.jl` before merging (we could have `isflat` for `AbstractHyperrectangle` too! (It has to check the `radius_hyperrectangle` is flat in any dimension..)

Keeping `\` as an alias for difference then and adding extra docs with the potential overlap with set `\` from `IntervalArithmetic.Interval`.

test/unit_Interval.jl
```diff
@@ -120,4 +120,13 @@ for N in [Float64, Float32, Rational{Int}]
     @test dAB == Interval(N(5), N(6))
     @test dAC == Interval(N(5), N(8))
     @test dAD == UnionSet(Interval(N(5), N(6)), Interval(N(7), N(8)))
+
+    # check if an interval is flat, i.e. if its endpoints coincide (to numerical precision)
+    ztol = LazySets.ABSZTOL(N) # pick up default absolute zero tolerance value
```
schillic: Should this not be a constant value? If we change `ABSZTOL` (see #1341), this test may break.

mforets: Hmm i don't think that it will break for another value, right? Precisely because `isflat(Interval(N(0), x))` holds for all `x <= ztol`, since it uses `isapproxzero`, while it  doesn't hold for bigger ones (eg. `2ztol`).

schillic: I see. That makes this test depend on the implementation, but I do not see a better way :+1:

mforets:
Squashed and merged, thanks a lot for the review!

Issue #335, opened 2018-05-16
Plot recipe for HPolytope/VPolytope in 2D 


mforets:
Proposal: given a `P::HPolytope` make `plot(P)` work and plot exactly by converting it to a `HPolygon`.

This applies to `VPolytope`.

mforets:
This functionality is already supported with 

```julia
@recipe function plot_polytope(P::AbstractPolytope;
                               color="blue", label="", grid=true, alpha=0.5)
```

The caveat is that it require one to load `Polyhedra`, since it calls `vertices_list`. I think we can either keep it as is, or preprocess the `P` with a call to `convert(HPolygon, P))`. But this will be weird/unnecessary if eg. say the `vertices_list(V)` is easy to get. For instance, if you are going to plot a `VPolytope` you don't want to call `convert(HPolygon, P))`

schillic:
This 2D issue is really unnerving. I would like to get this working, best with a general solution.

We could pull the `vertices_list` methods out of the `@require` blocks and add a switch like this (example for `HPolytope`, but the same idea applies for `VPolytope`):
```julia
if dim == 2
    return vertices_list(convert(HPolygon, P))
elseif !polyhedraloaded
    error("you need to load the package Polyhedra for working with HPolytopes")
end
# old code
```
(`polyhedraloaded` is the condition that `Polyhedra` is loaded.)

schillic:
This works now.

PR #1338, opened 2019-05-03
Variable algorithm for vertices list of HPolytope

mforets:
This change makes `vertices_list` work more efficiently with two-dimensional `HPolytope`s by default, making use of the `HPolygon` function.

A simple test shows a speedup of ~30x comparing the version in master and the one in this branch, for a polytope defined by four constraints.

The idea was originally given in: https://github.com/JuliaReach/LazySets.jl/issues/335

---

Requires https://github.com/JuliaReach/LazySets.jl/pull/1342/

### DIFF
--- a/src/HPolytope.jl
+++ b/src/HPolytope.jl
@@ -218,7 +218,7 @@ H-representation and then its `vertices_list` function is used. This ensures
 that, by default, the optimized two-dimensional methods are used.
 
 It is possible to use the `Polyhedra` backend in two-dimensions as well
-by passing a given backend, e.g. `backend=CDDLib.Library()`.
+by passing, e.g. `backend=CDDLib.Library()`.
 
 If the polytope is not two-dimensional, the concrete polyhedra manipulation
 library `Polyhedra` is used. The actual computation is performed by a given
--- a/src/HPolytope.jl
+++ b/src/HPolytope.jl
@@ -217,7 +217,7 @@ If the polytope is two-dimensional, the polytope is converted to a polygon in
 H-representation and then its `vertices_list` function is used. This ensures
 that, by default, the optimized two-dimensional methods are used.
 
-It is possible to use `Polyhedra` backend in two-dimensions as well; do this
+It is possible to use the `Polyhedra` backend in two-dimensions as well
 by passing a given backend, e.g. `backend=CDDLib.Library()`.
 
 If the polytope is not two-dimensional, the concrete polyhedra manipulation
--- a/src/HPolytope.jl
+++ b/src/HPolytope.jl
@@ -198,40 +198,49 @@ end # quote
 end # function load_polyhedra_hpolytope()
 
 """
-    vertices_list(P::HPolytope{N};
-                  [backend]=default_polyhedra_backend(P, N),
-                  [prunefunc]=removevredundancy!)::Vector{Vector{N}} where
-                  {N<:Real}
+    vertices_list(P::HPolytope{N}; [backend]=nothing)::Vector{Vector{N}} where {N<:Real}
 
 Return the list of vertices of a polytope in constraint representation.
 
 ### Input
 
-- `P`         -- polytope in constraint representation
-- `backend`   -- (optional, default: `default_polyhedra_backend(P, N)`)
-                  the polyhedral computations backend
-- `prunefunc` -- (optional, default: `removevredundancy!`) function to
-                 post-process the output of `vreps`
+- `P`       -- polytope in constraint representation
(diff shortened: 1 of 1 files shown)


src/HPolytope.jl
```diff
@@ -198,40 +198,49 @@ end # quote
...
 List of vertices.
 
-### Notes
+### Algorithm
+
+If the polytope is two-dimensional, the polytope is converted to a polygon in
+H-representation and then its `vertices_list` function is used. This ensures
+that, by default, the optimized two-dimensional methods are used.
+
+It is possible to use `Polyhedra` backend in two-dimensions as well; do this
```
schillic:
```suggestion
It is possible to use the `Polyhedra` backend in two-dimensions as well
```

src/HPolytope.jl
```diff
@@ -198,40 +198,49 @@ end # quote
...
+by passing a given backend, e.g. `backend=CDDLib.Library()`.
```
schillic:
```suggestion
by passing, e.g. `backend=CDDLib.Library()`.
```

src/HPolytope.jl
```diff
@@ -198,40 +198,49 @@ end # quote
...
 
+If the polytope is not two-dimensional, the concrete polyhedra manipulation
+library `Polyhedra` is used. The actual computation is performed by a given
+backend; for the default backend used in `LazySets` see `default_polyhedra_backend(N)`.
 For further information on the supported backends see
 [Polyhedra's documentation](https://juliapolyhedra.github.io/Polyhedra.jl/).
 """
-function vertices_list(P::HPolytope{N};
-                       backend=default_polyhedra_backend(P, N),
-                       prunefunc=removevredundancy!
```
schillic: Why remove the `prunefunc` option?

src/HPolytope.jl
```diff
@@ -198,40 +198,49 @@ end # quote
...
+
+    if dim(P) == 2 && backend == nothing
+        return vertices_list(convert(HPolygon, P))
+    else
+        @assert isdefined(@__MODULE__, :Polyhedra) "the function `vertices_list` "
+        "needs the package 'Polyhedra' to be loaded"
+        if backend == nothing
+            backend = default_polyhedra_backend(P, N)
+        end
+        P = polyhedron(P; backend=backend)
```
schillic: Rename to `Q` for clarity.

src/HPolytope.jl
mforets: I realized that this option is useless, since we got the `backend` option to decide which backend to apply. Moreover, `polyhedron(P; backend=backend)` is a `Polyhedra.Polyhedron` and we don't plan to work on those types here (better make a PR on Polyhedra directly).

schillic: > I realized that this option is useless

But now you always prune if `Polyhedra` is used and never prune otherwise, which is inconsistent. Maybe we should have a `prune::Bool` argument instead?

mforets: OK i agree it's a nice idea to add this boolean option to the function. So i'm putting https://github.com/JuliaReach/LazySets.jl/pull/1342/ as a requirement for this PR.

src/HPolytope.jl
```diff
@@ -198,40 +198,54 @@ end # quote
...
+backend; for the default backend used in `LazySets` see `default_polyhedra_backend(N)`.
 For further information on the supported backends see
 [Polyhedra's documentation](https://juliapolyhedra.github.io/Polyhedra.jl/).
 """
 function vertices_list(P::HPolytope{N};
-                       backend=default_polyhedra_backend(P, N),
-                       prunefunc=removevredundancy!
-                      )::Vector{Vector{N}} where {N<:Real}
+                       backend=nothing,
+                       prune=true)::Vector{Vector{N}} where {N<:Real}
```
schillic:
```suggestion
                       prune::Bool=true)::Vector{Vector{N}} where {N<:Real}
```

src/HPolytope.jl
```diff
@@ -198,40 +198,54 @@ end # quote
 end # function load_polyhedra_hpolytope()
 
 """
-    vertices_list(P::HPolytope{N};
-                  [backend]=default_polyhedra_backend(P, N),
-                  [prunefunc]=removevredundancy!)::Vector{Vector{N}} where
-                  {N<:Real}
+    vertices_list(P::HPolytope{N}; [backend]=nothing)::Vector{Vector{N}} where {N<:Real}
```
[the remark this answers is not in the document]
mforets: Arrgghh, thanks for pointing this out.

PR #1341, opened 2019-05-04
Increase default absolute zero tolerance in Floats

mforets:


schillic:
Any reason?

mforets:
Good question and the short answer is that *experiments suggest* that `sqrt(eps(Float64))` is too small for the discrete post. A "typical" calculation with the bouncing ball and octagon directions needs `~1e-8` to take the intersection properly, and since `sqrt(eps(Float64)) = 1.4901161193847656e-8` then it's not that far.

I plan to provide more details soon -- probably after we add the global way to handle the abstol, making it easier for tests in Reachability.

schillic:
Hm, okay.

PR #1342, opened 2019-05-04
Choose to prune=true by default

### LINKED ISSUE #1306: Vertices list of HPolygon from HPolytope conversion is empty
The iterative refinement returns a set that is correct (as can be checked plotting the two sets below; the return type of the first call is an `HPolytope`).

```julia
julia> X = Ball2([-1.17944, 0.555228], 1.3661522186891697) ∩
           Ball2([0.696509, 0.652333], 0.5595649284998497);

julia> dump(X)
Intersection{Float64,Ball2{Float64},Ball2{Float64}}
  X: Ball2{Float64}
    center: Array{Float64}((2,)) [-1.17944, 0.555228]
    radius: Float64 1.3661522186891697
  Y: Ball2{Float64}
    center: Array{Float64}((2,)) [0.696509, 0.652333]
    radius: Float64 0.5595649284998497
  cache: LazySets.IntersectionCache
    isempty: Int8 0

julia> length(vertices_list(overapproximate(X, PolarDirections(40))))
9

julia> length(vertices_list(convert(HPolygon, overapproximate(X, PolarDirections(40)))))
0
```

mforets:
Closes #1306.

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -229,7 +229,7 @@ Convert from 2D polytope in H-representation to polygon in H-representation.
 The 2D polytope represented as polygon.
 """
 function convert(::Type{HPOLYGON}, P::HPolytope{N};
-                prune=true) where {N<:Real, HPOLYGON<:AbstractHPolygon}
+                 prune::Bool=true) where {N<:Real, HPOLYGON<:AbstractHPolygon}
     @assert dim(P) == 2 "polytope must be two-dimensional for conversion"
     H = HPOLYGON{N}()
     for ci in constraints_list(P)
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -429,18 +429,16 @@ Convert from line segment to polygon in H-representation.
 A flat polygon in constraint representation with the minimal number of
 constraints (four).
 """
-function convert(::Type{HPOLYGON}, L::LineSegment{N};
-                 prune=false) where {N<:Real, HPOLYGON<:AbstractHPolygon}
+function convert(::Type{HPOLYGON}, L::LineSegment{N}) where {N<:Real, HPOLYGON<:AbstractHPolygon}
     H = HPOLYGON{N}()
     c = halfspace_left(L.p, L.q)
-    addconstraint!(H, c; prune=prune)
-    addconstraint!(H, LinearConstraint(-c.a, -c.b); prune=prune)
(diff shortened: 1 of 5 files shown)


src/convert.jl
```diff
@@ -233,7 +233,7 @@ function convert(::Type{HPOLYGON},
     H = HPOLYGON{N}()
     for ci in constraints_list(P)
         # guarantee that the edges are correctly sorted for storage
-        addconstraint!(H, ci; prune=false)
+        addconstraint!(H, ci; prune=true)
```
```suggestion
        addconstraint!(H, ci)
```
(`true` is the default.)

mforets: I have now added `prune` as an optional argument, and also updated docs where this argument was missing.

src/convert.jl
```diff
@@ -220,20 +220,20 @@ Convert from 2D polytope in H-representation to polygon in H-representation.
...
+- `prune` -- (optional, default: `true`) flag for removing redundant
+             constraints in the end
 ### Output
 
 The 2D polytope represented as polygon.
 """
-function convert(::Type{HPOLYGON},
-                 P::HPolytope{N}) where {N<:Real, HPOLYGON<:AbstractHPolygon}
+function convert(::Type{HPOLYGON}, P::HPolytope{N};
+                prune=true) where {N<:Real, HPOLYGON<:AbstractHPolygon}
```
```suggestion
                 prune::Bool=true) where {N<:Real, HPOLYGON<:AbstractHPolygon}
```

src/convert.jl
```diff
@@ -420,26 +420,27 @@ Convert from line segment to polygon in H-representation.
...
+             constraints in the end
 ### Output
 
 A flat polygon in constraint representation with the minimal number of
 constraints (four).
 """
-function convert(::Type{HPOLYGON}, L::LineSegment{N}
-                ) where {N<:Real, HPOLYGON<:AbstractHPolygon}
+function convert(::Type{HPOLYGON}, L::LineSegment{N};
+                 prune=false) where {N<:Real, HPOLYGON<:AbstractHPolygon}
```
schillic: Adding the `prune` argument in the conversion of a `LineSegment` does not make sense because there are always two constraints and none of them is ever redundant.

mforets: You are right, ill remove this argument.

PR #1343, opened 2019-05-04
WIP  - #1235 - Use CDDLib as default and update Project.toml

### LINKED ISSUE #1235: Use CDDLib as default
EDIT: See [this comment](https://github.com/JuliaReach/LazySets.jl/issues/1235#issuecomment-473852123) for the updated plan.

When using `Polyhedra`, we should automatically detect (via `Requires`) when `CDDLib` is loaded and use that as the default backend. That backend is more stable and faster.

(Note that in newer versions the constructor changed from `CDDLib.CDDLibrary()` to `CDDLib.Library()`.)

mforets:
Closes #1235.

Marked as WIP because it is still giving a warning after loading Polyhedra, even if i updated the Project.toml file and did `] resolve`  ¯\_(ツ)_/¯ 

```julia
julia> using LazySets
[ Info: Recompiling stale cache file /Users/forets/.julia/compiled/v1.1/LazySets/NjrGc.ji for LazySets [b4f0291d-fe17-52bc-9479-3d1a343d9043]

julia> using Polyhedra
┌ Warning: Package LazySets does not have CDDLib in its dependencies:
│   added CDDLib as a dependency but haven't updated your primary
└ Loading CDDLib into LazySets from project dependency, future warnings for LazySets are suppressed.
```

### DIFF
--- a/src/AbstractPolytope.jl
+++ b/src/AbstractPolytope.jl
@@ -163,7 +163,7 @@ else
                       convexhull,
                       hcartesianproduct, vcartesianproduct,
                       points
-    import CDDLib # default backend
+    import .CDDLib # default backend
 
     # to use Polyhedra's default library, set
     # default_library(LazySets.dim(P), Float64)
--- a/src/AbstractPolytope.jl
+++ b/src/AbstractPolytope.jl
@@ -162,15 +162,19 @@ else
                       intersect,
                       convexhull,
                       hcartesianproduct, vcartesianproduct,
-                      points,
-                      default_library
+                      points
+    import CDDLib # default backend
 
+    # to use Polyhedra's default library, set
+    # default_library(LazySets.dim(P), Float64)
     function default_polyhedra_backend(P, N::Type{<:AbstractFloat})
-        return default_library(LazySets.dim(P), Float64)
+        return CDDLib.Library()
     end
 
+    # to use Polyhedra's default library, set
+    # default_library(LazySets.dim(P), Rational{Int})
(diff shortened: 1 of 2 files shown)


src/AbstractPolytope.jl
```diff
@@ -162,15 +162,19 @@ else
                       intersect,
                       convexhull,
                       hcartesianproduct, vcartesianproduct,
-                      points,
-                      default_library
+                      points
+    import CDDLib # default backend
```
schillic: I do not like this. We should not load an optional package. I would rather update the docs by highly recommending to also load `CDDLib` (and maybe even printing a warning if `Polyhedra` is loaded first). Below we can use `CDDLib` as default depending on whether that package is loaded.

schillic:
```suggestion
    import .CDDLib # default backend
```

schillic:
If you add an easy way to change the backend to the old one, I'm in.

src/AbstractPolytope.jl
schillic: Okay, it does not work like that. New suggestion: We do *not* load `CDDLib` by default, but we set our default backend to it. When the `default_polyhedra_backend` function is called, we check if the backend is `CDDLib`, and if so, we check whether `CDDLib` is loaded. If it is not loaded, we print an error message, saying that the user should either load `CDDLib` or should change the backend actively.

mforets:
See discussion in https://github.com/JuliaReach/LazySets.jl/issues/1488 and https://github.com/JuliaReach/LazySets.jl/pull/1479.

PR #1345, opened 2019-05-04
#1344 - Efficient intersection with intervals

### LINKED ISSUE #1344: Efficient intersection with intervals
An `Interval` is one-dimensional. Intersection with another one-dimensional set can be efficiently evaluated by asking two support-function queries.
For the special cases of `HalfSpace` and `Hyperplane` the intersection can even be evaluated more efficiently. This is a common use case and so we should optimize for it.

schillic:
Closes #1344.

```julia

julia> i = Interval([1., 2.]);

julia> hs = HalfSpace([1.], 1.5);

julia> hp = Hyperplane([-1.], -1.5);

julia> b = Ball1([2.], 0.5);

julia> c = ConvexHull(Singleton([-5.]), Singleton([-1.]));

julia> @btime intersection(i, hs)  # master
  60.737 μs (210 allocations: 14.02 KiB)
HPolytope{Float64}(HalfSpace{Float64}[HalfSpace{Float64}([-1.0], -1.0), HalfSpace{Float64}([1.0], 1.5)])

julia> @which intersection(i, hs)  # uses generic intersection of polyhedra
intersection(P1::AbstractPolyhedron{N}, P2::AbstractPolyhedron{N}) where N<:Real in LazySets at LazySets/src/concrete_intersection.jl:345

julia> @btime intersection(i, hs)  # new
  338.991 ns (9 allocations: 160 bytes)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([1, 1.5])

julia> @btime intersection(i, hp)  # master
  81.807 μs (276 allocations: 18.55 KiB)
HPolytope{Float64}(HalfSpace{Float64}[HalfSpace{Float64}([-1.0], -1.5), HalfSpace{Float64}([1.0], 1.5)])

julia> @which intersection(i, hp)  # uses generic intersection of polyhedra
intersection(P1::AbstractPolyhedron{N}, P2::AbstractPolyhedron{N}) where N<:Real in LazySets at LazySets/src/concrete_intersection.jl:345

julia> @btime intersection(i, hp)  # new
  1.962 μs (5 allocations: 160 bytes)
Singleton{Float64,Array{Float64,1}}([1.5])

julia> @btime intersection(i, b)  # master
  86.117 μs (295 allocations: 19.34 KiB)
HPolytope{Float64}(HalfSpace{Float64}[HalfSpace{Float64}([1.0], 2.0), HalfSpace{Float64}([-1.0], -1.5)])

julia> @btime intersection(i, b)  # new
  309.758 ns (7 allocations: 608 bytes)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([1.5, 2])

julia> @btime intersection(i, c)  # master
ERROR: MethodError: no method matching intersection(::Interval{Float64,IntervalArithmetic.Interval{Float64}}, ::ConvexHull{Float64,Singleton{Float64,Array{Float64,1}},Singleton{Float64,Array{Float64,1}}})

julia> @btime intersection(i, c)  # new
  160.299 ns (2 allocations: 192 bytes)
EmptySet{Float64}()
```

### DIFF
--- a/src/concrete_intersection.jl
+++ b/src/concrete_intersection.jl
@@ -150,9 +150,8 @@ function intersection(H::AbstractHyperrectangle{N},
 end
 
 """
-    intersection(x::Interval{N},
-                 y::Interval{N}
-                 )::Union{Interval{N}, EmptySet{N}} where {N<:Real}
+    intersection(x::Interval{N}, y::Interval{N}
+                )::Union{Interval{N}, EmptySet{N}} where {N<:Real}
 
 Return the intersection of two intervals.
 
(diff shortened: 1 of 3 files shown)


mforets:
This is awesome! 🚀 

I started with the logic of `intersection(::Interval, HalfSpace)` which i agree. The variation considered below gives ~35% faster comparing the tests. Seems to me that it amounts to merging your `far_bound` and `close_bound` checks...?

```julia
function intersection_b(X::Interval{N}, hs::HalfSpace{N})::Union{Interval{N}, EmptySet{N}} where {N<:Real}
    @assert dim(hs) == 1 "cannot take the intersection between an interval " *
                         "and a $(dim(hs))-dimensional half-space"

    lo = min(X); hi = max(X)
    b = hs.b; a = hs.a[1]
    b_over_a = b / a
    
    if _geq(a, zero(N))
        if _leq(b_over_a, lo)
            return EmptySet()
        else
            return Interval(lo, min(b_over_a, hi))
        end
    else
        if _geq(b_over_a, hi)
            return EmptySet()
        else
            return Interval(max(b_over_a, l), hi)
        end
    end
end
```

src/concrete_intersection.jl
```diff
@@ -176,6 +174,148 @@ function intersection(x::Interval{N},
...
+### Output
+
+If the sets do not intersect, the result is the empty set.
+Otherwise the result is the interval that describes the intersection.
+"""
+function intersection(X::Interval{N}, Y::LazySet{N}
+                     )::Union{Interval{N}, EmptySet{N}} where {N<:Real}
+    lower = max(min(X), -ρ(N[-1], Y))
+    upper = min(max(X), ρ(N[1], Y))
+    if lower > upper
```
mforets: How about if we add robustness here too? What i have in mind is:

```julia
if lower < upper
    return Interval(lower, upper)
elseif _isapprox(lower, upper)
    return Singleton([lower])
else
    return EmptySet()
end
```

Otherwise, we can always return a one element interval (for which `isflat` could be true though, but it's not a big deal?),

```julia
if _leq(lower, upper)
    return Interval(lower, upper)
else
    return EmptySet()
end
```

schillic: I do not like to create `Interval`s whose lower bound is above the upper bound, even by a small amount.

schillic:
> `if _geq(a, zero(N))`

Using approximations for a sign check does not make sense to me.

mforets:
> Using approximations for a sign check does not make sense to me.

👍  it should be dropped

src/concrete_intersection.jl
```diff
@@ -176,6 +174,148 @@ function intersection(x::Interval{N},
...
+                     )::Union{Interval{N}, EmptySet{N}} where {N<:Real}
+    return invoke(intersection, Tuple{Interval{N}, LazySet{N}}, X, H)
+end
+function intersection(H::AbstractHyperrectangle{N}, X::Interval{N}
+                     )::Union{Interval{N}, EmptySet{N}} where {N<:Real}
+    return invoke(intersection, Tuple{Interval{N}, LazySet{N}}, X, H)
+end
+function intersection(X::Interval{N}, S::AbstractSingleton{N}
+                     )::Union{Interval{N}, EmptySet{N}} where {N<:Real}
+    return invoke(intersection, Tuple{typeof(S), LazySet{N}}, S, X)
```
```suggestion
    return invoke(intersection, Tuple{Interval{N}, LazySet{N}}, S, X)
```

mforets: To note that there is mismatch between the return type of the function in lines 310-312 (expects to return Interval or EmptySet) and the return type of the function that is invoked (expects Singleton or EmptySet).

schillic: I think the `intersection` method for a `Singleton` is more efficient.
But what is wrong is is the output type (should be `Union{Singleton{N}, EmptySet{N}}`).

src/concrete_intersection.jl
mforets: Ok and how about the first paragraph? Now we got:

```julia
julia> X
Interval{Float64,IntervalArithmetic.Interval{Float64}}([1, 2])

julia> S = Singleton([1.0]);

julia> intersection(X, S ⊕ [1.0 + LazySets.ABSZTOL(Float64)])
EmptySet{Float64}()

julia> intersection(X, S ⊕ [1.0 + 1e-15])
EmptySet{Float64}()

julia> intersection(X, S ⊕ [1.0 + 1e-16])
Interval{Float64,IntervalArithmetic.Interval{Float64}}([2, 2])
```
I think we want that for all values smaller than `LazySets.ABSZTOL(Float64)` it returns the interval.

schillic:
> The variation considered below gives ~35% faster comparing the tests.

I do not understand how the code works.

src/concrete_intersection.jl
schillic: `_isapprox` is only defined for `AbstractFloat`s. Is that on purpose?

schillic:
> The variation considered below gives ~35% faster comparing the tests.

I added a new version following your idea. It is, however, not as small as yours because there were problems with division by zero.

mforets:
[This code](https://github.com/JuliaReach/LazySets.jl/pull/1345#issuecomment-489995193) does divides in ` b_over_a = b / a`; if the normal vector was zero the half-space is either the universal set or the empty set, depending on `b`; i think that if we encounter those case in practice we can just an assertion in the `HalfSpace` constructor.

schillic:
Yes, I cared for that. I would say it can happen in practice that `a` is very close to zero.

src/concrete_intersection.jl
mforets: Yes, `_isapprox` was written for `AbstractFloat`, since that rationals can be compared exactly. The definition of [isapprox](https://github.com/JuliaLang/julia/blob/9a1cc63e7186b50cd080922105ba56b3d1b89b01/base/floatfuncs.jl#L273) in Julia uses `Number`. I agree that for methods that use `_isapprox` generically it make sense that it is defined for `Real`; this will use `isapproxzero(x::Real; kwargs...)` that uses the fallback `x == zero(x)` for everything which is not float.

src/concrete_intersection.jl
```diff
@@ -176,6 +174,177 @@ function intersection(x::Interval{N},
...
+            # half-space is empty
+            return EmptySet{N}()
+        end
+    end
+
+    # idea:
+    # ax ≤ b  <=>  x ≤ b/a  (if a > 0)
+    # ax ≤ b  <=>  x ≥ b/a  (if a < 0)
+    b_over_a = b / a
+
```
mforets: These are missing 👇 

```julia
 lo = min(X); hi = max(X)
```

Issue #1201, opened 2019-03-06
Add a dimension to EmptySet


schillic:
Computations with `EmptySet` may fail because the dimensions do not match (see for example #1097). Another example is that one cannot construct a `Universe` as the complement of an `EmptySet` because one cannot access the dimension (see #1152).
I say that we add a field for the dimension, as we did for `Universe`.

PR #1360, opened 2019-05-12
#1111 - Add AffineMap

### LINKED ISSUE #1111: Add lazy AffineMap operation
An affine map is a linear map `A` followed by a translation `b`, applied to some set `X`, so it represents the set `A X + b`. We already have the `LinearMap`, so we are basically missing the `Translation` (#882), and then we can immediately implement every operation for `AffineMap`s as the composition of the operations for `LinearMap` and `Translation`.

### Example

```julia
σ(d::Vector, am::AffineMap) = σ(d, Translation(am.b, σ(d, LinearMap(am.A, am.X))))
```

For efficiency, we may not want to create these wrapper types all the time and instead outsource the code for those types to some internal helper functions.

mforets:
Closes #1111.

~Requires: https://github.com/JuliaReach/LazySets.jl/issues/1364~

### DIFF
--- a/src/AffineMap.jl
+++ b/src/AffineMap.jl
@@ -1,0 +1,136 @@
+import Base: isempty
+
+export AffineMap,
+       an_element,
+       isempty
+
+"""
+    AffineMap{N<:Real, S<:LazySet{N},
+             NM, MAT<:AbstractMatrix{NM}, VN<:AbstractVector{NM}} <: LazySet{N}
+
+Type that represents an affine transformation ``M⋅X ⊕ b`` of a convex set ``X``.
+
+### Fields
+
+- `M` -- matrix/linear map
+- `b` -- vector
+- `X` -- convex set
+
+### Notes
+
+The affine map is the composition of a linear map and a translation.
+This type is parametric in the coefficients of the linear map, `NM`, which may be
+different from the numeric type of the wrapped set (`N`).
+"""
+struct AffineMap{N<:Real, S<:LazySet{N},
+                 NM, MAT<:AbstractMatrix{NM}, VN<:AbstractVector{NM}} <: LazySet{N}
+    M::MAT
+    b::VN
+    X::S
+
+    # default constructor with dimension match check
+    function AffineMap{N, S, NM, MAT, VN}(M::MAT, b::VN, X::S) where {N<:Real,
+                        S<:LazySet{N}, NM, MAT<:AbstractMatrix{NM}, VN}
+
+        @assert dim(X) == size(M, 2) "a matrix of size $(size(M)) cannot be "
+            "applied to a set of dimension $(dim(X))"
+
+        @assert size(M, 1) == length(b) "the output dimension of the map, $(size(M, 2)), "
+            "is incompatible with the dimension of the translation vector, $(dim(b))"
+
+        return new{N, S, NM, MAT, VN}(M, b, X)
+    end
+end
+
+# ============================ 
+# LazySet interface functions
+# ============================
+
+"""
+    dim(am::AffineMap)::Int
+
+Return the dimension of an affine map.
+
+### Input
+
+- `am` -- affine map
+
+### Output
+
+The dimension of an affine map.
+"""
+function dim(am::AffineMap)::Int
+    return length(am.b)
+end
+
+
+"""
+    σ(d::AbstractVector{N}, am::AffineMap{N}) where {N<:Real}
+
+Return the support vector of an affine map.
+
+### Input
+
+- `d`  -- direction
+- `am` -- affine map
+
+### Output
+
+The support vector in the given direction.
+"""
+function σ(d::AbstractVector{N}, am::AffineMap{N}) where {N<:Real}
+    return am.M * σ(_At_mul_B(am.M, d), am.X) + am.b
+end
+
+"""
+    ρ(d::AbstractVector{N}, am::AffineMap{N}) where {N<:Real}
+
+Return the support function of an affine map.
+
+### Input
+
+- `d`  -- direction
+- `am` -- affine map
+
+### Output
+
+The support function in the given direction.
+"""
+function ρ(d::AbstractVector{N}, am::AffineMap{N}) where {N<:Real}
+    return ρ(_At_mul_B(am.M, d), am.X) + dot(d, am.b)
+end
+
+"""
+    an_element(am::AffineMap)
+
+Return some element of an affine map.
+
+### Input
+
+- `am` -- affine map
+
+### Output
+
+An element of the affine map. It relies on the `an_element` function of the
+wrapped set.
+"""
+function an_element(am::AffineMap)
+    return am.M * an_element(am.X) + am.b
+end
+
+"""
+    isempty(am::AffineMap)::Bool
+
+Return whether an affine map is empty or not.
+
+### Input
+
+- `am` -- affine map
+
+### Output
+
+`true` iff the wrapped set is empty and the affine vector is empty.
+"""
+function isempty(am::AffineMap)::Bool
+    return isempty(am.X) && isempty(am.b)
+end
--- a/src/AffineMap.jl
+++ b/src/AffineMap.jl
@@ -11,15 +11,15 @@ export AffineMap,
 
 """
     AffineMap{N<:Real, S<:LazySet{N}, NM, MAT<:AbstractMatrix{NM},
-             VN<:AbstractVector{NM}} <: LazySet{N}
+              VN<:AbstractVector{NM}} <: LazySet{N}
 
-Type that represents an affine transformation ``M⋅X ⊕ b`` of a convex set ``X``.
+Type that represents an affine transformation ``M⋅X ⊕ v`` of a convex set ``X``.
 
 ### Fields
 
 - `M` -- matrix/linear map
-- `b` -- vector
 - `X` -- convex set
+- `v` -- transation vector
 
 ### Notes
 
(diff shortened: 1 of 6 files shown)


src/AffineMap.jl
```diff
@@ -0,0 +1,365 @@
...
+Return the linear map of a lazy affine map.
+
+### Input
+
+- `M`  -- matrix
+- `am` -- affine map
+
+### Output
+
+The polytope representing the linear map of the lazy affine map of a set.
```
schillic: Does it have to be a polytope?

test/unit_AffineMap.jl
```diff
@@ -0,0 +1,65 @@
...
+
+    # ==================================
+    # Type-specific methods
+    # ==================================
+
+    # the translation is the origin and the linear map is the identity => constraints remain unchanged
+    Id3 = Matrix(one(N) * I, 3, 3)
+    @test constraints_list(AffineMap(Id3, zeros(N, 3), B)) == constraints_list(B)
+
+    # two-dimensional case: inclusion check
```
schillic: I do not understand this comment here.

src/AffineMap.jl
```diff
@@ -0,0 +1,365 @@
...
+### Input
+
+- `am` -- affine map
+
+### Output
+
+`true` iff the wrapped set is empty and the affine vector is empty.
+"""
+function isempty(am::AffineMap)::Bool
+    return isempty(am.X) && isempty(am.b)
```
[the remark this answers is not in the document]
mforets: Oops, it was meant to be `iszero(am.b)`. But, you are right, the condition is not necessary.

src/AffineMap.jl
```diff
@@ -0,0 +1,365 @@
...
+- `x`  -- point/vector
+- `an` -- affine map of a convex set
+
+### Output
+
+`true` iff ``x ∈ am``.
+
+### Algorithm
+
+Note that ``x ∈ M⋅S ⊕ b`` iff ``M^{-1}⋅(x - b) ∈ S``.
```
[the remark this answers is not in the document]
schillic: Hm okay, you convinced me.

mforets:
> I was shortly thinking whether to swap the arguments (AffineMap(A, X, b)) because that would fit AX+b better, but I think it is fine.

Yes, i doubted about that; there is `Translation(X, v)` and `LinearMap(M, X)`. I don't have any particular preference for `AffineMap` so we can just go with `AffineMap(M, b, X)`. OTOH I don't know why i used `b` instead of `v` here, as in `Translation`. Using the same name field for both "translation" parts would be preferable.

src/AffineMap.jl
```diff
@@ -0,0 +1,365 @@
...
+the affine map, pass `apply_convex_hull=false` as a keyword argument.
+
+Note that we assume that the underlying set `X` is polyhedral, either concretely
+or lazily, i.e. there the function `vertices_list` should be applicable.
+"""
+function vertices_list(am::AffineMap{N};
+                       apply_convex_hull::Bool=true)::Vector{Vector{N}} where {N<:Real}
+    # for a zero map, the result is just the singleton containing the translation
+    if iszero(am.M)
+        return Singleton(am.b)
```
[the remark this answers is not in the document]
mforets: OK i would go for removing this case.

src/AffineMap.jl
mforets: Oh, good catch, my implementation is wrong here -- it is now returning a lazy translation, but the aim was that `linear_map` returns a concrete set..

mforets:
Based on the last comment and your [here](https://github.com/JuliaReach/LazySets.jl/pull/1360#pullrequestreview-236409894), i propose to rename the fields of the new type to be: `AffineMap(M, X, v)`.

Wrt the concrete linear map, i'm setting the [concrete linear map of a translation](https://github.com/JuliaReach/LazySets.jl/issues/1364) as a prerequisite.

schillic:
> Wrt the concrete linear map, i'm setting the concrete linear map of a translation as a prerequisite.

This is now resolved.

src/AffineMap.jl
```diff
@@ -359,3 +359,8 @@ A set corresponding to the linear map of the lazy affine map of a set.
 function linear_map(M::AbstractMatrix{N}, am::AffineMap{N}) where {N<:Real}
      return translate(linear_map(M * am.M, am.X), M * am.v)
 end
+
+# constructor from a linear map: perform the matrix multiplication immediately
+function LinearMap(M::AbstractMatrix, am::AffineMap)
+    return AffineMap(M * am.M, am.X, M * am.v)
```
schillic: This is a duplicate constructor.

test/unit_AffineMap.jl
```diff
@@ -0,0 +1,65 @@
...
+    am = AffineMap(M, B, v)
+
+    # dimension check
+    @test dim(am) == 3
+
+    # dimension assertion
+    @test_throws AssertionError AffineMap(M, B, N[0, 0])
+    @test_throws AssertionError AffineMap(M, B × B, v)
+
+    # linear map of an affine map is automatically simplified to a linear map
```
schillic:
```suggestion
    # linear map of an affine map is automatically simplified to an affine map
```

test/unit_AffineMap.jl
```diff
@@ -0,0 +1,65 @@
...
+
+    # support function
+    @test ρ(N[1, 0, 0], am) == N(2)
+
+    # boundedness
+    @test isbounded(am)
+    @test !isbounded(AffineMap(M, Universe{N}(3), v))
+
+    # function to get an element
+    @test (an_element(am) - am.v) ∈ (am.M * am.X) # an_element(am) ∈ am.M * am.X ⊕ am.b : requires #1358
```
schillic:
```suggestion
    @test (an_element(am) - am.v) ∈ (am.M * am.X) # an_element(am) ∈ am.M * am.X + am.v : requires #1358
```

test/unit_AffineMap.jl
```diff
@@ -0,0 +1,65 @@
...
+
+    # emptiness check
+    @test !isempty(am)
+
```
schillic:
```suggestion
    @test isempty(AffineMap(M, EmptySet{N}(), v))
```

test/unit_AffineMap.jl
```diff
@@ -0,0 +1,65 @@
...
+    # ==================================
+    # Type-specific methods
+    # ==================================
+
+    # the translation is the origin and the linear map is the identity => constraints remain unchanged
+    Id3 = Matrix(one(N) * I, 3, 3)
+    @test constraints_list(AffineMap(Id3, B, zeros(N, 3))) == constraints_list(B)
```
schillic: We could add a convenience constructor from `I` (the dimension can be obtained from `v`) (can be done in a new issue).

test/unit_AffineMap.jl
```diff
@@ -0,0 +1,65 @@
...
+
+    # two-dimensional case
+    B2 = BallInf(zeros(N, 2), N(1))
+    M = N[1 0; 0 2]
+    tl = N[-1, 0]
```
schillic: Is there a reason for calling this vector `tl` and not `v`?

test/unit_AffineMap.jl
mforets: ok, i've added

```julia
# convenience constructor from the identity
function AffineMap(M::UniformScaling, X::S, v::VN) where {N, NM, S<:LazySet{N}, VN<:AbstractVector{NM}}
    return AffineMap(Diagonal(ones(N, length(v))), X, v)
end
```

mforets: Hmm actually this case is a pure translation, so `Translation(X, v)` seems more appropriate.

test/unit_AffineMap.jl
mforets: There is a problem with this new test:

```julia
julia>     @test isempty(AffineMap(M, EmptySet{N}(), v))
Error During Test at REPL[31]:1
  Test threw exception
  Expression: isempty(AffineMap(M, EmptySet{N}(), v))
  AssertionError: a matrix of size (3, 3) cannot be applied to a set of dimension -1
```

I've kept it commented with a comment that it requires #1201

test/unit_AffineMap.jl
mforets: The commented line (requires #1358) could be removed as well and the test now passes.

test/unit_AffineMap.jl
schillic: True, sorry!

src/AffineMap.jl
```diff
@@ -0,0 +1,366 @@
...
+    AffineMap{N<:Real, S<:LazySet{N}, NM, MAT<:AbstractMatrix{NM},
+              VN<:AbstractVector{NM}} <: LazySet{N}
+
+Type that represents an affine transformation ``M⋅X ⊕ v`` of a convex set ``X``.
+
+### Fields
+
+- `M` -- matrix/linear map
+- `X` -- convex set
+- `v` -- transation vector
```
schillic:
```suggestion
- `v` -- translation vector
```

src/AffineMap.jl
```diff
@@ -0,0 +1,366 @@
...
+
+### Notes
+
+The affine map is the composition of a linear map and a translation. This type is
```
schillic:
```suggestion
An affine map is the composition of a linear map and a translation. This type is
```

src/AffineMap.jl
```diff
@@ -0,0 +1,366 @@
...
+                         affine map 
+
+### Output
+
+A list of vertices.
+
+### Algorithm
+
+This implementation computes all vertices of `X`, then transforms them through
+the affine map, i.e.  `x ↦ M*x + v` for each vertex `x` of `X`. By default, the
```
schillic:
```suggestion
the affine map, i.e. `x ↦ M*x + v` for each vertex `x` of `X`. By default, the
```

src/AffineMap.jl
```diff
@@ -0,0 +1,366 @@
...
+
+    # collect vertices list of the wrapped set
+    vlist_X = vertices_list(am.X)
+
+    # create resulting vertices list
+    vlist = Vector{Vector{N}}()
+    sizehint!(vlist, length(vlist_X))
+    for x in vlist_X
+        push!(vlist, am.M * x + am.v)
+    end
```
schillic: Shorter: `vlist = [am.M * x + am.v for x in vertices_list(am.X)]`

PR #1361, opened 2019-05-12
Some fixes in LinearMap docs

mforets:


src/LinearMap.jl
```diff
@@ -72,19 +72,20 @@ end
 
 """
 ```
-    *(a::N, X::LazySet{N}) where {N<:Real}
+    *(α::N, X::LazySet{N}) where {N<:Real}
```
schillic: If you rename `a` here, please do so in the code as well.

PR #1367, opened 2019-05-13
#1366 - Approximate equality test for `Rational`s

### LINKED ISSUE #1366: Approximate equality test for `Rational`s

schillic:
Closes #1366.

The last commit is unrelated, but I found it weird that we have different (and actually inconsistent) behavior for `Rational`s. I would actually argue that we should also remove the default for `_leq` for the same reason, but I did not have a closer look at it.

### DIFF
--- a/src/comparisons.jl
+++ b/src/comparisons.jl
@@ -47,7 +47,8 @@ arguments to a common numeric type, returning them as a tuple. The conversion
 is such that the common type to which the values are converted can represent
 them as faithfully as possible.
 """
-_leq(x::N, y::M; kwargs...) where {N<:Real, M<:Real} = _leq(promote(x, y)...; kwargs...)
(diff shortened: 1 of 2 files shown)


mforets:
Rationals can and should use exact comparisons; what i suggested [here](https://github.com/JuliaReach/LazySets.jl/pull/1345#discussion_r283413537) to share the interface, is to just let `isapproxzero(x::Real, ...)` handle it (but don't remove it!).

schillic:
Yes, but the other implementation uses exact comparison, so I argue that we can use the other method in all cases (which simplifies things). Furthermode, if one wants to change the tolerance for new number types, one can simply override `ABSZTOL` and it works automatically. This is not the case with the removed method.

mforets:
Oh, true, with `ABSZTOL(N::Type{Rational{INNER}}) where {INNER} = zero(N)` even `abs(x) <= ztol` has the same behavior as checking that `x` is zero.

Issue #1176, opened 2019-02-28
Lazy ReLU operation


schillic:
Given a set `X`, the set `Relu(X)` respresents the set `{max.(x, 0) | x ∈ X}`.
For a convex set `X`, `Relu(X)` is not necessarily convex (take `X` as the 2D line segment from `(0, -1)` to `(2, 1)`.

For a start, we only need the support vector. The first idea is to just reset all negative entries in the support vector of `X` to zero. We would need to check if that is actually correct.

mforets:
CC: @frehseg, @tomerarnon

schillic:
This is not correct. Take the line segment below.

![line_segment](https://user-images.githubusercontent.com/9656686/55250076-c30ed780-524d-11e9-8f9d-c4a916fcf93a.png)

The ReLU set is the following.

![relu](https://user-images.githubusercontent.com/9656686/55250085-c86c2200-524d-11e9-839f-2b64a0424955.png)

The support vector in direction `[1, 1]` would be the point to the right.

tomerarnon:
I'm not sure I understand the issue you are pointing out here.

>The support vector in direction [1, 1] would be the point to the right.

I.e. `[3, 0]` is the correct support vector in that direction? In that case, it seems it is correctly calculated with `ReLU(σ([1,1], LS))`

schillic:
The support vector `σ([1,1], LS)` is not unique. In fact, any point on the original line segment is correct. So in this example it would be luck to get a correct result.
Even worse, by slightly moving the upper end to the right, `σ([1,1], LS_skewed)` would be unique and the *top point*, so then it would even be always wrong.

tomerarnon:
Ah, now I see. I lack the intuition around support vectors to see these things easily 😅 

I imagine this is to do with the non convexity of the relu?

tomerarnon:
Actually, my guess here was also not correct. Although maybe in that case the non-convexity had something to do with it, that isn't the only issue. For example, if you take a Ball2 (that has a negative part, but remains convex after relu-ing) the resulting support vectors are incorrect. e.g. for a ball lying slightly above the x axis, `relu(σ([p, ball))` will give the point on the x axis in the direction `p`, not the coordinates of the nearest vertex as it should.

schillic:
Here are some new ideas/observations.
1. ReLU is conceptually simple for boxes, namely again a box. Let `Bi` be the input box. We construct the output box `Bo`:
[...]

PR #1368, opened 2019-05-13
Add 'rectify' helper function

schillic:
See #1176.

### DIFF
--- a/src/helper_functions.jl
+++ b/src/helper_functions.jl
@@ -669,3 +669,20 @@ function inner(x::AbstractVector{N}, A::AbstractMatrix{N}, y::AbstractVector{N}
               ) where {N}
     dot(x, A * y)
 end
+
+"""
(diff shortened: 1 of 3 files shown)


src/helper_functions.jl
```diff
@@ -669,3 +669,20 @@ function inner(x::AbstractVector{N}, A::AbstractMatrix{N}, y::AbstractVector{N}
...
+### Input
+
+- `x` -- vector
+
+### Output
+
+A copy of the vector where each negative entry is replaced by zero.
+"""
+function rectify(x::AbstractVector{N}) where {N<:Real}
+    return max.(x, zeros(N, length(x)))
```
mforets: Avoid the zeros vector allocation with `return map(xi -> max(xi, 0), x)`; i tested it and it is ~2x faster with random vectors.

(If this function will be used in performance-critical code we can offer a `rectify!` function that relies on Julia's `map!(function, destination, collection...)`.)

Issue #947, opened 2018-12-18
Issues with empty VPolytope


tomerarnon:
Empty (vertex-less) VPolytopes cause all kinds of problems:
```
julia> polyhedron(VPolytope())
ERROR: StackOverflowError:
Stacktrace:
 [1] similar at ./boot.jl:394 [inlined]
 [2] similar at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.0/LinearAlgebra/src/adjtrans.jl:144 [inlined]
 [3] Type at ./array.jl:498 [inlined]
 [4] MixedMatVRep(::LinearAlgebra.Adjoint{Any,Array{Any,1}}, ::Array{Any,2}, ::BitSet) at /Users/tomer/.julia/packages/Polyhedra/T1zRo/src/matrep.jl:111 (repeats 43667 times)
 [5] vrep(::LinearAlgebra.Adjoint{Any,Array{Any,1}}, ::Array{Any,2}, ::BitSet) at /Users/tomer/.julia/packages/Polyhedra/T1zRo/src/matrep.jl:79 (repeats 2 times)
 [6] vrep(::LinearAlgebra.Adjoint{Any,Array{Any,1}}) at /Users/tomer/.julia/packages/Polyhedra/T1zRo/src/matrep.jl:80
 [7] polyhedron(::VPolytope{Float64}, ::DefaultLibrary{Float64}) at /Users/tomer/.julia/packages/LazySets/XNXaI/src/VPolytope.jl:245
 [8] polyhedron(::VPolytope{Float64}) at /Users/tomer/.julia/packages/LazySets/XNXaI/src/VPolytope.jl:244
 [9] top-level scope at none:0

julia> tohrep(VPolytope())
ERROR: StackOverflowError:
Stacktrace:
 [1] similar at ./boot.jl:394 [inlined]
 [2] similar at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.0/LinearAlgebra/src/adjtrans.jl:144 [inlined]
 [3] Type at ./array.jl:498 [inlined]
 [4] MixedMatVRep(::LinearAlgebra.Adjoint{Any,Array{Any,1}}, ::Array{Any,2}, ::BitSet) at /Users/tomer/.julia/packages/Polyhedra/T1zRo/src/matrep.jl:111 (repeats 43666 times)
[...]

Issue #941, opened 2018-12-17
convex_hull of polytopes


tomerarnon:
The documentation mentions that convex_hull is used to compute the convex hull of a polygon (2D only). However, convex_hull appears to run on a VPolytope of any dimensionality. The following quick check shows that adding 2 more dimensions to a vector of 2D points does not change the convex hull, suggesting that the function is not applicable at >2D.
```
V1 = VPolytope([rand(4) for i in 1:200])
V2 = VPolytope([v[1:2] for v in V1.vertices])

length(convex_hull(V1.vertices)) == length(convex_hull(V2.vertices)) #returns true
```
If this is indeed the case, then perhaps convex_hull should contain an assertion for dimensionality to prevent erroneous usage.

Also, given that convex_hull should not be used on higher dimensional polytopes, is there another way to calculate a convex hull? I am aware of the ConvexHull and ConvexHullArray types, but they seem to only work lazily. Is there a way to concretize a ConvexHull into an actual array of points?

Thanks!

schillic:
> convex_hull should contain an assertion for dimensionality to prevent erroneous usage

Agreed.

> Is there a way to concretize a ConvexHull into an actual array of points?

If you only work with the `Polyhedra`-wrapper types (`VPolytope` and `HPolytope`), then we have `convex_hull(P)` which just calls `Polyhedra`'s `convexhull` function. Otherwise, I am afraid we do not support obtaining the vertices of a high-dimensional convex hull.

tomerarnon:
Thinking a bit more about this issue, I think it would be really great  to be able to concretize a `ConvexHull` or `ConvexHullArray` with a call to `convex_hull` (and quite possible). I can imagine cases (in fact, I have even encountered one! 😄) where building up a convex hull lazily and evaluating at the end could be much more efficient. This could work something like:
```
convex_hull(ch::ConvexHull) = convex_hull(ch.X, ch.Y)

function convex_hull(chA::ConvexHullArray)
    allverts = []
    for P in chA.array
        append!(allverts, vertices_list(P) # assumes the sets all have vertices_list defined
    end
    remove_redundancy!(polyhedron(VPolytope(allverts)))
end
```
Where the internal workings of the `ConvexHullArray` version could probably be made cleaner (maybe with greater familiarity with Polyhedra.jl).

schillic:
[...]

PR #1369, opened 2019-05-13
Improve binary convex hull of VPolytope

mforets:
This PR is related to https://github.com/JuliaReach/LazySets.jl/issues/941. 

Notes:

- the binary convex hull of two `VPolytope` now relies on the implementation of the concrete convex hull; the flow of that function is now similar to that of `VPolygon` (taking the union and calling convex hull)
- solves the `convex_hull` of empty `VPolytope` reported [here](https://github.com/JuliaReach/LazySets.jl/issues/947)
- this has the benefit that specific 1D and 2D implementations are used, while the higher-dim case uses `Polyhedra`

Some figures for a random polytope with 20 vertices:

-  in 2D: ~ 4us with Vpolygon; ~1.8ms with CDDLib; ~8.4ms with DefaultLibrary
- in 3D: ~2.2ms with CDDLib;  ~330ms with DefaultLibrary

Related to  https://github.com/JuliaReach/LazySets.jl/pull/1343 (make CDDLib default).

### DIFF
--- a/src/VPolytope.jl
+++ b/src/VPolytope.jl
@@ -337,12 +337,9 @@ for the `convex_hull`.
 """
 function convex_hull(P1::VPolytope{N}, P2::VPolytope{N};
                      backend=default_polyhedra_backend(P1, N)) where {N}
-    @assert isdefined(@__MODULE__, :Polyhedra) "the function `convex_hull` needs " *
-                                               "the package 'Polyhedra' to be loaded"
-    Pch = convexhull(polyhedron(P1; backend=backend),
-                     polyhedron(P2; backend=backend))
-    removevredundancy!(Pch)
-    return VPolytope(Pch)
+    vunion = [P1.vertices; P2.vertices]
+    convex_hull!(vunion; backend=backend)
+    return VPolytope(vunion)
 end
 
 """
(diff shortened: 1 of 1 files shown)


src/VPolytope.jl
```diff
@@ -322,27 +322,29 @@ Compute the convex hull of the set union of two polytopes in V-representation.
...
+                  computations backend
 
 ### Output
 
 The `VPolytope` obtained by the concrete convex hull of `P1` and `P2`.
 
 ### Notes
 
+This function takes the union of the vertices of each polytope and then relies
+on a concrete convex hull algorithm. For low dimensions, a specialied implementation
```
schillic:
```suggestion
on a concrete convex hull algorithm. For low dimensions, a specialized implementation
```

src/VPolytope.jl
```diff
@@ -322,27 +322,29 @@ Compute the convex hull of the set union of two polytopes in V-representation.
...
+for polygons is used. For higher dimensions, `convex_hull` relies on the polyhedral
+computations backend specified in the `backend` keyword argument, which defaults
+to `default_polyhedra_backend(P1, N)`.
```
schillic: You just changed it to `nothing`. Maybe just remove the comment about the default.

PR #1371, opened 2019-05-13
Halfspace representation of an empty VPolytope

mforets:
Fixes the corner case `tohrep(VPolytope())` reported [here](https://github.com/JuliaReach/LazySets.jl/issues/947).

Note: if this gets merged, for consistency we should use the same convention for `VPolygon`,  which gives a polygon with no constraints in `#master`:

```julia
julia> tohrep(VPolygon())
HPolygon{Float64}(HalfSpace{Float64}[])
```


schillic:
> if this gets merged, for consistency we should use the same convention for VPolygon

Let us do that here as well to avoid further confusion.

src/VPolytope.jl
```diff
@@ -426,6 +426,11 @@ For further information on the supported backends see
 """
 function tohrep(P::VPolytope{N};
                 backend=default_polyhedra_backend(P, N)) where {N<:Real}
+    vl = P.vertices
+    n = length(vl)
+    if n == 0
```
schillic: Shorter: `if isempty(vl)`

test/unit_Polygon.jl
```diff
@@ -396,4 +400,4 @@ end
 # default Float64 constructors
 @test HPolygon() isa HPolygon{Float64}
 @test HPolygonOpt() isa HPolygonOpt{Float64}
-@test VPolygon() isa VPolygon{Float64}
+@test VPolygon() isa VPolygon{Float64}
```
schillic: Why remove the empy line here?

mforets: (it was unintentional)

Thanks for the review. I'll update those changes and merge when CI passes.

PR #1372, opened 2019-05-13
#1176 - Lazy ReLU operation

### LINKED ISSUE #1176: Lazy ReLU operation

schillic:
First step for #1176.

### DIFF
--- a/src/Rectification.jl
+++ b/src/Rectification.jl
@@ -220,3 +220,20 @@ function ∈(x::AbstractVector{N}, r::Rectification{N})::Bool where {N<:Real}
     error("cannot determine membership of a vector with zero entries in a " *
           "lazy rectification")
 end
+
+"""
+    isempty(r::Rectification)::Bool
+
+Check whether a rectification is empty or not.
+
+### Input
+
+- `r` -- rectification
+
+### Output
+
+`true` iff the wrapped set is empty.
+"""
+function isempty(r::Rectification)::Bool
+    return isempty(r.X)
+end
--- a/src/Rectification.jl
+++ b/src/Rectification.jl
@@ -1,3 +1,5 @@
+import Base: ∈
+
 export Rectification
 
 """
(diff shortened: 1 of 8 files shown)


src/Rectification.jl
```diff
@@ -0,0 +1,355 @@
...
+```
+
+Clearly, ``\\text{rectify}(X ∩ O_j) = X`` if ``O_j`` is the nonnegative orthant.
+
+For example, consider a two-dimensional case and call the orthants
+``O_1, …, O_4`` in clockwise fashion, starting with the nonnegative orthant.
+We conclude that
+
+```math
+    \\text{rectify}(X) = (X ∩ O_1) ∪ \\text{rectify}(X ∩ O_2) ∩ \\text{rectify}(X ∩ O_3) ∩ \\text{rectify}(X ∩ O_4).
```
mforets:
```suggestion
    \\text{rectify}(X) = (X ∩ O_1) ∪ \\text{rectify}(X ∩ O_2) ∪ \\text{rectify}(X ∩ O_3) ∪ \\text{rectify}(X ∩ O_4).
```

src/Rectification.jl
```diff
@@ -0,0 +1,355 @@
...
+```
+
+The rectification of the intersection in the nonpositive orthant,
+``\\text{rectify}(X ∩ O_3)``, is either the empty set or the singleton
+containing the origin.
+The rectification of ``X ∩ O_2`` and ``X ∩ O_4`` both result in flat
+``n-1``-dimensional shapes on the corresponding hyperplane of ``O_1``.
```
mforets:
```suggestion
``1``-dimensional line segments on the corresponding hyperplane of ``O_1``.
```

PR #1373, opened 2019-05-13
Empty VPolytope to polyhedron conversion

mforets:
Related to https://github.com/JuliaReach/LazySets.jl/issues/947.

### DIFF
--- a/src/VPolytope.jl
+++ b/src/VPolytope.jl
@@ -508,6 +508,17 @@ Return an `VRep` polyhedron from `Polyhedra.jl` given a polytope in V-representa
 ### Output
 
 A `VRep` polyhedron.
+
+### Notes
+
+The *relative dimension* (or just *dimension*) refers to the dimension of the set
+independently of the ambient dimension. For example, a point has (relative) dimension
+one, and a line segment has (relative) dimension two, independently of the ambient
+space. In this library, `LazySets.dim` always returns the ambient dimension
+of the set, such that a point in two-dimensions has dimension 2, and a point in
+three dimensions has dimension three, etc. However, `Polyhedra.dim` will assign
+a flat line segment in two-dimensions a dimension one because it uses a different
+convention.
 """
 function polyhedron(P::VPolytope{N};
                     backend=default_polyhedra_backend(P, N),
--- a/src/VPolytope.jl
+++ b/src/VPolytope.jl
@@ -512,13 +512,13 @@ A `VRep` polyhedron.
 ### Notes
 
 The *relative dimension* (or just *dimension*) refers to the dimension of the set
-independently of the ambient dimension. For example, a point has (relative) dimension
-one, and a line segment has (relative) dimension two, independently of the ambient
-space. In this library, `LazySets.dim` always returns the ambient dimension
-of the set, such that a point in two-dimensions has dimension 2, and a point in
-three dimensions has dimension three, etc. However, `Polyhedra.dim` will assign
-a flat line segment in two-dimensions a dimension one because it uses a different
-convention.
+relative to itself, independently of the ambient dimension. For example, a point
+has (relative) dimension zero, and a line segment has (relative) dimension one.
+
+In this library, `LazySets.dim` always returns the ambient dimension of the set,
+such that a line segment in two dimensions has dimension two. However,
+`Polyhedra.dim` will assign a flat the same line segment a dimension equal to one,
+because it uses a different convention.
 """
 function polyhedron(P::VPolytope{N};
                     backend=default_polyhedra_backend(P, N),
--- a/src/VPolytope.jl
+++ b/src/VPolytope.jl
@@ -507,9 +507,13 @@ Return an `VRep` polyhedron from `Polyhedra.jl` given a polytope in V-representa
 A `VRep` polyhedron.
 """
 function polyhedron(P::VPolytope{N};
-                    backend=default_polyhedra_backend(P, N)) where {N<:Real}
-    V = hcat(vertices_list(P)...)'
-    return polyhedron(Polyhedra.vrep(V), backend)
+                    backend=default_polyhedra_backend(P, N),
+                    d::Int=dim(P)) where {N<:Real}
+    if d == -1 # empty polytope
+        error("cannot infer dimension of `VPolytope` constructed from no " *
+              "element, but it is required for `Polyhedra.polyhedron`")
+    end
+    return polyhedron(Polyhedra.vrep(P.vertices, d=d), backend)
 end
 
 end # quote
--- a/src/VPolytope.jl
+++ b/src/VPolytope.jl
@@ -501,6 +501,8 @@ Return an `VRep` polyhedron from `Polyhedra.jl` given a polytope in V-representa
 - `backend` -- (optional, default: `default_polyhedra_backend(P, N)`) the polyhedral
                computations backend, see [Polyhedra's documentation](https://juliapolyhedra.github.io/Polyhedra.jl/latest/installation.html#Getting-Libraries-1)
                for further information
+- `ambient_dimension` -- (default, optional: `dim(P)`) ambient dimension of the
+                         polytope; required if the polytope is empty
 
 ### Output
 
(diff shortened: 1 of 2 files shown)


src/VPolytope.jl
```diff
@@ -496,15 +496,22 @@ Return an `VRep` polyhedron from `Polyhedra.jl` given a polytope in V-representa
...
 
 A `VRep` polyhedron.
 """
 function polyhedron(P::VPolytope{N};
-                    backend=default_polyhedra_backend(P, N)) where {N<:Real}
-    V = hcat(vertices_list(P)...)'
-    return polyhedron(Polyhedra.vrep(V), backend)
+                    backend=default_polyhedra_backend(P, N),
+                    ambient_dimension::Int=dim(P)) where {N<:Real}
+    if ambient_dimension == -1 # empty polytope
```
schillic: This error message is not shown because `-1` is not the default value. Instead the method call will crash when it tries to execute `dim(P)`. Instead I suggest the following:
1. Remove the default value for `ambient_dimension`.
2. Only pass the dimension if the vertices list is empty (otherwise use the old implementation).
3. If the list is empty, test if `ambient_dimension` is defined. If not, throw the error you have here.

mforets: `-1` is the default value for an empty `VPolytope`, see [dim(::Vpolytope)](https://github.com/JuliaReach/LazySets.jl/blob/master/src/VPolytope.jl#L74).

schillic: I see :+1:

schillic: Should we rather add a field for the dimension (see #1201 for a related discussion) or assert in the constructor that there is at least one vertex? It seems a bit strange to have this workaround and keep the problem on our side.

mforets: I agree to add a new field for `EmptySet` as in [this comment](https://github.com/JuliaReach/LazySets.jl/issues/1201#issue-417734408). Are you suggesting to add a field for the dimension in `VPolytope`? We could but i dont think it is justified to handle this particular case (creating an empty set through a `VPolytope()`).  Besides, I don't think that the workaround of this PR poses problems on our side, does it?

OTOH i checked again the behavior of `Polyhedra` and calling this argument `ambient_dimension` is not consistent (compare to `LazySets` dim):

```julia
julia> p = polyhedron(vrep([rand(2) for _ in 1:2]))
Polyhedron DefaultPolyhedron{Float64,Polyhedra.Intersection{Float64,Array{Float64,1},Int64},Polyhedra.Hull{Float64,Array{Float64,1},Int64}}:
2-element iterator of Array{Float64,1}:
 [0.905471, 0.873191]
 [0.774891, 0.187635]

julia> dim(p)
1

julia> p = polyhedron(vrep(Vector{Vector{Float64}}(), d=2))
Polyhedron DefaultPolyhedron{Float64,Polyhedra.Intersection{Float64,Array{Float64,1},Int64},Polyhedra.Hull{Float64,Array{Float64,1},Int64}}

julia> dim(p)
-1
```

schillic: > Are you suggesting to add a field for the dimension in VPolytope?

I would rather forbid this case in the constructor.
The problem I see is that we call this method internally, and obviously we do not pass this argument. So you fixed the one problem where the user called `polyhedron`, but all other cases cannot be handled, so you would have to add this keyword argument everywhere. To me this PR just hides the problem a bit deeper.

> ```julia
> julia> dim(p)
> 1
> ```

Huh, why is the dimension `1` here?

> is not consistent

Why?

test/unit_Polytope.jl
```diff
@@ -180,6 +180,11 @@ for N in [Float64, Rational{Int}, Float32]
 
         # list of constraints of a VPolytope; calculates tohrep
         @test ispermutation(constraints_list(V), constraints_list(tohrep(V)))
+
+        # convert empty VPolytope to a polyhedron
+        Ve = VPolytope()
+        @test_throws ErrorException polyhedron(VPolytope()) # needs the ambient dim
+        Pe = polyhedron(VPolytope(), relative_dimension=2)
```
schillic: I guess you wanted to reuse `Ve` in these two lines...

src/VPolytope.jl
```diff
@@ -501,15 +501,26 @@ Return an `VRep` polyhedron from `Polyhedra.jl` given a polytope in V-representa
 - `backend` -- (optional, default: `default_polyhedra_backend(P, N)`) the polyhedral
                computations backend, see [Polyhedra's documentation](https://juliapolyhedra.github.io/Polyhedra.jl/latest/installation.html#Getting-Libraries-1)
                for further information
+- `relative_dimension` -- (default, optional: `nothing`) an integer representing
```
schillic: What is the *relative* dimension?

test/unit_Polytope.jl
```diff
@@ -180,6 +180,11 @@ for N in [Float64, Rational{Int}, Float32]
 
         # list of constraints of a VPolytope; calculates tohrep
         @test ispermutation(constraints_list(V), constraints_list(tohrep(V)))
+
+        # convert empty VPolytope to a polyhedron
+        Vempty = VPolytope()
+        @test_throws ErrorException polyhedron(Vempty) # needs the ambient dim
```
schillic: Does this comment need to be updated?

src/VPolytope.jl
mforets:
```julia
julia> p = polyhedron(Polyhedra.vrep([[0.0, 0.0], [1.0, 1.0]])) # a line segment
Polyhedron DefaultPolyhedron{Float64,Polyhedra.Intersection{Float64,Array{Float64,1},Int64},Polyhedra.Hull{Float64,Array{Float64,1},Int64}}:
2-element iterator of Array{Float64,1}:
 [0.0, 0.0]
 [1.0, 1.0]

julia> Polyhedra.dim(p)
1

julia> p = polyhedron(Polyhedra.vrep([[0.0, 0.0], [1.0, 1.0], [0.0, 2.0]])) # a triangle
Polyhedron DefaultPolyhedron{Float64,Polyhedra.Intersection{Float64,Array{Float64,1},Int64},Polyhedra.Hull{Float64,Array{Float64,1},Int64}}:
3-element iterator of Array{Float64,1}:
 [0.0, 0.0]
 [1.0, 1.0]
 [0.0, 2.0]

julia> Polyhedra.dim(p)
2
```
Here by "relative" i mean relative to the given set.

test/unit_Polytope.jl
mforets: Yes, thanks for pointing that out.

src/VPolytope.jl
schillic: I am afraid that this term is not universally understood. Especially since it is used in the error message. Maybe you can elaborate in the docs using this example and explicitly pointing out that a line segment has relative dimension one even if the ambient dimension is two.

src/VPolytope.jl
schillic: > Huh, why is the dimension 1 here?

Okay, I understood this part now with the discussion below.

I still think that this PR does not solve the problem but only hides it. But let us move on.

mforets: 👍  In Polyhedra.jl it is legit to create an empty polytope if you pass the kwarg `d`; this PR just adds this feature to LazySets.

schillic: But only for manually creating the `Polyhedra.polyhedron`. If you work with an empty `LazySets.VPolytope` and apply an operation that internally converts to `Polyhedra.polyhedron`, it will still crash.

mforets: Ok true, we should fix these when they appear (https://github.com/JuliaReach/LazySets.jl/pull/1371/ and https://github.com/JuliaReach/LazySets.jl/pull/1369 cover two such operations, that were breaking when the polytope is empty).

schillic: But you can only fix them by requiring to pass this additional argument (that was my point [here](https://github.com/JuliaReach/LazySets.jl/pull/1373#discussion_r283502911).) And in the application you mention, I doubt that you know the relative dimension in advance.

mforets: - In the case of `tohrep` of an empty `VPolytope`, again it is special-cased before the call to `polyhedron`, so we don't need to know its relative dimension.
- In the case of the `convex_hull`, before it was needed (because it was directly calling `Polyhedra.convex_hull`), but now it calls `LazySets.convex_hull` which operates on lists of vertices (and the empty case is handled separately) so we don't need to know the relative dimension.

I think this fixes the issues that have arised so far, but you are right, there may be more that we haven't identified yet.

src/VPolytope.jl
```diff
@@ -501,15 +501,37 @@ Return an `VRep` polyhedron from `Polyhedra.jl` given a polytope in V-representa
...
+
+### Notes
+
+The *relative dimension* (or just *dimension*) refers to the dimension of the set
+relative to itself, independently of the ambient dimension. For example, a point
+has (relative) dimension zero, and a line segment has (relative) dimension one.
+
+In this library, `LazySets.dim` always returns the ambient dimension of the set,
+such that a line segment in two dimensions has dimension two. However,
+`Polyhedra.dim` will assign a flat the same line segment a dimension equal to one,
```
schillic:
```suggestion
`Polyhedra.dim` will assign a dimension equal to one to a line segment
```

Issue #666, opened 2018-09-25
Concrete Minkowski sum for polygons


schillic:
See [this algorithm](https://en.wikipedia.org/wiki/Minkowski_addition#Two_convex_polygons_in_the_plane).

schillic:
Closed in #1377.

PR #1377, opened 2019-05-15
#666 - Concrete Minkowski sum for polygons

SebastianGuadalupe:
See #666

### DIFF
--- a/docs/src/lib/representations.md
+++ b/docs/src/lib/representations.md
@@ -415,6 +415,7 @@ constraints_list(::VPolygon{N}) where {N<:Real}
 translate(::VPolygon{N}, ::AbstractVector{N}) where {N<:Real}
 remove_redundant_vertices(::VPolygon{N}; ::String="monotone_chain") where {N<:Real}
 remove_redundant_vertices!(::VPolygon{N}; ::String="monotone_chain") where {N<:Real}
+minkowski_sum(::VPolygon{N}, ::VPolygon{N}) where {N<:Real}
 ```
 Inherited from [`LazySet`](@ref):
 * [`norm`](@ref norm(::LazySet, ::Real))
--- a/src/VPolygon.jl
+++ b/src/VPolygon.jl
@@ -5,7 +5,8 @@ export VPolygon,
        remove_redundant_vertices,
        remove_redundant_vertices!,
        convex_hull,
-       linear_map
+       linear_map,
+       minkowski_sum
 
 """
     VPolygon{N<:Real} <: AbstractPolygon{N}
@@ -552,4 +553,59 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
     @assert length(v) == dim(P) "cannot translate a $(dim(P))-dimensional " *
                                 "set by a $(length(v))-dimensional vector"
     return VPolygon([x + v for x in vertices_list(P)])
+end
+
+"""
+    minkowski_sum(P::VPolygon{N}, Q::VPolygon{N}) where {N<:Real}
+
+The Minkowski Sum of two polygon in vertex representation.
+
+### Input
+
+- `P` -- polygon in vertex representation
+- `Q` -- another polygon in vertex representation
+
+### Output
+
+A polygon in vertex representation.
+
+### Algorithm
+
+We treat each edge of the polygons as a vector, attaching them in polar order
+(attaching the tail of the next vector to the head of the previous vector). The
+resulting polygonal chain will be a polygon, which is the Minkowski sum of the 
+given polygons. This algorithm assumes that the vertices of P and Q are sorted 
+in counter-clockwise fashion and has linear complexity O(m+n) where m and n are 
+the number of vertices of P and Q respectively.
+
+"""
+function minkowski_sum(P::VPolygon{N}, Q::VPolygon{N}) where {N<:Real}
+    vlistP = vertices_list(P)
+    vlistQ = vertices_list(Q)
+    mP = length(vlistP)
+    mQ = length(vlistQ)
+    i = 1
+    k = 1
+    j = 1
+    R = Vector{Vector{N}}(undef, mP+mQ)
+    fill!(R, N[0, 0])
+    while i <= size(R, 1)
+        P₁, P₂ = vlistP[(k-1)%mP+1], vlistP[(k%mP+1)]
+        P₁P₂ = P₂ - P₁
+        Q₁, Q₂ = vlistQ[(j-1)%mQ+1], vlistQ[(j%mQ+1)]
+        Q₁Q₂ = Q₂ - Q₁
+        R[i] = P₁ + Q₁
+        turn = right_turn(P₁P₂, Q₁Q₂, N[0, 0])
+        if turn > 0
+            k += 1
+        elseif turn < 0
+            j += 1
+        else
+            pop!(R)
+            k += 1
+            j += 1
+        end
+        i += 1
+    end
+    return VPolygon(R)
 end
--- a/test/unit_Polygon.jl
+++ b/test/unit_Polygon.jl
@@ -169,6 +169,26 @@ for N in [Float64, Float32, Rational{Int}]
         @test translate(hp, N[1, 2]) == typeof(hp)(
             [HalfSpace(N[2, 2], N(18)), HalfSpace(N[-3, 3], N(9)),
              HalfSpace(N[-1, -1], N(-3)), HalfSpace(N[2, -4], N(-6))])
+        
+        # test for concrete minkowski sum
+        A = [N[4, 0], N[6, 2], N[4, 4]]
+        B = [N[-2, -2], N[2, 0], N[2, 2], N[-2, 4]]
+
+        P = VPolygon(A)
+        Q = VPolygon(B)
+        PQ = minkowski_sum(P, Q)
+        @test LazySets.ispermutation(PQ.vertices, [N[2, -2], N[6, 0], N[8, 2],
+                                                N[8, 4], N[6, 6], N[2, 8]])
+
+        #test for corner case of parallel edges in minkowski sum
+        C = [N[10, 5], N[10, 10], N[8, 10], N[5, 8], N[5, 5]]
+        D = [N[-1, -2], N[1, -2], N[-1, 2]]
+
+        R = VPolygon(C)
+        S = VPolygon(D)
+        RS = minkowski_sum(R, S)
+        @test LazySets.ispermutation(RS.vertices, [N[4, 3], N[11, 3], N[11, 8],
+                                                N[9,12], N[7, 12], N[4, 10]])
     end
 
     # concrete intersection of H-rep


src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
                                 "set by a $(length(v))-dimensional vector"
     return VPolygon([x + v for x in vertices_list(P)])
 end
+"""
```
mforets: Add an empty line between lines 556 and 557 to keep the different functions separated at least by one new line.

src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+    minkowski_sum(P::VPolygon{N}, Q::VPolygon{N}) where {N<:Real}
+
+The Minkowski Sum of two polygon in vertex representation.
```
mforets:
```suggestion
Compute the Minkowski sum of two polygons in vertex representation.
```

src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+
+A polygon in vertex representation.
+
+### Algorithm
+
+We treat each edge of the polygons as a vector, attaching them in polar order
+(attaching the tail of the next vector to the head of the previous vector). The
+resulting polygonal chain will be a polygon, which is the Minkowski sum of the given polygons.
+"""
+
```
mforets: Remove the empty line 576 (there should be no empty line between the docstring and the function declaration, see other examples in the file `VPolygon.jl`.

src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+function minkowski_sum(P::VPolygon{N}, Q::VPolygon{N}) where {N<:Real}
+    vlP = vertices_list(P)
```
mforets: For consistency with other functions of the library, we can rename `vlP` to `vlistP` and `vlQ` to `vlistQ`.

src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+    vlQ = vertices_list(Q)
+    mP = length(vlP);
```
mforets: no need for the semicolon

src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+    mQ = length(vlQ);
```
mforets: same here (semicolon is used to omit the output in the REPL)

src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+    i = 1
+    k = 1
+    j = 1
+    R = Vector{Vector{Float64}}(undef, mP+mQ)
```
mforets:
```suggestion
    R = Vector{Vector{N}}(undef, mP+mQ)
```

src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+    fill!(R, N[0, 0])
+    while i <= size(R, 1)
+        OP = vlP[(k%mP+1)]-vlP[(k-1)%mP+1]
+        OQ = vlQ[(j%mQ+1)]-vlQ[(j-1)%mQ+1]
+        R[i] = vlP[(k-1) % mP+1] + vlQ[(j-1) % mP+1]
+        is_rightTurn = LazySets.right_turn(OP, OQ, N[0, 0])
```
mforets:
```suggestion
        is_rightTurn = right_turn(OP, OQ, N[0, 0])
```

(since this function is inside the `LazySet` library, you don't need to write the module's name -- it is needed for external use of `right_turn` since this function is not exported; see [this part of the julia docs for examples](https://docs.julialang.org/en/v1/manual/modules/index.html#modules-1))

src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+        if is_rightTurn > 0
+            k += 1
+        elseif (is_rightTurn < 0)
```
mforets:
```suggestion
        elseif is_rightTurn < 0
```

src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+    vlQ = vertices_list(Q)
+    mP = length(vlP);
+    mQ = length(vlQ);
+    i = 1
+    k = 1
+    j = 1
+    R = Vector{Vector{Float64}}(undef, mP+mQ)
+    fill!(R, N[0, 0])
+    while i <= size(R, 1)
+        OP = vlP[(k%mP+1)]-vlP[(k-1)%mP+1]
```
mforets: it confuses me to call this variable "OP", since it does not represent the segment from the origin (O) to some point in the polygon `P`, it is a line segment from point `P₁` and `P₂`:

```julia
P₁, P₂ = vlP[(k-1)%mP+1], vlP[(k%mP+1)]
P₁P₂ = P₂ - P₁
```

src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+
+### Output
+
+A polygon in vertex representation.
+
+### Algorithm
+
+We treat each edge of the polygons as a vector, attaching them in polar order
+(attaching the tail of the next vector to the head of the previous vector). The
+resulting polygonal chain will be a polygon, which is the Minkowski sum of the given polygons.
```
mforets: i think that the algorithm assumes that the vertices of `P` and of `Q` are sorted in counter-clockwise fashion. if it is the case then it's a good idea to say it in this paragraph.

src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+        OP = vlP[(k%mP+1)]-vlP[(k-1)%mP+1]
+        OQ = vlQ[(j%mQ+1)]-vlQ[(j-1)%mQ+1]
+        R[i] = vlP[(k-1) % mP+1] + vlQ[(j-1) % mP+1]
+        is_rightTurn = LazySets.right_turn(OP, OQ, N[0, 0])
+        if is_rightTurn > 0
+            k += 1
+        elseif (is_rightTurn < 0)
+            j += 1
+        else
+            pop!(R)
```
mforets: is this final `else` there to cover the corner case that the line segments are parallel?

src/VPolygon.jl
```diff
@@ -553,3 +553,52 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+            k += 1
+            j += 1
+        end
+        i += 1
+    end
+    return VPolygon(R)
+end
```
mforets: please do not remove the empty line at the end of the file

src/VPolygon.jl
schillic: The name `is_rightTurn` suggests that the type of that variable is `Bool`. But `right_turn` returns a number. I admit that the documentation of this function is wrong (I created #1379 to fix that). To help intuition, I would use a different name (e.g., just `turn`) and instead add a comment to each case that explains what it corresponds to.

src/VPolygon.jl
SebastianGuadalupe: VPolygon's vertices are already sorted in a counterclockwise fashion (see the documentation for [VPolygon](https://juliareach.github.io/LazySets.jl/latest/lib/representations/#LazySets.VPolygon),  so I thought it was unnecessary, should I say it anyway?

mforets: Well, yes you are right! :) So here i suggest to add a small remark as a reminder, in the end of your paragraph: "This algorithm assumes that the vertices of ``P`` and ``Q`` are sorted in counter-clockwise fashion and has linear complexity ``O(m+n)`` where ``m`` and ``n`` are the number of vertices of ``P`` and ``Q`` respectively."

mforets: > it was unnecessary, should I say it anyway?

It goes in the direction of making the "Algorithm" notes self-contained so i think it's a good idea, even when the inputs are already `VPolygon`. (FWIW, `VPolygon` *can* be created with non-sorted vertices if you do `VPolygon(..., apply_convex_hull=false)`.)

src/VPolygon.jl
SebastianGuadalupe: Yes, when two line segments are parallel and are in the same direction in the polygons, between the resulting edges there is not a vertex. If it wasn't for the pop!(R), it would be a vertex at the origin.

docs/src/lib/representations.md
```diff
@@ -522,6 +522,7 @@ tovrep(::VPolytope)
 cartesian_product(::VPolytope{N}, ::VPolytope{N}) where N
 polyhedron(::VPolytope{N}) where {N<:Real}
 linear_map(::AbstractMatrix{N}, ::VPolytope{N}) where {N<:Real}
+minkowski_sum(::VPolytope{N},::VPolytope{N}) where {N<:Real}
```
mforets: This function is in the wrong place (sorry if i told you to add it here): it should go under the `VPolygon` section of `representations.md`, because the new function acts on polygons, `minkowski_sum(::VPolygon{N}, ::VPolygon{N}) where {N<:Real}`.

src/VPolygon.jl
```diff
@@ -553,3 +554,58 @@ function translate(P::VPolygon{N}, v::AbstractVector{N}) where {N<:Real}
...
+    k = 1
+    j = 1
+    R = Vector{Vector{N}}(undef, mP+mQ)
+    fill!(R, N[0, 0])
+    while i <= size(R, 1)
+        P₁, P₂ = vlistP[(k-1)%mP+1], vlistP[(k%mP+1)]
+        P₁P₂ = P₂ - P₁
+        Q₁, Q₂ = vlistQ[(j-1)%mQ+1], vlistQ[(j%mQ+1)]
+        Q₁Q₂ = Q₂ - Q₁
+        R[i] = vlistP[(k-1) % mP+1] + vlistQ[(j-1) % mP+1]
```
mforets: In the right-hand side, shouldn't you have `vlistQ[(j-1) % mQ+1]` instead?

If it is the case, it improves the readability to reuse the variables defined earlier:

```suggestion
        R[i] = P₁ + Q₁
```

PR #1381, opened 2019-05-19
#1176 - Support function for Rectification

schillic:
See #1176.

### DIFF
--- a/src/Rectification.jl
+++ b/src/Rectification.jl
@@ -6,9 +6,23 @@ export Rectification
     RectificationCache{N<:Real}
 
 Struct that is used as a cache for [`Rectification`](@ref)s.
+
+### Fields
+
+- `set`                -- set represented by the rectification (can be `nothing`
+                          if not computed yet)
+- `use_support_vector` -- flag indicating whether to use support-vector
+                          computations for the cached set
 """
-struct RectificationCache{N<:Real}
-    # empty for now
+mutable struct RectificationCache{N<:Real}
+    set::Union{LazySet{N}, UnionSetArray{N}, Nothing}
+    use_support_vector::Bool
+
+    # constructor without a set
+    RectificationCache{N}(::Nothing) where {N<:Real} = new{N}(nothing, false)
+
+    # constructor with a set
+    RectificationCache{N}(set::LazySet{N}) where {N<:Real} = new{N}(set, true)
 end
 
 """
@@ -80,7 +94,18 @@ struct Rectification{N<:Real, S<:LazySet{N}}
 
     # default constructor that initializes cache
     function Rectification{N, S}(X::S) where {N<:Real, S<:LazySet{N}}
-        cache = RectificationCache{N}()
+        if X isa AbstractHyperrectangle || X isa CartesianProduct ||
+                X isa CartesianProductArray
+            # set types with efficient support-vector computations
+            set = X
+        elseif dim(X) == 1 && isbounded(X)
+            # one-dimensional bounded sets are converted to Interval type
+            set = convert(Interval, X)
+        else
+            # do not pre-compute a set; computation is triggered later
+            set = nothing
+        end
+        cache = RectificationCache{N}(set)
         return new{N, S}(X, cache)
     end
 end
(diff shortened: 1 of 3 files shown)


schillic:
In principle this works. Compare the [example here](https://github.com/JuliaReach/LazySets.jl/issues/1176#issuecomment-490396060) with this result:
rectification
Here I plotted the four intersections/projections (the origin, two line segments, and the polygon) separately.

There are some precision problems due to lazy intersections. For instance, `ρ([-1., -1.], RP) ~ 0.1`, but it should clearly be 0. If I perform a concrete intersection instead, everything works.

src/Rectification.jl
```diff
@@ -353,3 +411,131 @@ function isbounded(r::Rectification{N})::Bool where {N<:Real}
...
+        for (i, d) in enumerate(mixed_dimensions)
+            if mixed_dimensions_enumeration[i]
+                # reset mixed index to one again because the set is nonnegative
+                v[d] = one(N)
+            end
+        end
+        return Diagonal(v) * set
+    end
+
+    # identify dimensions where set is negative and mixed (positive/negative)
```
mforets: Let's outsource the algorithm that receives a set and returns the indices where the set is negative (or is positive/negative) algorithm to a separate function.

src/Rectification.jl
```diff
@@ -353,3 +411,131 @@ function isbounded(r::Rectification{N})::Bool where {N<:Real}
...
+    negative_dimensions = Vector{Int}()
+    mixed_dimensions = Vector{Int}()
+    mixed_dimensions_enumeration = BitArray(undef, 0)
+    for i in 1:n
+        d = Approximations.UnitVector(i, n, -one(N))
```
mforets: i'd prefer to see `d = UnitVector(i, n, -one(N))` let's `using Approximations: UnitVector` if needed on top of this file

src/Rectification.jl
```diff
@@ -353,3 +411,131 @@ function isbounded(r::Rectification{N})::Bool where {N<:Real}
...
+        if ρ(d, X) > zero(N)
+            push!(negative_dimensions, i)
+            d = Approximations.UnitVector(i, n, one(N))
+            if ρ(d, X) > zero(N)
```
mforets: we can do `ρ(-d, X) > zero(N)` which is efficient and also returns a `SingleEntryVector` (https://github.com/JuliaReach/LazySets.jl/pull/1523)

src/Rectification.jl
```diff
@@ -353,3 +411,131 @@ function isbounded(r::Rectification{N})::Bool where {N<:Real}
...
+                push!(mixed_dimensions, i)
+            end
+        end
+    end
+
+    if isempty(mixed_dimensions)
+        # no mixed dimensions: no intersections needed
+        if isempty(negative_dimensions)
+            # no negative dimensions: set is unaffected by rectification
+            r.cache.set = X
```
mforets: The set `X` may be unaffected by the rectification. So if we change for `convert` we have to wrap it around `UnionSetArray`..

src/Rectification.jl
```diff
@@ -353,3 +411,131 @@ function isbounded(r::Rectification{N})::Bool where {N<:Real}
...
+            # project negative dimensions to zero
+            v = ones(N, n)
+            v[negative_dimensions] .= zero(N)
+            r.cache.set = Diagonal(v) * X
+        end
+    else
+        # apply intersections with half-spaces in every mixed dimension
+        m = length(mixed_dimensions)
+        i = m
+        projections = Vector{LazySet{N}}()
```
mforets: Aren't these projections instances of `LinearMap`?

src/Rectification.jl
```diff
@@ -353,3 +411,131 @@ function isbounded(r::Rectification{N})::Bool where {N<:Real}
...
+        sizehint!(projections, 2^m)
+        while true
+            # compute next projection
+            polyhedron = construct_constraints(mixed_dimensions,
+                                               mixed_dimensions_enumeration)
+            projection = construct_projection(X ∩ polyhedron,
```
mforets: if it's not too hard maybe add an option to compute this intersection either concretely or lazily

src/Rectification.jl
schillic: True, so I will not use `convert`.

PR #1397, opened 2019-05-24
#1374 - Zonotope overapproximation of convex hull of zonotopes

### LINKED ISSUE #1374: Zonotope overapproximation of convex hull of zonotopes
See [Ghorbal, Goubault, Putot - The zonotope abstract domain Taylor1+ (2009)](https://khalilghorbal.info/assets/pdf/papers/cav2009.pdf), Definition 1.

schillic:
Closes #1374.

~Apparently either the implementation or the method is not correct (our unit test discovers this).~
* [x] fix code
* [x] check whether "same number of generators" is actually required

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -192,7 +192,7 @@ Overapproximate the convex hull of two zonotopes.
 
 - `X`         -- convex hull of two zonotopes
 - `Zonotope`  -- type for dispatch
-- `algorithm` -- (optional; default: `"join"`) choice of algorithm; possible
+- `algorithm` -- (optional; default: `"mean"`) choice of algorithm; possible
                  values are `"mean"` and `"join"`
 
 ### Output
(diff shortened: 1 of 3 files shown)


schillic:
Test code:
```julia
julia> using LazySets, Plots

julia> Z1 = rand(Zonotope); Z2 = rand(Zonotope); C = CH(Z1, Z2);
       plot([Z1, Z2, overapproximate(C, Zonotope; algorithm="join"),
             overapproximate(C, Zonotope, algorithm="mean")])
```

schillic:
I set the default algorithm to the new method. There is no general empirical winner between the two methods, but I suspect that not having to reduce the order is a nice advantage.

mforets:
Some informal comments:

- some examples can be found in this notebook: https://nbviewer.jupyter.org/github/mforets/escritoire/blob/master/sets/ConvexHullZonotopes.ipynb

- if the order is small, "join" produces the box oa of the convex hull, and "mean" gives an overapproximating shape which is more similar to the shape of the exact convex hull.

- if the order is higher, "join" still produces a box-shaped set but some of its edges are diagonal.

- "mean" is faster than "join" by a factor of ~3..5, although i tried in a small set of instances.

- for reachability, seems to me that it may be preferable to use "mean", because it somehow keeps the information about the direction of the blow (while the box shape hides that information), even if the hausdorff distance is bigger

Issue #237, opened 2018-02-13
Clean up plot_recipes


schillic:
* [x] Use plural form for vector plots (like `plot_zonotope`/`plot_zonotopes`).
* [x] Make the variable names consistent (sometimes we use `arr` for vectors, sometimes not).
* [x] Fix the first line of `plot_zonotope`'s documentation.
* [x] Unify the `plot_lazyset` methods with and without `ε`. This way we also use the `:scatter` option from above also in the `ε` case. (see #1304)
* [x] Use the `:scatter` option from above also in the vector case (currently it does not work).
* [x] Unify the `plot_polygon` methods with a `Union` (or better: `AbstractPolygon`).
* [x] Can we even use `AbstractPolytope` above? This way we could remove the plot recipes for zonotopes and would also get more efficient plotting methods for the other polytopic sets for free.
* [x] ~Use `NaN`s for vectors.~ (not needed)
* [x] Rename `plot_linesegment` for (vectors of) `Interval`s.
* [x] `@assert` that the dimensionality is 2.
* [x] colors are ignored for some plot recipes (e.g., `LineSegment`) (see #1304 and docstring extension)
* [x] plot of 1-vertex `VPolygon`/`VPolytope` should use `Singleton` plot recipe (maybe simplify? see #614)
* [x] plot of 2-vertex `VPolygon`/`VPolytope` should use `LineSegment` plot recipe (maybe simplify? see #614).
* [x] `seriestype := norm(vlist[1, :] - vlist[2, :]) ≈ 0 ? :scatter : :shape` seems to check if the first and the second vertex are very close, but what about a flat but long polygon? I think we should check for all vertices. This can also be useful for the previous item (flat polygons).

schillic:
Also add the `NaN`s for vectors to get faster plots.

mforets:
i think we can check `@assert that the dimensionality is 2.`

schillic:
> i think we can check @assert that the dimensionality is 2.

The last six recipes do not check that.

mforets:
we don't need to check for intervals and linesegments because they are 2d.

schillic:
but for `Singleton`s

mforets:
> colors are ignored for some plot recipes (e.g., LineSegment and Singleton)

For `Singleton`s, colors work here.

[...]

PR #1400, opened 2019-05-25
#237 - Clean up plot_recipes

schillic:
See #237.

Most important changes:
* Plotting a vector of sets now falls back to the plot recipes for each individual set. This saves lots of code duplication and is generally preferable for both efficiency and consistency. In particular, this allows to plot  vectors containing lazy `Intersection`s.
* Arbitrary 1D sets can now be plotted.
* Common plotting options are unified.

### DIFF
--- a/src/plot_recipes.jl
+++ b/src/plot_recipes.jl
@@ -3,106 +3,32 @@ import RecipesBase.apply_recipe
 
 using LazySets.Approximations: overapproximate, PolarDirections
 
-function warn_empty_polytope()
-    @warn "received a polytope with no vertices during plotting"
-end
-
-# ====================================
-# Plot recipes for an abstract LazySet
-# ====================================
+# global values
+DEFAULT_COLOR = :auto
+DEFAULT_ALPHA = 0.5
+DEFAULT_LABEL = ""
+DEFAULT_GRID = true
+DEFAULT_ASPECT_RATIO = 1.0
+DEFAULT_EPSILON = 1e-3
+DEFAULT_POLAR_DIRECTIONS = 40
 
 """
-    plot_lazyset(X::LazySet{N}, [ε]::N=N(1e-3); ...) where {N<:Real}
+    plot_list(list::AbstractVector{VN}, [ε]::N=N(DEFAULT_EPSILON),
+              [Nφ]::Int=DEFAULT_POLAR_DIRECTIONS; ...) where {N<:Real,
+                                                              VN<:LazySet{N}}
 
-Plot a convex set in two dimensions.
+Plot a list of convex sets.
 
 ### Input
 
-- `X` -- convex set
-- `ε` -- (optional, default: `1e-3`) approximation error bound
-
-### Examples
-
-```julia
-julia> B = BallInf(ones(2), 0.1);
-
-julia> plot(2.0 * B)
-```
-
-An iterative refinement method is applied to obtain an overapproximation of `X`
-in constraint representation, which is then plotted. To improve the accuracy
-of the iterative refinement, use the second argument using a small value:
-
-```julia
-julia> B = Ball2(ones(2), 0.1);
-
-julia> plot(B, 1e-3);
-
-julia> plot(B, 1e-2); # faster than the previous try, but less accurate
-```
-
-### Algorithm
-
-In a first stage, an overapproximation of the given set to a polygon in constraint
-representation is computed. The second argument, `ε`, corresponds to the error
-in Hausdorff distance between the overapproximating set and `X`. The default
-value `1e-3` is chosen such that the unit ball in the 2-norm is plotted with
-reasonable accuracy. On the other hand, if you only want to produce a fast
-box-overapproximation of `X`, pass `ε=Inf`.
-
-In a second stage, the list of vertices of the overapproximation is computed
-with the `vertices_list` function of the polygon.
-
-A post-processing `convex_hull` is applied to the vertices list to ensure
-that the shaded area inside the convex hull of the vertices is covered
-correctly.
+- `list` -- list of convex sets (1D or 2D)
+- `ε`    -- (optional, default: `DEFAULT_EPSILON`) approximation error bound
+- `Nφ`   -- (optional, default: `DEFAULT_POLAR_DIRECTIONS`) number of polar
+            directions (used to plot lazy intersections)
 
 ### Notes
 
-This recipe detects if overapproximation is such that the first two vertices
-returned by `vertices_list` are the same. In that case, a scatter plot is used
-(instead of a shape plot). This corner case arises, for example, when lazy linear
-maps of singletons.
-"""
-@recipe function plot_lazyset(X::LazySet{N}, ε::N=N(1e-3);
-                              color=:auto, alpha=0.5) where {N<:Real}
-
-    @assert dim(X) == 2 "cannot plot a $(dim(X))-dimensional set using iterative refinement"
-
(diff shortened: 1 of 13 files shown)


src/plot_recipes.jl
```diff
@@ -411,67 +242,25 @@ julia> plot(L)
...
-The option `add_marker` (optional, default: `true`) can be used to specify if
-endpoint markers should be plotted or not.
-```julia
-julia> plot(L, add_marker=false, linecolor="red", linewidth=2.)
-```
 """
-@recipe function plot_linesegment(L::LineSegment{N}, ε::N=zero(N);
-                                  color=:auto, alpha=0.5, add_marker=true) where {N<:Real}
-
+@recipe function plot_linesegment(L::LineSegment{N}, ε::N=zero(N)
```
mforets: The argument `add_marker` was there to allow turning off the marker at the endpoints (The value `:none` does exist but it doesn't work..):

```julia
julia> plot(Interval(0.0, 1.0), markershape=:circle)  # ok

julia> plot(Interval(0.0, 1.0), markershape=:square)  # ok 

julia> plot(Interval(0.0, 1.0), markershape=:none)  # error
ERROR: KeyError: key :seriestype not found
Stacktrace:
 [1] getindex at ./dict.jl:478 [inlined]
 [2] preprocessArgs!(::Dict{Symbol,Any}) at /Users/forets/.julia/dev/Plots/src/args.jl:1006
 [3] #plot#140(::Base.Iterators.Pairs{Symbol,Symbol,Tuple{Symbol},NamedTuple{(:markershape,),Tuple{Symbol}}}, ::Function, ::Interval{Float64,IntervalArithmetic.Interval{Float
64}}) at /Users/forets/.julia/dev/Plots/src/plot.jl:52
 [4] (::getfield(RecipesBase, Symbol("#kw##plot")))(::NamedTuple{(:markershape,),Tuple{Symbol}}, ::typeof(plot), ::Interval{Float64,IntervalArithmetic.Interval{Float64}}) at
./none:0
 [5] top-level scope at none:0
```

src/plot_recipes.jl
```diff
@@ -482,252 +271,136 @@ Plot an interval.
...
 
 ### Output
 
-An error, since plotting a hyperplane is not implemented yet (see #576).
+An error, since plotting unbounded sets is not implemented yet (see
+[#576](https://github.com/JuliaReach/LazySets.jl/issues/576)).
 """
-@recipe function plot_hyperplane(::Hyperplane{N}, ε::N=zero(N)) where {N<:Real}
-    error("cannot plot a hyperplane")
+@recipe function plot_polyhedron(P::AbstractPolyhedron{N}, ε::N=zero(N)
```
mforets: This choice disables plotting some bounded sets without apparent reason:

```julia
julia> plot(convert(HPolyhedron, rand(Hyperrectangle)))
ERROR: cannot plot an unbounded set
Stacktrace:
 [2] apply_recipe(::Dict{Symbol,Any}, ::HPolyhedron{Float64}, ::Float64) at /Users/forets/.julia/packages/RecipesBase/Uz5AO/src/RecipesBase.jl:275
 [3] apply_recipe(::Dict{Symbol,Any}, ::HPolyhedron{Float64}) at /Users/forets/.julia/packages/RecipesBase/Uz5AO/src/RecipesBase.jl:269
 [4] _process_userrecipes(::Plots.Plot{Plots.GRBackend}, ::Dict{Symbol,Any}, ::Tuple{HPolyhedron{Float64}}) at /Users/forets/.julia/dev/Plots/src/pipeline.jl:83
 [5] _plot!(::Plots.Plot{Plots.GRBackend}, ::Dict{Symbol,Any}, ::Tuple{HPolyhedron{Float64}}) at /Users/forets/.julia/dev/Plots/src/plot.jl:178
 [6] #plot#140(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::HPolyhedron{Float64}) at /Users/forets/.julia/dev/Plots/src/plot.jl:57
 [7] plot(::HPolyhedron{Float64}) at /Users/forets/.julia/dev/Plots/src/plot.jl:51
 [8] top-level scope at none:0
```

Possible fix:

```julia
if !isbounded(P)
    error(..)
else
   # call series for general LazySet
end
```

src/plot_recipes.jl
schillic: I would say this is a bug in `Plots` because we even set `seriestype := :path` in the recipe. Explicitly defining `seriestype` as in `plot(Interval(0.0, 1.0), seriestype=:path, markershape=:none)` works. Alternatively, you can use `plot(Interval(0.0, 1.0), marker=0)` (which in `GR` adds tiny dots; in `PyPlot` there are not even dots).

One reason why I removed this parameter is that one cannot (at least I did not find out how) pass keyword arguments in recursive calls to plot recipes. So we would have to introduce new recipes for each set type. And it does not work for lists of sets as well.

If you still want this parameter, I can add it back (with mild protest).

mforets: > Explicitly defining seriestype as in plot(Interval(0.0, 1.0), seriestype=:path, markershape=:none) works

good 👍  

> If you still want this parameter, I can add it back (with mild protest).

No,we dont need it back (now or later) let's add a comment in the plot docstring explaining the previous trick.

src/plot_recipes.jl
schillic: I removed the plot recipe for `AbstractPolyhedron`, generalized the plot recipe for `AbstractPolytope` to `AbstractPolyhedron`, and added boundedness checks to the recipes of `LazySet` and `AbstractPolyhedron`.

schillic:
Sorry, I added the changes to the old commit, but they got a bit tangled with a rebase with `master`. [Here](https://github.com/JuliaReach/LazySets.jl/compare/67b0f9b458c4e42a0e7dd0763d0823b5948af6b7..6c5ee58fb6e92a8a6572ee00c294c5286be3aefb#diff-768a1ab02b1e7272858ea65c59e2a3a4) is the new file.

schillic:
Wow, building time of the docs increased from <17 minutes to 28 minutes here. What is so different? I even reduced the total number of methods here...
There seems to be is something specific with our way to include docs that `Documenter` is terribly slow with.

schillic:
I finally understood now what is going on with the plotting.

The part that got terribly slow was the zonotope algorithm in the manual. There we plotted 100 resp. 200 sets, which normally is not too much, but here these sets are lazy and every set is the previous set inside another nesting level. I was actually surprised that this even worked. 100 sets take 30 seconds in `master`. 200 sets take 300 seconds in `master`.

In this branch I had introduced helper functions that are called from the loop for individual handling of each set in the list. Julia tries to be smart and compiles a new version for every new set (recall that we build nested lazy sets in that zonotope code, so each set is different type-wise) and this costs a lot (0.05 seconds every time, but this sums up). In `master` we hardcoded the overapproximation/vertex conversion inside the loop; hence julia cannot compile efficiently; this means that the second run is potentially slower (i guess in this application it does not matter), but also that there is less compilation overhead in the first run (which is way too big here).

Conclusion: I finally added the hardcoded version as an option (deactivated by default because I like the individual plot recipes more and in our `Reachability` applications the sets always have the same type so this won't happen (and anyway we have our own plot recipe in `Reachability` - which we maybe want to remove?)); when we use the zonotope algorithm with concrete operations (#1413), this problem will also vanish, but it is good to keep this fast option for other cases that might pop up.

PR #1401, opened 2019-05-25
#310 - Generalize sparse matrix type

### LINKED ISSUE #310: Generalize sparse matrix type
We always use `SparseMatrixCSC`. There is also `AbstractSparseMatrix`. I think we actually want the latter.
If so, we should first check if it affects performance (recently we discovered that using the abstract type can do that, but admittedly it was a different setting).

schillic:
Closes #310.

### DIFF
--- a/src/ExponentialMap.jl
+++ b/src/ExponentialMap.jl
@@ -54,8 +54,8 @@ This type is provided for use with very large and very sparse matrices.
 The evaluation of the exponential matrix action over vectors relies on the
 [Expokit](https://github.com/acroy/Expokit.jl) package.
 """
-struct SparseMatrixExp{N} <: AbstractMatrix{N}
-    M::SparseMatrixCSC{N, Int}
+struct SparseMatrixExp{N, MN<:AbstractSparseMatrix{N}} <: AbstractMatrix{N}
+    M::MN
 end
 
 SparseMatrixExp(M::Matrix) =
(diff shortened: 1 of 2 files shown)


src/LinearMap.jl
```diff
@@ -5,7 +5,7 @@ export LinearMap,
        constraints_list
 
 """
-    LinearMap{N<:Real, S<:LazySet{N}, NM, MAT<:AbstractMatrix{NM}} <: LazySet{N}
+    LinearMap{N<:Real, S<:LazySet{N}, MN, MAT<:AbstractMatrix{MN}} <: LazySet{N}
```
mforets: It wasn't called `MN`, since `MN` is used for Matrices of numeric type `N` as used above in `ProjectionSparseMatrixExp` and [in the gudelines](https://juliareach.github.io/JuliaReachDevDocs/latest/guidelines/#Coding-Guidelines-1]. To change something, we could change `MAT` to `MN` but then need another letter for the numeric type of the matrix.

schillic: Okay, I will drop the second commit then.

src/ExponentialMap.jl
```diff
@@ -406,10 +407,11 @@ Type that represents the projection of a sparse matrix exponential, i.e.,
 - `E` -- sparse matrix exponential
 - `R` -- right multiplication matrix
 """
-struct ProjectionSparseMatrixExp{N<:Real}
-    L::SparseMatrixCSC{N, Int}
+struct ProjectionSparseMatrixExp{N<:Real, MN1<:AbstractSparseMatrix{N},
```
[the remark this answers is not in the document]
mforets: > Let's also concretely type spmexp, as in spmexp::SparseMatrixExp{N, MN}.

I'm not 100% sure that this is convenient to do.

PR #1407, opened 2019-05-26
#1326 - Travis terminates building the docs

### LINKED ISSUE #1326: Travis terminates building the docs
Building the docs currently takes 700 seconds on Travis. However, Travis terminates builds that do not write any output within 600 seconds.
Currently there are some warnings about `ExponentialMap` docs missing (because we do not load the optional package `Expokit` for building the docs, which that type depends on). This is a dirty hack and makes the docs incomplete.
We should try to add an output when building docs instead (and load `Expokit`).

schillic:
Closes #1326.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,5 +19,5 @@ jobs:
       os: linux
       script:
         - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate();'
-        - julia --project=docs/ docs/make.jl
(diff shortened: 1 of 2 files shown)


schillic:
Apparently the warning comes too late :disappointed:
I have to think about a new solution then.

schillic:
@mforets found the *correct* solution to this: We increase the tolerance on Travis as described [here](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) :confetti_ball:

PR #1409, opened 2019-05-28
#378 - Add isapprox between sets

### LINKED ISSUE #378: Implement comparison operator for set types
We should support `X == Y` for set types `X` and `Y`. See [here](https://docs.julialang.org/en/stable/stdlib/math/#Base.:==) for the documentation. Particularly, the fallback is `===`, i.e., object identity.

There are two possible semantics:
1. The objects have the same type and we check equality recursively for all fields. This is also what the Julia docs suggest (`Collections should generally implement == by calling == recursively on all contents.`).
2. The objects represent the same set. This is very hard to achieve and up to numerical issues. I suggest that, if we want to have this feature, we use a different function name for it (like `isequivalent`).

Similarly, we should support the `≈` operator.

schillic:
Closes #378.

### DIFF
--- a/src/LazySet.jl
+++ b/src/LazySet.jl
@@ -1,4 +1,4 @@
-import Base: ==, copy
+import Base: ==, ≈, copy
 import Random.rand
 
 export LazySet,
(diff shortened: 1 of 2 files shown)


src/LazySet.jl
```diff
@@ -341,6 +345,61 @@ function ==(X::LazySet, Y::LazySet)
...
+```
+"""
+function ≈(X::LazySet, Y::LazySet)
+    # if the common supertype of X and Y is abstract, they cannot be compared
+    if isabstracttype(promote_type(typeof(X), typeof(Y)))
+        return false
+    end
+
+    for f in fieldnames(typeof(X))
+        if !(getfield(X, f) ≈ getfield(Y, f))
```
```suggestion
        if !_isapprox(getfield(X, f), getfield(Y, f))
```
so that it uses the defaults defined in `comparisons.jl` (we will add a feature so they can be tuned, eventually..)

schillic: This does not work because `_isapprox` is only defined between numbers.

PR #1410, opened 2019-05-28
#476 - Better document AbstractHPolygon

### LINKED ISSUE #476: Better document AbstractHPolygon
See #475.

schillic:
Closes #476.

### DIFF
--- a/src/HPolygon.jl
+++ b/src/HPolygon.jl
@@ -10,7 +10,8 @@ are sorted in counter-clockwise fashion with respect to their normal directions.
 
 ### Fields
 
-- `constraints`       -- list of linear constraints, sorted by the angle
+- `constraints`       -- list of linear constraints, sorted by the angle in
+                         counter-clockwise fashion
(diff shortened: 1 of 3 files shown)


src/HPolygon.jl
```diff
@@ -10,7 +10,8 @@ are sorted in counter-clockwise fashion with respect to their normal directions.
 
 ### Fields
 
-- `constraints`       -- list of linear constraints, sorted by the angle
+- `constraints`       -- list of linear constraints, sorted by the angle in
```
mforets: this was there before, `.. sorted by the angle` -- is this correct? if the algorithm may start in any position, then the constraints will not be sorted by the angle (by "the angle" i understand `a1 < a2 < ... < an` where `ai` is the angle)

schillic: True, I think we agreed that we want to give up this guarantee. Note that our sorting mechanisms will still guarantee this, but we do not make use of that property, so we should not require (and hence not guarantee) it.

PR #1411, opened 2019-05-29
#1383 - Convex hull of a triangle

### LINKED ISSUE #1383: Convex hull of a triangle
The convex hull for 3 points can be special cased.
See [here](https://stackoverflow.com/a/2122620) for an implementation.

### LINKED ISSUE #1418: Add convex_hull tests
We have a [file with unit tests](https://github.com/JuliaReach/LazySets.jl/blob/873bbf1d7104e8baa3c752ce2841cc9b1939c44f/test/unit_convex_hull.jl) that is not loaded in `runtests.jl`.
In fact, some of the tests would fail because `ispermutation` is not used correctly.

SebastianGuadalupe:
Closes #1383.
Closes #1418.

### DIFF
--- a/src/concrete_convex_hull.jl
+++ b/src/concrete_convex_hull.jl
@@ -28,8 +28,8 @@ The convex hull as a list of vectors with the coordinates of the points.
 
 ### Algorithm
 
-A pre-processing step treats the cases with `0`, `1` and `2` points in any dimension.
-For more than `3` points, the algorithm used depends on the dimension.
+A pre-processing step treats the cases with `0`, `1`, `2` and `3` points in any dimension.
+For `4` or more points, the algorithm used depends on the dimension.
 
 For the one-dimensional case we return the minimum and maximum points, in that order.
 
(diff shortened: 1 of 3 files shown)


mforets:
LGTM. Can you please add a unit test for the new special case?

On a separate note, seems reasonable to add [here](https://github.com/JuliaReach/LazySets.jl/pull/1411/files#diff-16eb30cefbe8ef8192a0feb0bbb50197R136) a robust check too (for another PR).

src/concrete_convex_hull.jl
```diff
@@ -28,8 +28,8 @@ The convex hull as a list of vectors with the coordinates of the points.
 
 ### Algorithm
 
-A pre-processing step treats the cases with `0`, `1` and `2` points in any dimension.
-For more than `3` points, the algorithm used depends on the dimension.
+A pre-processing step treats the cases with `0`, `1`, `2` and `3` points in any dimension.
+For more than `4` points, the algorithm used depends on the dimension.
```
schillic: This was ambiguous before, but we can take the chance and improve it now.
```suggestion
For `4` or more points, the algorithm used depends on the dimension.
```

src/concrete_convex_hull.jl
```diff
@@ -99,27 +99,10 @@ function convex_hull!(points::Vector{VN};
...
-                nothing
-            else
-                points[1], points[2] = p2, p1
-            end
-        end
-        return points
+    if m == 2 # two points
+        return _two_points_2d!(points)
+    elseif m == 3 # three points
+        return _three_points_2d!(points)
```
schillic: This is switch block is currently not correct (you call the functions with arbitrary dimension). I would suggest to move up the dimension checks (on `n`) and integrate this switch on `m` in the `n == 2` case.

src/concrete_convex_hull.jl
```diff
@@ -136,6 +119,48 @@ function convex_hull!(points::Vector{VN};
     end
 end
 
+function _two_points_2d!(points)
+
+    n = length(first(points))
+
+    if n == 1
```
schillic: Remove the case split on `n` because the function name suggests that you already checked that the points are two-dimensional.

src/concrete_convex_hull.jl
```diff
@@ -136,6 +119,48 @@ function convex_hull!(points::Vector{VN};
...
+    return points
+end
+
+function _three_points_2d!(points)
+    A, B, C = points[1], points[2], points[3]
+    turn = (A[2] - B[2]) * C[1] + (B[1] - A[1]) * C[2] + (A[1] * B[2] - B[1] * A[2])
+
+    if isapproxzero(turn)
+        # ABC are collinear
+        points[1], points[2] = points[argmin([A[1], B[1], C[1]])], points[argmax([A[1], B[1], C[1]])]
```
schillic: I think the following is equivalent.
```suggestion
        points[1], points[2] = min([A[1], B[1], C[1]]), max([A[1], B[1], C[1]])
```
However, I do not understand why it is sufficient to sort by x1 coordinate. What if they all lie on the line `x1 = c` for some `c`?

src/concrete_convex_hull.jl
```diff
@@ -136,6 +119,48 @@ function convex_hull!(points::Vector{VN};
...
+        pop!(points)
+        _two_points_2d!(points)
+    elseif turn < 0
```
schillic:
```suggestion
    elseif turn < zero(N)
```
For this to work, you need to annotate the type by writing `points::AbstractVector{<:AbstractVector{N}}) where {N<:Real}`.

test/unit_convex_hull.jl
```diff
@@ -33,6 +33,20 @@ for N in [Float64, Rational{Int}]
     convex_hull!(points_2D) # check in-place version
     @test points_2D == [[N(-1), N(-1)], [N(1), N(0)], [N(1), N(1)], [N(0), N(1)]]
 
+    # three vertex case in 2 dimension
```
schillic:
```suggestion
    # three vertex case in 2 dimensions
```

test/unit_convex_hull.jl
```diff
@@ -33,6 +33,20 @@ for N in [Float64, Rational{Int}]
...
+    ccw_points = [[N(1), N(1)], [N(-1), N(1)], [N(-1), N(0)]]
```
schillic: Simpler (all over this file):
```suggestion
    ccw_points = [N[1, 1], N[-1, 1], N[-1, 0]]
```

test/unit_convex_hull.jl
```diff
@@ -33,6 +33,20 @@ for N in [Float64, Rational{Int}]
...
+    @test  ccw_p == ccw_expr || ccw_p == circshift(ccw_expr, 1) || ccw_p == circshift(ccw_expr, 2) # points sorted in a counter-clockwise fashion
+    
+    cw_points = [[N(-1), N(1)], [N(1), N(1)], [N(-1), N(0)]]
+    cw_p = convex_hull!(cw_points)
+    cw_expr = [[N(-1), N(1)], [N(1), N(1)], [N(-1), N(0)]]
+    @test  cw_p == cw_expr || cw_p == circshift(cw_expr, 1) || cw_p == circshift(cw_expr, 2) # points sorted in clockwise fashion
+
+    aligned_points = [[N(1), N(1)], [N(2), N(2)], [N(3), N(3)]]
+    @test convex_hull!(aligned_points) == ispermutation([[N(1), N(1)], [N(2), N(2)]]) # points aligned
+
```
schillic: Please add unit tests for the problems mentioned above (1-dimensional case, 3-dimensional case, three points on a vertical line).

src/concrete_convex_hull.jl
schillic: You still need to add this type information.

src/concrete_convex_hull.jl
```diff
@@ -99,41 +99,64 @@ function convex_hull!(points::Vector{VN};
...
+    p1, p2 = points[1], points[2]
+    if p1 == p2  # check for redundancy
+        pop!(points)
+    elseif p1[1] > p2[1]
+        points[1], points[2] = p2, p1
+    end
+    return points
+end
+
+function _two_points_2d!(points)
```
schillic: I think the implementation of `_two_points_2d!` also works in the 1D case.
```julia
julia> [2.] <= [2.]
true

julia> [2.1] <= [2.]
false
```
So I think you can remove the 1D case and call the same function.

src/concrete_convex_hull.jl
schillic: Sorry, I was wrong here. Can you undo again? :wink:

schillic: I am also still concerned that this is incorrect if the x1 value is identical (then you probably get the same point twice).

src/concrete_convex_hull.jl
mforets: LazySets overrides the `<=` [fallback for vectors](https://github.com/JuliaLang/julia/blob/master/base/abstractarray.jl#L1759) so it doesn't work:
 
```julia
julia> [2.] <= [2.]
true

julia> using LazySets

julia> [2.] <= [2.]
ERROR: BoundsError: attempt to access 1-element Array{Float64,1} at index [2]
Stacktrace:
 [1] getindex at ./array.jl:729 [inlined]
 [2] <=(::Array{Float64,1}, ::Array{Float64,1}) at /Users/forets/.julia/dev/LazySets/src/AbstractPolygon.jl:199
 [3] top-level scope at none:0
```

I would suggest to keep this code split (another approach would be to use a branch in the `<=` code, but i don't support this idea at least not before checking if it can be done without regression).

src/concrete_convex_hull.jl
```diff
@@ -99,41 +99,76 @@ function convex_hull!(points::Vector{VN};
...
+                pop!(points)
+                pop!(points)
+                return points
+            else
+                points[1], points[2] = points[argmin([A[2], B[2], C[2]])], points[argmax([A[2], B[2], C[2]])]
+            end
+        else
+            points[1], points[2] = points[argmin([A[1], B[1], C[1]])], points[argmax([A[1], B[1], C[1]])]
+        end
+        pop!(points)
```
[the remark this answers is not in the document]
schillic: Okay, I take that back.

test/unit_convex_hull.jl
```diff
@@ -33,6 +33,19 @@ for N in [Float64, Rational{Int}]
     convex_hull!(points_2D) # check in-place version
     @test points_2D == [[N(-1), N(-1)], [N(1), N(0)], [N(1), N(1)], [N(0), N(1)]]
 
+    # three vertex case in 2 dimensions
+    ccw_points = [N[1, 1], N[-1, 1], N[-1, 0]]
+    ccw_p = convex_hull!(ccw_points)
+    ccw_expr = [N[1, 1], N[-1, 1], N[-1, 0]]
+    @test  ccw_p == ccw_expr || ccw_p == circshift(ccw_expr, 1) || ccw_p == circshift(ccw_expr, 2) # points sorted in a counter-clockwise fashion
```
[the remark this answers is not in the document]
mforets: It is deterministic, but this test reflects the weaker condition that implementations respect CCW ordering only.

schillic:
Have we actually evaluated whether this implementation is faster?

mforets:
I tried it, it was ~7x faster than the default algorithm.

PR #1415, opened 2019-05-31
#571 - Restrict iterative refinement to AbstractFloat

### LINKED ISSUE #571: Revise numeric type in iterative_refinement.jl
For some arguments we currently do not require `N`.
In general the algorithms rely on `norm`, which exists for `Float64` and `Float32`, but not for `Rational` types, so I suggest that we restrict `N <: AbstractFloat`.

schillic:
Closes #571.

I had first added a type restriction on the input, but we do use iterative refinement with `ε = Inf` sometimes in the tests. So I decided to instead just raise an informative error message if one tries to use a smaller value for `ε`.

### DIFF
--- a/src/Approximations/iterative_refinement.jl
+++ b/src/Approximations/iterative_refinement.jl
@@ -67,7 +67,7 @@ end
 
 """
     new_approx(S::LazySet, p1::Vector{N}, d1::Vector{N}, p2::Vector{N},
-               d2::Vector{N}) where {N<:Real}
+               d2::Vector{N}) where {N<:AbstractFloat}
 
 Create a `LocalApproximation` instance for the given excerpt of a polygonal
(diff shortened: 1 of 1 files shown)


mforets:
There are other functions in `iterative_refinement.jl` that have `{N<:Real}`, including the struct `LocalApproximation`. My proposal is to restrict `{N<:AbstractFloat}` in this file. The case `ε = Inf` is a case handled correctly in [this branch](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Approximations/overapproximate.jl#L43) so we don't need new type assertions in `iterative_refinement.jl`.

schillic:
I tried that first, but then several tests failed (but I cannot explain why). I guess I should investigate...

PR #1419, opened 2019-06-04
Decomposed linear map of CartesianProductArray

### LINKED ISSUE #1091: Decomposed linear map of CartesianProductArray

kostakoida:
Closes #1091.

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -334,7 +334,7 @@ function overapproximate(X::LazySet{N},
 end
 
 """
-    overapproximate(S::LazySet{N}, ::Union{Type{Interval}, Type{Interval{N}}}) where {N<:Real}
+    overapproximate(S::LazySet{N}, ::Type{<:Interval}) where {N<:Real}
 
 Return the overapproximation of a real unidimensional set with an interval.
 
@@ -352,7 +352,7 @@ An interval.
 The method relies on the exact conversion to `Interval`. Two support
 function evaluations are needed in general.
 """
-function overapproximate(S::LazySet{N}, ::Union{Type{Interval}, Type{Interval{N}}}) where {N<:Real}
+function overapproximate(S::LazySet{N}, ::Type{<:Interval}) where {N<:Real}
     @assert dim(S) == 1 "cannot overapproximate a $(dim(S))-dimensional set with an `Interval`"
     return convert(Interval, S)
 end
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -520,55 +520,51 @@ end
     overapproximate(lm::LinearMap{N, <:CartesianProductArray{N,
                             <:LazySet{N}}}, ::Type{<:CartesianProductArray}, oa=Hyperrectangle)  where {N}
 
-Overapproximating a lazy linear map of cartesian product array with template directions for each block.
+Decompose a lazy linear map of a cartesian product array.
 
 ### Input
 
 - `lm`   -- lazy linear map of cartesian product array
-- Type{<:CartesianProductArray} -- type to specify decomposed overapproximation
-- `oa` -- (concrete) direction representation
+- `CartesianProductArray` -- type for dispatch
+- `oa` -- approximation option for decomposition
 
 ### Output
 
-An `CartesianProductArray` with overapproximation of the each block with the directions from `dir`.
+A `CartesianProductArray` representing the decomposed linear map.
 """
 function overapproximate(lm::LinearMap{N, <:CartesianProductArray{N,
                             <:LazySet{N}}}, ::Type{<:CartesianProductArray}, oa=Hyperrectangle) where {N}
 
     @assert size(lm.M, 2) == dim(lm.X) "matrix needs to be commensurate with the cartesian product"
 
-    cp = lm.X
+    cpa = lm.X
     M = lm.M
 
     col_end_ind = 0
 
-    return decomposed_overapproximation(cp, M, col_end_ind, oa)
-end
-
-#template directions
-function decomposed_overapproximation(cpa::CartesianProductArray{N,<:LazySet{N}},
-                                      M::AbstractMatrix{N},
-                                      col_end_ind::Int, oa) where {N}
     array = allocate_result(oa, N)
     sizehint!(array,length(cpa.array))
 
-    for i in 1:size(M, 2)
-        ms = overapproximate_row_blocks(cpa, M, i)
+    i, n = 1, 0
+    for bi in cpa.array
+        n += dim(bi)
+        ms = blocks_linear_map(cpa, M, i, n)
         push!(array, overapproximate(ms, oa))
+        i += n
     end
 
     result = CartesianProductArray(array)
     return result
 end
 
-#overapproximation of linear map to each block
-function overapproximate_row_blocks(cpa::CartesianProductArray{N,<:LazySet{N}},
-                                    M::AbstractMatrix{N}, i::Int) where {N}
+
+function blocks_linear_map(cpa::CartesianProductArray{N,<:LazySet{N}},
+                                    M::AbstractMatrix{N}, row_start_ind::Int, row_end_ind::Int) where {N}
     col_start_ind, col_end_ind = 1, 0
     h_min_sum = MinkowskiSumArray()
     for bi in cpa.array
         col_end_ind += dim(bi)
-        push!(h_min_sum.array, LinearMap(M[i : i, col_start_ind : col_end_ind], bi))
+        push!(h_min_sum.array, LinearMap(M[row_start_ind : row_end_ind, col_start_ind : col_end_ind], bi))
         col_start_ind = col_end_ind + 1
     end
 
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -561,7 +561,7 @@ end
 function blocks_linear_map(cpa::CartesianProductArray{N,<:LazySet{N}},
                                     M::AbstractMatrix{N}, row_start_ind::Int, row_end_ind::Int) where {N}
     col_start_ind, col_end_ind = 1, 0
-    h_min_sum = MinkowskiSumArray()
+    h_min_sum = MinkowskiSumArray(length(cpa.array), N)
     for bi in cpa.array
         col_end_ind += dim(bi)
         push!(h_min_sum.array, LinearMap(M[row_start_ind : row_end_ind, col_start_ind : col_end_ind], bi))
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -535,14 +535,14 @@ An `CartesianProductArray` with overapproximation of the each block with the dir
 function overapproximate(lm::LinearMap{N, <:CartesianProductArray{N,
                             <:LazySet{N}}}, ::Type{<:CartesianProductArray}, oa=Hyperrectangle) where {N}
 
-    @assert !(size(lm.M, 2) == dim(lm.X)) "matrix needs to be commensurate with the cartesian product"
+    @assert size(lm.M, 2) == dim(lm.X) "matrix needs to be commensurate with the cartesian product"
 
     cp = lm.X
     M = lm.M
 
     col_end_ind = 0
 
-    return decomposed_overapproximation(cp, M, col_st_index, col_end_ind, oa)
+    return decomposed_overapproximation(cp, M, col_end_ind, oa)
 end
 
 #template directions
@@ -553,11 +553,11 @@ function decomposed_overapproximation(cp::CartesianProductArray{N,<:LazySet{N}},
     sizehint!(array,length(cp.array))
 
     for bi in cp.array
-        v_block, col_end_ind = overapproximate_block(bi, col_end_ind, M)
+        v_block, col_end_ind = overapproximate_lm_block(bi, col_end_ind, M)
         push!(array, Approximations.overapproximate(v_block, oa))
     end
 
-    result = CartesianProductArray{N}(array)
+    result = CartesianProductArray(array)
     return result
 end
 
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -593,7 +593,7 @@ function overapproximate_lm_block(bi:: LazySet{N}, col_st_index::Int, col_end_in
     matrices = Vector{Matrix{N}}()
 
     while end_ind <= size(M, 2)
-        push!(matrices, M[col_st_index : col_end_ind, start_ind : end_ind])
+        push!(matrices, @view M[col_st_index : col_end_ind, start_ind : end_ind])
         start_ind = end_ind + 1
         end_ind += n
     end
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -517,8 +517,8 @@ function overapproximate(cap::Intersection{N,
 end
 
 """
-    overapproximate(lm::LinearMap{N, <:CartesianProductArray{N, <:LazySet{N}}},
-                         dir::Type{<:AbstractDirections})::CartesianProductArray{N, HPolytope{N}} where {N}
+    overapproximate(lm::LinearMap{N, <:CartesianProductArray{N,
+                            <:LazySet{N}}}, ::Type{<:CartesianProductArray}, oa=Hyperrectangle)  where {N}
 
 Overapproximating a lazy linear map of cartesian product array with template directions for each block.
 
@@ -533,76 +533,61 @@ Overapproximating a lazy linear map of cartesian product array with template dir
 An `CartesianProductArray` with overapproximation of the each block with the directions from `dir`.
 """
 function overapproximate(lm::LinearMap{N, <:CartesianProductArray{N,
-                            <:LazySet{N}}}, ::Type{<:CartesianProductArray}, oa::Union{Type{<:AbstractDirections},Type{<:Hyperrectangle}, Type{<:Interval}}=Hyperrectangle) where {N}
+                            <:LazySet{N}}}, ::Type{<:CartesianProductArray}, oa=Hyperrectangle) where {N}
 
-    if !(size(lm.M, 2) == dim(lm.X))
-        error("Matrix needs to be commensurate with the cartesian product")
-    end
+    @assert !(size(lm.M, 2) == dim(lm.X)) "matrix needs to be commensurate with the cartesian product"
 
     cp = lm.X
     M = lm.M
 
-    col_st_index, col_end_ind = 0, 0
+    col_end_ind = 0
 
     return decomposed_overapproximation(cp, M, col_st_index, col_end_ind, oa)
 end
 
 #template directions
-function decomposed_overapproximation(cp::CartesianProductArray{N,<:LazySet{N}}, M::AbstractArray{N},
-            col_st_index::Int, col_end_ind::Int, oa::Type{<:AbstractDirections}) where {N}
-
-    array =  Vector{HPolytope{N}}()
-
-    sizehint!(array,length(cp.array))
-
-    for bi in cp.array
-        v_block, col_st_index, col_end_ind = overapproximate_block(bi, col_st_index, col_end_ind, M)
-
-        push!(array, Approximations.overapproximate(v_block, oa{N}(dim(v_block))))
-    end
-
-    result = CartesianProductArray{N, HPolytope{N}}(array)
-    return result
-end
-
-#Interval Hull
-function decomposed_overapproximation(cp::CartesianProductArray{N,<:LazySet{N}}, M::AbstractArray{N},
-            col_st_index::Int, col_end_ind::Int, oa::Union{Type{<:Hyperrectangle}, Type{<:Interval}}) where {N}
-
-    array =  Vector{oa{N}}()
+function decomposed_overapproximation(cp::CartesianProductArray{N,<:LazySet{N}}, M::AbstractMatrix{N},
+             col_end_ind::Int, oa) where {N}
+    array = allocate_result(oa, N)
 
     sizehint!(array,length(cp.array))
 
     for bi in cp.array
-        v_block, col_st_index, col_end_ind = overapproximate_block(bi, col_st_index, col_end_ind, M)
-
+        v_block, col_end_ind = overapproximate_block(bi, col_end_ind, M)
         push!(array, Approximations.overapproximate(v_block, oa))
     end
 
-    result = CartesianProductArray{N, oa{N}}(array)
+    result = CartesianProductArray{N}(array)
     return result
 end
 
+function allocate_result(oa, N)
+    return Vector{LazySet{N}}()
+end
+
+function allocate_result(oa::Type{<:LazySet}, N)
+    return Vector{oa{N}}()
+end
+
+function allocate_result(oa::Type{<:AbstractDirections}, N)
+    return Vector{HPolytope{N}}()
+end
+
 #overapproximation of linear map to each block
-function overapproximate_lm_block(bi:: LazySet{N}, col_st_index::Int, col_end_ind::Int, M::AbstractArray{N}) where {N}
+function overapproximate_lm_block(bi::LazySet{N}, col_end_ind::Int, M::AbstractMatrix{N}) where {N}
     col_st_index = col_end_ind + 1
     n = dim(bi)
     col_end_ind += n
 
-    start_ind, end_ind = 1, n
+    row_start_ind, row_end_ind = 1, n
     matrices = Vector{Matrix{N}}()
 
-    while end_ind <= size(M, 2)
-        push!(matrices, @view M[col_st_index : col_end_ind, start_ind : end_ind])
-        start_ind = end_ind + 1
-        end_ind += n
-    end
-
     v_block = MinkowskiSumArray()
-
-    for m in matrices
-        push!(v_block.array,LinearMap(m, bi))
+    while row_end_ind <= size(M, 2)
+        push!(v_block.array, LinearMap(M[col_st_index : col_end_ind, row_start_ind : row_end_ind], bi))
+        row_start_ind = row_end_ind + 1
+        row_end_ind += n
     end
 
-    return v_block, col_st_index, col_end_ind
+    return v_block, col_end_ind
 end
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -517,8 +517,8 @@ function overapproximate(cap::Intersection{N,
 end
 
 """
-    overapproximate(lm::LinearMap{N, <:CartesianProductArray{N,
-                            <:LazySet{N}}}, ::Type{<:CartesianProductArray}, oa=Hyperrectangle)  where {N}
+    overapproximate(lm::LinearMap{N, <:CartesianProductArray},
+                      ::Type{<:CartesianProductArray}, oa=Hyperrectangle)  where {N}
 
 Decompose a lazy linear map of a cartesian product array.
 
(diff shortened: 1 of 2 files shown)


src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,93 @@ function overapproximate(cap::Intersection{N,
...
+    M = lm.M
+
+    col_st_index, col_end_ind = 0, 0
+
+    return decomposed_overapproximation(cp, M, col_st_index, col_end_ind, oa)
+end
+
+#template directions
+function decomposed_overapproximation(cp::CartesianProductArray{N,<:LazySet{N}}, M::AbstractArray{N},
+            col_st_index::Int, col_end_ind::Int, oa::Type{<:AbstractDirections}) where {N}
```
schillic: There is no need for different methods, so you can move that code back to the main function. Just replace `oa{N}(dim(v_block))` by `oa` and `CartesianProductArray{N, HPolytope{N}}(array)` by `CartesianProductArray(array)` below.

One downside is that you cannot preallocate the `array`. So maybe you want to outsource this particular line:
```julia
function overapproximate(...)
    ...
    array = allocate_result(oa, N)
    ...
end

function allocate_result(oa, N)
    return Vector{LazySet{N}}()
end

function allocate_result(oa::Type{<:LazySet}, N)
    return Vector{oa{N}}()
end

function allocate_result(oa::oa::Type{<:AbstractDirections}, N)
    return Vector{HPolytope{N}}()
end
```

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,93 @@ function overapproximate(cap::Intersection{N,
...
+
+        push!(array, Approximations.overapproximate(v_block, oa))
+    end
+
+    result = CartesianProductArray{N, oa{N}}(array)
+    return result
+end
+
+#overapproximation of linear map to each block
+function overapproximate_lm_block(bi:: LazySet{N}, col_st_index::Int, col_end_ind::Int, M::AbstractArray{N}) where {N}
```
```suggestion
function overapproximate_lm_block(bi::LazySet{N}, col_end_ind::Int, M::AbstractMatrix{N}) where {N}
```
There is no need to pass `col_st_index` if you overwrite its value immediately.

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,93 @@ function overapproximate(cap::Intersection{N,
...
+    col_st_index = col_end_ind + 1
+    n = dim(bi)
+    col_end_ind += n
+
+    start_ind, end_ind = 1, n
```
```suggestion
    row_start_ind, row_end_ind = 1, n
```
(helps in understanding)

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,93 @@ function overapproximate(cap::Intersection{N,
...
+        push!(matrices, @view M[col_st_index : col_end_ind, start_ind : end_ind])
+        start_ind = end_ind + 1
+        end_ind += n
+    end
+
+    v_block = MinkowskiSumArray()
+
+    for m in matrices
+        push!(v_block.array,LinearMap(m, bi))
+    end
```
schillic: Simpler and probably better:
```julia
v_block = MinkowskiSumArray([LinearMap(m, bi) for m in matrices])
```
On second thought, it does not make sense to store all these matrices only to run through them in a second loop.
So better create the array and then store the `LinearMap`s in the loop above already.

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,93 @@ function overapproximate(cap::Intersection{N,
...
+    while end_ind <= size(M, 2)
+        push!(matrices, @view M[col_st_index : col_end_ind, start_ind : end_ind])
+        start_ind = end_ind + 1
+        end_ind += n
+    end
+
+    v_block = MinkowskiSumArray()
+
+    for m in matrices
+        push!(v_block.array,LinearMap(m, bi))
```
schillic: You cannot use `bi` here. You need to run through all sets in the original `CPA`.

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,93 @@ function overapproximate(cap::Intersection{N,
...
+    end
+
+    return v_block, col_st_index, col_end_ind
+end
```
schillic: Add at least one unit test for `oa == Hyperrectangle` and one for `oa == BoxDirections`. Choose the input set so that these tests uncover the error with `bi` above.

src/Approximations/overapproximate.jl
kostakoida: Why do we need:

```
function allocate_result(oa, N)
    return Vector{LazySet{N}}()
end
```
?

schillic: This is a fallback for other overapproximations.

src/Approximations/overapproximate.jl
kostakoida: Why?

schillic: Look at the formula on the top of page 11 in the paper. The Minkowski sum sums up terms where `X_j` changes.

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,93 @@ function overapproximate(cap::Intersection{N,
...
+### Output
+
+An `CartesianProductArray` with overapproximation of the each block with the directions from `dir`.
+"""
+function overapproximate(lm::LinearMap{N, <:CartesianProductArray{N,
+                            <:LazySet{N}}}, ::Type{<:CartesianProductArray}, oa::Union{Type{<:AbstractDirections},Type{<:Hyperrectangle}, Type{<:Interval}}=Hyperrectangle) where {N}
+
+    if !(size(lm.M, 2) == dim(lm.X))
+        error("Matrix needs to be commensurate with the cartesian product")
+    end
```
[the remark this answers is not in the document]
schillic: Note that you have to invert the condition!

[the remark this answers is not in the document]
kostakoida: > Note that you have to invert the condition!

It's still under WIP. I'll update title once I'll think that you can review again.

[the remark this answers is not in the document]
schillic: Okay; you can also just re-request a review.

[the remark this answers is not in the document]
kostakoida: Ok

test/unit_overapproximate.jl
```diff
@@ -129,4 +129,22 @@ for N in [Float64, Float32]
     Y_zonotope = overapproximate(Y, Zonotope) # overapproximate with a zonotope
     @test Y_polygon ⊆ Y_zonotope
     @test !(Y_zonotope ⊆ Y_polygon)
+
+    #Decomposed approximation of lazy linear map of CartesianProductArray
+    i1 = Interval([0, 1])
+    h = Hyperrectangle(low=[3., 3.], high=[5., 5.])
```
schillic: `h` should not have the same x1 and x2 coordinates to see the difference in the formula.

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,73 @@ function overapproximate(cap::Intersection{N,
...
+"""
+    overapproximate(lm::LinearMap{N, <:CartesianProductArray{N,
+                            <:LazySet{N}}}, ::Type{<:CartesianProductArray}, oa=Hyperrectangle)  where {N}
+
+Overapproximating a lazy linear map of cartesian product array with template directions for each block.
+
+### Input
+
+- `lm`   -- lazy linear map of cartesian product array
+- Type{<:CartesianProductArray} -- type to specify decomposed overapproximation
```
```suggestion
- `CartesianProductArray` -- type for dispatch
```

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,73 @@ function overapproximate(cap::Intersection{N,
     return overapproximate(swap(cap), dir; kwargs...)
 end
 
+"""
+    overapproximate(lm::LinearMap{N, <:CartesianProductArray{N,
+                            <:LazySet{N}}}, ::Type{<:CartesianProductArray}, oa=Hyperrectangle)  where {N}
+
+Overapproximating a lazy linear map of cartesian product array with template directions for each block.
```
```suggestion
Decompose a lazy linear map of a cartesian product array.
```

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,73 @@ function overapproximate(cap::Intersection{N,
...
+
+### Input
+
+- `lm`   -- lazy linear map of cartesian product array
+- Type{<:CartesianProductArray} -- type to specify decomposed overapproximation
+- `oa` -- (concrete) direction representation
```
```suggestion
- `oa` -- approximation option for decomposition
```

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,73 @@ function overapproximate(cap::Intersection{N,
...
+
+### Output
+
+An `CartesianProductArray` with overapproximation of the each block with the directions from `dir`.
```
```suggestion
A `CartesianProductArray` representing the decomposed linear map.
```

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,73 @@ function overapproximate(cap::Intersection{N,
...
+    cp = lm.X
+    M = lm.M
+
+    col_end_ind = 0
+
+    return decomposed_overapproximation(cp, M, col_end_ind, oa)
+end
+
+#template directions
+function decomposed_overapproximation(cpa::CartesianProductArray{N,<:LazySet{N}},
```
schillic: Merge this function with the one above.

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,73 @@ function overapproximate(cap::Intersection{N,
...
+                                      M::AbstractMatrix{N},
+                                      col_end_ind::Int, oa) where {N}
+    array = allocate_result(oa, N)
+    sizehint!(array,length(cpa.array))
+
+    for i in 1:size(M, 2)
+        ms = overapproximate_row_blocks(cpa, M, i)
```
schillic: Choose a different function name (because the overapproximation happens later).

schillic: You cannot iterate over all integers. You need to iterate over all blocks. In general the blocks are not necessarily one-dimensional.
I am surprised that this was not detected by the test.

test/unit_overapproximate.jl
```diff
@@ -129,4 +129,22 @@ for N in [Float64, Float32]
     Y_zonotope = overapproximate(Y, Zonotope) # overapproximate with a zonotope
     @test Y_polygon ⊆ Y_zonotope
     @test !(Y_zonotope ⊆ Y_polygon)
+
+    #Decomposed approximation of lazy linear map of CartesianProductArray
+    i1 = Interval([0, 1])
```
schillic: Use `N[0, 1]` (also for the other examples). This adds tests for different numeric types.

test/unit_overapproximate.jl
```diff
@@ -129,4 +129,22 @@ for N in [Float64, Float32]
...
+    oa = overapproximate(lm, Hyperrectangle)
+    oa_box = overapproximate(lm, Approximations.BoxDirections)
+    d_oa_d_hp = overapproximate(lm, CartesianProductArray)
+    d_oa_d_box = overapproximate(lm, CartesianProductArray, Approximations.BoxDirections)
+    oa_d_hp = overapproximate(d_oa_d_hp)
+    oa_d_box = overapproximate(d_oa_d_box, Approximations.BoxDirections)
+
+    @test oa == oa_d_hp
+    @test oa_box == oa_d_box
+
```
schillic: The tests are good, but you also need to add tests for the concrete results.
```julia
for (oax, set_type) in [(d_oa_d_hp, Hyperrectangle), (d_oa_d_box, HPolytope)]
    @test oax isa CartesianProductArray
    arr = array(oa)
    @test length(arr) == 2 && dim(arr[1]) = 1 && dim(arr[2]) = 2
    @test all(X -> X isa st_type, arr)
end
```

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,73 @@ function overapproximate(cap::Intersection{N,
...
+        push!(array, overapproximate(ms, oa))
+    end
+
+    result = CartesianProductArray(array)
+    return result
+end
+
+#overapproximation of linear map to each block
```
```suggestion
# overapproximation of Minkowski sum of linear maps for each block in the row block
```

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,69 @@ function overapproximate(cap::Intersection{N,
...
+- `lm`   -- lazy linear map of cartesian product array
+- `CartesianProductArray` -- type for dispatch
+- `oa` -- approximation option for decomposition
+
+### Output
+
+A `CartesianProductArray` representing the decomposed linear map.
+"""
+function overapproximate(lm::LinearMap{N, <:CartesianProductArray{N,
+                            <:LazySet{N}}}, ::Type{<:CartesianProductArray}, oa=Hyperrectangle) where {N}
```
schillic: You can remove the type annotations from `CartesianProductArray (`{N, <:LazySet{N}}` (there is no information gain from having it, so it makes it just harder to read) (also update the comment).

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,69 @@ function overapproximate(cap::Intersection{N,
...
+
+    @assert size(lm.M, 2) == dim(lm.X) "matrix needs to be commensurate with the cartesian product"
```
schillic: We can actually remove this assertion because the constructor of `LinearMap` already asserts it.

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,69 @@ function overapproximate(cap::Intersection{N,
...
+
+    cpa = lm.X
+    M = lm.M
+
+    col_end_ind = 0
```
schillic: not used

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,69 @@ function overapproximate(cap::Intersection{N,
...
+
+    array = allocate_result(oa, N)
+    sizehint!(array,length(cpa.array))
+
+    i, n = 1, 0
```
schillic: We use `n` for the dimension of a set. Just call them `row_start_ind` and `row_end_ind`?

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,69 @@ function overapproximate(cap::Intersection{N,
...
+
+    @assert size(lm.M, 2) == dim(lm.X) "matrix needs to be commensurate with the cartesian product"
+
+    cpa = lm.X
+    M = lm.M
+
+    col_end_ind = 0
+
+    array = allocate_result(oa, N)
+    sizehint!(array,length(cpa.array))
```
```suggestion
    sizehint!(array, length(cpa.array))
```

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,69 @@ function overapproximate(cap::Intersection{N,
...
+
+    result = CartesianProductArray(array)
+    return result
+end
+
+
+function blocks_linear_map(cpa::CartesianProductArray{N,<:LazySet{N}},
+                                    M::AbstractMatrix{N}, row_start_ind::Int, row_end_ind::Int) where {N}
+    col_start_ind, col_end_ind = 1, 0
+    h_min_sum = MinkowskiSumArray(length(cpa.array), N)
```
schillic: I would prefer to allocate a `Vector{LazySet{N}}` with a `sizehint!` and in the end return a `MinkowskiSumArray(h_min_sum)`, just as you do it for the `CPA` above.

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,69 @@ function overapproximate(cap::Intersection{N,
...
+
+    array = allocate_result(oa, N)
+    sizehint!(array,length(cpa.array))
+
+    i, n = 1, 0
+    for bi in cpa.array
+        n += dim(bi)
+        ms = blocks_linear_map(cpa, M, i, n)
+        push!(array, overapproximate(ms, oa))
+        i += n
```
schillic: Make this line consistent with the symmetric one in the loop below (either `i += n` or `i = n + 1` in both cases).

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,69 @@ function overapproximate(cap::Intersection{N,
...
+end
+
+
+function blocks_linear_map(cpa::CartesianProductArray{N,<:LazySet{N}},
+                                    M::AbstractMatrix{N}, row_start_ind::Int, row_end_ind::Int) where {N}
+    col_start_ind, col_end_ind = 1, 0
+    h_min_sum = MinkowskiSumArray(length(cpa.array), N)
+    for bi in cpa.array
+        col_end_ind += dim(bi)
+        push!(h_min_sum.array, LinearMap(M[row_start_ind : row_end_ind, col_start_ind : col_end_ind], bi))
```
schillic: `row_start_ind:row_end_ind` is constant in the loop, so you can pull it out.

schillic:
@mforets: Should we rename the function to `decompose`?

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,67 @@ function overapproximate(cap::Intersection{N,
...
+    end
+
+    result = CartesianProductArray(array)
+    return result
+end
+
+
+function blocks_linear_map(cpa::CartesianProductArray,
+                           M::AbstractMatrix{N},
+                           row_range::UnitRange{Int64}) where {N}
```
```suggestion
                           row_range::UnitRange{Int}) where {N}
```

kostakoida: Always forget about it:(

schillic:
Here is another test case.
```julia
julia> lm = N[1 2; 0 1] * CartesianProductArray([Interval(N[0, 1]), Interval(N[2, 3])])
julia> oa = overapproximate(lm, CartesianProductArray)

julia> plot([lm, oa])
```
decomposed_map

kostakoida:
> Here is another test case.
> 
> ```julia
> julia> lm = N[1 2; 0 1] * CartesianProductArray([Interval(N[0, 1]), Interval(N[2, 3])])
> julia> oa = overapproximate(lm, CartesianProductArray)
> 
> julia> plot([lm, oa])
> ```
> 
> decomposed_map

kostakoida:
> Here is another test case.
> 
> ```julia
> julia> lm = N[1 2; 0 1] * CartesianProductArray([Interval(N[0, 1]), Interval(N[2, 3])])
> julia> oa = overapproximate(lm, CartesianProductArray)
> 
> julia> plot([lm, oa])
> ```
> 
> decomposed_map

do you want me to add this test and check that lm is a subset of oa?

schillic:
It was more meant as a sanity check for me because the three-dimensional test you have is hard to imagine. But sure, you can add it.

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,67 @@ function overapproximate(cap::Intersection{N,
...
+    result = CartesianProductArray(array)
+    return result
+end
+
+
+function blocks_linear_map(cpa::CartesianProductArray,
+                           M::AbstractMatrix{N},
+                           row_range::UnitRange{Int}) where {N}
+    col_start_ind, col_end_ind = 1, 0
+    array = Vector{LazySet{N}}()
```
schillic: Since we always put a `LinearMap` in this array, we can change the type to `Vector{LazySet{N}}`

mforets: I would try to use the type information of `cpa` as well.

kostakoida: In general, each block of `cpa` might have different type.

mforets: Actually why don't you make `array` a `MinkowskiSumArray`?

kostakoida: See https://github.com/JuliaReach/LazySets.jl/pull/1419#discussion_r290609335

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +516,67 @@ function overapproximate(cap::Intersection{N,
...
+    row_start_ind, row_end_ind = 1, 0
+    for bi in cpa.array
+        row_end_ind += dim(bi)
+        ms = blocks_linear_map(cpa, M, row_start_ind : row_end_ind)
+        push!(array, overapproximate(ms, oa))
+        row_start_ind += row_end_ind
+    end
+
+    result = CartesianProductArray(array)
+    return result
```
mforets: Return directly? `return CartesianProductArray(array)`

src/Approximations/overapproximate.jl
mforets: It *is* possible to create a CPA with different set types, but i think it is not the correct approach -- with performance in mind -- to write the `overapproximate` functions in a type-agnostic way straight from the beginning. Moreover, in `Reachability.jl` benchmarks we have not used different set types for different blocks that i remember.

mforets: For instance in `Reachability.jl`, i think we should move in the direction of writing type-stable versions of the decomposed algo and the hybrid scheme.

schillic: I think the default decomposition using the block structure `[2,  1]` will choose the types `[Hyperrectangle, Interval]`, which is mixed.

mforets: Let me reformulate: there are cases where the set type is mixed (the `cpa` is of type `CartesianProductArray{N, LazySet{N}}`, and others where it isn't (the `cpa` is of type `CartesianProductArray{N, Hyperrectangle{N}}` or other concrete type).

This implementation covers the first case, but ignores the information of the second one when there is some.

schillic: So if the input is `CartesianProductArray{N, T}`, then we should create a `Vector{LinearMap{N, T}}`?

mforets: Yes,

```julia
function blocks_linear_map(cpa::CartesianProductArray{N, T},
                           M::AbstractMatrix{N},
                           row_range::UnitRange{Int}) where {N, T<:LazySet{N}}
    col_start_ind, col_end_ind = 1, 0
    array = Vector{LinearMap{N, T}}()
...
```

test/unit_overapproximate.jl
```diff
@@ -47,12 +47,36 @@ for N in [Float64, Rational{Int}, Float32]
...
-    
+
     # overapproximation of the lazy linear map of a hyperrectangular set
     H = Hyperrectangle(N[0, 0], N[1/2, 1])
     M = Diagonal(N[2, 2])
     OA = overapproximate(M*H, Hyperrectangle)
     @test OA isa Hyperrectangle && OA.center == N[0, 0] && OA.radius == N[1, 2]
+
+    #overapproximation of Minkowski sum of linear maps for each block in the row block
+    i1 = Interval(N[0, 1])
```
mforets: Can you please add the case of the CPA of three intervals? `CartesianProductArray([i1, i1, i1])` approximated with a `CartesianProductArray` of intervals.

src/Approximations/overapproximate.jl
mforets: Below is the full code. This implementation receives a typed output set type, such as the cartesian product array of intervals, or hyperrectangles, or `AbstractHyperrectangle`. There is a second signature that allows to receive a third argument, the template direction. 


```julia
function overapproximate(lm::LinearMap{N, CartesianProductArray{N, T}},
                           ::Type{CartesianProductArray{N, O}}) where {N, T<:LazySet{N}, O<:LazySet{N}}

    M, cpa = lm.M, lm.X
    array = Vector{O}(undef, length(cpa.array))
    return _overapproximate_lm_cpa!(M, cpa, array, O)
end

function overapproximate(lm::LinearMap{N, CartesianProductArray{N, T}},
                           ::Type{CartesianProductArray{N, HPolytope{N}}},
                        dir::AbstractDirections{N}) where {N, T<:LazySet{N}}
    M, cpa = lm.M, lm.X
    array = Vector{HPolytope{N}}(undef, length(cpa.array))
    return _overapproximate_lm_cpa!(M, cpa, array, O)
end

function _overapproximate_lm_cpa!(M, cpa, array, O)
    row_start_ind, row_end_ind = 1, 0
    @inbounds for (i, bi) in enumerate(cpa.array)
        row_end_ind += dim(bi)
        ms = blocks_linear_map(cpa, M, row_start_ind : row_end_ind)
        array[i] = overapproximate(ms, O)
        row_start_ind += row_end_ind
    end
    return CartesianProductArray(array)
end

function blocks_linear_map(cpa::CartesianProductArray{N, T},
                           M::AbstractMatrix{N},
                           row_range::UnitRange{Int}) where {N, T<:LazySet{N}}
    array = Vector{LinearMap{N, <:T}}(undef, length(cpa.array))

    col_start_ind, col_end_ind = 1, 0
    @inbounds for (i, bi) in enumerate(cpa.array)
        col_end_ind += dim(bi)
        array[i] = LinearMap(M[row_range, col_start_ind : col_end_ind], bi)
        col_start_ind += col_end_ind
    end

    return MinkowskiSumArray(array)
end
```

For the tests this will require to write for example,

```julia
d_oa_d_hp = overapproximate(lm, CartesianProductArray{N, Hyperrectangle{N}})
```

A "lazy" signature  `d_oa_d_hp = overapproximate(lm, CartesianProductArray)` which decomposes into the CPA of hyperrectangles by default is (maybe) doable without too much changes, perhaps with `invoke`, but i don't know if we should bother.

schillic:
> Should we rename the function to decompose?

I changed my mind. We compute what we call the decomposed map, which is generally an overapproximation of the operation sequence 1) apply the high-dimensional map and 2) decompose.

src/Approximations/overapproximate.jl
kostakoida: I believe this code doesn't cover mixed set type.

schillic:
> > Should we rename the function to decompose?
>
> I changed my mind. We compute what we call the decomposed map, which is generally an overapproximation of the operation sequence 1) apply the high-dimensional map and 2) decompose.

Actually, I changed my mind again. Is there a difference? I rather think these things are equivalent.

src/Approximations/overapproximate.jl
schillic: > I believe this code doesn't cover mixed set type.

Why? If you are worried about the type `T`: that is (by definition) a common supertype of all elements in `cpa.array`.

src/Approximations/overapproximate.jl
```diff
@@ -352,11 +352,10 @@ An interval.
 The method relies on the exact conversion to `Interval`. Two support
 function evaluations are needed in general.
 """
-function overapproximate(S::LazySet{N}, ::Type{Interval}) where {N<:Real}
+function overapproximate(S::LazySet{N}, ::Union{Type{Interval}, Type{Interval{N}}}) where {N<:Real}
```
schillic: The correct way should be `Type{<:Interval}`.

test/unit_overapproximate.jl
```diff
@@ -47,12 +47,56 @@ for N in [Float64, Rational{Int}, Float32]
...
+    end
+
+    i1 = Interval(N[0, 1])
+    i2 = Interval(N[2, 3])
+    i3 = Interval(N[1, 4])
+    M = N[1 2; 0 1]
+    cpa = CartesianProductArray([i1, i2])
+    lm = M * cpa
+    d_oa = overapproximate(lm, CartesianProductArray{N, Interval{N}})
+    oa = overapproximate(lm, 1e-8)
```
schillic: This does not work for `Rational{Int}`, so you have to either move it this block of tests to the next loop or use a different overapproximation method (e.g., with template directions).

src/Approximations/overapproximate.jl
schillic: > I believe this code doesn't cover mixed set type.

The best way is to give a counterexample in the form of a test case.

mforets:
Do you propose to merge `decompose` and `overapproximate` functions? They have close similarities, indeed, although `decompose` let's you play with the partition, while `overapproximate` assumes that the partition is fixed. We can continue the discussion in another issue..

src/Approximations/overapproximate.jl
schillic: We actually have a test with mixed set types.

PR #1420, opened 2019-06-04
Update README

schillic:


README.md
```diff
@@ -19,18 +19,19 @@
...
-Pkg.add("LazySets")
+pkg> add LazySets
 ```
 
-If you want to install the latest development version, do:
+To install the latest development version, instead use:
 
 ```julia
-Pkg.clone("https://github.com/JuliaReach/LazySets.jl.git")
+pkg> dev LazySets
```
mforets:
```suggestion
pkg> add LazySets#master
```

schillic: What is the difference?

mforets: We can add a link to https://docs.julialang.org/en/v1/stdlib/Pkg/index.html for the different uses (and `dev` in particular).

schillic: I would say that `dev` is the preferred way to develop a package (because it gets moved to your `dev` folder).
I can add a link to [adding registered packages](https://julialang.github.io/Pkg.jl/v1/managing-packages/#Adding-registered-packages-1) and [developing packages](https://julialang.github.io/Pkg.jl/v1/managing-packages/#Developing-packages-1).
But generally I prefer to keep the instructions short and simple. If one does not know how to use Julia, one has to consult a tutorial.

mforets: Indeed, `dev` clones the full git repo for development purpose (i don't think this is the case when using `add LazySets#master`).

The updated version looks good to me 👍

PR #1421, opened 2019-06-05
Decomposed intersection for CartesianProductArray

kostakoida:
Example of the algorithm:

Assume we have a cartesian product array: `I_1 \times I_2 \times I_3 \cap Y` where `Y = Halfspace([1.,1., 0.], 1.)`. It means we take union of I_1 and I_2. Basically for union I run `HPolytope(constraints_list(CartesianProductArray(I_1, I_2)))`. Resulted set I intersect with projection of `Y` onto variables 1 and 2.  Now we need to decompose it back to the original blocks`\hat{I_1} \times \hat{I_2} `.  This is the place where we cannot get exact result and need an overapproximation. 
In last step we take `\hat{I_1} \times \hat{I_2} ` and merge it with unused blocks (here it is `I_3`). So the final set is `\hat{I_1} \times \hat{I_2} \times I_3`

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -1,3 +1,4 @@
+using LazySets: block_to_dimension_indices, substitute_blocks, get_constrained_lowdimset
 """
     overapproximate(X::S, ::Type{S}) where {S<:LazySet}
 
@@ -910,4 +911,61 @@ function _overapproximate_lm_cpa!(arr, M, cpa, overapprox_option)
     end
 
     return CartesianProductArray(arr)
+end
+
+"""
+    overapproximate(cap::Intersection{N,
+                            <:CartesianProductArray{N},
+                            <:AbstractPolyhedron{N}},
+                       ::Type{CartesianProductArray}, oa) where {N}
+
+Return the intersection of the cartesian product of a finite number of convex sets and a polyhedron.
+
+### Input
+
+ - `cap` -- Lazy intersection of cartesian product array and polyhedron
+ - `CartesianProductArray` -- type for dispatch
+ - `oa`  -- template directions
+
+### Output
+
+The intersection between `cpa` and `P` with overapproximation in each constrained block.
+
+### Algorithm
+
+The intersection is only needed to be taken in the elements of the cartesian product array (subsets of
+variables, or "blocks") which are constrained in `P`.
+Hence we first search for constrained blocks and then take the intersection of
+a lower-dimensional Cartesian product of these blocks with the projection of `Y`
+onto the variables of these blocks. (This projection is syntactic and exact.)
+The result is a `CartesianProductArray` with the same block structure as in `X`.
+However, when we decompose back our set we overapproximate during projection operation,
+therefore we need to specify overapproximation strategy (it is Hyperrectangle by default)
+"""
+function overapproximate(cap::Intersection{N,
+                                            <:CartesianProductArray{N},
+                                            <:AbstractPolyhedron{N}},
+                            ::Type{CartesianProductArray}, oa) where {N}
+
+    cpa, P = cap.X, cap.Y
+
+    cpa_low_dim, vars, block_structure, blocks = get_constrained_lowdimset(cpa, P)
+
+    low_intersection = intersection(cpa_low_dim, project(P, vars))
+
+    if isempty(low_intersection)
+        return EmptySet{N}()
+    end
+
+    decomposed_low_set = decompose(low_intersection, block_structure, oa)
+
+    return substitute_blocks(decomposed_low_set, cpa, blocks)
+end
+
+# symmetric method
+function overapproximate(cap::Intersection{N,
+                                            <:AbstractPolyhedron{N},
+                                            <:CartesianProductArray{N}},
+                            ::Type{CartesianProductArray}, oa) where {N}
+    overapproximate(Intersection(cap.Y, cap.X), oa)
 end
--- a/src/CartesianProduct.jl
+++ b/src/CartesianProduct.jl
@@ -3,7 +3,8 @@ import Base: *, ∈, isempty
 export CartesianProduct,
        CartesianProductArray,
        CartesianProduct!,
-       array
+       array,
+       same_block_structure
 
 """
     CartesianProduct{N<:Real, S1<:LazySet{N}, S2<:LazySet{N}} <: LazySet{N}
@@ -561,4 +562,146 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
         end
     end
     return true
+end
+
+"""
+    block_structure(cpa::CartesianProductArray{N}) where {N}
+
+Returns an array containing the dimension ranges of each block in a `CartesianProductArray`.
+
+### Input
+
+- `cpa` -- Cartesian product array
+
+### Output
+
+A vector of ranges
+
+### Example
+
+```jldoctest
+julia> using LazySets: block_structure
+
+julia> cpa = CartesianProductArray([BallInf(zeros(n), 1.0) for n in [3, 1, 2]]);
+
+julia> block_structure(cpa)
+3-element Array{UnitRange{Int64},1}:
+ 1:3
+ 4:4
+ 5:6
+```
+"""
+function block_structure(cpa::CartesianProductArray{N}) where {N}
+    result = Vector{UnitRange{Int}}(undef, length(array(cpa)))
+    start_index = 1
+    @inbounds for (i, bi) in enumerate(array(cpa))
+        end_index = start_index + dim(bi) - 1
+        result[i] = start_index:end_index
+        start_index = end_index + 1
+    end
+    return result
+end
+
+
+"""
+    block_to_dimension_indices(cpa::CartesianProductArray{N}, vars::Vector{Int}) where {N}
+
+Returns a vector mapping block index `i`
+to tuple `(f, l)` such that either `f = l = -1` or `f` is the first dimension index and `l` is the last dimension
+index of the `i`-th block, depending on whether one of the block's dimension indices is specified in `vars`.
+
+### Input
+
+- `cpa` -- Cartesian product array
+- `vars` -- list containing the variables of interest, sorted in ascending order
+
+### Output
+
+(i) A vector of tuples, where values in tuple relate to range of dimensions in the i-th block.
+(ii) Number of constrained blocks
+
+### Example
+
+```jldoctest
+julia> using LazySets: block_to_dimension_indices
+
+julia> cpa = CartesianProductArray([BallInf(zeros(n), 1.0) for n in [1, 3, 2, 3]]);
+
+julia> block_to_dimension_indices(cpa, [2, 4, 8])
+(Tuple{Int64,Int64}[(-1, -1), (2, 4), (-1, -1), (7, 9)], 2)
+```
+This vector represents the mapping "second block from dimension 2 to dimension 4,
+fourth block from dimension 7 to dimension 9."
+These blocks contain the dimensions specified in `[2, 4, 8]`.
+Number of constrained variables here is 2 (2nd and 4th blocks)
+"""
+function block_to_dimension_indices(cpa::CartesianProductArray{N}, vars::Vector{Int}) where {N}
+    result = fill((-1, -1), length(array(cpa)))
+    non_empty_length = 0
+    start_index, end_index = 1, 0
+    v_i = 1
+    @inbounds for i in 1:length(cpa.array)
+        end_index += dim(cpa.array[i])
+        if v_i <= length(vars) && vars[v_i] <= end_index
+            result[i] = (start_index, end_index)
+            non_empty_length += 1
+            while v_i <= length(vars) && vars[v_i] <= end_index
+                v_i += 1
+            end
+        end
+        if v_i > length(vars)
+            break
+        end
+        start_index = end_index + 1
+    end
+    return result, non_empty_length
+end
+
+#same but for all variables
+function block_to_dimension_indices(cpa::CartesianProductArray{N}) where {N}
+    result = Vector{Tuple{Int, Int}}(undef, length(cpa.array))
+
+    start_index, end_index = 1, 0
+    @inbounds for i in 1:length(cpa.array)
+        end_index += dim(cpa.array[i])
+        result[i] = (start_index, end_index)
+        start_index = end_index + 1
+    end
+    return result, length(cpa.array)
+end
+
+"""
+    substitute_blocks(low_dim_cpa::CartesianProductArray{N},
+                         orig_cpa::CartesianProductArray{N},
+                           blocks::Vector{Tuple{Int,Int}}) where {N}
+
+Return merged Cartesian Product Array between original CPA and some low-dimensional CPA,
+which represents updated subset of variables in specified blocks.
+
+### Input
+
+- `low_dim_cpa` -- low-dimensional cartesian product array
+- `orig_cpa` -- original high-dimensional Cartesian product array
+- `blocks` -- index of the first variable in each block of `orig_cpa`
+
+### Output
+
+Merged cartesian product array
+"""
+function substitute_blocks(low_dim_cpa::CartesianProductArray{N},
+                           orig_cpa::CartesianProductArray{N},
+                           blocks::Vector{Tuple{Int,Int}}) where {N}
+
+    array = Vector{LazySet{N}}(undef, length(orig_cpa.array))
+    index = 1
+    for bi in 1:length(orig_cpa.array)
+        start_ind, end_index = blocks[bi]
+        if start_ind == -1
+            array[bi] = orig_cpa.array[bi]
+        else
+            array[bi] = low_dim_cpa.array[index]
+            index += 1
+        end
+    end
+    return CartesianProductArray(array)
 end
--- a/src/LazySets.jl
+++ b/src/LazySets.jl
@@ -89,7 +89,13 @@ include("Rectification.jl")
 # =============================
 # Conversions between set types
 # =============================
+include("intersection_helper.jl")
 include("convert.jl")
+
+# =====================
+# Approximations module
+# =====================
+include("Approximations/Approximations.jl")
 
 # ===========================
 # Concrete operations on sets
@@ -103,11 +109,6 @@ include("difference.jl")
 # Aliases
 # =======
 include("aliases.jl")
-
-# =====================
-# Approximations module
-# =====================
-include("Approximations/Approximations.jl")
 
 # ==========================
 # Parallel algorithms module
--- a/src/Translation.jl
+++ b/src/Translation.jl
@@ -10,7 +10,7 @@ export Translation,
 
 Type that represents a lazy translation.
 
-The translation of set `X` along vector `v` is the map: 
+The translation of set `X` along vector `v` is the map:
 
 ```math
 x ↦ x + v,\\qquad x ∈ X
@@ -142,7 +142,7 @@ struct Translation{N<:Real, VN<:AbstractVector{N}, S<:LazySet{N}} <: LazySet{N}
     # default constructor with dimension check
     function Translation(X::S, v::VN) where {N, VN<:AbstractVector{N}, S<:LazySet{N}}
         @assert dim(X) == length(v) "cannot create a translation of a set of dimension $(dim(X)) " *
-                                    "along a vector of length $(length(v))" 
+                                    "along a vector of length $(length(v))"
         return new{N, VN, S}(X, v)
     end
 end
@@ -180,7 +180,7 @@ Unicode alias constructor ⊕ (`oplus`) for the lazy translation operator.
 # translation from the left
 ⊕(v::AbstractVector, X::LazySet) = Translation(X, v)
 
-# ============================ 
+# ============================
 # LazySet interface functions
 # ============================
 
@@ -278,7 +278,7 @@ An element in the translation.
 ### Notes
 
 This function first asks for `an_element` function of the wrapped set, then
-translates this element according to the given translation vector. 
+translates this element according to the given translation vector.
 """
 function an_element(tr::Translation)
     return an_element(tr.X) + tr.v
--- a/src/concrete_intersection.jl
+++ b/src/concrete_intersection.jl
@@ -678,7 +678,7 @@ Return the intersection of a lazy linear map and a convex set.
 
 - `L` -- linear map
 - `S` -- convex set
-  
+
 ### Output
 
 The polytope obtained by the intersection of `l.M * L.X` and `S`.
@@ -781,4 +781,37 @@ end
 function intersection(rm::ResetMap{N, <:AbstractPolytope},
                       P::AbstractPolyhedron{N}) where {N<:Real}
     return intersection(P, rm)
+end
+
+function intersection(U::Universe{N}, X::CartesianProductArray{N}) where {N<:Real}
+    return X
+end
+
+# symmetric method
+function intersection(X::CartesianProductArray{N}, U::Universe{N}) where {N<:Real}
+    return intersection(U, X)
+end
+
+"""
+        intersection(X::CartesianProductArray{N}, Y::CartesianProductArray{N})
+
+Return the intersection between cartesian products of a finite number of convex sets.
+
+### Input
+
+ - `X` -- cartesian product of a finite number of convex sets
+ - `Y` -- cartesian product of a finite number of convex sets
+
+### Output
+
+The decomposed set which represents concrete intersection between `X` and `Y`
+
+### Algorithm
+
+This algorithm intersect corresponding blocks between sets.
+"""
+function intersection(X::CartesianProductArray{N}, Y::CartesianProductArray{N}) where {N<:Real}
+    @assert same_block_structure(array(X), array(Y)) "block structure has to be the same"
+
+    return CartesianProductArray([intersection(array(X)[i], array(Y)[i]) for i in eachindex(array(X))])
 end
--- a/src/helper_functions.jl
+++ b/src/helper_functions.jl
@@ -542,9 +542,9 @@ julia> subtypes(AbstractPolytope, true)
 function subtypes(interface, concrete::Bool)
 
     subtypes_to_test = subtypes(interface)
-    
+
     # do not seek the concrete subtypes further
-    if !concrete 
+    if !concrete
         return sort(subtypes_to_test, by=string)
     end
 
--- a/src/intersection_helper.jl
+++ b/src/intersection_helper.jl
@@ -1,0 +1,49 @@
+"""
+    get_constrained_lowdimset(cpa::CartesianProductArray{N, S}, P::AbstractPolyhedron{N}) where {N<:Real, S<:LazySet{N}}
+
+Preprocess step for intersection between cartesian product array and H-polyhedron.
+Returns low-dimensional HPolytope in constrained dimensions of original cpa,
+constrained variables and variables in corresponding blocks, original block structure
+of low-dimensional set and list of constrained blocks.
+
+### Input
+
+- `cpa` -- Cartesian product array of convex sets
+- `P` -- polyhedron
+
+### Output
+
+A tuple of low-dimensional set, list of constrained dimensions, original block
+structure of low-dimensional set and corresponding blocks indices.
+"""
+function get_constrained_lowdimset(cpa::CartesianProductArray{N, S}, P::AbstractPolyhedron{N}) where {N<:Real, S<:LazySet{N}}
+
+    if isbounded(P)
+        blocks, non_empty_length = block_to_dimension_indices(cpa)
+    else
+        constrained_vars = constrained_dimensions(P)
+        blocks, non_empty_length = block_to_dimension_indices(cpa, constrained_vars)
+    end
+
+    array = Vector{S}()
+    sizehint!(array, non_empty_length)
+    vars = Vector{Int}()
+    block_structure = Vector{UnitRange{Int}}()
+    sizehint!(block_structure, non_empty_length)
+
+    last_var = 1
+    for i in 1:length(blocks)
+        start_index, end_index = blocks[i]
+        block_end = last_var + end_index - start_index
+        if start_index != -1
+            push!(array, cpa.array[i])
+            append!(vars, start_index : end_index)
+            push!(block_structure, last_var : block_end)
+            last_var = block_end + 1
+        end
+    end
+
+    cpa_low_dim = HPolytope(constraints_list(CartesianProductArray(array)));
+
+    return cpa_low_dim, vars, block_structure, blocks
+end
--- a/src/is_intersection_empty.jl
+++ b/src/is_intersection_empty.jl
@@ -1,3 +1,4 @@
+using LazySets.Approximations: project
 export is_intersection_empty,
        isdisjoint
 
@@ -1364,4 +1365,58 @@ function is_intersection_empty(X::LazySet{N},
                               )::Union{Bool, Tuple{Bool, Vector{N}}} where
                                   {N<:Real}
     return is_intersection_empty(C, X, witness)
+end
+
+"""
+    is_intersection_empty(X::CartesianProductArray{N, S},
+                          P::HPolyhedron{N}) where {N<:Real, S<:LazySet{N}}
+
+Check whether a Cartesian product array intersects with a H-polyhedron.
+
+### Input
+
+- `X` -- Cartesian product array of convex sets
+- `P` -- H-polyhedron
+
+### Output
+
+`true` iff ``X ∩ Y = ∅ ``.
+"""
+function is_intersection_empty(X::CartesianProductArray{N},
+                               P::HPolyhedron{N}) where {N<:Real}
+    cpa_low_dim, vars, block_structure = get_constrained_lowdimset(X, P)
+
+    return isdisjoint(cpa_low_dim, project(P, vars))
+end
+
+# symmetric method
+function is_intersection_empty(P::HPolyhedron{N},
+                               X::CartesianProductArray{N, S}) where {N<:Real, S<:LazySet{N}}
+        is_intersection_empty(X, P)
+end
+
+"""
+    is_intersection_empty(X::CartesianProductArray{N}, Y::CartesianProductArray{N}) where {N}
+
+Check whether two Cartesian products of a finite number of convex sets do not intersect.
+
+### Input
+
+- `X` -- Cartesian product array of convex sets
+- `Y` -- Cartesian product array of convex sets
+
+### Output
+
+`true` iff ``X ∩ Y = ∅ ``.
+"""
+function is_intersection_empty(X::CartesianProductArray{N}, Y::CartesianProductArray{N}) where {N}
+    @assert same_block_structure(array(X), array(Y)) "block structure has to be the same"
+
+    for i in 1:length(X.array)
+        if isdisjoint(X.array[i], Y.array[i])
+            return true
+        end
+    end
+
+    return false
 end
--- a/test/unit_CartesianProduct.jl
+++ b/test/unit_CartesianProduct.jl
@@ -1,3 +1,5 @@
+using LazySets.Approximations: overapproximate
+
 for N in [Float64, Float32, Rational{Int}]
     # Cartesian Product of a centered 1D BallInf and a centered 2D BallInf
     # Here a 3D BallInf
@@ -261,7 +263,7 @@ for N in [Float64, Float32, Rational{Int}]
     Hcp = convert(CartesianProductArray{N, Interval{N}}, H)
     @test Hcp isa CartesianProductArray &&
           all([array(Hcp)[i] == Interval(N(-1), N(2*i-1)) for i in 1:3])
-              
+
     # ================
     # common functions
     # ================
@@ -272,4 +274,22 @@ for N in [Float64, Float32, Rational{Int}]
     @test absorbing(CartesianProduct) == absorbing(CartesianProductArray) ==
           EmptySet
     @test b × e == e × b == cpa × e == e × cpa == e × e == e
+end
+
+for N in [Float64, Float32]
+    # is_intersection_empty
+    i1 = Interval(N[0, 1])
+    i2 = Interval(N[2, 3])
+    h1 = Hyperrectangle(low=N[3, 4], high=N[5, 7])
+    h2 = Hyperrectangle(low=N[5, 5], high=N[6, 8])
+    cpa1 = CartesianProductArray([i1, i2, h1])
+    cpa2 = CartesianProductArray([i1, i2, h2])
+    G = HPolyhedron([HalfSpace(N[1, 0, 0, 0], N(1))])
+    G_empty = HPolyhedron([HalfSpace(N[1, 0, 0, 0], N(-1))])
+	cpa1_box = overapproximate(cpa1)
+	cpa2_box = overapproximate(cpa2)
+
+    @test is_intersection_empty(cpa1, cpa2) == is_intersection_empty(cpa1_box, cpa2_box) == false
+    @test is_intersection_empty(cpa1, G_empty) == is_intersection_empty(cpa1_box, G_empty) == true
+    @test is_intersection_empty(cpa1, G) == is_intersection_empty(Approximations.overapproximate(cpa1), G) == false
 end
--- a/test/unit_overapproximate.jl
+++ b/test/unit_overapproximate.jl
@@ -1,3 +1,5 @@
+using LazySets.Approximations: project
+
 for N in [Float64, Rational{Int}, Float32]
     # overapproximating a set of type T1 with an unsupported type T2 is the
     # identity if T1 = T2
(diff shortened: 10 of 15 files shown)


src/helper_functions.jl
```diff
@@ -4,8 +4,8 @@ export _At_mul_B
 const DEFAULT_COND_TOL = 1e6
 
 # matrix-matrix multiplication and division
-@inline _At_mul_B(A, B) = transpose(A) * B
-@inline _At_ldiv_B(A, B) = transpose(A) \ B
+@inline _At_mul_B(A, B) = transpose(Matrix(A)) * B
+@inline _At_ldiv_B(A, B) = transpose(Matrix(A)) \ B
```
schillic: why?

src/helper_functions.jl
```diff
@@ -159,13 +159,13 @@ We check whether the matrix is square and whether the
 `cond(M)` is below some prescribed tolerance.
 """
 function isinvertible(M::Matrix; cond_tol::Number=DEFAULT_COND_TOL)
-    return issquare(M) && cond(M) < cond_tol
+    return cond(M) < cond_tol
```
schillic: why? A non-square matrix cannot be invertible.

kostakoida: Actually it is a good question how to fix it. 
The problem is when we are doing a projection of `HalfSpace` we have a sparse matrix, which is not a square by itself.

src/helper_functions.jl
kostakoida: I believe it was an old workaround for using projection (and Linear Map) for `Halfspaces` on specific variables. We discussed it at some point with @mforets. Maybe now he has some better ideas on how to fix it.

schillic: `transpose` works for sparse matrices. Do you have a use case?

src/helper_functions.jl
schillic: I do not see the problem. If it is not square, then it is not invertible. End of the story. You cannot just change `isinvertible` and hope that this makes the matrix invertible :wink:

schillic: Furthermore, this method is not used for sparse matrices.

src/helper_functions.jl
kostakoida: We can transpose standard `SparceArray` with type `SparseMatrixCSC{Float64,Int64}`:
```
julia> sp = spzeros(1,4)
1×4 SparseMatrixCSC{Float64,Int64} with 0 stored entries

julia> sp[1,1] = 1
1

julia> transpose(sp)
4×1 LinearAlgebra.Transpose{Float64,SparseMatrixCSC{Float64,Int64}}:
 1.0
 0.0
 0.0
 0.0
```
However, inside helper function we somehow get sparce matrix from `LinearAlgebra` package with the type `SuiteSparse.SPQR.QRSparse{Float64,Int64}` 
```
julia> lspar = qr(sp)
SuiteSparse.SPQR.QRSparse{Float64,Int64}
Q factor:
1×1 SuiteSparse.SPQR.QRSparseQ{Float64,Int64}:
 1.0
R factor:
1×4 SparseMatrixCSC{Float64,Int64} with 1 stored entry:
  [1, 1]  =  1.0
Row permutation:
1-element Array{Int64,1}:
 1
Column permutation:
4-element Array{Int64,1}:
 1
 2
 3
 4

julia> transpose(lspar)
ERROR: MethodError: no method matching transpose(::SuiteSparse.SPQR.QRSparse{Float64,Int64})
```

schillic: Then the Julia paradigm is to define new specific methods for these cases instead of making all other cases suffer:
```julia
@inline _At_mul_B(A::Factorization, B) = transpose(Matrix(A)) * B
@inline _At_ldiv_B(A::Factorization, B) = transpose(Matrix(A)) \ B
```
I found the type for `A` like this:
```julia
julia> supertype(typeof(lspar))
Factorization{Float64}
```

kostakoida: Actually now I am confused. It seems that we actually have standard sparce matrix, but the error says opposite:
```
julia> d_int_g = intersection(cpa1, G)
HPolytope{Float64}
LinearMap{Float64,HPolyhedron{Float64},Float64,SparseMatrixCSC{Float64,Int64}}
Type of M is.. SparseMatrixCSC{Float64,Int64}

ERROR: MethodError: no method matching transpose(::SuiteSparse.SPQR.QRSparse{Float64,Int64})
``` 
(I have added my own output to debug...)

kostakoida: Fot testing purposes you can use:

```
i1 = Interval(N[0, 1])
i2 = Interval(N[2, 3])
h1 = Hyperrectangle(low=N[3, 4], high=N[5, 7])
cpa1 = CartesianProductArray([i1, i2, h1])
G = HPolyhedron([HalfSpace(N[1, 0, 0, 0], N(1))])
d_int_g = intersection(cpa1, G)
```

schillic: My only guess is that you do not `transpose` this matrix directly. Best look at the line where it occurs.

kostakoida: So as I can see in linalg.jl, since matrix M is not squared it calls `LinearAlgebra`'s sparce matrix:
https://github.com/JuliaLang/julia/blob/master/stdlib/SparseArrays/src/linalg.jl#L1400

src/is_intersection_empty.jl
```diff
@@ -1365,3 +1365,82 @@ function is_intersection_empty(X::LazySet{N},
...
+        # no free variables
+        blocks = block_indices(X)
+    else
+        constrained_vars = constrained_dimensions(Y)
+        blocks = block_indices(X, constrained_vars)
+    end
+
+    result = CartesianProductArray(length(X.array), N)
+
+    low_set = CartesianProductArray(length(blocks), N)
```
schillic: As in the other PR, better create the `Vector{<:S}(undef, length(blocks))` here and create the CPA after the loop.

src/is_intersection_empty.jl
```diff
@@ -1365,3 +1365,82 @@ function is_intersection_empty(X::LazySet{N},
...
+    vars = Vector{Int}()
+    blocks = sort(blocks)
+
+    for bi in keys(blocks)
+        push!(low_set.array, X.array[bi])
```
schillic: You should not `push!` here because you already allocated the array. Either use `sizehint!` above or (preferred) write to the next index `i` (to be defined) (`low_set.array[i] = ...`).

src/is_intersection_empty.jl
```diff
@@ -1365,3 +1365,82 @@ function is_intersection_empty(X::LazySet{N},
...
+        append!(vars, variable_indices(X, bi, blocks[bi]))
+    end
+
+    approx_low_set = HPolytope(constraints_list(low_set));
+
+    if isdisjoint(approx_low_set, Approximations.project(Y, vars))
```
schillic:
```julia
if cond
    return true
end
return false
```
can be simplified to
```julia
return cond
```

src/is_intersection_empty.jl
```diff
@@ -1365,3 +1365,82 @@ function is_intersection_empty(X::LazySet{N},
...
+# symmetric method
+function is_intersection_empty(Y::HPolyhedron{N},
+                               X::CartesianProductArray{N, S}) where {N<:Real, S<:LazySet{N}}
+        is_intersection_empty(X, Y)
+end
+
+"""
+    is_intersection_empty(X::CartesianProductArray{N}, Y::CartesianProductArray{N}) where {N}
+
+Check whether 2 cartesian product arrays intersection is not empty.
```
schillic:
```suggestion
Check whether two Cartesian products of a finite number of convex sets do not intersect.
```

src/concrete_intersection.jl
```diff
@@ -778,3 +778,42 @@ function intersection(rm::ResetMap{N, <:AbstractPolytope},
...
+The decomposed set which represents concrete intersection between `X` and `Y`
+
+### Algorithm
+
+This algorithm intersect corresponding blocks between sets.
+"""
+function intersection(X::CartesianProductArray{N}, Y::CartesianProductArray{N}) where {N<:Real}
+    @assert same_block_structure(array(X), array(Y)) "block structure has to be the same"
+
+    result = CartesianProductArray(length(X.array), N)
```
schillic: The same comments as above apply. Here is a better, one-line version:
```julia
return CartesianProductArray([intersection(array(X)[i], array(Y)[i]) for i in eachindex(X.array)])
```

src/helper_functions.jl
```diff
@@ -165,7 +165,7 @@ end
 # cond is not available for sparse matrices; see JuliaLang#6485 and related issues
 function isinvertible(M::AbstractSparseMatrix;
                       cond_tol::Number=DEFAULT_COND_TOL)
-    return issquare(M) && isinvertible(Matrix(M), cond_tol=cond_tol)
+     return issquare(M) && isinvertible(Matrix(M), cond_tol=cond_tol)
```
schillic:
```suggestion
    return issquare(M) && isinvertible(Matrix(M), cond_tol=cond_tol)
```

src/is_intersection_empty.jl
```diff
@@ -1365,3 +1365,82 @@ function is_intersection_empty(X::LazySet{N},
...
+- `Y` -- Cartesian product array of convex sets
+
+### Output
+
+`true` iff ``X ∩ Y = ∅ ``.
+"""
+function is_intersection_empty(X::CartesianProductArray{N}, Y::CartesianProductArray{N}) where {N}
+    @assert same_block_structure(array(X), array(Y)) "block structure has to be the same"
+
+    result = CartesianProductArray(length(X.array), N)
```
schillic: unused

test/unit_CartesianProduct.jl
```diff
@@ -272,4 +272,23 @@ for N in [Float64, Float32, Rational{Int}]
     @test absorbing(CartesianProduct) == absorbing(CartesianProductArray) ==
           EmptySet
     @test b × e == e × b == cpa × e == e × cpa == e × e == e
+
+
+
```
schillic: remove empty lines

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +517,84 @@ function overapproximate(cap::Intersection{N,
...
+                      Y::AbstractPolyhedron{N},
+                      blocks::Dict{Int,Int},
+                      oa=Hyperrectangle) where {N}
+
+    low_set, vars, block_structure = initialize_vars(X, blocks)
+
+    low_intersection = intersection(low_set, Approximations.project(Y, vars))
+
+    if isempty(low_intersection)
+        return EmptySet()
```
```suggestion
        return EmptySet{N}()
```

src/CartesianProduct.jl
```diff
@@ -562,3 +563,158 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
...
+- `cpa` -- Cartesian product array
+- `block_index` -- block index
+- `start_dim` -- index of the first variable in the block, according to the whole system
+
+### Output
+
+List of the variables in the block
+
+"""
+function variable_indices(cpa::CartesianProductArray{N}, block_index::Int, start_dim::Int) where {N}
```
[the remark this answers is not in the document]
schillic: The method is now unused.

src/is_intersection_empty.jl
```diff
@@ -1365,3 +1365,78 @@ function is_intersection_empty(X::LazySet{N},
...
+### Input
+
+- `X` -- Cartesian product array of convex sets
+- `Y` -- H-polyhedron
+
+### Output
+
+`true` iff ``X ∩ Y = ∅ ``.
+"""
+function is_intersection_empty(X::CartesianProductArray{N, S},
```
schillic: This method is *very* similar to `overapproximate`. Can you outsource the common parts (say, `intersection_helper`)?

src/CartesianProduct.jl
```diff
@@ -562,3 +563,119 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
...
+                              blocks::Dict{Int,Int}) where {N}
+
+Return merged Cartesian Product Array between original CPA and some low-dimensional CPA,
+which represents updated subset of variables in specified blocks.
+
+### Input
+
+- `low_dim_cpa` -- low-dimensional cartesian product array
+- `orig_cpa` -- original high-dimensional product array
+- `blocks` -- index of the first variable in the block, according to the whole system
```
schillic:
```suggestion
- `blocks` -- index of the first variable in each block of `orig_cpa`
```

src/CartesianProduct.jl
```diff
@@ -562,3 +563,119 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
...
+"""
+function combine_resulted_cpa(low_dim_cpa::CartesianProductArray{N},
+                              orig_cpa::CartesianProductArray{N},
+                              blocks::Vector{Tuple{Int,Int}}) where {N}
+
+    array = Vector{LazySet{N}}(undef, length(orig_cpa.array))
+    index = 1
+    for bi in 1:length(orig_cpa.array)
+        start_ind, end_index = blocks[bi]
+        if start_ind != -1
```
schillic: I find it easier to swap the cases and check for the positive condition.

src/is_intersection_empty.jl
```diff
@@ -1365,3 +1365,78 @@ function is_intersection_empty(X::LazySet{N},
...
+# symmetric method
+function is_intersection_empty(Y::HPolyhedron{N},
+                               X::CartesianProductArray{N, S}) where {N<:Real, S<:LazySet{N}}
+        is_intersection_empty(X, Y)
+end
+
+"""
+    is_intersection_empty(X::CartesianProductArray{N}, Y::CartesianProductArray{N}) where {N}
+
+    Check whether two Cartesian products of a finite number of convex sets do not intersect.
```
schillic:
```suggestion
Check whether two Cartesian products of a finite number of convex sets do not intersect.
```

src/Approximations/overapproximate.jl
```diff
@@ -516,3 +517,90 @@ function overapproximate(cap::Intersection{N,
...
+function initialize_vars(X::CartesianProductArray{N, S}, blocks::Vector{Tuple{Int,Int}}, non_empty_length::Int) where {N<:Real, S<:LazySet{N}}
+    array = Vector{S}(undef, non_empty_length)
+    vars = Vector{Int}()
+    block_structure = Vector{UnitRange{Int}}()
+
+    last_var = 1
+    for i in 1:length(blocks)
+        start_index, end_index = blocks[i]
+        if (start_index != -1)
+            array[i] = X.array[i]
```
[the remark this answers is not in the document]
kostakoida: Yes, but I know the size of array in advance.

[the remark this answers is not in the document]
schillic: Think about what `i` iterates over. You need a different index.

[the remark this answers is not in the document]
kostakoida: Indeed. I was cleaning extra indices from the code and introduced this bug.

src/CartesianProduct.jl
```diff
@@ -562,3 +563,142 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
...
+
+### Example
+
+```jldoctest
+julia> cpa = CartesianProductArray([BallInf(zeros(n), 1.0) for n in [1, 3, 2, 3]]);
+
+julia> block_to_dimension_indices(cpa, [2, 4, 8])
+(Tuple{Int64,Int64}[(-1, -1), (2, 4), (-1, -1), (7, 9)], 2)
+```
+This vector represents the mapping "second block from dimension 2 to dimension 4, fourth block from dimension 7 to dimension 9."
```
schillic: line break

src/intersection_helper.jl
```diff
@@ -0,0 +1,54 @@
...
+    array = Vector{S}()
+    sizehint!(array, non_empty_length)
+    vars = Vector{Int}()
+    block_structure = Vector{UnitRange{Int}}()
+
+    last_var = 1
+    for i in 1:length(blocks)
+        start_index, end_index = blocks[i]
+        block_end = last_var + end_index - start_index
+        if (start_index != -1)
```
schillic:
```suggestion
        if start_index != -1
```

src/intersection_helper.jl
```diff
@@ -0,0 +1,54 @@
...
+            push!(array, X.array[i])
+            append!(vars, start_index : end_index)
+            push!(block_structure, last_var:(block_end))
+        end
+        last_var += block_end + 1
```
schillic:
```suggestion
        last_var = block_end + 1
```
(`last_var` is already added to `block_end`.)

src/intersection_helper.jl
```diff
@@ -0,0 +1,54 @@
...
+    block_structure = Vector{UnitRange{Int}}()
+
+    last_var = 1
+    for i in 1:length(blocks)
+        start_index, end_index = blocks[i]
+        block_end = last_var + end_index - start_index
+        if (start_index != -1)
+            push!(array, X.array[i])
+            append!(vars, start_index : end_index)
+            push!(block_structure, last_var:(block_end))
```
schillic:
```suggestion
            push!(block_structure, last_var:block_end)
```

src/intersection_helper.jl
```diff
@@ -0,0 +1,54 @@
...
+structure of low-dimensional set and corresponding blocks indicies
+"""
+function get_constrained_subset(X::CartesianProductArray{N, S}, Y::HPolyhedron{N}) where {N<:Real, S<:LazySet{N}}
+
+    blocks, non_empty_length = get_constrained_blocks(X, Y)
+
+    array = Vector{S}()
+    sizehint!(array, non_empty_length)
+    vars = Vector{Int}()
+    block_structure = Vector{UnitRange{Int}}()
```
schillic: Add the `sizehint!` for `block_structure` as well.

src/LazySets.jl
```diff
@@ -90,24 +90,25 @@ include("Rectification.jl")
 # =============================
 include("convert.jl")
 
+# =====================
+# Approximations module
+# =====================
+include("Approximations/Approximations.jl")
```
schillic: This change should not be necessary. There should not be any dependency on the `Approximations` module.

src/intersection_helper.jl
```diff
@@ -0,0 +1,54 @@
...
+
+- `X` -- Cartesian product array of convex sets
+- `Y` -- H-polyhedron
+
+### Output
+
+A tuples of low-dimensional set, list of constrained dimensions, original block
+structure of low-dimensional set and corresponding blocks indicies
+"""
+function get_constrained_subset(X::CartesianProductArray{N, S}, Y::HPolyhedron{N}) where {N<:Real, S<:LazySet{N}}
```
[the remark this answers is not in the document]
kostakoida: I thought it is a good idea to have separate file as this particular function is common for `is_intersection_empty.jl` and `overapproximate.jl`. Moreover, if we think that we are going to add some optimisations (like saving block structure, constrained blocks, etc.) we will need eventually this file.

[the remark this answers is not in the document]
schillic: > we will need eventually this file

I do not see why. You can write all the code in a single file :wink:
But we can keep it.

src/LazySets.jl
kostakoida: We added `using LazySets.Approximations: projection` in `is_empty_intersection.jl` and therefore it requires module`Approximation` to build LazySets.

schillic: I see. I created #1429 to fix this later.

src/Approximations/overapproximate.jl
```diff
@@ -597,3 +598,61 @@ function blocks_linear_map(cpa::CartesianProductArray{N, T},
...
+                            <:AbstractPolyhedron{N}},
+                       ::Type{CartesianProductArray}, oa) where {N}
+
+Return the intersection of the cartesian product of a finite number of convex sets and a polyhedron.
+
+### Input
+
+ - `cap` -- Lazy intersection of cartesian product array and polyhedron
+ - `Y`   -- polyhedron
+ - `Type{CartesianProductArray}` -- argument to specify decomposed overapproximation
```
```suggestion
 - `CartesianProductArray` -- type for dispatch
```
(This is what we usually write there - don't ask :wink:)

src/CartesianProduct.jl
```diff
@@ -562,3 +563,147 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
...
+- `cpa` -- Cartesian product array
+
+### Output
+
+A vector of ranges
+
+### Example
+
+```jldoctest
+julia> using LazySets: block_structure
```
schillic: Just a comment: Instead you can write `LazySets.block_structure` below (as you prefer).

src/CartesianProduct.jl
```diff
@@ -562,3 +563,147 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
...
+        start_index = end_index + 1
+    end
+    return result
+end
+
+
+"""
+    block_to_dimension_indices(cpa::CartesianProductArray{N}, vars::Vector{Int}) where {N}
+
+Return the vector with ranges of dimensions of blocks with specified blocksReturns a vector mapping block index `i`
```
schillic:
```suggestion
Returns a vector mapping block index `i`
```

src/CartesianProduct.jl
```diff
@@ -562,3 +563,147 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
...
+    start_index, end_index = 1, 0
+    v_i = 1
+    for i in 1:length(cpa.array)
+        end_index += dim(cpa.array[i])
+        while v_i <= length(vars) && vars[v_i] <= end_index
+            if vars[v_i] ∈ (start_index : end_index)
+                result[i] = (start_index, end_index)
+                non_empty_length += 1
+                while v_i <= length(vars) && vars[v_i] <= end_index
+                    v_i +=1
```
schillic:
```suggestion
                    v_i += 1
```

src/CartesianProduct.jl
```diff
@@ -562,3 +563,147 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
...
+- `blocks` -- index of the first variable in each block of `orig_cpa`
+
+### Output
+
+Merged cartesian product array
+
+"""
+function substitute_blocks(low_dim_cpa::CartesianProductArray{N},
+                              orig_cpa::CartesianProductArray{N},
+                              blocks::Vector{Tuple{Int,Int}}) where {N}
```
schillic: indentation (also in the comment header above)

src/CartesianProduct.jl
```diff
@@ -562,3 +563,147 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
...
+### Input
+
+- `low_dim_cpa` -- low-dimensional cartesian product array
+- `orig_cpa` -- original high-dimensional Cartesian product array
+- `blocks` -- index of the first variable in each block of `orig_cpa`
+
+### Output
+
+Merged cartesian product array
+
```
schillic:
```suggestion
```

src/intersection_helper.jl
```diff
@@ -0,0 +1,49 @@
...
+        block_end = last_var + end_index - start_index
+        if start_index != -1
+            push!(array, cpa.array[i])
+            append!(vars, start_index : end_index)
+            push!(block_structure, last_var : block_end)
+        end
+        last_var = block_end + 1
+    end
+
+    cpa_low_dim = HPolytope(constraints_list(CartesianProductArray(array)));
```
schillic: The following is much cheaper:
```suggestion
    cpa_low_dim = CartesianProductArray(array)
```
However, it will create infinite loops because the callers would call themselves again. Still, I find it cleaner if this helper method really just returns a CPA and the callers do the conversion to `HPolytope` so it is more explicit.

This also requires that the two main methods (`overapproximate` and `is_intersection_empty`) receive a polytopic CPA (at least in the constrained blocks). Please add `polytopic` to the `Input` documentations.

test/unit_CartesianProduct.jl
```diff
@@ -273,3 +273,19 @@ for N in [Float64, Float32, Rational{Int}]
...
+    h1 = Hyperrectangle(low=N[3, 4], high=N[5, 7])
+    h2 = Hyperrectangle(low=N[5, 5], high=N[6, 8])
+    cpa1 = CartesianProductArray([i1, i2, h1])
+    cpa2 = CartesianProductArray([i1, i2, h2])
+    G = HPolyhedron([HalfSpace(N[1, 0, 0, 0], N(1))])
+    G_empty = HPolyhedron([HalfSpace(N[1, 0, 0, 0], N(-1))])
+
+    @test is_intersection_empty(cpa1, cpa2) == is_intersection_empty(Approximations.overapproximate(cpa1), Approximations.overapproximate(cpa2))
+    @test is_intersection_empty(cpa1, G_empty) == is_intersection_empty(Approximations.overapproximate(cpa1), G_empty)
+    @test is_intersection_empty(cpa1, G) == is_intersection_empty(Approximations.overapproximate(cpa1), G)
```
schillic: * Add `using LazySets.Approximations: overapproximate` to the top.
* Pull out `cpa1_box = overapproximate(cpa1)`.
* Add the expected `Bool` value.

test/unit_CartesianProduct.jl
```diff
@@ -273,3 +273,19 @@ for N in [Float64, Float32, Rational{Int}]
           EmptySet
     @test b × e == e × b == cpa × e == e × cpa == e × e == e
 end
+
+for N in [Float64, Float32]
+    #is_intersection_empty
```
schillic:
```suggestion
    # is_intersection_empty
```

test/unit_overapproximate.jl
```diff
@@ -165,6 +165,24 @@ for N in [Float64, Float32]
     Y_zonotope = overapproximate(Y, Zonotope) # overapproximate with a zonotope
     @test Y_polygon ⊆ Y_zonotope
     @test !(Y_zonotope ⊆ Y_polygon)
+
+    #decomposed intersection between cartesian product array and abstract polyhedron
```
schillic:
```suggestion
    # decomposed intersection between cartesian product array and abstract polyhedron
```

test/unit_overapproximate.jl
```diff
@@ -165,6 +165,24 @@ for N in [Float64, Float32]
...
+    h2 = Hyperrectangle(low=N[5, 5], high=N[6, 8])
+    cpa1 = CartesianProductArray([i1, i2, h1])
+    cpa2 = CartesianProductArray([i1, i2, h2])
+    G = HPolyhedron([HalfSpace(N[1, 0, 0, 0], N(1))])
+    G_empty = HPolyhedron([HalfSpace(N[1, 0, 0, 0], N(-1))])
+
+    d_int_g = Intersection(cpa1, G)
+    l_int_g = Intersection(cpa1, G)
+    d_int_cpa = intersection(cpa1, cpa2)
+    l_int_cpa = Intersection(cpa1, cpa2)
```
schillic: `d_int_g == l_int_g` and `d_int_cpa == l_int_cpa`!

src/CartesianProduct.jl
kostakoida: I believe for the comment header above it is good. Should it be aligned to the semicolon?

schillic: It should be aligned with the `(`.

test/unit_CartesianProduct.jl
```diff
@@ -273,3 +273,19 @@ for N in [Float64, Float32, Rational{Int}]
...
+    i1 = Interval(N[0, 1])
+    i2 = Interval(N[2, 3])
+    h1 = Hyperrectangle(low=N[3, 4], high=N[5, 7])
+    h2 = Hyperrectangle(low=N[5, 5], high=N[6, 8])
+    cpa1 = CartesianProductArray([i1, i2, h1])
+    cpa2 = CartesianProductArray([i1, i2, h2])
+    G = HPolyhedron([HalfSpace(N[1, 0, 0, 0], N(1))])
+    G_empty = HPolyhedron([HalfSpace(N[1, 0, 0, 0], N(-1))])
```
schillic: If you do the change to `AbstractPolyhedron`, this wrapping in `HPolyhedron` is obsolete.

src/intersection_helper.jl
kostakoida: > `However, it will create infinite loops because the callers would call themselves again.`

I think `overapproximate` shouldn't call itself again, since we can call it without `CartesianProductArray` type as an argument, but we will get this problem for `is_intersection_empty.jl`

test/unit_CartesianProduct.jl
kostakoida: I am getting an error:
```
ERROR: MethodError: LazySets.is_intersection_empty(::CartesianProductArray{Float64,AbstractHyperrectangle{Float64}}, ::HPolyhedron{Float64}) is ambiguous. Candidates:
  is_intersection_empty(X::CartesianProductArray{N,S} where S<:LazySet{N}, Y::AbstractPolyhedron{N}) where N<:Real in LazySets at /Users/kpotomkin/.julia/dev/LazySets/src/is_intersection_empty.jl:1387
  is_intersection_empty(X::CartesianProductArray{N,S}, Y::AbstractPolyhedron{N}) where {N<:Real, S<:LazySet{N<:Real}} in LazySets at /Users/kpotomkin/.julia/dev/LazySets/src/is_intersection_empty.jl:1387
  is_intersection_empty(X::LazySet{N}, P::Union{AbstractPolytope{N}, HPolyhedron{N}}) where N<:Real in LazySets at /Users/kpotomkin/.julia/dev/LazySets/src/is_intersection_empty.jl:973
```

test/unit_overapproximate.jl
kostakoida: Do you mean that they are equal and I can just remove duplication?

src/intersection_helper.jl
schillic: > I think overapproximate shouldn't call itself again, since we can call it without CartesianProductArray type as an argument

I was referring to the version *after* the change. There the first argument would be a CPA again. But indeed you call `intersection` instead of `overapproximate`, so this would not be a loop. But there is another problem: We do not have `intersection(::CPA, ::AbstractPolyhedron)` :smile:

test/unit_overapproximate.jl
schillic: I do not know what you intended to test, so this is just the observation that they are equal and that I have the feeling that you may not have wanted this.

test/unit_CartesianProduct.jl
schillic: The first two lines look like you define several methods that do the same. Still, you may have to add more methods to resolve the ambiguities. If you want I can do the generalization to `AbstractPolyhedron` in a follow-up PR.

schillic:
Here are the missing things to address before merging:
* [ ] https://github.com/JuliaReach/LazySets.jl/pull/1421#discussion_r291972675
* [ ] https://github.com/JuliaReach/LazySets.jl/pull/1421#discussion_r292321408
* [ ] item 2: https://github.com/JuliaReach/LazySets.jl/pull/1421#pullrequestreview-248577464
* [ ] item 3: https://github.com/JuliaReach/LazySets.jl/pull/1421#pullrequestreview-248577464
* [ ] https://github.com/JuliaReach/LazySets.jl/pull/1421#discussion_r292785972


Collection of things for follow-up PR:
* https://github.com/JuliaReach/LazySets.jl/pull/1421#discussion_r292775970
* item 1: https://github.com/JuliaReach/LazySets.jl/pull/1421#pullrequestreview-248577464
* https://github.com/JuliaReach/LazySets.jl/pull/1421#discussion_r292850031

kostakoida:
> * Pelase add a test for each method where you have to combine several blocks.

I don't understand why I cannot pass one of the tests:
```
julia> i1 = Interval(N[0, 1])
Interval{Float64,IntervalArithmetic.Interval{Float64}}([0, 1])

julia> i2 = Interval(N[2, 3])
Interval{Float64,IntervalArithmetic.Interval{Float64}}([2, 3])

julia> h1 = Hyperrectangle(low=N[3, 4], high=N[5, 7])
Hyperrectangle{Float64}([4.0, 5.5], [1.0, 1.5])

julia> cpa1 = CartesianProductArray([i1, i2, h1])
CartesianProductArray{Float64,AbstractHyperrectangle{Float64}}(AbstractHyperrectangle{Float64}[Interval{Float64,Interval{Float64}}([0, 1]), Interval{Float64,Interval{Float64}}([2, 3]), Hyperrectangle{Float64}([4.0, 5.5], [1.0, 1.5])])

julia> G_comb = HPolyhedron([HalfSpace(N[1, 1, 0, 0], N(2.5))])
HPolyhedron{Float64}(HalfSpace{Float64,VN} where VN<:AbstractArray{Float64,1}[HalfSpace{Float64,Array{Float64,1}}([1.0, 1.0, 0.0, 0.0], 2.5)])

julia> d_int_g_comb = Intersection(cpa1, G_comb)
Intersection{Float64,CartesianProductArray{Float64,AbstractHyperrectangle{Float64}},HPolyhedron{Float64}}(CartesianProductArray{Float64,AbstractHyperrectangle{Float64}}(AbstractHyperrectangle{Float64}[Interval{Float64,Interval{Float64}}([0, 1]), Interval{Float64,Interval{Float64}}([2, 3]), Hyperrectangle{Float64}([4.0, 5.5], [1.0, 1.5])]), HPolyhedron{Float64}(HalfSpace{Float64,VN} where VN<:AbstractArray{Float64,1}[HalfSpace{Float64,Array{Float64,1}}([1.0, 1.0, 0.0, 0.0], 2.5)]), LazySets.IntersectionCache(-1))

julia> o_d_int_g_comb = overapproximate(d_int_g_comb, CartesianProductArray, Hyperrectangle)
CartesianProductArray{Float64,LazySet{Float64}}(LazySet{Float64}[Hyperrectangle{Float64}([0.25], [0.25]), Hyperrectangle{Float64}([2.25], [0.25]), Hyperrectangle{Float64}([4.0, 5.5], [1.0, 1.5])])

julia> o_bd_d_int_g_comb = overapproximate(d_int_g_comb, BoxDirections)
HPolytope{Float64}(HalfSpace{Float64,VN} where VN<:AbstractArray{Float64,1}[HalfSpace{Float64,UnitVector{Float64}}([1.0, 0.0, 0.0, 0.0], 0.5), HalfSpace{Float64,UnitVector{Float64}}([0.0, 1.0, 0.0, 0.0], 2.5), HalfSpace{Float64,UnitVector{Float64}}([0.0, 0.0, 1.0, 0.0], 5.0), HalfSpace{Float64,UnitVector{Float64}}([0.0, 0.0, 0.0, 1.0], 7.0), HalfSpace{Float64,UnitVector{Float64}}([0.0, 0.0, 0.0, -1.0], -4.0), HalfSpace{Float64,UnitVector{Float64}}([0.0, 0.0, -1.0, 0.0], -3.0), HalfSpace{Float64,UnitVector{Float64}}([0.0, -1.0, 0.0, 0.0], -2.0), HalfSpace{Float64,UnitVector{Float64}}([-1.0, 0.0, 0.0, 0.0], 1.30096e-16)])

julia> overapproximate(o_d_int_g_comb)
Hyperrectangle{Float64}([0.25, 2.25, 4.0, 5.5], [0.25, 0.25, 1.0, 1.5])

julia> overapproximate(o_bd_d_int_g_comb)
Hyperrectangle{Float64}([0.25, 2.25, 4.0, 5.5], [0.25, 0.25, 1.0, 1.5])

julia> @test overapproximate(o_d_int_g_comb) == overapproximate(o_bd_d_int_g_comb)
Test Failed at none:1
  Expression: overapproximate(o_d_int_g_comb) == overapproximate(o_bd_d_int_g_comb)
   Evaluated: Hyperrectangle{Float64}([0.25, 2.25, 4.0, 5.5], [0.25, 0.25, 1.0, 1.5]) == Hyperrectangle{Float64}([0.25, 2.25, 4.0, 5.5], [0.25, 0.25, 1.0, 1.5])
ERROR: There was an error during testing

julia> overapproximate(o_d_int_g_comb) == overapproximate(o_bd_d_int_g_comb)
false
```

src/CartesianProduct.jl
```diff
@@ -562,3 +563,119 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
...
+    result = Vector{Tuple{Int, Int}}(undef, length(cpa.array))
+
+    fill!(result, (-1, -1))
+    non_empty_length = 0
+    start_index, end_index = 1, 0
+    v_i = 1
+    for i in 1:length(cpa.array)
+        end_index += dim(cpa.array[i])
+        while v_i <= length(vars) && vars[v_i] <= end_index
+            if in(vars[v_i], start_index : end_index)
```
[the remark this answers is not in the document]
kostakoida: I hope I understood correctly what you have meant here:)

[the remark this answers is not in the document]
schillic: Indeed! And I hope I was right :)

schillic:
> I don't understand why I cannot pass one of the tests:

They are *almost* identical. You can see that if you use the long output of the vectors.

```julia
julia> overapproximate(o_d_int_g_comb).center
4-element Array{Float64,1}:
 0.25
 2.25
 4.0 
 5.5 

julia> overapproximate(o_bd_d_int_g_comb).center
4-element Array{Float64,1}:
 0.2500000001848634
 2.2500000001848637
 4.0               
 5.5
```

The imprecision probably comes from the line-search algorithm in the intersection.
You can rescue the test by using `≈`.

test/unit_overapproximate.jl
kostakoida: I removed duplication. The idea is that I am calling different `overapproximate`. Old one and new decomposed version. Then I am checking whether results are the same.

test/unit_overapproximate.jl
```diff
@@ -1,3 +1,5 @@
+using Optim
```
schillic: This is not needed (we have this command in `runtests.jl` already). Your test does not work with `Float32`, so you have to move this block of tests out of the loop (into the last loop).

kostakoida: Thank you!

kostakoida:
> * Please add a test for each method where the constrained dimensions are not in the first block.
> * Please add a test for each method where you have to combine several blocks.

Tests are there now.

test/unit_overapproximate.jl
```diff
@@ -194,4 +216,15 @@ for N in [Float64] # due to sparse vectors: a = sparse(Float32[1 -1; 1 1];); a \
...
+    # decomposed intersection between cartesian product array and abstract polyhedron
+    i1 = Interval(N[0, 1])
+    i2 = Interval(N[2, 3])
+    h1 = Hyperrectangle(low=N[3, 4], high=N[5, 7])
+    cpa = CartesianProductArray([i1, i2, h1])
+    G_comb = HPolyhedron([HalfSpace(N[1, 1, 0, 0], N(2.5))])
+    int_g_comb = Intersection(cpa1, G_comb)
+    o_d_int_g_comb = overapproximate(int_g_comb, CartesianProductArray, Hyperrectangle)
+    o_bd_int_g_comb = overapproximate(int_g_comb, BoxDirections)
+    @test overapproximate(o_d_int_g_comb) ≈ overapproximate(o_bd_int_g_comb)
```
schillic:
```suggestion
    @test overapproximate(o_d_int_g_comb) ≈ overapproximate(o_bd_int_g_comb)
    projection = project(o_bd_int_g_comb, [1, 2], OctDirections)
    @test vertices_list(projection) ≈ [[0., 2.5], [0., 2.], [0.5, 2.]]
```

kostakoida: Projection of `x+y<=n ∩ Hyperrectangle` onto *2 blocks* x and y won't be a triangel, it will be two intervals, so the result will be an interval hull or, in this case, `Hyperrectangle`

schillic: Indeed, I had locally reassigned the variable. I meant this:
```julia
projection = project(overapproximate(int_g_comb, OctDirections), [1, 2], OctDirections)
```

kostakoida:
> Please squash the commits and then let's wait for @mforets.

Can I squash without commiting to master?

@mforets Do you have any comments?

src/Approximations/overapproximate.jl
```diff
@@ -911,3 +912,63 @@ function _overapproximate_lm_cpa!(arr, M, cpa, overapprox_option)
...
+# symmetric method
+function overapproximate(cap::Intersection{N,
+                                            <:AbstractPolyhedron{N},
+                                            <:CartesianProductArray{N}},
+                            ::Type{CartesianProductArray}, oa) where {N}
+    overapproximate(Intersection(cap.Y, cap.X), oa)
+end
+
+
+end
```
```suggestion
```
This `end` should be removed.

src/CartesianProduct.jl
```diff
@@ -562,3 +563,143 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
...
+index of the `i`-th block, depending on whether one of the block's dimension indices is specified in `vars`.
+
+### Input
+
+- `cpa` -- Cartesian product array
+- `vars` -- list containing the variables of interest, sorted in ascending order
+
+### Output
+
+A vector of tuples, where values in tuple relate to range of dimensions in the i-th block.
```
mforets: The text doesn't correspond to the output (see line `630` below). The vector of tuples is only the first component of the output. Please define "non_empty_length" and add it in the docstring (or remove it if it's not needed).

src/CartesianProduct.jl
```diff
@@ -562,3 +563,143 @@ function same_block_structure(x::AbstractVector{S1}, y::AbstractVector{S2}
...
+        if v_i > length(vars)
+            break
+        end
+        start_index = end_index + 1
+    end
+    return result, non_empty_length
+end
+
+#same but for all variables
+function block_to_dimension_indices(cpa::CartesianProductArray{N}) where {N}
```
mforets: Do you have an example where it is worth (faster?) using the version with second argument `vars` instead of this one?

Also, add the `@inbounds` macro in front of the loop, as in` @inbounds for i in ...`, to eliminate array bounds checking and improve performance (a bit), since all accesses are known to be in bounds.

src/Translation.jl
```diff
@@ -364,3 +365,29 @@ This implementation relies on the set membership function for the wrapped set
...
+### Output
+
+A concrete set corresponding to the linear map.
+The type of the result depends on the type of the set wrapped by `tr`.
+
+### Algorithm
+
+We compute `translate(linear_map(M, tr.X), M * tr.v)`.
+"""
+function linear_map(M::AbstractMatrix{N}, tr::Translation{N}) where {N<:Real}
```
mforets: This function is unrelated to this PR 😕

src/concrete_intersection.jl
```diff
@@ -782,3 +782,36 @@ function intersection(rm::ResetMap{N, <:AbstractPolytope},
...
+The decomposed set which represents concrete intersection between `X` and `Y`
+
+### Algorithm
+
+This algorithm intersect corresponding blocks between sets.
+"""
+function intersection(X::CartesianProductArray{N}, Y::CartesianProductArray{N}) where {N<:Real}
+    @assert same_block_structure(array(X), array(Y)) "block structure has to be the same"
+
+    return CartesianProductArray([intersection(array(X)[i], array(Y)[i]) for i in eachindex(X.array)])
```
```suggestion
    return CartesianProductArray([intersection(array(X)[i], array(Y)[i]) for i in eachindex(array(X))])
```

test/unit_overapproximate.jl
```diff
@@ -165,6 +167,78 @@ for N in [Float64, Float32]
...
+    i2 = Interval(N[2, 3])
+    M = N[1 2; 0 1]
+    cpa = CartesianProductArray([i1, i2])
+    lm = M * cpa
+    d_oa = overapproximate(lm, CartesianProductArray{N, Interval{N}})
+    oa = overapproximate(lm, OctDirections)
+    @test oa ⊆ d_oa
+
+    # =======================================
+    # Zonotope overapprox. of a Taylor model
```
mforets: Again `Zonotope overapprox. of a Taylor model` does not belong to this PR.

src/CartesianProduct.jl
kostakoida: > Do you have an example where it is worth (faster?) using the version with second argument `vars` instead of this one?

We don't need information about all blocks usually, therefore it is useful to use version with `vars` as the second parameter in functions, which call `block_to_dimension_indices`.

mforets: Ok. I was asking because i benchmarked these two functions in two cases that i created (one small dim and another high dim), and the general one was slightly *faster* in both cases.

PR #1425, opened 2019-06-06
Simplifications in overapproximate

schillic:
some tweaks for #1419

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -516,84 +516,103 @@ function overapproximate(cap::Intersection{N,
 end
 
 """
-    overapproximate(lm::LinearMap{N, CartesianProductArray{N, T}},
-                      ::Type{CartesianProductArray{N, O}}) where {N, T<:LazySet{N}, O<:LazySet{N}}
+    overapproximate(lm::LinearMap{N, <:CartesianProductArray{N}},
+                    ::Type{CartesianProductArray{N, S}}
+                   ) where {N, S<:LazySet{N}}
 
-Decompose a lazy linear map of a cartesian product array keeping original block structure.
+Decompose a lazy linear map of a cartesian product array while keeping the
+original block structure.
 
 ### Input
 
-- `lm`   -- lazy linear map of cartesian product array
+- `lm` -- lazy linear map of cartesian product array
 - `CartesianProductArray` -- type for dispatch
 
 ### Output
 
 A `CartesianProductArray` representing the decomposed linear map.
 """
-function overapproximate(lm::LinearMap{N, CartesianProductArray{N, T}},
-                           ::Type{CartesianProductArray{N, O}}) where {N, T<:LazySet{N}, O<:LazySet{N}}
-
-    M, cpa = lm.M, lm.X
-    array = Vector{O}(undef, length(cpa.array))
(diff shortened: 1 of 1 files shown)


src/Approximations/overapproximate.jl
```diff
@@ -516,84 +516,103 @@ function overapproximate(cap::Intersection{N,
...
 
 A `CartesianProductArray` representing the decomposed linear map.
 """
-function overapproximate(lm::LinearMap{N, CartesianProductArray{N, T}},
-                           ::Type{<:CartesianProductArray},
-                         dir::Type{<:AbstractDirections}) where {N, T<:LazySet{N}}
-    M, cpa = lm.M, lm.X
-    array = Vector{HPolytope{N}}(undef, length(cpa.array))
-    return _overapproximate_lm_cpa!(M, cpa, array, dir)
+function overapproximate(lm::LinearMap{N, <:CartesianProductArray{N}},
```
mforets: This change is such that if the second argument is, for example `CartesianProductarray{N, HPolygon{N}}`, the `HPolygon` is ignored. Can you add this to the docstring?

src/Approximations/overapproximate.jl
```diff
@@ -516,84 +516,103 @@ function overapproximate(cap::Intersection{N,
...
+
+- `lm`  -- lazy linear map of a cartesian product array
+- `CartesianProductArray` -- type for dispatch
+- `set_type` -- set type for overapproximation
+
+### Output
+
+A `CartesianProductArray` representing the decomposed linear map.
+"""
+function overapproximate(lm::LinearMap{N, <:CartesianProductArray{N}},
```
mforets: This function is a repetition of https://github.com/JuliaReach/LazySets.jl/pull/1425/files#diff-eb4fb67940b1678df4462dad2b59a25cR535 , with a different signature (sounds like it's not a good idea to have many of these methods, but OK), i suggest to add a comment of this fact in the docstring.

kostakoida:
> LGTM.
> 
> @kostakoida do you have something else to comment?

Looks good to me too.

src/Approximations/overapproximate.jl
schillic: No, there is no functional change here. You could not specify the output type before either. The `T` parameter was just not used, so I removed it. We *always* allocate a `Vector{HPolytope{N}}` because template-direction overapproximation returns a `HPolytope{N}`.

src/Approximations/overapproximate.jl
schillic: Again, I did not change functionality here.

> sounds like it's not a good idea to have many of these methods

This is the Julia way, I'm sorry.

src/Approximations/overapproximate.jl
mforets: You're right! It is ignored in both cases. I opened https://github.com/JuliaReach/LazySets.jl/issues/1430 as follow up.

src/Approximations/overapproximate.jl
mforets: It doesn't prevent from improving the docstring (same for previous comment). In any case, i opened https://github.com/JuliaReach/LazySets.jl/issues/1431 as follow-up.

mforets:
Thanks!

PR #1426, opened 2019-06-07
Set default aspect ratio in plot recipes to :none

schillic:
EDIT: outdated
I used `:auto` because it seems to be an intuitive value, but `Plots` does not recognize it when using `plot!`. So I had to add these `if`s. However, when I use the value `:none` instead, `plot!` overwrites the value. So I think we need these `if`s anyway, and then I prefer `:auto`.

### DIFF
--- a/src/plot_recipes.jl
+++ b/src/plot_recipes.jl
@@ -8,7 +8,7 @@ DEFAULT_COLOR = :auto
 DEFAULT_ALPHA = 0.5
 DEFAULT_LABEL = ""
 DEFAULT_GRID = true
-DEFAULT_ASPECT_RATIO = 1.0
+DEFAULT_ASPECT_RATIO = :none
 PLOT_PRECISION = 1e-3
 PLOT_POLAR_DIRECTIONS = 40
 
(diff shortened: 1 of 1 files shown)


mforets:
> I used :auto because it seems to be an intuitive value, but Plots does not recognize it when using plot!

It is not recognised at all (so i discourage to use it); in recent versions of Plots only `:equal` or `:none` (or a number) are allowed, see [here](https://github.com/JuliaPlots/Plots.jl/blob/master/src/args.jl#L341). E.g. in `master`

```julia
julia> plot(rand(Zonotope), aspect_ratio=:auto)
Error showing value of type Plots.Plot{Plots.GRBackend}:

ERROR: MethodError: no method matching /(::Float64, ::Symbol)
...
``` 


> However, when I use the value :none instead, plot! overwrites the value.

I cannot reproduce, do you have an example?

```julia
DEFAULT_ASPECT_RATIO = :none

...

julia> plot(rand(Zonotope))

julia> plot!(rand(Zonotope))
```

schillic:
> I cannot reproduce, do you have an example?

In `master` edit `plot_recipes.jl` as follows:
```julia
DEFAULT_ASPECT_RATIO = :none
```
Then do:
```julia
julia> Z = Zonotope([-1.19275, 1.6019], [0.229994 -0.231331; -1.17255 -0.997225]);

julia> plot(Z, aspectratio=1)

julia> plot!(Singleton(σ([1.,0], Z)))
```
The picture is first brought to `aspectratio=1` but then `plot!` overwrites it.

PR #1428, opened 2019-06-10
#1355 - Remove redundant vertices of LinearMap

### LINKED ISSUE #1355: Remove redundant vertices of LinearMap
Proposal: add an optional argument (default: `true`) for the `vertices_list` of a lazy linear map that takes the convex hull.

Possible test:

```julia
julia> X = rand(Interval) × ZeroSet(1);

julia> M = [0 1.; 0 2];

julia> vertices_list(M * X)  # should return only [[0.0, 0.0]]
 [0.0, 0.0]
 [0.0, 0.0]
```

SebastianGuadalupe:
Closes #1355.

### DIFF
--- a/src/LinearMap.jl
+++ b/src/LinearMap.jl
@@ -316,13 +316,14 @@ function isempty(lm::LinearMap)::Bool
 end
 
 """
-    vertices_list(lm::LinearMap{N})::Vector{Vector{N}} where {N<:Real}
+    vertices_list(lm::LinearMap{N}; prune::Bool=true)::Vector{Vector{N}} where {N<:Real}
 
 Return the list of vertices of a (polyhedral) linear map.
 
 ### Input
 
 - `lm` -- linear map
+- `prune` -- (optional, default: `true`) if true removes redundant vertices
 
 ### Output
 
@@ -333,7 +334,7 @@ A list of vertices.
 We assume that the underlying set `X` is polyhedral.
 Then the result is just the linear map applied to the vertices of `X`.
 """
-function vertices_list(lm::LinearMap{N})::Vector{Vector{N}} where {N<:Real}
+function vertices_list(lm::LinearMap{N}; prune::Bool=true)::Vector{Vector{N}} where {N<:Real}
     # for a zero map, the result is just the list containing the origin
     if iszero(lm.M)
         return [zeros(N, dim(lm))]
@@ -349,7 +350,7 @@ function vertices_list(lm::LinearMap{N})::Vector{Vector{N}} where {N<:Real}
         push!(vlist, lm.M * v)
     end
 
-    return vlist
+    return prune ? convex_hull(vlist) : vlist
 end
 
 """
--- a/test/unit_LinearMap.jl
+++ b/test/unit_LinearMap.jl
@@ -84,6 +84,11 @@ for N in [Float64, Rational{Int}, Float32]
     vlist = vertices_list(LinearMap(M, b))
     @test vlist == [zeros(N, 2)]
 
+    # test that redundant vertices are removed by default (see #1355)
+    X = Interval(N(0), N(1)) × ZeroSet{N}(1)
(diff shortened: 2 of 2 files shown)


mforets:
Thanks for your contribution! 

Can you add the example in https://github.com/JuliaReach/LazySets.jl/issues/1355#issue-442393136 as a new test? It can go in the `test/unit_LinearMap.jl` file.

src/LinearMap.jl
```diff
@@ -333,7 +334,7 @@ A list of vertices.
 We assume that the underlying set `X` is polyhedral.
 Then the result is just the linear map applied to the vertices of `X`.
 """
-function vertices_list(lm::LinearMap{N})::Vector{Vector{N}} where {N<:Real}
+function vertices_list(lm::LinearMap{N}; remove_redundant::Bool=true)::Vector{Vector{N}} where {N<:Real}
```
schillic: Can we rename `remove_redundant` to `prune`? This is shorter and used for the H-polyhedron already. I just checked and we do not have `remove_redundant` elsewhere.

test/unit_LinearMap.jl
```diff
@@ -83,6 +83,9 @@ for N in [Float64, Rational{Int}, Float32]
     M = zeros(N, 2, 2)
     vlist = vertices_list(LinearMap(M, b))
     @test vlist == [zeros(N, 2)]
+    X = rand(Interval) × ZeroSet(1)
```
mforets: In principle you'd use `rand(Interval, N=N)` to create a random interval of numeric type `N` (such that it matches the numeric type of `M` -- because the constructor of `M * X` doesn't handle different numeric types, although being supported by the lazy linear map).

However, `rand(Interval, N=Rational{Int})` doesn't work. So i would suggest to pick any interval, say `X = Interval(N(0), N(1))`.

src/LinearMap.jl
```diff
@@ -349,7 +350,11 @@ function vertices_list(lm::LinearMap{N})::Vector{Vector{N}} where {N<:Real}
         push!(vlist, lm.M * v)
     end
 
-    return vlist
+    if prune
+        return convex_hull(vlist)
+    else
+        return vlist
+    end
```
mforets: In cases such as these, the [ternary operator](https://docs.julialang.org/en/v1/manual/control-flow/index.html#man-conditional-evaluation-1) simplifies a bit and i would recommend it: `return prune ? convex_hull(vlist) : vlist`

test/unit_LinearMap.jl
```diff
@@ -83,6 +83,9 @@ for N in [Float64, Rational{Int}, Float32]
     M = zeros(N, 2, 2)
     vlist = vertices_list(LinearMap(M, b))
     @test vlist == [zeros(N, 2)]
+    X = X = Interval(N(0), N(1)) × ZeroSet(1)
```
mforets:
```suggestion
    X = Interval(N(0), N(1)) × ZeroSet(1)
```

mforets:
```suggestion

    # test that redundant vertices are removed by default (see #1355)
    X = Interval(N(0), N(1)) × ZeroSet(1)
```

test/unit_LinearMap.jl
```diff
@@ -84,6 +84,11 @@ for N in [Float64, Rational{Int}, Float32]
     vlist = vertices_list(LinearMap(M, b))
     @test vlist == [zeros(N, 2)]
 
+    # test that redundant vertices are removed by default (see #1355)
+    X = Interval(N(0), N(1)) × ZeroSet(1)
```
schillic:
```suggestion
    X = Interval(N(0), N(1)) × ZeroSet{N}(1)
```

schillic: After that fix, can you squash the commits, please?

mforets:
Thanks a lot!

PR #1433, opened 2019-06-12
Pr/1408

### LINKED ISSUE #1395: Approximation of linear taylor models with zonotopes
As a first step we can add the conversion for *linear* TMs in one or several variables, using the idea in Xin Chen's thesis page 52 and Lemma 2.4.8 

The normalization in the symmetric interval `[-1, 1]` used in zonotopes is solved in `TaylorSeries.normalize_taylor`.

Our code can use `Require`s with `TaylorModels`s as an optional dependency. In the linear case the approximation is exact so we can also add corresponding `convert` methods. (Side note: In LazySets we keep `convert` to be used for "exact" conversions, and `overapproximate` for those cases where the conversion *may* add overapproximation error.)

We can do the non-linear case as a follow-up issue; a "simple but effective way" is described in Xin Chen's thesis page 112, subsection "Zonotopic over-approximation". It consists of linearizing the TM and using the approximation from this issue.

```julia
function overapproximate(vTM::Vector{TaylorModel1{T, S}}}, ::Type{Zonotope}) where {T, S}
   @assert (that the vector of TMs is linear, or explain in the docs until this assumption is lifted)
   ...
end

function overapproximate(vTM::Vector{TaylorModelN{N, T, S}}}, ::Type{Zonotope}) where {N, T, S}
     @assert (that the vector of TMs is linear, or explain in the docs until this assumption is lifted)
      ....
end
```

---

CC: @Aadesha, @lbenet

### LINKED ISSUE #1396: Approximation of nonlinear TMs with zonotopes
Follow up of https://github.com/JuliaReach/LazySets.jl/issues/1395.

mforets:
Closes #1395.
Closes #1396.


test/unit_Polytope.jl
```diff
@@ -271,7 +271,7 @@ if test_suite_polyhedra
         @test !isdisjoint(P, Q; algorithm="sufficient")
         res, w = isdisjoint(Q, P, true)
         @test !isdisjoint(Q, P) && !res && w ∈ P && w ∈ Q
-        res, w = isdisjoint(Q, R, true)
+        res, w =isdisjoint(Q, R, true)
```
schillic: unrelated and unnecessary change

mforets:
FWIW, tests pass in my machine for any `N`. See

schillic:
The `IntervalBox` had type `Float64`. Seems like this part of the example does not depend on `N`.

mforets:
I get

```julia
using IntervalArithmetic; const IA = IntervalArithmetic

for N in [Float32, Float64]
    Dx₁ = IA.Interval(N(1.0), N(3.0))
    Dx₂ = IA.Interval(N(-1.0), N(1.0))
    Dx₃ = IA.Interval(N(-1.0), N(0.0))
    D = Dx₁ × Dx₂ × Dx₃   # domain
    x0 = IntervalBox(IA.mid.(D)...)
    @show typeof(x0)
end

typeof(x0) = IntervalBox{3,Float32}
typeof(x0) = IntervalBox{3,Float64}
```

schillic:
I get `IntervalBox{3,Float64}` in both cases. I use the release version of `IntervalArithmetic`.

PR #1434, opened 2019-06-13
#1384 - Add four point case for convex hull

### LINKED ISSUE #1384: Convex hull of four points
Investigate if the particular case of the convex hull of four points in the plane can be done more efficiently than the current (generic) implementation.

See also: https://stackoverflow.com/questions/2122305/convex-hull-of-4-points

SebastianGuadalupe:
Closes #1384.

### DIFF
--- a/src/concrete_convex_hull.jl
+++ b/src/concrete_convex_hull.jl
@@ -190,7 +190,7 @@ function _three_points_2d!(points::AbstractVector{<:AbstractVector{N}}) where {N
     return points
 end
 
-function _collinear_case!(points::Vector{VN}, A::VN, B::VN, C::VN, D::VN)::Vector{VN} where {N<:Real, VN<:AbstractVector{N}}
+function _collinear_case!(points::Vector{VN}, A::VN, B::VN, C::VN, D::VN) where {N<:Real, VN<:AbstractVector{N}}
     # A, B and C collinear, D is the extra point
     if isapprox(A[1], B[1]) && isapprox(B[1], C[1]) && isapprox(C[1], A[1])
         # points are approximately equal in their first component
--- a/src/concrete_convex_hull.jl
+++ b/src/concrete_convex_hull.jl
@@ -190,6 +190,31 @@ function _three_points_2d!(points::AbstractVector{<:AbstractVector{N}}) where {N
     return points
 end
 
+function _collinear_case!(points::Vector{VN}, A::VN, B::VN, C::VN, D::VN)::Vector{VN} where {N<:Real, VN<:AbstractVector{N}}
+    # A, B and C collinear, D is the extra point
+    if isapprox(A[1], B[1]) && isapprox(B[1], C[1]) && isapprox(C[1], A[1])
+        # points are approximately equal in their first component
+        if isapprox(A[2], B[2]) && isapprox(B[2], C[2]) && isapprox(C[2], A[2])
+            # the three points are approximately equal
+            points[1], points[2] = A, D
+            pop!(points)
+            pop!(points)
+            return _two_points_2d!(points)
+        else
+            # assign the points with max and min value in their second component to the
+            # firsts points and the extra point to the third place, then pop the point that was in the middle
+            points[1], points[2], points[3] = points[argmin([A[2], B[2], C[2]])], points[argmax([A[2], B[2], C[2]])], D
+            pop!(points)
+        end
+    else
+        # assign the points with max and min value in their first component to the
+        # firsts points and the extra point to the third place, then pop the point that was in the middle
+        points[1], points[2], points[3] = points[argmin([A[1], B[1], C[1]])], points[argmax([A[1], B[1], C[1]])], D
+        pop!(points)
+    end
+    return _three_points_2d!(points)
+end
+
 function _four_points_2d!(points::AbstractVector{<:AbstractVector{N}}) where {N<:Real}
     A, B, C, D = points[1], points[2], points[3], points[4]
     tri_ABC = right_turn(A, B, C)
@@ -210,42 +235,17 @@ function _four_points_2d!(points::AbstractVector{<:AbstractVector{N}}) where {N<
         key = key + 1
     end
     
-    function collinear_case(A, B, C, D)
-        # A, B and C collinear, D is the extra point
-        if isapprox(A[1], B[1]) && isapprox(B[1], C[1]) && isapprox(C[1], A[1])
-            # points are approximately equal in their first component
-            if isapprox(A[2], B[2]) && isapprox(B[2], C[2]) && isapprox(C[2], A[2])
-                # the three points are approximately equal
-                points[1], points[2] = A, D
-                pop!(points)
-                pop!(points)
-                return _two_points_2d!(points)
-            else
-                # assign the points with max and min value in their second component to the
-                # firsts points and the extra point to the third place, then pop the point that was in the middle
-                points[1], points[2], points[3] = points[argmin([A[2], B[2], C[2]])], points[argmax([A[2], B[2], C[2]])], D
-                pop!(points)
-            end
-        else
-            # assign the points with max and min value in their first component to the
-            # firsts points and the extra point to the third place, then pop the point that was in the middle
-            points[1], points[2], points[3] = points[argmin([A[1], B[1], C[1]])], points[argmax([A[1], B[1], C[1]])], D
-            pop!(points)
-        end
-        return _three_points_2d!(points)
-    end
-    
     if isapproxzero(tri_ABC)
-        return collinear_case(A, B, C, D)
+        return _collinear_case!(points, A, B, C, D)
     end
     if isapproxzero(tri_ABD)
-        return collinear_case(A, B, D, C)
+        return _collinear_case!(points, A, B, D, C)
     end
     if isapproxzero(tri_BCD)
-        return collinear_case(B, C, D, A)
+        return _collinear_case!(points, B, C, D, A)
     end
     if isapproxzero(tri_CAD)
-        return collinear_case(C, A, D, B)
+        return _collinear_case!(points, C, A, D, B)
     end
     
     # ABC  ABD  BCD  CAD  hull
--- a/src/concrete_convex_hull.jl
+++ b/src/concrete_convex_hull.jl
@@ -28,8 +28,10 @@ The convex hull as a list of vectors with the coordinates of the points.
 
 ### Algorithm
 
-A pre-processing step treats the cases with `0`, `1`, `2` and `3` points in any dimension.
-For `4` or more points, the algorithm used depends on the dimension.
+A pre-processing step treats the cases with `0`, `1` and `2` points for one
+dimension, `0`, `1`, `2`, `3` and `4` for two dimensions, if there is more
+points, its used a general algorithm for one and two dimensions.
+For higher dimension there is a general algorithm.
 
 For the one-dimensional case we return the minimum and maximum points, in that order.
 
@@ -114,6 +116,9 @@ function convex_hull!(points::Vector{VN};
         elseif m == 3
             # three points case in 2d
             return _three_points_2d!(points)
+        elseif m == 4
+            # four points case in 2d
+            return _four_points_2d!(points)
         else
             # general case in 2d
             return _convex_hull_2d!(points, algorithm=algorithm)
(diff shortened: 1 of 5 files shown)


schillic:
There is a problem with the sorting of vertices (hence the error):
```julia
julia> using LazySets
julia> B = Ball1(zeros(2), 1.);

# this branch:
julia> P = convert(VPolygon, B)
VPolygon{Float64,Array{Float64,1}}(Array{Float64,1}[[0.0, -1.0], [-1.0, 0.0], [0.0, 1.0], [1.0, 0.0]])

# master:
julia> P = convert(VPolygon, B)
VPolygon{Float64,Array{Float64,1}}(Array{Float64,1}[[-1.0, 0.0], [0.0, -1.0], [1.0, 0.0], [0.0, 1.0]])
```
Please add a test case that detects this.

src/concrete_convex_hull.jl
```diff
@@ -186,6 +191,126 @@ function _three_points_2d!(points::AbstractVector{<:AbstractVector{N}}) where {N
...
+    elseif key == 1011
+        points[1], points[2], points[3], points[4] = A, D, B, C #  +    -    +    +   ADBC
+    elseif key == 1010
+        points[1], points[2], points[3] = B, C, D #  +    -    +    -   BCD
+        pop!(points)
+    elseif key == 1001
+        points[1], points[2], points[3] = C, A, D #  +    -    -    +    CAD
+        pop!(points)
+    elseif key == 1000
+        #  +    -    -    -   [should not happen]
```
schillic: Instead of empty branches, add an `else` case at the end with this body:
```julia
@assert false "unexpected case in convex_hull"
```

src/concrete_convex_hull.jl
```diff
@@ -186,6 +191,126 @@ function _three_points_2d!(points::AbstractVector{<:AbstractVector{N}}) where {N
...
+    #  +    -    -    -   [should not happen]
+    #  -    +    +    +   [should not happen]
+    #  -    +    +    -   ACD
+    #  -    +    -    +   DCB
+    #  -    +    -    -   DBCA
+    #  -    -    +    +   ADB
+    #  -    -    +    -   ACDB
+    #  -    -    -    +   ADCB
+    #  -    -    -    -   ACB
+    if key == 1111
```
schillic: A slightly faster implementation (with a bit more to write) is the following:
```julia
# +    +    +    +   ABC
function _four_points_2d_helper!(points, key::Val{1111}, A, B, C, D)
    points[1], points[2], points[3] = A, B, C
    pop!(points)
end
# +    +    +    -   ABCD
function _four_points_2d_helper!(points, key::Val{1110}, A, B, C, D)
     points[1], points[2], points[3], points[4] = A, B, C, D
end
...
# fallback
function _four_points_2d_helper!(points, key, A, B, C, D)
    @assert false "unexpected case in convex_hull"
end

_four_points_2d_helper!(points, key, A, B, C, D)
```
Rationale: The compiler uses a jump table to choose the correct method instead of running along this long if-then-else chain.

These helper methods can also be added inside the function `_four_points_2d!` which may allow to omit the arguments apart from `key` (not sure if this has better, worse, or the same performance).

src/concrete_convex_hull.jl
```diff
@@ -28,8 +28,10 @@ The convex hull as a list of vectors with the coordinates of the points.
 
 ### Algorithm
 
-A pre-processing step treats the cases with `0`, `1`, `2` and `3` points in any dimension.
-For `4` or more points, the algorithm used depends on the dimension.
+A pre-processing step treats the cases with `0`, `1` and `2` points for one
+dimension, `0`, `1`, `2`, `3` and `4` for two dimensions, if there is more
+points, its used a general algorithm for one and two dimensions.
+For higher dimension there is a general algorithm.
```
schillic:
```
A pre-processing step treats the cases with up to two points for one dimension
and up to four points for two dimensions.
For more points in one resp. two dimensions, we use more general algorithms.
```

src/concrete_convex_hull.jl
```diff
@@ -186,6 +191,126 @@ function _three_points_2d!(points::AbstractVector{<:AbstractVector{N}}) where {N
     return points
 end
 
+function _four_points_2d!(points::AbstractVector{<:AbstractVector{N}}) where {N<:Real}
+    A, B, C, D = points[1], points[2], points[3], points[4]
+    tri_ABC = (A[2] - B[2]) * C[1] + (B[1] - A[1]) * C[2] + (A[1] * B[2] - B[1] * A[2])
```
mforets: Define a helper function and use the factored form:

```julia
@inline _orientation(A, B, C) = (A[2] - B[2]) * (C[1] - A[1]) + (B[1] - A[1]) * (C[2] - A[2])
```

mforets: .. or (better) use [right_turn](https://github.com/JuliaReach/LazySets.jl/blob/master/src/concrete_convex_hull.jl#L243) 😄

src/concrete_convex_hull.jl
schillic: I would still add an `else` case at the end.

src/concrete_convex_hull.jl
schillic: Hm, the three patterns do not match. They might still be equivalent, but it is not obvious to me.

schillic:
```julia
LazySets.ConvexHull: Test Failed at /home/travis/build/JuliaReach/LazySets.jl/test/unit_ConvexHull.jl:105
  Expression: convex_hull!(points, algorithm="")
    Expected: ErrorException
  No exception thrown
```
This error is a bug in our tests:
https://github.com/JuliaReach/LazySets.jl/blob/a5545aeed10c2ee91b7f022c2e2a7cf5c8ae05fb/test/unit_ConvexHull.jl#L103-L105
The tests should instead either use the non-`!` function or use `copy(points)`, e.g.:
```julia
    @test convex_hull!(copy(points), algorithm="monotone_chain") == sorted
    @test convex_hull!(copy(points), algorithm="monotone_chain_sorted") == sorted
    @test_throws ErrorException convex_hull!(copy(points), algorithm="")
```

src/concrete_convex_hull.jl
schillic: After benchmarking, this turned out to be slower.

schillic:
I wonder why your tests did not catch the first error. Can you add a test for that case?

test/unit_convex_hull.jl
```diff
@@ -38,23 +38,70 @@ for N in [Float64, Rational{Int}]
     convex_hull!(points_2D) # check in-place version
     @test points_2D == [[N(-1), N(-1)], [N(1), N(0)], [N(1), N(1)], [N(0), N(1)]]
 
-    # three vertex case in 2 dimensions
-    function iscounterclockwise(result, correct_expr)
+    function iscounterclockwise_four_points(result, correct_expr)
```
mforets: Define `iscounterclockwise` using `any(... circshift(correct_expr, i))` so it applies to any number of points.

schillic: I think this is a helpful method to define in `src/helper_functions.jl`. You can then include it here:
https://github.com/JuliaReach/LazySets.jl/blob/894fda37e6b0dc2cf9746c1accdebc5d8757f814/test/runtests.jl#L14

src/helper_functions.jl
```diff
@@ -645,3 +645,21 @@ A copy of the vector where each negative entry is replaced by zero.
 function rectify(x::AbstractVector{N}) where {N<:Real}
     return map(xi -> max(xi, zero(N)), x)
 end
+
+"""
+    iscounterclockwise(result, correct_expr) where {N<:Real}
+
+Returns a boolean, i.e., take the element-wise maximum with zero.
```
schillic: I think this description is not informative (it might even be the wrong function?).

src/helper_functions.jl
```diff
@@ -645,3 +645,21 @@ A copy of the vector where each negative entry is replaced by zero.
...
+
+### Input
+
+- `result` -- vector
+- `correct_expr` -- vector with it's elements in ccw order
```
schillic:
```suggestion
- `correct_expr` -- paragon vector with elements in ccw order
```

src/helper_functions.jl
```diff
@@ -645,3 +645,21 @@ A copy of the vector where each negative entry is replaced by zero.
...
+
+### Output
+
+A boolean indicating if the elements of the result are in the same order as correct_expr or any of it's cyclic permutations
```
schillic:
```suggestion
A boolean indicating if the elements of `result` are in the same order as
`correct_expr` or any of its cyclic permutations.
```

test/unit_convex_hull.jl
```diff
@@ -55,6 +51,47 @@ for N in [Float64, Rational{Int}]
...
+    points = [A, B, C, D]
+    @test  iscounterclockwise(convex_hull!(points), expr) # ABCD
+    points = [A, D, C, B]
+    @test  iscounterclockwise(convex_hull!(points), expr) # ADCB
+    points = [A, B, D, C]
+    @test  iscounterclockwise(convex_hull!(points), expr) # ABDC
+    points = [A, D, B, C]
+    @test  iscounterclockwise(convex_hull!(points), expr) # ADBC
+    points = [D, B, C, A]
+    @test  iscounterclockwise(convex_hull!(points), expr) # DBCA
```
schillic: This test is redundant (covered by ADBC) and instead it should test the case DACB (which corresponds to the bug we saw).

src/concrete_convex_hull.jl
mforets: Let:
```julia
    tri_ABC = (A[2] - B[2]) * C[1] + (B[1] - A[1]) * C[2] + (A[1] * B[2] - B[1] * A[2]) 
... adding and substracting:  
= (A[2] - B[2]) * C[1] + (B[1] - A[1]) * C[2] + A[1] * B[2] - B[1] * A[2] + A[2]*A[1] - A[2]*A[1]
= (A[2] - B[2]) * C[1] + (B[1] - A[1]) * C[2] - (A[2] - B[2])(A[1]) - (B[1] - A[1])*(A[2])
= (A[2] - B[2])*(C[1] - A[1]) + (B[1] - A[1]) * (C[2] - A[2])
```
while:
```
right_turn(O, A, B) = (A[1] - O[1]) * (B[2] - O[2]) - (A[2] - O[2]) * (B[1] - O[1])
=>
right_turn(A, B, C) = (B[1] - A[1]) * (C[2] - A[2]) - (B[2] - A[2]) * (C[1] - A[1])
                               =  (A[2] - B[2]) * (C[1] - A[1]) + (B[1] - A[1]) * (C[2] - A[2])  
```
=> `tri_ABC = right_turn(A, B, C)`.

src/concrete_convex_hull.jl
```diff
@@ -186,6 +190,124 @@ function _three_points_2d!(points::AbstractVector{<:AbstractVector{N}}) where {N
     return points
 end
 
+function _collinear_case!(points::Vector{VN}, A::VN, B::VN, C::VN, D::VN)::Vector{VN} where {N<:Real, VN<:AbstractVector{N}}
```
schillic:
```suggestion
function _collinear_case!(points::Vector{VN}, A::VN, B::VN, C::VN, D::VN) where {N<:Real, VN<:AbstractVector{N}}
```

mforets: Note that the outer signature is already typed to be a `Vector` (in and out), what is kept arbitrary is the representation of each point:

```julia
function convex_hull!(points::Vector{VN};
                      algorithm=default_convex_hull_algorithm(points),
                      backend=nothing)::Vector{VN} where {N<:Real, VN<:AbstractVector{N}}
...
```

schillic: True, and in full generality, each element could have a different concrete type. The most general version would thus be:
```julia
function _collinear_case!(points::AbstractVector{N}, A::AbstractVector{N},
    B::AbstractVector{N}, C::AbstractVector{N}, D::AbstractVector{N}) where {N<:Real}
```
I think if you do not use this version, the method will crash if you use, e.g., `SparseVector`s.

schillic: Do you need the type annotations or can you just remove them?

mforets: After discussion on gitter, @SebastianGuadalupe  would you please remove type annotations in `_collinear_case!`?

schillic: And then squash :+1:

mforets: After discussion on gitter, @SebastianGuadalupe  would you please remove type annotations in `_collinear_case!`?

mforets:
> after addressing the open comment

i'll handle it, plus some other minor tweaks in the same file that i planned to do, in a follow-up PR.

> squashing

done (from github).

Issue #1270, opened 2019-04-02
AbstractZonotope interface


schillic:
New interface: `AbstractZonotope <: AbstractCentrallySymmetricPolytope`
Subtypes: `Zonotope`, `LineSegment`, `AbstractHyperrectangle`
Necessary methods: `center` (already there), `generators` (missing but trivial)
Benefit: We can generalize many algorithms currently only written for `Zonotope`s and we do not need to convert hyperrectangles to zonotopes anymore*.

The existing interface `AbstractCentrallySymmetricPolytope` is more general. In particular, [the `Ball1` is a centrally symmetric polytope but not a zonotope](http://page.math.tu-berlin.de/~henk/preprints/henk%20linke%20wills&minimal%20zonotopes%20containing%20the%20crosspolytope.pdf).

*However, it may still be more efficient if we access the generators more than once.

PR #1444, opened 2019-06-16
#1413 - Add concrete version of zonotope algorithm to manual

### LINKED ISSUE #1413: Add concrete version of zonotope algorithm to manual
Plotting in [this part of the manual](https://github.com/JuliaReach/LazySets.jl/blob/master/docs/src/man/reach_zonotopes.md) takes 30 seconds for the first plot and 5 minutes for the second plot ([final result](https://juliareach.github.io/LazySets.jl/latest/man/reach_zonotopes)) .
Since the second plot is not very pretty anyway, we can save valuable time by reducing the number of sets that we plot to ~20%.
Furthermore, we should add an alternative that uses concrete operations because that is much more efficient. Then we can actually plot only the efficient version.

schillic:
Closes #1413.

This requires:
* [x] ~#1270 to define the `minkowski_sum` of a `Zonotope` and a `BallInf`~
* [x] ~#1450~

### DIFF
--- a/src/AbstractZonotope.jl
+++ b/src/AbstractZonotope.jl
@@ -136,3 +136,28 @@ An integer representing the number of generators.
 function ngens(Z::AbstractZonotope)::Int
     return length(generators(Z))
 end
+
+"""
+    minkowski_sum(Z1::AbstractZonotope{N}, Z2::AbstractZonotope{N})
+        where {N<:Real}
+
+Concrete Minkowski sum of a pair of zonotopic sets.
+
+### Input
+
+- `Z1` -- zonotopic set
+- `Z2` -- zonotopic set
+
+### Output
+
+A `Zonotope` corresponding to the concrete Minkowski sum of `Z1` and `Z2`.
+
+### Algorithm
+
+The resulting zonotope is obtained by summing up the centers and concatenating
+the generators of `Z1` and `Z2`.
+"""
+function minkowski_sum(Z1::AbstractZonotope{N},
+                       Z2::AbstractZonotope{N}) where {N<:Real}
+    return Zonotope(center(Z1) + center(Z2), [genmat(Z1) genmat(Z2)])
+end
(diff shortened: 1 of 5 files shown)


src/AbstractZonotope.jl
```diff
@@ -136,3 +136,55 @@ An integer representing the number of generators.
...
+A `Zonotope` corresponding to the concrete Minkowski sum of `Z1` and `Z2`.
+
+### Algorithm
+
+The resulting zonotope is obtained by summing up the centers and concatenating
+the generators of `Z1` and `Z2`.
+"""
+function minkowski_sum(Z1::AbstractZonotope{N},
+                       Z2::AbstractZonotope{N}) where {N<:Real}
+    return Zonotope(center(Z1) + center(Z2), [genmat(Z1) genmat(Z2)])
```
mforets: Interesting, this adds quite a bunch of concrete M-sums which are missing, given that `AbstractHyperrectangle <: AbstractZonotope`.

src/AbstractZonotope.jl
```diff
@@ -136,3 +136,55 @@ An integer representing the number of generators.
...
+### Output
+
+A zonotope corresponding to the concrete linear map `M` applied to `Z`.
+
+### Algorithm
+
+The resulting zonotope is obtained by applying the linear map `M` to the center
+and the generators of `Z`.
+"""
+function linear_map(M::AbstractMatrix{N},
```
mforets: Dispatching `linear_map` on the `AbstractZonotope` interface has effects on other other sets and we should consider it before making this change (given that `AbstractZonotope <: AbstractPolyhedron`, cf. [this function](https://github.com/JuliaReach/LazySets.jl/blob/master/src/AbstractPolyhedron_functions.jl#L300)).

For instance, in master `linear_map(M, H)` for `H` a hyperrectangle returns a new hyperrectangle, while this branch returns a zonotope. This is not necessarily bad, since linear maps on zonotopes are very fast, maybe even faster than the current implementation ( ;) ), which needs to shift the constraints according to `M`. However, returning an hyperrectangle seems better IMO (independently of the implementation).

Can you dispatch on `Zonotope` here and let the generalization for later?

schillic: Given that you already did the benchmarks, do you still think we should postpone?

schillic: The change in `linear_map` was moved to #1461.

schillic:
This makes the docs build 6 minutes faster (which affects every PR) :tada:

PR #1451, opened 2019-06-18
#1450 - Wrong constraints_list of zonotope with non-full-rank generator matrix

### LINKED ISSUE #1450: Wrong constraints_list of zonotope with non-full-rank generator matrix
```julia
julia> Z = Zonotope([0., 0.], [2. 3.; 0. 0.]);

julia> constraints_list(Z)
4-element Array{LazySets.HalfSpace{Float64,Array{Float64,1}},1}:
 LazySets.HalfSpace{Float64,Array{Float64,1}}([0.0, -1.0], 0.0)
 LazySets.HalfSpace{Float64,Array{Float64,1}}([0.0, -1.0], 0.0)
 LazySets.HalfSpace{Float64,Array{Float64,1}}([-0.0, 1.0], 0.0)
 LazySets.HalfSpace{Float64,Array{Float64,1}}([-0.0, 1.0], 0.0)

# for comparison, the following is the correct list
julia> P = convert(VPolygon, Z)
VPolygon{Float64,Array{Float64,1}}(Array{Float64,1}[[5.0, 0.0], [-5.0, 0.0]])

julia> constraints_list(P)
4-element Array{LazySets.HalfSpace{Float64,VN} where VN<:AbstractArray{Float64,1},1}:
 LazySets.HalfSpace{Float64,Array{Float64,1}}([10.0, -0.0], 50.0) 
 LazySets.HalfSpace{Float64,Array{Float64,1}}([0.0, 10.0], 0.0)   
 LazySets.HalfSpace{Float64,Array{Float64,1}}([-10.0, 0.0], 50.0) 
 LazySets.HalfSpace{Float64,Array{Float64,1}}([-0.0, -10.0], -0.0)
```

(Side note: The array type in the last command is suboptimal.)

schillic:
Closes #1450.

### DIFF
--- a/src/Zonotope.jl
+++ b/src/Zonotope.jl
@@ -600,7 +600,9 @@ Return the list of constraints defining a zonotope.
 
 ### Input
 
-- `Z` -- zonotope
+- `Z`               -- zonotope
+- `check_full_rank` -- (optional; default: `true`) flag for checking whether the
+                       generator matrix has full rank
 
 ### Output
 
@@ -613,8 +615,8 @@ The result has ``2 \\binom{p}{n-1}`` (with ``p`` being the number of generators
 and ``n`` being the ambient dimension) constraints, which is optimal under this
 assumption.
 
-If ``p < n``, we fall back to the (slower) computation based on the vertex
-representation.
+If ``p < n`` or the generator matrix is not full rank, we fall back to the
+(slower) computation based on the vertex representation.
 
 ### Algorithm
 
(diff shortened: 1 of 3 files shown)


schillic:
Invertibility is not the correct criterion because the matrix needs not be square. I will try "zero row", but this is just guessing.

mforets:
The constraints list algorithm has a precondition that the `p` generators are independent. `[2. 0]` and `[3. 0]` are not.

So perhaps we can check if the matrix of generators is [full rank](https://en.wikipedia.org/wiki/Rank_(linear_algebra)), and if it is not, remove the linearly dependent generators with the QR factorization and apply the algorithm for the independent case.

schillic:
I outsourced the efficient handling to #1453 to make this a quick bug fix only.

src/Zonotope.jl
```diff
@@ -625,15 +625,17 @@ Reachable Sets of Hybrid Systems Using a Combination of Zonotopes and Polytopes.
...
-function constraints_list(Z::Zonotope{N}) where {N<:AbstractFloat}
+function constraints_list(Z::Zonotope{N}; check_full_rank::Bool=true
+                         ) where {N<:AbstractFloat}
+    G = genmat(Z)
     p = ngens(Z)
     n = dim(Z)
-    if p < n
+
+    # use fallback implementation if order < 1 or matrix is not full rank
+    if p < n || (check_full_rank && rank(G) < size(G, 2))
```
mforets:
```suggestion
    if p < n || (check_full_rank && rank(G) < n)
```
The condition for full rank is `rank(G) == min(n, p)`. In the second clause, we can assume that `p >= n` so `G` is not full rank whenever `rank(G)` is smaller than `n`.

src/Zonotope.jl
```diff
@@ -625,15 +625,17 @@ Reachable Sets of Hybrid Systems Using a Combination of Zonotopes and Polytopes.
 The one-dimensional case is not covered by that algorithm; we manually handle
 this case, assuming that there is only one generator.
 """
-function constraints_list(Z::Zonotope{N}) where {N<:AbstractFloat}
+function constraints_list(Z::Zonotope{N}; check_full_rank::Bool=true
```
mforets: `check_full_rank` is missing in the docstring header and in the `Input` section.

PR #1452, opened 2019-06-19
#616 - Outsource matrix and vector machinery to Arrays module

### LINKED ISSUE #616: Refactoring of UnitVector
By the name `UnitVector` i would expect that the vector has norm one. But the constructor is more general: 

```julia
julia> LazySets.Approximations.UnitVector(1, 2, 2.0)
2-element LazySets.Approximations.UnitVector{Float64}:
 2.0
 0.0
```

Proposal: for many use cases we will need the default "1.0" entry, so that `LazySets.Approximations.UnitVector(1, 2)` corresponds to `LazySets.Approximations.UnitVector(1, 2, 1.0)`

On the other hand, maybe this type should be pulled out of the `Approximations` module.

schillic:
Closes #616.

- rename `UnitVector` -> `SingleEntryVector`
- rename `delete_zero_columns` -> `delete_zero_columns!`
- create `Arrays` module
- move matrix and vector machinery to `Arrays` module


src/Arrays/vector_operations.jl
```diff
@@ -0,0 +1,241 @@
...
+function rectify(x::AbstractVector{N}) where {N<:Real}
+    return map(xi -> max(xi, zero(N)), x)
+end
+
+"""
+    iscounterclockwise(result::AbstractVector,
+                       correct_expr::AbstractVector) where {N<:Real}
+
+Returns a boolean, true if the elements in `result` are ordered in
+a couter-clockwise fashion and in the same order as `correct_expr`.
```
mforets:
```suggestion
a counter-clockwise fashion and in the same order as `correct_expr`.
```

schillic: Since I copied this only and this is going to change in #1463, I rule that this stays unchanged :crown:

Issue #600, opened 2018-09-02
Specialized support function computations


schillic:
For some set types the support function can be computed much more efficiently than computing the support vector and taking the dot product.
See [Propositions 1 and 2 here](https://pdfs.semanticscholar.org/8fc5/cf0aadb3a450451896dec345d00c1b9a5643.pdf).

These are the current set types where we use the fallback definition:
* [x] `AbstractHyperrectangle` (#1454)
* [x] `AbstractSingleton` (#1454)
* [x] `Interval` (#1454)
* [x] `Ball1` (#3274)
* [x] `Ball2`
* [x] `Ballp` (#3274)
* [x] `EmptySet` (#1454)
* [x] `ExponentialProjectionMap` (#3280)
* ~~`HPolygon(Opt)`~~ (the default is already optimized because it uses binary search)
* [x] `Line`
* [x] `LineSegment` (#3275)
* [x] `Rectification` (#1381)
* [x] `SymmetricIntervalHull`
* ~~`VPolygon`~~ (the default is already optimized because it uses binary search)
* [x] `VPolytope` (#3276)
* [x] `ZeroSet` (#1454)

schillic:
The remaining items are all addressed in PRs now.

PR #1454, opened 2019-06-20
#600 - Specialized support function computations

schillic:
See #600.

```julia
julia> @btime ρ($[1., 0.], $(BallInf(zeros(2), 1.)))
  7.835 ns (0 allocations: 0 bytes)  # this branch
  87.346 ns (2 allocations: 192 bytes)  # master
```

### DIFF
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -263,6 +263,7 @@ This interface defines the following functions:
 
 ```@docs
 σ(::AbstractVector{N}, ::AbstractSingleton{N}) where {N<:Real}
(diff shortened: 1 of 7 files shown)


mforets:
Cool stuff 🎉 For `BallInf`, a similar "trick" can be applied to the support vector: instead of using `radius_hyperrectangle(H)` which allocates (because it has a `fill`), `radius_hyperrectangle(H)` is better since it can directly use `B.radius`.

schillic:
That is why I used `radius_hyperrectangle(H, i)`. I even used `BallInf` in my example above :wink:

Issue #1457, opened 2019-06-20
Unsound witness in isdisjoint for hyperrectangles


schillic:
```julia
julia> b1 = BallInf(2*ones(2), 0.5);
julia> b2 = BallInf(ones(2), 1.);
julia> isdisjoint(b1, b2, true)
(false, [1.0, 1.0])
```

PR #1458, opened 2019-06-20
#641 - Check for intersection of CartesianProductArray and high-dimensional Cartesian set

### LINKED ISSUE #641: Check for intersection of CartesianProductArray and high-dimensional Cartesian set
The idea is to check for intersection of each low-dimensional set in the `CartesianProductArray` with a suitable projection of the high-dimensional set.
Generally, this is approximate unless the set is Cartesian (e.g., `AbstractHyperrectangle`).

The default implementation can just pick the blocks from top to bottom.
Can we use heuristics to prefer those dimensions where the intersection is less likely? I do not see a simple criterion.
The individual checks are also trivially parallelizable.

schillic:
Closes #641.

~Requires #1457 for the tests to pass.~


mforets:
CC: @kostakoida

schillic:
[The new part](https://github.com/JuliaReach/LazySets.jl/compare/70d917f2861455785c41ee590358a717b1285dc5..105408a5e725e0cfa6d84f50a4b17190d9e08e37#diff-cf5d6cb9db1d1e2cb50a051651882d17R1451) is this:
```
### Algorithm

The sets `cpa` and `H` are disjoint if and only if at least one block of `cpa`
and the corresponding projection of `H` are disjoint.
We perform these checks sequentially.
```

PR #1467, opened 2019-06-22
#593 - Faster vertices_list implementation for flat hyperrectangles

### LINKED ISSUE #593: Faster vertices_list implementation for flat hyperrectangles
Follow-up of #92.
For each zero in the radius we can (and should) cut the number of vertices in half.
Example (current status):
```julia
julia> vertices_list(Hyperrectangle([1., 1.], [0., 1.]))
 [1.0, 2.0]
 [1.0, 2.0]
 [1.0, 0.0]
 [1.0, 0.0]
```
expected:
```julia
julia> vertices_list(Hyperrectangle([1., 1.], [0., 1.]))
 [1.0, 2.0]
 [1.0, 0.0]
```

schillic:
Closes #593.

The implementation is actually faster even if the radius does not contain zeros. The old implementation is only faster if the radius is all zero (because we had hard-coded that case; but I think this is such a rare case that we should not optimize for it).
(In the code piece below, `vertices_list2` is the new implementation.)

```julia
# no zeros in radius

julia> H = rand(Hyperrectangle, dim=4)
Hyperrectangle{Float64}([-1.03944, 0.108362, 1.35677, 0.492468], [0.0479897, 0.60482, 1.01331, 0.0135837])

julia> @btime vertices_list($H);
  1.661 μs (29 allocations: 2.91 KiB)

julia> @btime vertices_list2($H);
  748.148 ns (19 allocations: 2.16 KiB)


# many zeros in radius

julia> H = Hyperrectangle(H.center, [0., 0., 1., 0.])
Hyperrectangle{Float64}([-1.03944, 0.108362, 1.35677, 0.492468], [0.0, 0.0, 1.0, 0.0])

julia> length(vertices_list(H))
16

julia> length(vertices_list2(H))
2

julia> @btime vertices_list($H);
  1.676 μs (29 allocations: 2.91 KiB)

julia> @btime vertices_list2($H);
  177.610 ns (5 allocations: 528 bytes)


# radius is zero (was hard-coded before, hence the old code was faster)

julia> H = Hyperrectangle(H.center, zeros(4))
Hyperrectangle{Float64}([-1.03944, 0.108362, 1.35677, 0.492468], [0.0, 0.0, 0.0, 0.0])

julia> @btime vertices_list($H);
  79.719 ns (2 allocations: 208 bytes)

julia> @btime vertices_list2($H);
  138.946 ns (4 allocations: 416 bytes)
```

### DIFF
--- a/src/AbstractHyperrectangle.jl
+++ b/src/AbstractHyperrectangle.jl
@@ -117,7 +117,8 @@ end
 
 
 """
-    vertices_list(H::AbstractHyperrectangle{N})::Vector{Vector{N}} where {N<:Real}
+    vertices_list(H::AbstractHyperrectangle{N}
+                 )::Vector{Vector{N}} where {N<:Real}
 
 Return the list of vertices of a hyperrectangular set.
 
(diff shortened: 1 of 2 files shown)


src/AbstractHyperrectangle.jl
```diff
@@ -128,19 +128,52 @@ Return the list of vertices of a hyperrectangular set.
...
+        if iszero(ri)
+            trivector[i] = Int8(0)
+        else
+            v[i] += ri
+            trivector[i] = Int8(1)
+            m *= 2
+        end
+    end
+
+    # create vertices by modifying the three-valued vector and constructing the
```
mforets: Can you pull this information to a new `Algorithm` section? And maybe extend it, explaining the purpose of the three-valued vector.

schillic: Indeed I had to think about it myself again :smile:

PR #1470, opened 2019-06-25
Revise benchmarks folder

mforets:


schillic:
I am not convinced that adding random numbers is good for benchmarking unless the `judge` call uses the same objects. Do you know?

mforets:
In the calls to `rand`, the seed `rng` is fixed for each `n`, so the values of `d` and `X` are reproducible.

mforets:
I didn't remove it; it's been renamed them to `SUITE["ρ"]["LinearMap of $(string(set_type))", "dense", n] = @benchmarkable ρ($d, $Y)` in lines 22-23.

In follow-up PR i plan to add other versions of similar tests, such as sparse vectors. Adding other set types is also interesting.

schillic:
Indeed, sorry!

PR #1472, opened 2019-06-25
#1471 - No sorting for two-points convex hull

### LINKED ISSUE #1471: 2-point convex hull shouldn't sort vertices by default
`_two_points_2d!`  and `_two_points_1d!` currently sort the vertices in increasing angle wrt ccw ordering with zero being the positive "x" axis. However, we've seen that this condition is isn't required for support function / vector evaluations.

--

CC: @SebastianGuadalupe

mforets:
Closes #1471.

Requires: ~~#1484.~~

---

I hesitated to keep the current version and use dispatch on values, and finally decided against it. If we have an application that requires sorting, it is easy to reimplement it.

### DIFF
--- a/test/unit_convex_hull.jl
+++ b/test/unit_convex_hull.jl
@@ -6,7 +6,7 @@ for N in [Float64, Rational{Int}]
     # corner cases in dimension 1
     @test convex_hull([Vector{N}(undef, 0)]) == [Vector{N}(undef, 0)]
     @test convex_hull([[N(0)]]) == [[N(0)]]
-    @test convex_hull([[N(2)], [N(1)]]) == [[N(2)], [N(1)]]
+    @test ispermutation(convex_hull([N[2], N[1]]), [N[2], N[1]])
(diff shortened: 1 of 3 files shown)


test/unit_convex_hull.jl
```diff
@@ -6,7 +6,7 @@ for N in [Float64, Rational{Int}]
     # corner cases in dimension 1
     @test convex_hull([Vector{N}(undef, 0)]) == [Vector{N}(undef, 0)]
     @test convex_hull([[N(0)]]) == [[N(0)]]
-    @test convex_hull([[N(2)], [N(1)]]) == [[N(1)], [N(2)]]
+    @test convex_hull([[N(2)], [N(1)]]) == [[N(2)], [N(1)]]
```
schillic:
```suggestion
    @test ispermutation(convex_hull([N[2], N[1]]), [N[2], N[1]])
```
Or do you want to require that the implementation does not change the order?

PR #1473, opened 2019-06-25
#1471 - remove sorting of vertices by default in 2-point convex hull  

### LINKED ISSUE #1471: 2-point convex hull shouldn't sort vertices by default

SebastianGuadalupe:
Closes #1471.

### DIFF
--- a/src/concrete_convex_hull.jl
+++ b/src/concrete_convex_hull.jl
@@ -133,23 +133,16 @@ function _two_points_1d!(points)
     if p1 == p2
         # check for redundancy
         pop!(points)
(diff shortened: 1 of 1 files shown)


mforets:
Oh, seems like both sent the same PRs, sorry about that.. I should have made it clearer that i planned to send this PR. I pinged you since we've discussed this change previously.

PR #1474, opened 2019-06-25
#1462 - Fix minkowski sum for VPolygons

### LINKED ISSUE #1462: minkowski_sum for VPolygons depends on order of vertices
I just looked at the code from #1377 more closely and found that the implementation is unsound. 

https://github.com/JuliaReach/LazySets.jl/blob/3e9d8370bb295fe3a4007af77b5c9eb53e16c11c/src/VPolygon.jl#L582-L611

Here is a counterexample: `P2` is equivalent to `P` but with a different starting point in the vertices. The result is the same except for the last vertex.
```julia
julia> P = VPolygon([[4., 0], [6., 2], [4., 4]];)
julia> Q = VPolygon([[-2., -2], [2., 0], [2., 2], [-2., 4]]);
julia> PQ = minkowski_sum(P, Q)
VPolygon{Float64,Array{Float64,1}}(Array{Float64,1}[
[2.0, -2.0], [6.0, 0.0], [8.0, 2.0], [8.0, 4.0], [6.0, 6.0], [2.0, 8.0]])

julia> P2 = VPolygon([[4., 4], [4., 0], [6., 2]]);
julia> P2Q = minkowski_sum(P2, Q)
VPolygon{Float64,Array{Float64,1}}(Array{Float64,1}[
[2.0, -2.0], [6.0, 0.0], [8.0, 2.0], [8.0, 4.0], [6.0, 6.0], [2.0, 2.0]])
```
PQ

P2Q

The algorithm ([p. 295](https://people.inf.elte.hu/fekete/algoritmusok_msc/terinfo_geom/konyvek/Computational%20Geometry%20-%20Algorithms%20and%20Applications,%203rd%20Ed.pdf)) relies on the sorting of the vertices based on a common start, which the `VPolygon` type does not guarantee. Excerpt from the book:
> The lists of vertices are assumed to be in counter-clockwise order, with v1 and w1 being the vertices with smallest y-coordinate (and smallest x-coordinate in case of ties).

So in a preprocessing we need to identify the right value for `j` and `k`. (I suppose that it does not really matter that we choose the smallest y coordinate but rather that we align the two starting vertices, so it might be sufficient to adapt `j` to `k` or vice versa. But this might be over-engineering.)

SebastianGuadalupe:
Closes #1462.


src/VPolygon.jl
```diff
@@ -585,8 +585,8 @@ function minkowski_sum(P::VPolygon{N}, Q::VPolygon{N}) where {N<:Real}
     mP = length(vlistP)
     mQ = length(vlistQ)
     i = 1
-    k = 1
-    j = 1
+    k = findfirst(==(σ(N[1, 0], P)), vlistP)
+    j = findfirst(==(σ(N[1, 0], Q)), vlistQ)
```
schillic: Very neat :+1:
I was a bit surprised that `σ(N[1, 0], Q)` seems to be computed only once. Are we sure this is the case?

test/unit_Polygon.jl
```diff
@@ -177,17 +177,23 @@ for N in [Float64, Float32, Rational{Int}]
...
         Q = VPolygon(B)
         PQ = minkowski_sum(P, Q)
-        @test LazySets.ispermutation(PQ.vertices, [N[2, -2], N[6, 0], N[8, 2],
+        @test is_cyclic_permutation(PQ.vertices, [N[2, -2], N[6, 0], N[8, 2],
                                                 N[8, 4], N[6, 6], N[2, 8]])
+        
+        # test for different starting points in vertices_list of minkowski sum
+        P = VPolygon([N[4, 0], N[6, 2], N[4, 4]];)
+        P2 = VPolygon([N[4, 4], N[4, 0], N[6, 2]]);
+        Q = VPolygon([N[-2, -2], N[2, 0], N[2, 2], N[-2, 4]]);
```
schillic: Remove the `;`.

test/unit_Polygon.jl
```diff
@@ -177,17 +177,23 @@ for N in [Float64, Float32, Rational{Int}]
...
-        #test for corner case of parallel edges in minkowski sum
+        # test for corner case of parallel edges in minkowski sum
         C = [N[10, 5], N[10, 10], N[8, 10], N[5, 8], N[5, 5]]
         D = [N[-1, -2], N[1, -2], N[-1, 2]]
 
         R = VPolygon(C)
         S = VPolygon(D)
         RS = minkowski_sum(R, S)
-        @test LazySets.ispermutation(RS.vertices, [N[4, 3], N[11, 3], N[11, 8],
+        @test LazySets.is_cyclic_permutation(RS.vertices, [N[4, 3], N[11, 3], N[11, 8],
```
schillic:
```suggestion
        @test is_cyclic_permutation(RS.vertices, [N[4, 3], N[11, 3], N[11, 8],
```

src/VPolygon.jl
```diff
@@ -585,8 +585,10 @@ function minkowski_sum(P::VPolygon{N}, Q::VPolygon{N}) where {N<:Real}
     mP = length(vlistP)
     mQ = length(vlistQ)
     i = 1
-    k = 1
-    j = 1
+    σP = σ(N[1, 0], P)
+    σQ = σ(N[1, 0], Q)
+    k = findfirst(==(σP), vlistP)
```
mforets: :warning: Equality check doesn't play well with floating point. In this case it won't hurt since we get a vertex from `VPolygon`'s list of vertices [here](https://github.com/JuliaReach/LazySets.jl/blob/master/src/VPolygon.jl#L264) by asking the support vector, but this is very specific to polygons.

So i think a cleaner (and slightly more efficient) approach would be to get directly the corresponding index from `VPolygon`'s support vector function. It suffices to outsource the algorithm in https://github.com/JuliaReach/LazySets.jl/blob/master/src/VPolygon.jl#L256 to an independent helper function.

PR #1479, opened 2019-06-25
#1477 - Concrete minkowski sum for polytopes

### LINKED ISSUE #1477: Concrete minkowski sum of VPolytopes
```julia
julia> V1 = rand(VPolytope, dim=3);

julia> V2 = rand(VPolytope, dim=3);

julia> minkowski_sum(V1, V2)
ERROR: MethodError: no method matching minkowski_sum(::VPolytope{Float64}, ::VPolytope{Float64})
Stacktrace:
 [1] top-level scope at none:0
```

mforets:
Closes #1477.

Requires ~#1483 (for this application it is crucial to use the `CDDLib` backend)~ , ~#1484.~

### DIFF
--- a/docs/src/lib/binary_functions.md
+++ b/docs/src/lib/binary_functions.md
@@ -49,6 +49,12 @@ convex_hull(::VPolytope{N}, ::VPolytope{N}) where {N<:Real}
 convex_hull(::VPolygon{N}, ::VPolygon{N}) where {N<:Real}
 ```
 
+## Minkowski sum
+
+```@docs
+minkowski_sum(::AbstractPolytope{N}, ::AbstractPolytope{N}) where {N<:Real}
+```
+
 ## Intersection of two sets
 
 ```@docs
--- a/docs/src/lib/binary_functions.md
+++ b/docs/src/lib/binary_functions.md
@@ -49,12 +49,6 @@ convex_hull(::VPolytope{N}, ::VPolytope{N}) where {N<:Real}
 convex_hull(::VPolygon{N}, ::VPolygon{N}) where {N<:Real}
 ```
 
-## Minkowski sum
-
-```@docs
-minkowski_sum(::AbstractPolytope{N}, ::AbstractPolytope{N}) where {N<:Real}
-```
-
 ## Intersection of two sets
 
 ```@docs
--- a/src/AbstractPolytope.jl
+++ b/src/AbstractPolytope.jl
@@ -3,7 +3,8 @@ import Base.isempty
 export AbstractPolytope,
        vertices_list,
        singleton_list,
-       isempty
+       isempty,
+       minkowski_sum
 
 """
     AbstractPolytope{N<:Real} <: AbstractPolyhedron{N}
@@ -126,6 +127,39 @@ end
     else
         return HPolytope(constraints)
     end
+end
+
+"""
+    minkowski_sum(V1::AbstractPolytope{N}, V2::AbstractPolytope{N};
+                  [backend]=default_polyhedra_backend(V1, N)) where {N<:Real}
+
+Compute the Minkowski sum between two polytopes using their vertex representation.
+
+### Input
+
+- `V1`      -- polytope
+- `V2`      -- another polytope
+- `backend` -- (optional, default: `default_polyhedra_backend(V1, N)`) the
+               backend for polyhedral computations used to post-process with a
+               convex hull
+
+### Output
+
+A new polytope in vertex representation whose vertices are the convex hull of
+the sum of all possible sums of vertices of `V1` and `V2`.
+"""
+function minkowski_sum(V1::AbstractPolytope{N}, V2::AbstractPolytope{N};
+                       backend=default_polyhedra_backend(V1, N)) where {N<:Real}
+    vlist1, vlist2 = vertices_list(V1), vertices_list(V2)
+    n, m = length(vlist1), length(vlist2)
+    Vout = Vector{Vector{N}}()
+    sizehint!(Vout, n + m)
+    for vi in vlist1
+        for vj in vlist2
+            push!(Vout, vi + vj)
+        end
+    end
+    return VPolytope(convex_hull!(Vout, backend=backend))
 end
 
 # =============================================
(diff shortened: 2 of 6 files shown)


test/unit_Polytope.jl
```diff
@@ -350,6 +350,12 @@ if test_suite_polyhedra
         @test BallInf(N[0, 0], N(1)) ⊆ P
         @test !(BallInf(N[0, 0], N(1.01)) ⊆ P)
 
+        # concrete minkowski sum
+        B = convert(VPolytope, BallInf(N[0, 0, 0], N(1)))
+        X = minkowski_sum(B, B, backend=CDDLib.Library())
```
schillic:
```suggestion
        X = minkowski_sum(B, B)
```

test/unit_Polytope.jl
```diff
@@ -350,6 +350,12 @@ if test_suite_polyhedra
...
+        twoB = 2.0*B
+        @test X ⊆ twoB && twoB ⊆ X
```
schillic: What about comparing the vertices approximately instead?

mforets: This test passes here, i think that we can keep the double inclusion. On the other hand, without converting to `VPolytope` it fails; i reported it [here](https://github.com/JuliaReach/LazySets.jl/issues/1481).

test/unit_Polytope.jl
mforets: It doesn't work with the default library unfortunately (here the command runs forever). If you have another idea to substitute this one let's change it, although the BallInf is already quite "simple".

mforets:
> I would actually define the method for two AbstractPolytopes to extend the functionality of the library.

Makes sense, i'll move it to the abstract polytope interface.

docs/src/lib/binary_functions.md
```diff
@@ -49,6 +49,12 @@ convex_hull(::VPolytope{N}, ::VPolytope{N}) where {N<:Real}
 convex_hull(::VPolygon{N}, ::VPolygon{N}) where {N<:Real}
 ```
 
+## Minkowski sum
+
+```@docs
+minkowski_sum(::AbstractPolytope{N}, ::AbstractPolytope{N}) where {N<:Real}
```
schillic: Hm, I do not like to have two places in the manual for this. Let us put it in `interfaces.md` only and then move all `minkowski_sum`s to this file in another PR.

src/AbstractPolytope.jl
```diff
@@ -128,6 +129,39 @@ end
     end
 end
 
+"""
+    minkowski_sum(V1::AbstractPolytope{N}, V2::AbstractPolytope{N};
```
schillic: Rename to `P1` and `P2`?

src/AbstractPolytope.jl
```diff
@@ -128,6 +129,39 @@ end
...
+    vlist1, vlist2 = vertices_list(V1), vertices_list(V2)
+    n, m = length(vlist1), length(vlist2)
+    Vout = Vector{Vector{N}}()
+    sizehint!(Vout, n + m)
+    for vi in vlist1
+        for vj in vlist2
+            push!(Vout, vi + vj)
+        end
+    end
+    return VPolytope(convex_hull!(Vout, backend=backend))
```
schillic: Maybe add an option for using the `convex_hull!`.

test/unit_Polytope.jl
schillic: Can you add another test where the vertices of the two operands are not identical (to catch more bugs)?
```julia
P1 = VPolytope([N[0, 0], N[0, 1]]
P2 = VPolytope([N[0, 0], N[1, 0]]
Q = minkowski_sum(P1, P2)
@test ispermutation(vertices_list(Q), [N[0, 0], N[0, 1], N[1, 0], N[1, 1]])
```

src/AbstractPolytope.jl
```diff
@@ -130,27 +130,31 @@ end
...
 A new polytope in vertex representation whose vertices are the convex hull of
-the sum of all possible sums of vertices of `V1` and `V2`.
+the sum of all possible sums of vertices of `P1` and `P2`.
 """
-function minkowski_sum(V1::AbstractPolytope{N}, V2::AbstractPolytope{N};
-                       backend=default_polyhedra_backend(V1, N)) where {N<:Real}
-    vlist1, vlist2 = vertices_list(V1), vertices_list(V2)
+function minkowski_sum(P1::AbstractPolytope{N}, P2::AbstractPolytope{N};
+                       apply_convex_hull::Bool=true,
+                       backend=default_polyhedra_backend(P1, N)) where {N<:Real}
```
schillic: To not require `Polyhedra` if `apply_convex_hull=false`, you need to set the backend to `nothing` and reassign in the body:
```julia
...
if apply_convex_hull
    if backend == nothing
        backend=default_polyhedra_backend(P1, N)
    end
    convex_hull!(Vout, backend=backend)
end
```

src/AbstractPolytope.jl
```diff
@@ -128,6 +129,46 @@ end
...
+
+### Output
+
+A new polytope in vertex representation whose vertices are the convex hull of
+the sum of all possible sums of vertices of `P1` and `P2`.
+"""
+function minkowski_sum(P1::AbstractPolytope{N}, P2::AbstractPolytope{N};
+                       apply_convex_hull::Bool=true,
+                       backend=default_polyhedra_backend(P1, N)) where {N<:Real}
+    vlist1, vlist2 = vertices_list(P1), vertices_list(P2)
```
schillic: Maybe `@assert dim(P1) == dim(P2)`?

src/AbstractPolytope.jl
```diff
@@ -128,6 +129,46 @@ end
...
+    n, m = length(vlist1), length(vlist2)
+    Vout = Vector{Vector{N}}()
```
schillic: Note that the code does not work if `n == 0` or `m == 0`.

mforets: With the assertion, say if `n = 0` and `m > 0`, now it gives an error that the dimensions do not match.

PR #1484, opened 2019-06-27
#1439 - Approximate membership test for vectors and permutations

### LINKED ISSUE #1439: Approximate version of ispermutation
Wanted: `ispermutation` but not with `==` but with `≈`

mforets:
Closes #1439.


src/Arrays/vector_operations.jl
```diff
@@ -84,6 +83,19 @@ function ispermutation(u::AbstractVector{T}, v::AbstractVector)::Bool where {T}
...
+    return in(x, itr)
+end
+
+function _in(x::AbstractVector{T}, itr) where {T<:AbstractFloat}
+    for y in itr
+        if LazySets._isapprox(x, y)
+            return true
+        end
+    end
+    return false
```
schillic: Does something like `return all(y -> LazySets._isapprox(x, y), itr)` not work?

src/comparisons.jl
```diff
@@ -142,6 +142,30 @@ function _isapprox(x::N, y::N;
     end
 end
 
+# generic "dense"
+function _isapprox(x::AbstractVector{N}, y::AbstractVector{N};
+                   rtol::Real=Base.rtoldefault(N),
+                   ztol::Real=ABSZTOL(N),
+                   atol::Real=zero(N)) where {N<:Real}
+    n = length(x)
+    @assert n == length(y)
```
schillic: I would rather just `return false` if the vectors have different lengths.

src/Arrays/vector_operations.jl
mforets: You probably mean `any(..)`. Yes, it is a good idea to make that change.

schillic: I actually meant `all` with the negated condition, but that was still wrong here :smile:

PR #1490, opened 2019-07-02
#1488 - Pass solver option to prune polytopes

### LINKED ISSUE #1488: Pass solver to removevredundancy of polytopes
`Polyhedra.removevredundancy` has improved, but we are not using it correctly. We should pass an (LP) solver.

mforets:
Closes #1488.


Project.toml
```diff
@@ -33,6 +37,7 @@ TaylorModels = "≥ 0.1.1"
 CDDLib = "3391f64e-dcde-5f30-b752-e11513730f60"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Expokit = "a1e7a1ef-7a5d-5822-a38c-be74e1bb89f4"
+GLPK = "60bf3e95-4087-53dc-ae20-288a0d20c6a6"
```
schillic:
```suggestion
```

src/AbstractPolytope.jl
```diff
@@ -198,5 +205,12 @@ end
 function default_polyhedra_backend(P, N::Type{<:Rational})
     return default_library(LazySets.dim(P), Rational{Int})
 end
+
+import JuMP, GLPK
```
schillic: I would add these `import`s to the top of the `load_polyhedra_abstractpolytope` block.

src/VPolytope.jl
```diff
@@ -385,16 +386,25 @@ Return the polytope obtained by removing the redundant vertices of the given pol
...
+                                   solver=default_lp_solver(N))::VPolytope{N} where {N<:Real}
     require(:Polyhedra; fun_name="remove_redundant_vertices")
+
     Q = polyhedron(P; backend=backend)
-    removevredundancy!(Q)
+    if Polyhedra.supportssolver(typeof(Q))
+        vQ = Polyhedra.vrep(Q)
+        Polyhedra.setvrep!(Q, Polyhedra.removevredundancy(vQ, solver))
+    else
+        Polyhedra.removevredundancy!(Q)
```
schillic: I do not understand what is going on here. Can you add some documentation?

src/VPolytope.jl
```diff
@@ -385,16 +386,25 @@ Return the polytope obtained by removing the redundant vertices of the given pol
...
-                                   backend=default_polyhedra_backend(P, N))::VPolytope{N} where {N<:Real}
+                                   backend=default_polyhedra_backend(P, N),
+                                   solver=default_lp_solver(N))::VPolytope{N} where {N<:Real}
     require(:Polyhedra; fun_name="remove_redundant_vertices")
+
     Q = polyhedron(P; backend=backend)
-    removevredundancy!(Q)
+    if Polyhedra.supportssolver(typeof(Q))
+        vQ = Polyhedra.vrep(Q)
+        Polyhedra.setvrep!(Q, Polyhedra.removevredundancy(vQ, solver))
```
schillic: Does this accept `solver == nothing`? Would it not make more sense to catch this case and use the default LP solver?

src/concrete_convex_hull.jl
```diff
@@ -304,13 +309,15 @@ function _convex_hull_1d!(points::Vector{VN})::Vector{VN} where {N<:Real, VN<:Ab
...
 
 function _convex_hull_nd!(points::Vector{VN};
-                          backend=nothing
+                          backend=nothing,
+                          solver=nothing
                           )::Vector{VN} where {N<:Real, VN<:AbstractVector{N}}
     V = VPolytope(points)
     if backend == nothing
         backend = default_polyhedra_backend(V, N)
+        solver = default_lp_solver(N)
```
schillic: `backend == nothing` does not imply `solver == nothing`. Maybe just keep it as it is and resolve the case `solver == nothing` in the other place (see my previous comment).

src/VPolytope.jl
mforets: The `solver` option is mandatory in ` Polyhedra.removevredundancy(vQ, solver)`; the function that i intend to use here is [this one](https://github.com/JuliaPolyhedra/Polyhedra.jl/blob/master/src/redundancy.jl#L211). The argument in `remove_redundant_vertices` is a VPolytope and the logic is to either:

- attach the GLPK solver (by default, but any other LP solver supported by JuMP is ok)  if `Q` supports it and call the 2 argument version of `removevredundancy`

or

- do not attach any solver and use the solver's `removevredundancy!` function (line 406 below)

src/VPolytope.jl
mforets: I have added the description in the docstring.

src/concrete_convex_hull.jl
mforets: That's correct, the user could have passed a solver but not a backend. Defaults for both options (`backend` and `solver`) are handled by `remove_redundant_vertices` so i opt for that case and remove the check in `_convex_hull_nd!` altogether.

src/VPolytope.jl
schillic: This does not answer my question.

src/concrete_convex_hull.jl
schillic: > that is not possible.

Why not? And why set the `solver` only if `backend == nothing`?

schillic:
Apparently you need to add `JuMP` to `extras` as well. Then my intuition seems to be right: add all packages there that you need to load during testing and that have not been loaded with the main package already.

src/VPolytope.jl
mforets: > Does this accept solver == nothing?

What is "this"? The particular invocation `Polyhedra.removevredundancy`  in line 404 does not accept `nothing` as solver, that explains why i made the check in line 402.

> Would it not make more sense to catch this case and use the default LP solver?

Which case, if `solver` is nothing? It is not always problematic (for example if the backend is CDDLib, `solver` is not needed).

src/concrete_convex_hull.jl
mforets: > Why not?

It is not possible without changing `remove_redundant_vertices`, sure. 

> And why set the solver only if backend == nothing?

It is a choice if the `backend` is not set then one needs a solver because we are using the default backend, but the converse is not always true (for example if the backend was chosen to be `CDDLib`).

src/VPolytope.jl
schillic: > that explains why i made the check in line 402

Line 402 does not look at `solver`, so it cannot fix the problem.

> Which case, if solver is nothing?

Yes.

> It is not always problematic

This somewhat implies that it is problematic *sometimes*. We should solve these cases.

src/concrete_convex_hull.jl
schillic: Sounds like we should pass `backend` to `default_lp_solver` then.

src/VPolytope.jl
mforets: The line number changed after some commit, i meant the check `if Polyhedra.supportssolver(typeof(Q))`.

src/concrete_convex_hull.jl
schillic: Generally we should not set the default backend/solver in different functions (unless there is a good reason why some functions use a different default than others). There should be a central place where we do that, and only immediately before we actually use it. This makes it easier to avoid inconsistencies and make changes in the future.

src/VPolytope.jl
schillic: I know, but there is no `solver` in that line!

mforets: > This somewhat implies that it is problematic sometimes. We should solve these cases

This function `remove_redundant_vertices` is an attempt to use `Polyhedra.removevredundancy` but with a `VPolytope` input, which is of course not understood by Polyhedra, so one has to make some decision regarding conversion to a type that is understood by it.

What we've done in `AbstractPolytope.jl` is to create a default polyhedron without solver, and pair to one solve later in this function (but this is just a choice, as solver can be stored in the type [here](https://github.com/JuliaPolyhedra/Polyhedra.jl/blob/master/src/defaultlibrary.jl#L9)). I think it is clear to do it this way because for certain applications in LazySets, the `solver` is not needed.

Issue #1078, opened 2019-01-27
Generalize methods to AbstractPolyhedron


schillic:
Follow-up of #780/#1077:

* [x] generalize `linear_map` to `AbstractPolyhedron` (#1093)
* [x] generalize `polyhedron` to `AbstractPolyhedron`
  This is also needed for the following refactorings:
  * [ ] generalize `isempty` to `AbstractPolyhedron`
  There are two optional implementations. The second one would just work. For the first one, we would need the `polyhedron` method above.
  * [ ] generalize `isempty` to list of constraints
  This allows to finish [this comment](https://github.com/JuliaReach/LazySets.jl/issues/780#issuecomment-456308999) by removing the `HPolyhedron` wrapper in [this line](https://github.com/JuliaReach/LazySets.jl/blob/master/src/is_intersection_empty.jl#L870).
* [x] replace `Union{AbstractPolytope, HPolyhedron}` by `AbstractPolyhedron` (#1501)
* [x] check which of the `AbstractPolytope` methods can be generalized
* [x] move the `Polyhedra` import code from `AbstractPolytope.jl` (#1501)

PR #1501, opened 2019-07-04
#1078 - Generalize methods to AbstractPolyhedron

schillic:
See #1078.

I also moved the `Polyhedra` imports to `init.jl` and removed some of them to make it explicit in the code where we use `Polyhedra`.

---

- fixes bug in intersection of universal set with halfspace/hyperplane


schillic:
> i would prefer to keep the Polyhedra initialization in its own file

I do not see the reason for that, but I outsourced the code to new files.

> let's fix these bugs

Thanks, I fixed them and added tests.

> i think that we can remove function is_intersection_empty(U::Universe{N}, hp::Union{Hyperplane{N}, Line{N}}, altogether, and the same for those that dispatch on HalfSpace

No, they are needed for disambiguation.

Issue #1485, opened 2019-06-29
Add test for changing rtol


mforets:
It remains to add a test for changing the tolerance.

mforets:
`Tolernace` has been added in #1510. We should add a test and review if the docs are good or can be improved.

PR #1510, opened 2019-07-20
Tolerance setting

mforets:
See #1485.

This PR is an idea for optional tolerance setting such that it extends to all functions that rely on `comparisons.jl`. It adds getter and setter functions for `Float64` since it's the type that we use in applications, and `Rational{Int}`. For other numeric types, the behavior is left unchanged (one cannot set defaults).

### DIFF
--- a/src/comparisons.jl
+++ b/src/comparisons.jl
@@ -2,7 +2,49 @@
 #
 export set_rtol, set_ztol, set_atol
 
-# struct to contain the tolerances for a given numberic type
+"""
+    Tolerance{N<:Number}
+
+Type that represents the tolerances for a given numeric type.
+
+### Fields
+
+- `rtol` -- relative tolerance
+- `ztol` -- zero tolerance or absolute tolerance for comparison against zero
+- `atol` -- absolute tolerance
+
+### Notes
+
+The type `Tolerance`, parametric in the numeric type `N`, is used to store default
+values for numeric comparisons. It is mutable and setting the value of a field
+affects the getter functions hence it can be used to fix the tolerance globally
+in `LazySets`.
+
+Default values are defined for the most commonly used numeric types, and for those
+cases when other numeric types are needed one can extend the default values
+as explained next.
+
+The cases `Float64` and `Rational` are special in the sense that they are the most
+commonly used types in applications. Getting and setting default tolerances
+is achieved with the functions `_rtol` and `set_rtol` (and similarly for the other
+tolerances); the implementation creates an instance of `Tolerance{Float64}`
+(resp. `Tolerance{Rational}`) and sets some default values. Again since `Tolerance`
+is mutable, setting a value is possible e.g. `set_rtol(Type{Float64}, ε)` for some
+floating-point `ε`.
+
+For all other cases, a dictionary mapping numeric types to instances of `Tolerance`
+for that numeric type is used. For floating-point types, a default value has been
+defined through `default_tolerance` as follows:
+
+```julia
+default_tolerance(N::Type{<:AbstractFloat}) = Tolerance(Base.rtoldefault(N), sqrt(eps(N)), zero(N))
+```
+Hence to set a single tolerance (either `rtol`, `ztol` or `atol`) for a given
+floating-point type, use the corresponding `set_rtol` function, while the values
+which have not been set will be pulled from `default_tolerance`. If you would like
+to define the three default values at once, or are computing with a non floating-point
+numeric type, you can just extend `default_tolerance(N::Type{<:Number})`.
+"""
 mutable struct Tolerance{N<:Number}
     rtol::N
     ztol::N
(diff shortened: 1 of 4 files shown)


src/comparisons.jl
```diff
@@ -1,7 +1,42 @@
...
+# struct to contain the tolerances for a given numberic type
+mutable struct TOL{N<:Number}
+    rtol::N
+    ztol::N
+    atol::N
+end
+
+# global Float64 tolerances
+const F64 = Float64
+const tF64 = Type{Float64}
```
schillic: I would not introduce these constants for just a few lines. They make reading the code harder.

src/comparisons.jl
```diff
@@ -1,7 +1,42 @@
-# This file is inspired from Polyhedra.jl
+# Some functions in this file are inspired from Polyhedra.jl
 
-const ABSZTOL(N::Type{<:AbstractFloat}) = N(10) * sqrt(eps(N))
-const ABSZTOL(N::Type{Rational{INNER}}) where {INNER} = zero(N)
+# struct to contain the tolerances for a given numberic type
+mutable struct TOL{N<:Number}
```
schillic: Let us not use capital letters for types.
```suggestion
mutable struct Tolerance{N<:Number}
```

src/comparisons.jl
```diff
@@ -1,7 +1,42 @@
...
+end
+
+# global Float64 tolerances
+const F64 = Float64
+const tF64 = Type{Float64}
+const _TOL_F64 = TOL(Base.rtoldefault(F64), F64(10)*sqrt(eps(F64)), zero(F64))
+
+get_rtol(N::tF64) = _TOL_F64.rtol
+get_ztol(N::tF64) = _TOL_F64.ztol
+get_atol(N::tF64) = _TOL_F64.atol
```
schillic: Instead of having a few hardcoded collections like `_TOL_F64`, I suggest to define a dictionary `TOL_N`:
```julia
TOL_N = Dict{Type{<:Number}, Tolerance}()

rtol(N::Type{<:Number}) = get!(TOL_N, N, default_tolerance(N)).rtol
...
xx(N::Type{<:Number}) = get!(TOL_N, N, default_tolerance(N)).xx

set_rtol(N::Type{<:Number}, val) = get!(TOL_N, N, default_tolerance(N))[rtol] = val
...
set_xx(N::Type{<:Number}, val) = get!(TOL_N, N, default_tolerance(N))[xx] = val

default_tolerance(N::Type{Float64}) = Tolerance(..., ..., ...)
...
```

src/comparisons.jl
```diff
@@ -1,7 +1,42 @@
...
+
+set_rtol(N::tF64, ε::F64) = _TOL_F64.rtol = ε
+set_ztol(N::tF64, ε::F64) = _TOL_F64.ztol = ε
+set_atol(N::tF64, ε::F64) = _TOL_F64.atol = ε
+
+# global rational tolerances
+const RAT = Rational{Int}
```
schillic: Let us use `Rational` instead so it automatically works for other `Rational` instantiations as well.

src/comparisons.jl
mforets: I disagree, since these constants are repeated several times, i added them to make the code look simpler. But i won't insist, it would be genuine bikeshedding 😆

src/comparisons.jl
mforets: Good, i changed it.

src/comparisons.jl
mforets: Fixed; i used `Type{<:Rational}` where possible.

schillic: Note that `typeof(zero(Rational)) == Rational{Int}`.

src/comparisons.jl
schillic: Note that your approach only allows to set the bounds for `Float64` and `Rational{Int}`. I find this very restrictive.

mforets: The idea of a dictionary is appealing, since it allows users to define a new global set of default tolerances without having to manually extend `comparisons.jl` inside LazySets for this purpose (*).

I'm not opposed to using a dictionary whose keys are the numeric types and whose values are instance of `Tolerance`. But i'd like to understand: is there any difference performance-wise between the two approaches? The dictionary cannot be strictly typed (because one cannot know at compile time which numeric types will be added).

(*) of course, `rtol`, `atol` and `ztol` can still be passed as keyword arguments, the feature added in this PR is to set them globally for the numeric type that we use in practice.

schillic: One can also combine the approaches: use a fixed object for `Float64` and use the dictionary as fallback for all other types.

src/comparisons.jl
```diff
@@ -1,7 +1,40 @@
-# This file is inspired from Polyhedra.jl
+# Some functions in this file are inspired from Polyhedra.jl
+#
+export set_rtol
```
schillic: Any reason why `set_ztol` and `set_atol` are not exported?

mforets: No; let's export those as well 👍

PR #1514, opened 2019-07-22
#1379 - Revise (is_)right_turn

### LINKED ISSUE #1379: FIx documentation of right_turn (internal function)
The docs suggest that the function `right_turn` determines whether an angle is a right turn.

https://github.com/JuliaReach/LazySets.jl/blob/923cf4ea5f893bde598cc152b3c0d9d1a95688dd/src/concrete_convex_hull.jl#L169-L197

However, this is the job of `is_right_turn`.

https://github.com/JuliaReach/LazySets.jl/blob/923cf4ea5f893bde598cc152b3c0d9d1a95688dd/src/AbstractPolygon.jl#L116-L135

While we are at it, I suggest to move both functions to `helper_functions.jl`.

schillic:
Closes #1379.

We had `right_turn(O, u, v)` and `is_right_turn(u, v)`. Here I added the two missing counterparts `right_turn(u, v)` and `is_right_turn(O, u, v)`.

I shortly considered to rename `right_turn` → `Base.cross`, but then decided against it.

### DIFF
--- a/src/concrete_convex_hull.jl
+++ b/src/concrete_convex_hull.jl
@@ -229,20 +229,17 @@ function _four_points_2d!(points::AbstractVector{<:AbstractVector{N}}) where {N<
     if tri_CAD > zero(N)
         key = key + 1
     end
-    
+
     if isapproxzero(tri_ABC)
         return _collinear_case!(points, A, B, C, D)
-    end
-    if isapproxzero(tri_ABD)
+    elseif isapproxzero(tri_ABD)
         return _collinear_case!(points, A, B, D, C)
-    end
-    if isapproxzero(tri_BCD)
+    elseif isapproxzero(tri_BCD)
         return _collinear_case!(points, B, C, D, A)
-    end
-    if isapproxzero(tri_CAD)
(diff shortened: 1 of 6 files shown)


src/concrete_convex_hull.jl
```diff
@@ -398,11 +365,11 @@ For further details see
 function monotone_chain!(points::Vector{VN}; sort::Bool=true
                         )::Vector{VN} where {N<:Real, VN<:AbstractVector{N}}
 
-    @inline function build_hull!(semihull, iterator, points, zero_N)
+    @inline function build_hull!(semihull, iterator, points)
         @inbounds for i in iterator
             while length(semihull) >= 2 &&
                     (right_turn(semihull[end-1], semihull[end], points[i])
-                         <= zero_N)
+                         <= zero(N))
```
mforets: i wonder if the `=` is needed or we can just we `!is_right_turn(...)`

schillic: Mathematically the `=` is needed. In practice, with floating-point numbers it might not be needed, but if we have rationals, it is needed again.

PR #1518, opened 2019-07-23
#786 - Revise tests of polyhedra

### LINKED ISSUE #786: Replace empty polytopes in unit_Polytope
Some of the polytopes are empty, which invalidates all the checks.

schillic:
Closes #786.

### DIFF
--- a/test/unit_Polyhedron.jl
+++ b/test/unit_Polyhedron.jl
@@ -59,14 +59,14 @@ for N in [Float64, Rational{Int}, Float32]
     # constrained dimensions
     @test constrained_dimensions(p) == [1, 2]
     @test constrained_dimensions(
-        HPolyhedron{N}([LinearConstraint(N[1, 0], N(1))])) == [1]
+        HPolyhedron([LinearConstraint(N[1, 0], N(1))])) == [1]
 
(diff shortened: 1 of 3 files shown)


test/unit_Polyhedron.jl
```diff
@@ -112,27 +112,26 @@ if test_suite_polyhedra
...
 
         # intersection with polytope
         A = [N(1) N(0); N(0) N(1); N(-1) N(-1)]
         b = N[1, 1, -1]
         p = HPolyhedron(A, b)
         b = BallInf(N[1, 1], N(0.5))
-        cap = intersection(p, b)
-        cap = intersection(b, p)
+        intersection(p, b)
+        intersection(b, p)
```
mforets: the intention was to test something here or just to check that `intersection` can dispatch on `HPolyhedron` and `BallInf`?

schillic: I do not know. I just removed the return value because we never used it :wink:

mforets: 😆  we could check that the obtained sets are the same, for instance

schillic: Okay, see 1531.

PR #1520, opened 2019-07-23
#811 - Check low .< high in hyperrectangle constructor

### LINKED ISSUE #811: Check low .< high in hyperrectangle constructor
(sometimes i fall into this trap creating random values for low and high.. if we add this check, it should be possible deactivate it).

schillic:
Closes #811.

### DIFF
--- a/src/Hyperrectangle.jl
+++ b/src/Hyperrectangle.jl
@@ -59,7 +59,7 @@ Hyperrectangle(center::Vector{N}, radius::Vector{N}) where {N<:Real} =
 function Hyperrectangle(;
                         high::AbstractVector{N},
                         low::AbstractVector{N}) where {N<:Real}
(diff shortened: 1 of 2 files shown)


src/Hyperrectangle.jl
```diff
@@ -59,6 +59,7 @@ Hyperrectangle(center::Vector{N}, radius::Vector{N}) where {N<:Real} =
 function Hyperrectangle(;
                         high::AbstractVector{N},
                         low::AbstractVector{N}) where {N<:Real}
+    @assert all(low .<= high) "lower bound must be lower than upper bound"
```
mforets: more efficient check:
```suggestion
    @assert all(i -> low[i] <= high[i], eachindex(low)) "lower bound must be lower than upper bound"
```

PR #1525, opened 2019-07-23
#1116 - SymmetricIntervalHull of unbounded set

### LINKED ISSUE #1116: SymmetricIntervalHull of unbounded set
```julia
julia> S = SymmetricIntervalHull(HPolyhedron([HalfSpace([1., 1.], 1.)]));

julia> ρ([1., 0.], S)
Inf

julia> isbounded(S)
true

julia> vertices_list(S)
 [Inf, Inf]  
 [-Inf, Inf] 
 [Inf, -Inf] 
 [-Inf, -Inf]
```
The problem is that `SymmetricIntervalHull` is an `AbstractHyperrectangle`, but this is not true if the wrapped set is unbounded.
I see the following ways to fix this:
* Let `SymmetricIntervalHull` not inherit from this interface anymore. This implies that we lose many convenient implementations.
* (my favorite) Clarify in the documentation that the results are undefined if the wrapped set is unbounded.

schillic:
Closes #1116.

### DIFF
--- a/src/SymmetricIntervalHull.jl
+++ b/src/SymmetricIntervalHull.jl
@@ -4,11 +4,11 @@ export SymmetricIntervalHull,
 """
     SymmetricIntervalHull{N<:Real, S<:LazySet{N}} <: AbstractHyperrectangle{N}
 
-Type that represents the symmetric interval hull of a convex set.
+Type that represents the symmetric interval hull of a compact convex set.
 
 ### Fields
 
-- `X`     -- convex set
+- `X`     -- compact convex set
 - `cache` -- partial storage of already computed bounds, organized as mapping
     from dimension to tuples `(bound, valid)`, where `valid` is a flag
     indicating if the `bound` entry has been computed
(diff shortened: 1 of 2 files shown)


src/SymmetricIntervalHull.jl
```diff
@@ -25,13 +25,17 @@ However, if one asks for many support vectors in a loop, the number of
...
+
+The set `X` must be compact.
 """
 struct SymmetricIntervalHull{N<:Real, S<:LazySet{N}} <: AbstractHyperrectangle{N}
     X::S
     cache::Vector{N}
 
     # default constructor that initializes cache
     function SymmetricIntervalHull{N, S}(X::S) where {N<:Real, S<:LazySet{N}}
+        @assert isbounded(X) "the symmetric interval hull is only defined for" *
```
mforets:
```suggestion
        @assert isbounded(X) "the symmetric interval hull is only defined for " *
```

PR #1530, opened 2019-07-24
#1233 - Efficient support vector for VPolytope

### LINKED ISSUE #1233: VPolytope's support vector goes to hrep
The implementation of the support vector for `VPolytope` is this:
https://github.com/JuliaReach/LazySets.jl/blob/9c30b2407363cb73b7ba45b1d1526519af4f57a7/src/VPolytope.jl#L93-L103

If this is called in a loop, e.g., from an overapproximation, this repeated `tohrep` computation is very costly. It would be good to let `overapproximate` first call `tohrep`.

We should add a new `algorithm` option that just computes the support function for each vertex. This has the optimal complexity for a single query.

schillic:
Closes #1233.

We might even want to drop the other algorithm because it is always inferior.

```julia
using LazySets, BenchmarkTools, Polyhedra

function benchmark(n)
    P = rand(VPolytope, dim=n, num_vertices=10)
    d = rand(n)
    @btime σ($d, $P, algorithm=$"hrep")
    @btime σ($d, $P, algorithm=$"sfun")
end

for n in 1:8
    benchmark(n)
end
```
```julia
  39.861 μs (187 allocations: 8.53 KiB)
  165.692 ns (0 allocations: 0 bytes)
  1.182 ms (13276 allocations: 828.88 KiB)
  168.392 ns (0 allocations: 0 bytes)
  4.208 ms (71792 allocations: 5.22 MiB)
  179.587 ns (0 allocations: 0 bytes)
  14.202 ms (248627 allocations: 21.41 MiB)
  194.296 ns (0 allocations: 0 bytes)
  42.561 ms (659218 allocations: 57.89 MiB)
  202.488 ns (0 allocations: 0 bytes)
  22.158 ms (346066 allocations: 32.52 MiB)
  208.631 ns (0 allocations: 0 bytes)
  28.534 ms (438832 allocations: 40.59 MiB)
  217.820 ns (0 allocations: 0 bytes)
  34.010 ms (490169 allocations: 47.85 MiB)
  224.552 ns (0 allocations: 0 bytes) 
```
The new implementation scales very well with the dimension (165.692 ns .. 224.552 ns) while the old implementation is heavily influenced by that (39.861 μs .. 34.010 ms; note the units).
Speedup in dim 8: x150,000 :tada:

### DIFF
--- a/src/VPolytope.jl
+++ b/src/VPolytope.jl
@@ -75,34 +75,27 @@ function dim(P::VPolytope)::Int
 end
 
 """
-    σ(d::AbstractVector{N}, P::VPolytope{N}; algorithm="sfun") where {N<:Real}
+    σ(d::AbstractVector{N}, P::VPolytope{N}) where {N<:Real}
 
-Return the support vector of a polyhedron (in V-representation) in a given
+Return the support vector of a polytope in V-representation in a given
 direction.
 
 ### Input
 
-- `d`         -- direction
-- `P`         -- polyhedron in V-representation
-- `algorithm` -- (optional, default: `"sfun"`) method to compute the support
-                 vector; available options:
-    * `"sfun"` - query the support function for every vertex and return the
-                 maximizing vertex
-    * `"hrep"` - convert to constraint representation
+- `d` -- direction
+- `P` -- polytope in V-representation
 
 ### Output
 
 The support vector in the given direction.
 
-### Notes
+### Algorithm
 
 A support vector maximizes the support function.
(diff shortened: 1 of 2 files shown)


mforets:
> We might even want to drop the other algorithm because it is always inferior.

Agreed about dropping the other algorith -- for the reason evoked in https://github.com/JuliaReach/LazySets.jl/issues/1233#issue-421947806 and that `σ(d, tohrep(P))` is a conversion that seems ok to be left to the user

PR #1532, opened 2019-07-28
#1511 - Bump Documenter version

### LINKED ISSUE #1511: Bump Documenter version
The `DocTestSetup` in `@meta` blocks should now be defined in `make.jl`. See https://github.com/JuliaDocs/Documenter.jl/blob/v0.23.0/CHANGELOG.md#version-v0230

schillic:
Closes #1511.

I had to downgrade one doctest in `docs/src/man/convex_hulls.md` to a normal REPL output because I could not figure out how to make it work with doctests. Using an `@example` block worked in the doctests but failed when building the docs.

I did not understand why `subtypes` was not found (in the REPL it works), so I included it via a `setup` keyword argument.


Project.toml
```diff
@@ -21,7 +21,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 julia = "≥ 1.0.0"
 CDDLib = "≥ 0.5.2"
 Distributions = "≥ 0.20.0"
-Documenter = "^0.22.6"
+Documenter = "≥ 0.23.0"
```
mforets: Should we use `Documenter = "^ 0.23.0" ? To make sure that LazySets builds do not start failing suddenly due to a new major release of Documenter, which may include breaking changes. Related: https://juliadocs.github.io/Documenter.jl/dev/man/hosting/#.travis.yml-Configuration-1

schillic: We do not do that for other packages as well, so I tend not to. It is usually easy to detect such cases.

mforets: Ok.

PR #1538, opened 2019-07-30
#1536 - Function to compute the Hausdorff distance between two sets

### LINKED ISSUE #1536: Function to compute the Hausdorff distance between two sets
At least in the infinity norm this is doable.

schillic:
Closes #1536.

Requires:
* ~The inclusion check fails because we do not have `constraints_list(::MinkowskiSum)`. We can start with #1563.~
* ~bugfix for `minkowski_sum` (#1592)~

### DIFF
--- a/src/Approximations/hausdorff_distance.jl
+++ b/src/Approximations/hausdorff_distance.jl
@@ -87,7 +87,7 @@ function hausdorff_distance_two_sided(X, Y, δ, n, p)
         hausdorff_distance_one_sided(Y, X, δ, n, p)
 end
 
-function hausdorff_distance_one_sided(X, Y, δ, n, p)
+function hausdorff_distance_one_sided(X::LazySet{N}, Y, δ, n, p) where {N}
     return X ⊆ Y + Ballp(p, zeros(N, n), δ)
 end
 
--- a/src/Approximations/hausdorff_distance.jl
+++ b/src/Approximations/hausdorff_distance.jl
@@ -53,6 +53,7 @@ we use a lazy `MinkowskiSum` with a `Ballp` centered in the origin.
 """
 function hausdorff_distance(X::LazySet{N}, Y::LazySet{N}; p::N=N(Inf),
                             ε::N=N(1e-3)) where {N<:Real}
+    @assert ε > zero(N) "the value ε must be positive"
     @assert isbounded(X) && isbounded(Y) "the Hausdorff distance is only " *
         "defined for compact sets"
 
--- a/src/Approximations/hausdorff_distance.jl
+++ b/src/Approximations/hausdorff_distance.jl
@@ -1,0 +1,102 @@
+export hausdorff_distance
+
+"""
+    hausdorff_distance(X::LazySet{N}, Y::LazySet{N}; [p]::N=N(Inf),
+                       [ε]::N=N(1e-3)) where {N<:Real}
+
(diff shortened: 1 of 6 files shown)


src/Approximations/hausdorff_distance.jl
```diff
@@ -0,0 +1,102 @@
...
+    end
+    return δ_upper
+end
+
+function hausdorff_distance_two_sided(X, Y, δ, n, p)
+    return hausdorff_distance_one_sided(X, Y, δ, n, p) &&
+        hausdorff_distance_one_sided(Y, X, δ, n, p)
+end
+
+function hausdorff_distance_one_sided(X, Y, δ, n, p)
```
mforets: `N` is missing from the function's signature (either pass it as a new variable or read it from the type of `X` and `Y` ?)

src/Approximations/hausdorff_distance.jl
```diff
@@ -0,0 +1,102 @@
...
+        if hausdorff_distance_two_sided(X, Y, δ, n, p)
+            δ_upper = δ
+        else
+            δ_lower = δ
+        end
+    end
+    return δ_upper
+end
+
+function hausdorff_distance_two_sided(X, Y, δ, n, p)
```
mforets: the function's name does not reflect that the answer is a boolean. shall we change this to `_issubset_within_δ` or so?

src/Approximations/hausdorff_distance.jl
```diff
@@ -0,0 +1,103 @@
...
+To find an upper bound ``u``, we start with the heuristics of taking the biggest
+distance in the axis-parallel directions.
+As long as this bound does not work, we increase the bound by ``2``.
+
+Given a value ``δ``, to check whether the sets are within Hausdorff distance
+``δ``, we simply check the inclusions given above, where on the right-hand side
+we use a lazy `MinkowskiSum` with a `Ballp` centered in the origin.
+"""
+function hausdorff_distance(X::LazySet{N}, Y::LazySet{N}; p::N=N(Inf),
+                            ε::N=N(1e-3)) where {N<:Real}
```
mforets: let's add an `MAX_ITER` termination thresold (in case the user lets epsilon = 0)

src/Approximations/hausdorff_distance.jl
```diff
@@ -0,0 +1,103 @@
...
+                  N(1e-3))  # this initial bound should be strictly positive
+    # verify that this is an upper bound
+    while !_mutual_issubset_in_δ_bloating(X, Y, δ_upper, n, p)
+        δ_upper *= N(2)
+    end
+
+    # phase 2: perform binary search between lower bound (initially 0) and upper
+    # bound until convergence
+    δ_lower = N(0)
+    while δ_upper - δ_lower > ε
```
mforets: we can also add a termination criterion that compares the width of the interval wrt its previous value in the loop and if it is too small then stop (meaning that progress is stuck) -- or just compares two consecutive values of δ

src/Approximations/hausdorff_distance.jl
schillic: I instead added the requirement `ε > 0`.

src/Approximations/hausdorff_distance.jl
schillic: This is already happening.

PR #1550, opened 2019-07-31
#1021 - Remove redundancies in zonotope's vertices list

### LINKED ISSUE #1021: Zonotope has redundant vertices
```julia
julia> Z = Zonotope([0., 0.], [1. 0. 1.; 0. 1. 1.]);

julia> vertices_list(Z)
8-element Array{Array{Float64,1},1}:
 [2.0, 2.0]  
 [0.0, 2.0]  
 [2.0, 0.0]  
 [0.0, 0.0]  
 [0.0, 0.0]  
 [-2.0, 0.0] 
 [0.0, -2.0] 
 [-2.0, -2.0]
```
The origin is both a duplicate and is even inside.
It turns out that `2^p` is only an upper bound (and probably a lower bound is `2p`).
Related: #593 and #561

mforets:
Closes #1021.

### DIFF
--- a/src/Zonotope.jl
+++ b/src/Zonotope.jl
@@ -195,7 +195,7 @@ center with some linear combination of generators, where the combination
 factors are ``ξ_i ∈ \\{-1, 1\\}``.
 
 There are at most ``2^p`` distinct vertices. Use the flag `apply_convex_hull` to
-control if a convex hull algorithm is applied to the vertices computed by this
+control whether a convex hull algorithm is applied to the vertices computed by this
 method; otherwise, redundant vertices may be present.
 """
 function vertices_list(Z::Zonotope{N};
--- a/src/Zonotope.jl
+++ b/src/Zonotope.jl
@@ -49,6 +49,7 @@ A two-dimensional zonotope with given center and set of generators:
 ```jldoctest zonotope_label
 julia> Z = Zonotope([1.0, 0.0], [0.1 0.0; 0.0 0.1])
 Zonotope{Float64}([1.0, 0.0], [0.1 0.0; 0.0 0.1])
+
 julia> dim(Z)
 2
 ```
@@ -61,8 +62,8 @@ julia> vertices_list(Z)
 4-element Array{Array{Float64,1},1}:
  [1.1, 0.1]
  [0.9, 0.1]
- [1.1, -0.1]
  [0.9, -0.1]
+ [1.1, -0.1] 
 ```
 
 The support vector along a given direction can be computed using `σ`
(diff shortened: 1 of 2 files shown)


test/unit_Zonotope.jl
```diff
@@ -221,4 +221,9 @@ for N in [Float64]
         P = tovrep(HPolygon(constraints_list(Z)))
         @test ispermutation(vertices_list(P), [N[5, 0], [-5, 0]])
     end
+
+    # test that redundant vertices are removed by default (#1021)
+    Z = Zonotope([0., 0.], [1. 0. 1.; 0. 1. 1.])
+    vlistZ = vertices_list(Z)
+    @test ispermutation(vlistZ, [N[-2, -2], N[0, -2], N[2, 0], N[2, 2], N[0, 2], N[-2, 0]])
```
schillic: Can you add the same test with `apply_convex_hull=false` to see the difference?

src/Zonotope.jl
```diff
@@ -172,30 +173,33 @@ end
...
-### Notes
+If the zonotope has ``p`` generators, each vertex is the result of summing the
+center with some linear combination of generators, where the combination
+factors are ``ξ_i ∈ \\{-1, 1\\}``.
 
-For high dimensions, it would be preferable to develop a `vertex_iterator`
-approach.
+There at most ``2^p`` distinct vertices. Use the flag `apply_convex_hull` to
+control if a convex hull algorithm is applied to the vertices computed by this
+method; otherwise, redundant vertices may be present.
```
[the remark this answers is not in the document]
schillic:
```julia
julia> Z = Zonotope(zeros(2), [1. 1. 0.; 1. -1. 1.]);

julia> vertices_list(Z)
8-element Array{Array{Float64,1},1}:
 [2.0, 1.0]  
 [0.0, -1.0] 
 [0.0, 3.0]  
 [-2.0, 1.0] 
 [2.0, -1.0] 
 [0.0, -3.0] 
 [0.0, 1.0]  
 [-2.0, -1.0]
```

zonotope

Obviously the generators of this zonotope are not redundant (you cannot express this shape with just two generators). What I probably meant was linear dependence.

test/unit_Zonotope.jl
```diff
@@ -225,5 +225,10 @@ for N in [Float64]
     # test that redundant vertices are removed by default (#1021)
     Z = Zonotope([0., 0.], [1. 0. 1.; 0. 1. 1.])
     vlistZ = vertices_list(Z)
+    @test length(vlistZ) == 6
```
schillic: I think this is redundant because `ispermutation` checks this already.

Issue #1392, opened 2019-05-23
Remove redundant vertices in VPolytope constructor


schillic:
> the constructor of VPolytope does not remove redundant vertices.

Here is a proposal for a default constructor, but we have to make sure that we pass `false` internally whenever we know that no redundant vertices exist.

```julia
    function VPolytope{N}(vertices::Vector{Vector{N}};
                          removeredundancy::Bool=true) where {N<:Real}
        if removeredundancy
            @assert isdefined(@__MODULE__, :Polyhedra) "redundancy removal " *
                "needs the package 'Polyhedra' loaded"
            P = polyhedron(Polyhedra.vrep(vertices))
            removevredundancy!(P)
            vertices = P.vrep.points.points
        end
        return new{N}(vertices)
    end

# convenience constructor without type parameter
VPolytope(vertices::Vector{Vector{N}}) where {N<:Real} = VPolytope{N}(vertices)
```

_Originally posted by @schillic in https://github.com/JuliaReach/LazySets.jl/issues/561#issuecomment-448146681_

schillic:
Since `removeredundancy` requires `Polyhedra`, we should not use it by default.

PR #1551, opened 2019-07-31
#601 - Add underapproximate function

### LINKED ISSUE #601: Polyhedral underapproximation
Analogous to polyhedral overapproximations, we can easily underapproximate a **convex** set by sampling support vectors `v_1, ..., v_k` and taking the convex hull of those (in our terminology, `VPolytope([v_1, ..., v_k])`).
There might also be a procedure like our iterative refinement for epsilon-close underapproximation.

mforets:
Closes #601.

### DIFF
--- a/src/Approximations/underapproximate.jl
+++ b/src/Approximations/underapproximate.jl
@@ -1,0 +1,20 @@
+"""
+    underapproximate(X::LazySet{N}, dirs::AbstractDirections) where {N<:Real}
+
+Compute the underapproximation of a convex set by sampling support vectors.
+
+### Input
+
+- `X`    -- set
+- `dirs` -- directions
+
+### Output
+
+The `VPolytope` obtained by taking the convex hull of the support vectors of `X`
+along the directions determined by `dirs`.
+"""
+function underapproximate(X::LazySet{N}, dirs::AbstractDirections) where {N<:Real}
+    @assert dim(X) == dim(dirs)
+    vinner = [σ(di, X) for di in dirs]
+    return VPolytope(vinner)
+end
--- a/src/Approximations/underapproximate.jl
+++ b/src/Approximations/underapproximate.jl
@@ -1,20 +1,29 @@
 """
-    underapproximate(X::LazySet{N}, dirs::AbstractDirections) where {N<:Real}
+    underapproximate(X::LazySet{N}, dirs::AbstractDirections;
+                    [apply_convex_hull]::Bool=false) where {N<:Real}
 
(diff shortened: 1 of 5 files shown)


mforets:
Yes. Actually the `VPolytope` constructor does not compute the convex hull ATM (this is https://github.com/JuliaReach/LazySets.jl/issues/1392). So my impementation is wrong and we should either return a `ConvexHullArray` of `Singleton` or just apply `convex_hull` to `vinner`.

schillic:
> So my impementation is wrong

I do not see why. It may just contain redundant vertices.

src/Approximations/underapproximate.jl
```diff
@@ -0,0 +1,29 @@
...
+                         the support vectors with a convex hull operation
+### Output
+
+The `VPolytope` obtained by taking the convex hull of the support vectors of `X`
+along the directions determined by `dirs`.
+"""
+function underapproximate(X::LazySet{N}, dirs::AbstractDirections;
+                          apply_convex_hull::Bool=false) where {N<:Real}
+    @assert LazySets.dim(X) == LazySets.dim(dirs)
+    vinner = Vector{Vector{N}}(undef, length(dirs)) # TODO could be generalized to eltype(dirs)
```
mforets: The reason that `eltype(dirs)` doesn't currently work here is that `eltype` of directions is other than `Vector` in general, while `VPolytope`'s `vertices` fields is a `vertices::Vector{Vector{N}}`. That should be eventually generalized.

src/Approximations/underapproximate.jl
```diff
@@ -1,20 +1,29 @@
...
 The `VPolytope` obtained by taking the convex hull of the support vectors of `X`
 along the directions determined by `dirs`.
 """
-function underapproximate(X::LazySet{N}, dirs::AbstractDirections) where {N<:Real}
-    @assert dim(X) == dim(dirs)
-    vinner = [σ(di, X) for di in dirs]
+function underapproximate(X::LazySet{N}, dirs::AbstractDirections;
+                          apply_convex_hull::Bool=false) where {N<:Real}
+    @assert LazySets.dim(X) == LazySets.dim(dirs)
+    vinner = Vector{Vector{N}}(undef, length(dirs)) # TODO could be generalized to eltype(dirs)
```
schillic: Not sure about the TODO. The result type of `σ` also depends on `X`.

src/Approximations/underapproximate.jl
schillic: Not sure about the TODO. The result type of σ also depends on X.

PR #1574, opened 2019-08-06
Install external dependency for GLFW.jl

schillic:
This is an attempt to fix the recent build errors with `FFMPEG` and `GLFW`.


schillic:
It seems that the errors are rather random.

PR #1577, opened 2019-08-07
#1576 - Robust set membership for AbstractHyperrectangle

### LINKED ISSUE #1576: Robust set membership test for AbstractHyperrectangle
The support vector of a hyperrectangle does not belong to it (due to default floating-point tolerances):

```julia
julia> H = rand(Hyperrectangle, dim=5);

julia> σ(rand(5), H) ∈ H
false
```
We should instead use a robust check via `!_leq(...)`.

mforets:
Closes #1576.

### DIFF
--- a/src/AbstractHyperrectangle.jl
+++ b/src/AbstractHyperrectangle.jl
@@ -400,17 +400,17 @@ Then ``x ∈ H`` iff ``|c_i - x_i| ≤ r_i`` for all ``i=1,…,n``.
 function ∈(x::AbstractVector{N},
            H::AbstractHyperrectangle{N})::Bool where {N<:Real}
     @assert length(x) == dim(H)
-    for i in eachindex(x)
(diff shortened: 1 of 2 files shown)


src/AbstractHyperrectangle.jl
```diff
@@ -400,18 +400,18 @@ Then ``x ∈ H`` iff ``|c_i - x_i| ≤ r_i`` for all ``i=1,…,n``.
 function ∈(x::AbstractVector{N},
            H::AbstractHyperrectangle{N})::Bool where {N<:Real}
     @assert length(x) == dim(H)
-    for i in eachindex(x)
-        if abs(center(H)[i] - x[i]) > radius_hyperrectangle(H, i)
+    c = center(H)
```
schillic: This requires to compute the full `SymmetricIntervalHull` even if you can find a counterexample in the first dimension.

PR #1580, opened 2019-08-09
Remove cp array from cp constructor

mforets:
This constructor exists since earlier versions prior to `xxxArray`, but it is no longer used (and its use is not recommended), so we may as well remove it in library code.


schillic:
It needs to be removed from the docs of `CartesianProduct` as well.

mforets:
Seems to be good because the docs use the general signature

```julia
```@docs
CartesianProduct
...
```

schillic:
I meant this:
https://github.com/JuliaReach/LazySets.jl/blob/0534718d77cfb19ba585cff612c6de94585f48cf/src/CartesianProduct.jl#L28-L34

Issue #452, opened 2018-08-01
Remove duplicate exports


schillic:
For instance, we have `export an_element` in `LazySet.jl`. This is sufficient.
However, it can also be helpful if we always write all exported functions at the top of each file in order to see what is public API and what are private functions. In that case we should be consistent and export *all* newly defined functions in each file.

If we go for removing the `export`s, should we also remove `export`s of methods that are defined in different concrete types, like `halfspace_left`? Here it would not be clear where to keep it and where to remove it.

Yet another idea is to collect all `export`s in one file, e.g., `LazySets.jl` in order to see all `export`s in one place.

mforets:
> However, it can also be helpful if we always write all exported functions at the top of each file in order to see what is public API and what are private functions. In that case we should be consistent and export all newly defined functions in each file.

Yes, IMO in `LazySets` this approach looks sufficiently compact and understandable. The most common method that we omit is `export σ`, because all `LazySets` should implement it anyway, but it doesn't hurt to add it..

mforets:
in other words, i think that some cases already under`LazySet.jl` can stay there, such as `export σ`, and so on, maybe `an_element` as well, since they are generic for all sets.

schillic:
> in other words, i think that some cases already underLazySet.jl can stay there, such as export σ, and so on, maybe an_element as well, since they are generic for all sets.

Okay, so omit those defined in `LazySet.jl`. Does the same apply for other interfaces?

> Yet another idea is to collect all exports in one file, e.g., LazySets.jl in order to see all exports in one place.

And what about this one, in addition? I think this can be helpful.

schillic:
See also the duplicate #874.

PR #1598, opened 2019-08-13
Fix typo

schillic:
This is why I prefer to have #452.


mforets:
how this doesn't produce an error? 😕

schillic:
You just export a function name to Julia's lookup mechanism. It does not have to be defined.

PR #1601, opened 2019-08-13
Minkowski sum of VPolygon with small number of vertices

### LINKED ISSUE #1600: Minkowski sum of VPolygon's with small number of vertices
```julia
julia> v1 = rand(VPolygon, num_vertices=2);

julia> v2 = rand(VPolygon, num_vertices=2);

julia> minkowski_sum(v1, v2)
ERROR: AssertionError: n > 2
Stacktrace:
 [1] _binary_support_vector(::Array{Float64,1}, ::VPolygon{Float64,Array{Float64,1}}) at C:\Users\marce\.julia\dev\LazySets\src\VPolygon.jl:277
 [2] minkowski_sum(::VPolygon{Float64,Array{Float64,1}}, ::VPolygon{Float64,Array{Float64,1}}) at C:\Users\marce\.julia\dev\LazySets\src\VPolygon.jl:632
 [3] top-level scope at none:0
```

mforets:
Closes #1600.

### DIFF
--- a/src/VPolygon.jl
+++ b/src/VPolygon.jl
@@ -258,7 +258,7 @@ function σ(d::AbstractVector{N}, P::VPolygon{N}) where {N<:Real}
 end
 
 # heuristic to define the method used to compute the support vector of a polygon
-# in vertex representation; if the number of vertices of the polygon is smallar
+# in vertex representation; if the number of vertices of the polygon is smaller
 # than this value, the brute force method is used; otherwise binary search is used
 const binary_or_brute_force = 10
 
--- a/src/VPolygon.jl
+++ b/src/VPolygon.jl
@@ -257,7 +257,7 @@ function σ(d::AbstractVector{N}, P::VPolygon{N}) where {N<:Real}
     return P.vertices[_σ_helper(d, P)]
 end
 
-# heuristic to defines the method used to compute the support vector of a polygon
+# heuristic to define the method used to compute the support vector of a polygon
 # in vertex representation; if the number of vertices of the polygon is smallar
 # than this value, the brute force method is used; otherwise binary search is used
 const binary_or_brute_force = 10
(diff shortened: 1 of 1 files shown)


src/VPolygon.jl
```diff
@@ -254,11 +254,20 @@ based on an algorithm described in [1].
 """
 function σ(d::AbstractVector{N}, P::VPolygon{N}) where {N<:Real}
     @assert !isempty(P.vertices) "the polygon has no vertices"
-    binary_or_brute_force = 10;
+    return P.vertices[_σ_helper(d, P)]
+end
+
+# heuristic to defines the method used to compute the support vector of a polygon
```
schillic:
```suggestion
# heuristic to define the method used to compute the support vector of a polygon
```

src/VPolygon.jl
```diff
@@ -254,11 +254,20 @@ based on an algorithm described in [1].
...
+# in vertex representation; if the number of vertices of the polygon is smallar
```
schillic:
```suggestion
# in vertex representation; if the number of vertices of the polygon is smaller
```

src/VPolygon.jl
```diff
@@ -254,11 +254,20 @@ based on an algorithm described in [1].
...
+# than this value, the brute force method is used; otherwise binary search is used
+const binary_or_brute_force = 10
+
+# return the index of a support vector of P along d from the list of vertices of P
+function _σ_helper(d::AbstractVector{N}, P::VPolygon{N}) where {N <: Real}
     if length(P.vertices) > binary_or_brute_force
```
schillic: I wonder if we should change this:
```suggestion
     if length(P.vertices) >= binary_or_brute_force
```

mforets: The benchmarks showed that the transition happened between 9 and 11 iirc.

cc: @SebastianGuadalupe

schillic: This is not about the number. You can change the constant as well. It is more like what I would expect to happen if I see this constant.

PR #1612, opened 2019-08-16
#1589 - Rejection Sampling

### LINKED ISSUE #1589: Rejection sampling
[Rejection sampling](https://en.wikipedia.org/wiki/Rejection_sampling) is good as a (very general) fallback, since it only needs `LazySets.Approximations.box_approximation`.

_Originally posted by @mforets in https://github.com/JuliaReach/LazySets.jl/issues/917#issuecomment-509178607_

cc: @ueliwechsler

ueliwechsler:
Closes #1589.

Not anything is documented yet because I am not sure about the general structure. 
I created a new type `Sampler`, in order to have more flexibility and also use the Sampler outside of LazySets without recomputing the `Sampler.sampler` variable again ... maybe you have an idea for a better name there ;)
I am looking forward to your feedback.


src/sample_from_set.jl
```diff
@@ -0,0 +1,100 @@
...
+- `P`           -- lazyset
+- `n_samples`    -- number of random samples
+- `rng`         -- (optional, default: `GLOBAL_RNG`) random number generator
+- `seed`        -- (optional, default: `nothing`) seed for reseeding
+
+### Output
+
+A vector of `n_samples` vectors.
+"""
+function Base.rand(P::LazySet{N}, n_samples::Int;
```
schillic: @mforets does not like `rand`.

src/sample_from_set.jl
```diff
@@ -0,0 +1,100 @@
+function load_distributions_sample_from_set()
```
schillic: Maybe the file name `sample_from_set.jl` could be shortened to `sample.jl`. @mforets?

src/sample_from_set.jl
```diff
@@ -0,0 +1,100 @@
...
+"""
+function canonical_length(P::LazySet{N}) where {N<:Real}
+    dims = dim(P)
+    x = Array{N}(undef, dims, 2)
+    for j=1:dims
+        ej = Arrays.SingleEntryVector(j, dims, 1.0)
+        x[j,:] = [-ρ(-ej, P), ρ(ej, P)]
+    end
+    return x
+end
```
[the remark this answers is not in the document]
mforets: Well it depends, box approximation can (and should) dispatch on the type of X hence using it is preferable. For instance, if X is an hyperrectangle the operation should be a no-op.

src/sample_from_set.jl
ueliwechsler: I like the use of `rand` in julia in general because then you don't have to think about what function to use to get an instance or a sample of a specific type. But I am open to suggestions. Would `sample` better?

src/sample_from_set.jl
ueliwechsler: There is also a file `samples.jl` should I merge this to files?

schillic: Yes!

src/sample_from_set.jl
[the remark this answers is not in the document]
schillic: True, I was only considering the creation of the matrix via `hcat` and the fact that our `Hyperrectangle` has a different representation, so it has to compute center/radius and convert back. But yes, this is in the order of milliseconds, so efficiency is really not an argument here.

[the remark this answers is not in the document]
ueliwechsler: Okay. Where should this function be located?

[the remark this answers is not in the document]
schillic: It could just replace your function (remove the `2` in the name of course).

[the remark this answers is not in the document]
ueliwechsler: Sorry, I was unclear. I was just wondering if this new function should be in the `samples.jl`  file or somewhere else.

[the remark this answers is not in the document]
schillic: Ah. It is a general function, but we never needed it so far. The name is not clear for a general purpose (but I cannot think of a good name). So I would
1. either put it in `samples.jl` (and maybe add a leading `_` which we use to mark internal functions)
2. or put it in `AbstractHyperrectangle.jl` if you find a name that clearly indicates what this function does.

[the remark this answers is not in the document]
ueliwechsler: Okay. I will go with the first version.

Just as a little caveat. With the new function the test for an unbounded `Polyhedra` breaks. 

```julia
P3 = HPolyhedron([0.0 0.1], [3.0])
@show box_approximation(P3)
@test LazySets.canonical_length(P3) == [-Inf Inf; -Inf 30.0]
```
```julia
a = box_approximation(P3) = Hyperrectangle{Float64}([NaN, -Inf], [Inf, Inf])
Evaluated: [NaN NaN; -Inf NaN] == [-Inf Inf; -Inf 30.0]
```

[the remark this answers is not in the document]
schillic: Oh, true! Then you should of course keep your version :+1:

src/sample_from_set.jl
```diff
@@ -0,0 +1,9 @@
+function load_distributions_sample_from_set()
```
schillic: This file can now be removed.

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+
+mutable struct Sampler
+    X
+    sampler
+    distribution
+end
+
+function Sampler(X, distribution=Uniform)
+    box_lengths = _canonical_length(X)
+    sampler = [distribution(length...) for length in eachrow(box_lengths)]
```
schillic: This throws `LoadError: UndefVarError: eachrow not defined` in Julia v1.0.

test/runtests.jl
```diff
@@ -143,6 +143,11 @@ if test_suite_basic
     # ====================================
     include("check_method_implementation.jl")
     @time @testset "LazySets.interfaces" begin include("unit_interfaces.jl") end
+
+    # Also loads Distributions
```
schillic:
```suggestion
```

test/unit_samples.jl
```diff
@@ -0,0 +1,31 @@
+# using Pkg
+# Pkg.activate("LazySets")
+# using Test
+
+using LazySets
+using Distributions
+using Polyhedra
```
schillic: Eventually remove these lines above (they are not needed).

src/sample_from_set.jl
mforets: Indeed, we had discussion about this and the functions are so different that keeping the names separate is better IMO, provided that:

- `rand` receives a set *type* and returns a LazySet instance -- (i.e. an infinite number of points, and you draw from the distribution on the type given some constraints...)
- `sample` receives a LazySet *instance* and returns a sample of points in the set -- (a finite number of points and you draw from a distribution which is eg. uniform over the set)

src/samples.jl
```diff
@@ -1,7 +1,7 @@
 function load_distributions_samples()
 return quote
 
-using .Distributions: Uniform, Normal
+using .Distributions: Distribution, Uniform, Normal
```
schillic: You never used `Distribution`... :confused:

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+## Rejection Sampling
+
+"""
+    _canonical_length(X::LazySet)
+
+Get the canonical length of the box-approximation of the convex set `X`.
+
+### Inputs
+
+- `X`   - lazyset
```
schillic:
```suggestion
- `X` - convex set
```

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+
+### Outputs
+
+Matrix with two columns and `n=dims(X)` rows. Each column stands for
```
schillic:
```suggestion
A matrix with `n = dims(X)` rows and two columns. Each row stands for
```

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+one dimension of `X` whereas the first column is the minimum and the second
+column the maximum value of the corresponding dimension.
```
schillic:
```suggestion
column is the maximum value of the corresponding dimension.
```

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+"""
+function _canonical_length(X::LazySet{N}) where {N<:Real}
+    dims = dim(X)
+    x = Matrix{N}(undef, dims, 2)
+    for j=1:dims
```
schillic:
```suggestion
    for j = 1:dims
```

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+    end
+    return x
+end
+
+"""
+    rand(X::LazySet{N}, n_samples::Int;
+           rng::AbstractRNG=GLOBAL_RNG,
+           seed::Union{Int, Nothing}=nothing) where {N}
+
+Rejection sampling of an arbitrary LazySet `X` for which the support value
```
schillic:
```suggestion
Rejection sampling of an arbitrary `LazySet` `X`.
```
(The support function is a requirement of this library.)

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+function is defined.
```
schillic:
```suggestion
```

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+
+Draw a sample `x` from a uniform distribution of a box-overapproximation of the
+original set `X` in all `n` dimensions. The function rejects a drawn sample `x`
+and redraws as long as the sample is not contained in the original set `X`,
```
schillic: Move this paragraph to a new section `### Algorithm` at the end of this docstring.

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+i.e., `x ∉ X`.
+
+### Input
+
+- `X`           -- lazyset
+- `n_samples`    -- number of random samples
```
schillic:
```suggestion
- `n_samples`   -- number of random samples
```

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+function is defined.
+
+Draw a sample `x` from a uniform distribution of a box-overapproximation of the
+original set `X` in all `n` dimensions. The function rejects a drawn sample `x`
+and redraws as long as the sample is not contained in the original set `X`,
+i.e., `x ∉ X`.
+
+### Input
+
+- `X`           -- lazyset
```
schillic:
```suggestion
- `X`           -- convex set
```

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+
+### Output
+
+A vector of `n_samples` vectors.
+"""
+function Base.rand(X::LazySet{N}, n_samples::Int;
+                rng::AbstractRNG=GLOBAL_RNG,
+                seed::Union{Int, Nothing}=nothing) where {N<:Real}
+    rng = reseed(rng, seed)
+    @assert isbounded(X) "this function requires that the set `X` is bounded, but it is not"
```
schillic: Huh, why should the set be bounded?

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+    return D
+end
+
+function Base.rand(X::LazySet{N};
+                rng::AbstractRNG=GLOBAL_RNG,
+                seed::Union{Int, Nothing}=nothing) where {N<:Real}
+    return rand(X,1)[1]
+end
+
+mutable struct Sampler
```
schillic: Maybe call it `BoxSampler`?

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+    X
+    sampler
+    distribution
```
schillic: `distribution` is actually never used. This does not necessarily mean that it is not good to store this information, though.

src/samples.jl
```diff
@@ -117,5 +117,100 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+    box_lengths = _canonical_length(X)
+    sampler = [distribution(length...) for length in eachrow(box_lengths)]
+    return Sampler(X, sampler, distribution)
+end
+
+function sample!(D::Vector{Vector{N}},
+               sampler::Sampler;
+               rng::AbstractRNG=GLOBAL_RNG) where {N}
+    @inbounds for i in 1:length(D)
+        while true
```
schillic: Just a comment: I find the following loop easier to read:
```julia
i = 0
n = @inbounds while i < length(D)
    w = rand.(Ref(rng), sampler.sampler)
    if w ∈ sampler.X
        i += 1
        D[i] = w
    end
end
```

test/unit_samples.jl
```diff
@@ -0,0 +1,31 @@
...
+
+n = 5
```
schillic:
```suggestion
```

test/unit_samples.jl
```diff
@@ -0,0 +1,31 @@
...
+P1 = rand(BallInf, dim=n)
+A = [1.0  0.0;
+     0.0  1.0;
+    -1.0  0.0;
+     0.0 -1.0]
+b = [1.0, 2.0, 3.0, 4.0]
+P2 = HPolyhedron(A, b)
+P3 = HPolyhedron([0.0 0.1], [3.0])
+P4 = rand(Ball2, dim=n)
```
schillic:
```suggestion
P4 = Ball2{Float64}([0.2, -0.3, -1.1, 0.6, -0.7], 0.4)
```
I prefer to have deterministic tests where possible for reproducibility.
(Of course the sampling itself can be random.)

test/unit_samples.jl
```diff
@@ -0,0 +1,31 @@
+# using Pkg
+# Pkg.activate("LazySets")
+# using Test
+
+using LazySets
+using Distributions
+using Polyhedra
+
+n = 5
+P1 = rand(BallInf, dim=n)
```
schillic:
```suggestion
P1 = BallInf{Float64}([0.6, 0.1, -1.3, -0.4, 0.2], 0.6)
```

src/samples.jl
ueliwechsler: What is the reason for using `n = `?

src/samples.jl
ueliwechsler: Right. But I think, it is not necessary. The distribution information is already stored in `sampler`.

src/samples.jl
ueliwechsler: Right, in the beginning I had
```julia
mutable struct Sampler 
    X::LazySet
    sampler::Vector{<:Distribution}
end
```
Should I add this? What is best practice?

src/sample_from_set.jl
ueliwechsler: 👍
Should I import `sample` from `Distributions`? Because I got the warning message
```WARNING: both Distributions and LazySets export "sample"; uses of it in module Main must be qualified```

File: src/samples.jl (hunk 5 above)
Comment (ueliwechsler) [reply src/samples.jl]: It should work now.

File: src/sample_from_set.jl (hunk 1 above)
Comment (schillic) [reply src/sample_from_set.jl]: Yes, you should.

File: src/samples.jl (hunk 21 above)
Comment (schillic) [reply src/samples.jl]: The is a relict of a previous version where I precomputed `length(D)`, but I decided that this is not worth adding a line. Just remove it :wink:

File: test/unit_samples.jl (hunk 7 above)
Comment (schillic) [reply test/unit_samples.jl]: There are some commented lines on top that can also be removed.

File: src/samples.jl (hunk 8 above)
Comment (schillic) [reply src/samples.jl]: I think the best is to add type parameters.
```julia
struct Sampler{S<:LazySet, D<:Distribution}
    X::S
    sampler::Vector{D}
end
```
Note that the struct needs not be `mutable` if you only ever modify the `Vector`'s contents.

test/unit_samples.jl
```diff
@@ -0,0 +1,28 @@
...
+@test sum(p1_samples .∈ Ref(P1)) == length(p1_samples)
+
+# test _canonical_length
+@test LazySets._canonical_length(P1+-P1.center) ≈ radius(P1)*[-ones(5) ones(5)]
+@test LazySets._canonical_length(P2) == [-b[3:4] b[1:2]]
+#does not work with box_approximation in _canonical_length at the moment
+@test LazySets._canonical_length(P3) == [-Inf Inf; -Inf 30.0]
+@test LazySets._canonical_length(P4+-P4.center) ≈ radius(P4)*[-ones(5) ones(5)]
+
+@test LazySets.SetSampler(P2).sampler == [Uniform(-3.0,1.0), Uniform(-4.0,2.0)]
```
schillic: You need to add `using Distributions: Uniform` to the top of this file.

src/samples.jl
mforets: > What is best practice?

Indeed, for library code it is preferable to consider using type parameters; see the discussion in https://docs.julialang.org/en/v1/manual/performance-tips/#Type-declarations-1

src/samples.jl
```diff
@@ -117,5 +117,98 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
     return D
 end
 
+## Rejection Sampling
+
+"""
+    _canonical_length(X::LazySet)
+
+Get the canonical length of the box-approximation of the convex set `X`.
```
mforets: I had to look at the code to understand the meaning of "canonical length"; maybe say that the function `Returns the support function of the given set along the positive and negative canonical directions.`

src/samples.jl
```diff
@@ -117,5 +117,98 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+- `X`           -- lazyset
+- `n_samples`    -- number of random samples
+- `rng`         -- (optional, default: `GLOBAL_RNG`) random number generator
+- `seed`        -- (optional, default: `nothing`) seed for reseeding
+
+### Output
+
+A vector of `n_samples` vectors.
+"""
+function sample(X::LazySet{N}, n_samples::Int;
```
mforets: Let's call the second argument `num_samples` (compare to `num_vertices`, or `num_constraints`).

src/samples.jl
mforets: Rejection sampling proceeds by taking a box overapproximation of the set, and sample inside that box. If the set is unbounded, the box will have infinities. So in that case we  should sample from a distribution with unbounded support. It's ok to restrict to bounded sets in this issue (and open a new issue for unbounded sets so we don't forget).

src/samples.jl
```diff
@@ -117,5 +117,98 @@ function _sample_unit_nball_muller!(D::Vector{Vector{N}}, n::Int, p::Int;
...
+                rng::AbstractRNG=GLOBAL_RNG,
+                seed::Union{Int, Nothing}=nothing) where {N<:Real}
+    @assert isbounded(X) "this function requires that the set `X` is bounded, but it is not"
+
+    D = Vector{Vector{N}}(undef, n_samples) # preallocate output
+    sample!(D, SetSampler(X); rng=rng, seed=seed)
+    return D
+end
+
+function sample(X::LazySet{N}; kwargs...) where {N<:Real}
```
mforets: This is ok but i don't see a compelling reason to not give the default value, 1, to the function defined in line 171. The second argument can still have a default value without being a keyword argument. What do you think?

src/samples.jl
schillic: @mforets: Just a comment: in https://github.com/JuliaReach/LazySets.jl/blob/c4e52db11b6c90543026a34d08571e7ae537bdf5/src/samples.jl#L98 you called the variable `p` (which I find arbitrary).

src/samples.jl
schillic: I was referring to [this comment](https://github.com/JuliaReach/LazySets.jl/pull/1612#discussion_r314941390). The only reason why we kept the function `_canonical_length` was because it additionally supports unbounded sets. And this assertion makes that case break.

Note that `rand` is never tested for unbounded sets here (and I bet it would fail).

test/unit_samples.jl
ueliwechsler: Will do that. I left them there because I am still not able to run the `runtest.jl` file due to problem with `CDDLib` on Windows.

src/samples.jl
ueliwechsler: Thanks for the explanation. I will add your version.

src/samples.jl
ueliwechsler: For me both versions are not that readable, to be honest ;)
I will add a clearer but slightly more verbose version and like to hear your opinion.

src/samples.jl
ueliwechsler: I agree with you in general. But I did this to avoid having a `Vector{Vector}` as a result when I only want one sample from a set. Therefore the `[1]` in `sample(X, 1; kwargs...)[1]`. Is there a better way to handle this?

src/sample_from_set.jl
ueliwechsler: I did this. But there is still the warning.
`WARNING: import of Distributions.sample into LazySets conflicts with an existing identifier; ignored.`
I think this is due to the require in https://github.com/JuliaReach/LazySets.jl/blob/f3e5fbadee35a13c26199d72f289055ad89bf15c/src/Ball2.jl#L266

Should we move the sample function outside of the `samples.jl` file?

src/samples.jl
schillic: Looks good!

src/sample_from_set.jl
schillic: I found that just doing `import .Distributions` instead of `import .Distributions: sample` works.

src/samples.jl
schillic: I disagree. Type instabilities are not good.

schillic: You can add the default value `1` to the other method, but the return type should *always* be a vector of vectors. For convenience it is useful to have this other method that only returns a single vector instead.

schillic: True, then forget about the default value `1`.

schillic: I misunderstood your comment completely, sorry. I agree with what you said :+1:

ueliwechsler:
I added some more documentation and changed some function names. I hope this does not make it worse. Is there anything to do? @schillic I think you were also saying, that it does not make to use `_canonical_length` when the only case it would be "better" than the `box_approximation` version is momentary not support (because `rand(Uniform(Inf,3))` does not work).

schillic:
> Is there anything to do?

* There are still some comments to resolve (they are hidden because there are so many; you need to click "Load more") [here](https://github.com/JuliaReach/LazySets.jl/pull/1612#discussion_r314948323), [here](https://github.com/JuliaReach/LazySets.jl/pull/1612#discussion_r314948329), [here](https://github.com/JuliaReach/LazySets.jl/pull/1612#discussion_r314948368), and [here](https://github.com/JuliaReach/LazySets.jl/pull/1612#discussion_r314948387).
* I reopened a discussion [here](https://github.com/JuliaReach/LazySets.jl/pull/1612#discussion_r314948444) (the name `Sampler` sounds general, but the implementation only works for boxes; if we ever plan to add other samplers, we would have to rename this type).
* You need to fix the merge conflicts (for instance by rebasing with master: `git rebase master`).
* After that, we do the next iteration :repeat: :smiling_imp:

schillic:
> I think you were also saying, that it does not make to use _canonical_length when the only case it would be "better" than the box_approximation version is momentary not support (because rand(Uniform(Inf,3)) does not work).

True, I would prefer to remove this function (mainly because it is unclear from the name what it does and we might get confused about it).

mforets:
> You need to fix the merge conflicts (for instance by rebasing with master: git rebase master).

Conflicts resolved.

mforets:
@ueliwechsler  Thanks again for contributing with this PR that adds rejection sampling for an arbitrary lazyset 🎉 👍  The PR looks good and is functional, and is covered with unit tests. Let me address the remaining review comments in a follow-up PR.

PR #1616, opened 2019-08-17
#1603 - List of LazySet types that implement a set-operation

### LINKED ISSUE #1603: List of LazySet types that implement a set-operation
Given a set operation `op` (e.g., `linear_map`), I want to know all set types `T` such that `op(::T)` is applicable. This should be easy via `applicable` and `subtypes(LazySet)`. We can start with unary operations.
Likewise, I want to know the complement of these set types (i.e., those for which an implementation is missing). So the function should return two lists.

proposed name: `implementing_sets(::Symbol)`

#### Bonus challenge
Find out whether the implementation is specific to a concrete set type (e.g., `BallInf`) or inherited from an interface (e.g., `AbstractHyperrectangle`). So the bonus version should return three lists: specific implementations, inherited implementations, and missing implementations.

#### Applications
- You want to automatically generate a table of existing set operations.
- You find out that an operation is missing for your set type. Now you want to look at available set types to check whether their implementation can be extended.
- You want to know which implementations are missing to add a new one.
- You want to know which implementations could possibly be improved by adding a specific implementation (bonus version).

schillic:
Closes #1603.

### DIFF
--- a/docs/src/lib/utils.md
+++ b/docs/src/lib/utils.md
@@ -64,11 +64,12 @@ StrictlyIncreasingIndices
 
 ```@docs
 LazySets.subtypes(::Any, ::Bool)
+implementing_sets
 ```
 
 ## Sampling
 
 ```@docs
-LazySets._sample_unit_nsphere_muller!
-LazySets._sample_unit_nball_muller!
+_sample_unit_nsphere_muller!
+_sample_unit_nball_muller!
 ```
--- a/src/helper_functions.jl
+++ b/src/helper_functions.jl
@@ -262,3 +262,180 @@ function subtypes(interface, concrete::Bool)
     end
     return sort(result, by=string)
 end
+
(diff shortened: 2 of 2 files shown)


mforets:
The set complement of "available" and "specific" are the types which use the interface-level implementation. Then `implementing_sets` helps to quickly find if a given operation uses the interface-level implementation or a specific one. Can we add an option, or another function, that returns the interface the is invoked?

schillic:
> Can we add an option, or another function, that returns the interface-level implementation that is invoked?

Can you give an example?

mforets:
> Can you give an example?

Think of any function that admits more than one interface-level implementation. One can easily find the called method but a function similar to `implementing_sets` would be useful to compare different types at once.

schillic:
Is this not what `@which` does?

PR #1618, opened 2019-08-17
#1609 - linear_map for polyhedral CartesianProductArray

### LINKED ISSUE #1609: linear_map for polyhedral CartesianProductArray

schillic:
Closes #1609.

### DIFF
--- a/src/CartesianProduct.jl
+++ b/src/CartesianProduct.jl
@@ -279,6 +279,10 @@ In both cases, we then call `linear_map` on the resulting polytope.
 """
 function linear_map(M::AbstractMatrix{N}, cp::CartesianProduct{N}
                    ) where {N<:Real}
+    return linear_map_cartesian_product(M, cp)
+end
+
+function linear_map_cartesian_product(M, cp)
     @assert dim(cp) == size(M, 2) "a linear map of size $(size(M)) cannot " *
                                   "be applied to a set of dimension $(dim(cp))"
 
@@ -293,10 +297,10 @@ function linear_map(M::AbstractMatrix{N}, cp::CartesianProduct{N}
     return linear_map(M, P)
 end
 
-
 # ======================================
 #  Cartesian product of an array of sets
 # ======================================
+
 """
     CartesianProductArray{N<:Real, S<:LazySet{N}} <: LazySet{N}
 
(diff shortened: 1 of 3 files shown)


src/CartesianProduct.jl
```diff
@@ -727,3 +765,42 @@ function substitute_blocks(low_dim_cpa::CartesianProductArray{N},
...
+A polytope.
+
+### Algorithm
+
+We check if the matrix is invertible.
+If so, we convert the Cartesian product to constraint representation.
+Otherwise, we convert the Cartesian product to vertex representation.
+In both cases, we then call `linear_map` on the resulting polytope.
+"""
+function linear_map(M::AbstractMatrix{N}, cpa::CartesianProductArray{N}
```
mforets: This function looks similar to the one for `CartesianProduct`. Using a `Union` type in the previous function may avoid this code dup.

src/CartesianProduct.jl
```diff
@@ -256,6 +256,44 @@ function vertices_list(cp::CartesianProduct{N}
...
+A polytope.
+
+### Algorithm
+
+We check if the matrix is invertible.
+If so, we convert the Cartesian product to constraint representation.
+Otherwise, we convert the Cartesian product to vertex representation.
+In both cases, we then call `linear_map` on the resulting polytope.
+"""
+function linear_map(M::AbstractMatrix{N}, cp::CartesianProduct{N}
```
mforets: This function is OK but it does not exploit the structure of the Cartesian product nor the possible dispatch of `linear_map` on the types of `cp.X` and `cp.Y`. I raise the question if it can be done differently, or we can provide another algorithm.

For instance, if `cp.X` and `cp.Y` are zonotopes, the linear map can be computed efficiently for each sub-block, as well as the minkowski sum of the constituent parts. Then, the result is again a cartesian product of two zonotopes.

src/CartesianProduct.jl
schillic: True. I was thinking about it, but we usually created two versions here. The main reason why I created two methods even if the code is identical was that we would have to decide where to put the docs.
But I could use meta programming to create the two functions in a loop. How does that sound?

src/CartesianProduct.jl
schillic: Yes, I thought about how to exploit the structure, but I think it is usually not possible. The linear map of a Cartesian product is in general not a Cartesian product of the same block structure again. I think it only preserves the structure if the map is decomposed (i.e., block diagonal) or the sets are singletons; maybe there are other corner cases.
We can create a follow-up PR to exploit the matrix structure if you want, but this PR was only supposed to bring a quick implementation for https://github.com/JuliaReach/Reachability.jl/pull/650.

src/CartesianProduct.jl
schillic: I tried but could not figure out how to document *methods* programmatically. [The Julia docs](https://docs.julialang.org/en/v1/manual/documentation/#Advanced-Usage-1) only explain how to document *functions*. [The text here](https://github.com/JuliaLang/julia/blob/b5c4e63db08a56f442b35ffd85a419c54c9c0cc1/base/docs/Docs.jl#L55-L59) seems to indicate that this is not actually possible. So I ended up outsourcing the code to a helper function instead.

PR #1627, opened 2019-08-20
#1606 - Revise lazy-map constructors

### LINKED ISSUE #1606: Move dispatch from aliases to underlying functions
We define this nice method:
https://github.com/JuliaReach/LazySets.jl/blob/9c2f52556c53a1509fd46afef5b433fead6a9f33/src/LinearMap.jl#L122-L125
However, this only applies to `*`. If we use `LinearMap` instead, this method will not help:
```julia
julia> M = [1. 2.; 3 4];

julia> M * ∅
EmptySet{Float64}()

julia> LinearMap(M, ∅)
ERROR: AssertionError: a linear map of size (2, 2) cannot be applied to a set of dimension -1
```
The same applies to the `ExponentialMap`.
For `AffineMap` and `ResetMap` we do not define this method (but we should).

schillic:
Closes #1606.

* [x] `LinearMap`
* [x] `AffineMap`
* [x] `ResetMap`
* [x] `ExponentialMap`
* I skipped `ExponentialProjectionMap` because it is not used and does not have the bug.

### DIFF
--- a/docs/src/lib/operations.md
+++ b/docs/src/lib/operations.md
@@ -324,8 +324,6 @@ linear_map(::AbstractMatrix{N}, ::Translation{N}) where {N<:Real}
 
 ```@docs
 AffineMap
-*(::AbstractMatrix{N}, ::AffineMap{N}) where {N<:Real}
-*(::N, ::AffineMap{N}) where {N<:Real}
 dim(::AffineMap)
 σ(::AbstractVector{N}, ::AffineMap{N}) where {N<:Real}
 ρ(::AbstractVector{N}, ::AffineMap{N}) where {N<:Real}
--- a/docs/src/lib/operations.md
+++ b/docs/src/lib/operations.md
@@ -224,9 +224,7 @@ Inherited from [`LazySet`](@ref):
 
 ```@docs
 LinearMap
-*(::AbstractMatrix{N}, ::LazySet{N}) where {N<:Real}
-*(::N, ::LazySet{N}) where {N<:Real}
-*(::N, ::LinearMap{N}) where {N<:Real}
+*(::Union{AbstractMatrix, UniformScaling, AbstractVector, Real}, ::LazySet)
 dim(::LinearMap)
 ρ(::AbstractVector{N}, ::LinearMap{N}) where {N<:Real}
 σ(::AbstractVector{N}, ::LinearMap{N}) where {N<:Real}
--- a/src/AffineMap.jl
+++ b/src/AffineMap.jl
@@ -118,17 +118,12 @@ function AffineMap(α::N, X::LazySet, v::AbstractVector) where {N<:Real}
 end
 
 # simplification for a LinearMap
-function LinearMap(map, am::AffineMap)
-    return AffineMap(map * am.M, am.X, map * am.v)
-end
-
-# disambiguation
-function LinearMap(M::AbstractMatrix, am::AffineMap)
-    return AffineMap(M * am.M, am.X, M * am.v)
-end
-
-function LinearMap(α::Real, am::AffineMap)
-    return AffineMap(α * am.M, am.X, α * am.v)
+for MAP in (:AbstractMatrix, :Real)
+    @eval begin
+        function LinearMap(map::$MAP, am::AffineMap)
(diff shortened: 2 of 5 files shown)


docs/src/lib/operations.md
```diff
@@ -223,9 +223,7 @@ Inherited from [`LazySet`](@ref):
 
 ```@docs
 LinearMap
-*(::AbstractMatrix{N}, ::LazySet{N}) where {N<:Real}
-*(::N, ::LazySet{N}) where {N<:Real}
-*(::N, ::LinearMap{N}) where {N<:Real}
+*(::Any, ::LazySet)
```
mforets: I think we shouldn't be using `Any` for this operation. It is used for matrices, numeric types or sets.

src/LinearMap.jl
```diff
@@ -21,106 +21,165 @@ independent of the numeric type of the wrapped set (`N`).
...
 ```
 
-Return the linear map of a convex set.
+Alias to create a `LinearMap` object.
 
 ### Input
 
-- `M` -- matrix/linear map
-- `X` -- convex set
+- `map` -- linear map
```
mforets: I don't understand the change from `M::AbstractMatrix`  to "map". This function is meant to produce a new lazy linear map form a given matrix and set type. Other uses should be added through dispatch as usual (or eventually with a small `Union` type).

schillic: I recently saw a Julia talk about dispatch. One of the messages was that not adding types is helpful for new implementations to "just work". This was my motivation. I can remove these methods again if you prefer.

schillic: I had a closer look: the correct way to do things here is to add dispatch methods of `LinearMap`. We currently have them for `*`. Now we only need a single method for `*`, so the old ones become redundant. While they do not hurt, but I prefer to have a smaller code base. So I had to combine all of them into a single method. I did not want to have a big `Union` argument, but changed it now as per your request.

src/AffineMap.jl
```diff
@@ -35,71 +89,66 @@ struct AffineMap{N<:Real, S<:LazySet{N}, NM, MAT<:AbstractMatrix{NM},
...
+function AffineMap(α::N, X::LazySet, v::AbstractVector) where {N<:Real}
+    if α == one(N)
+        return Translation(X, v)
+    end
+    return AffineMap(Diagonal(fill(α, length(v))), X, v)
+end
 
-### Output
+# simplification for a LinearMap
+function LinearMap(map, am::AffineMap)
```
mforets: we can merge this function and the 2 ones below with a `Union` type

src/AffineMap.jl
```diff
@@ -35,71 +89,66 @@ struct AffineMap{N<:Real, S<:LazySet{N}, NM, MAT<:AbstractMatrix{NM},
...
-
-### Output
+    return Singleton(v)
+end
 
-The scaled affine map.
-"""
-function *(α::N, am::AffineMap{N}) where {N<:Real}
-    return AffineMap(α * am.M, am.X, α * am.v)
+# EmptySet is absorbing for LinearMap
```
mforets:
```suggestion
# EmptySet is absorbing for AffineMap
```

src/LinearMap.jl
```diff
@@ -21,106 +21,169 @@ independent of the numeric type of the wrapped set (`N`).
...
-function *(M::AbstractMatrix{N}, X::LazySet{N}) where {N<:Real}
-    return LinearMap(M, X)
+function *(map::Union{AbstractMatrix, UniformScaling, AbstractVector, Real}, X::LazySet)
+    return LinearMap(map, X)
 end
 
-function *(M::AbstractVector{N}, X::LazySet{N}) where {N<:Real}
-    return LinearMap(reshape(M, length(M), 1), X)
+# convenience constructor from a vector
+function LinearMap(v::AbstractVector, X::LazySet)
```
mforets: we can keep it, but tthinking again of this constructor, it is admittedly rather strange given that `v` is a *vector* so the interpretation of `vX` could vary; i dont remember why we added it in the first place.

schillic: If you do not like it anymore, I suggest to remove it later to keep the changes in this PR limited.

PR #1632, opened 2019-08-23
#1014 - Add parallelotope in constraint form

### LINKED ISSUE #1014: Add parallelotope
For background and use cases see:

- https://people.eecs.berkeley.edu/~tommasodreossi/papers/hscc2016.pdf [broken link]
- https://arxiv.org/pdf/1607.02200.pdf
- https://mediatum.ub.tum.de/doc/1287515/893367.pdf

SebastianGuadalupe:
Closes #1014.


schillic:
The build currently breaks because you do not implement the support vector for `Parallelotope`, which is a requirement for any `LazySet`. If you wait for #1635, you will get all the (potentially suboptimal) implementations for general zonotopes for free. Of course if you know a more efficient implementation, you should still either add it or open an issue.

schillic:
> If you wait for #1635, you will get all the (potentially suboptimal) implementations for general zonotopes for free.

This is now available in `master`, so you can `rebase` this branch.

src/Parallelotope.jl
```diff
@@ -0,0 +1,125 @@
...
+Type that represents a parallelotope.
+
+### Fields
+
+- `offset`     -- vector; each element is the offset of the corresponding constraint
+- `directions` -- matrix ``D``; each row is the direction of two corresponding
+                  constraints
+
+### Notes
+``D_{i} = -D_{i+n}`` for ``i = 1, ..., n`` with ``n`` rows in ``D``
```
schillic:
```suggestion

``D_{i} = -D_{i+n}`` for ``i = 1, ..., n`` with ``n`` rows in ``D``
```

src/Parallelotope.jl
```diff
@@ -0,0 +1,125 @@
...
+    # default constructor with dimension check
+    function Parallelotope{N, VN, MN}(o::VN, D::MN, c::VN) where {N<:Real,
+         VN<:AbstractVector{N}, MN<:AbstractMatrix{N}}
+        @assert length(c) == checksquare(D)
+        return new{N, VN, MN}(o, D, c)
+    end
+end
+
+# convenience constructor without type parameter
+Parallelotope(o::VN, D::MN, c::VN) where {N<:Real,
```
schillic: I think these constructors are a relic of previous Julia versions in this repository. You should be able to remove the type parameters from the inner constructor and then remove this constructor here.

src/Parallelotope.jl
schillic: Can you add some explanation or a reference? I do not understand how this set representation works.

src/Parallelotope.jl
```diff
@@ -0,0 +1,125 @@
...
+
+### Input
+
+- `P` -- parallelotope
+
+### Output
+
+The center of the parallelotope.
+"""
+function center(P::Parallelotope{N})::Vector{N} where {N<:Real}
```
schillic:
```suggestion
function center(P::Parallelotope{N}) where {N<:Real}
```
(same in the documentations and the other getters)

src/Parallelotope.jl
```diff
@@ -0,0 +1,125 @@
...
+A matrix with the directions of the parallelotope.
+"""
+function directions(P::Parallelotope{N}) where {N<:Real}
+    return P.directions
+end
+
+"""
+    genmat(P::Parallelotope{N}) where {N<:Real}
+
+Return the generator wit matrix of a parallelotope.
```
schillic:
```suggestion
Return the generator matrix of a parallelotope.
```

src/Parallelotope.jl
```diff
@@ -0,0 +1,125 @@
...
+
+### Input
+
+- `P` -- parallelotope
+
+### Output
+
+A matrix where each row represents one generator of the parallelotope `P`.
```
schillic:
```suggestion
A matrix where each column represents one generator of the parallelotope `P`.
```

src/Parallelotope.jl
schillic: You can also keep them for now since you already wrote it. I created #1657 to remove them later again if we want to.

src/Parallelotope.jl
```diff
@@ -0,0 +1,121 @@
...
+    center::VN
+    offset::VN
+    directions::MN
+
+    # default constructor with dimension check
+    function Parallelotope(c::VN, D::MN, o::VN) where {N<:Real,
+             VN<:AbstractVector{N}, MN<:AbstractMatrix{N}}
+        @assert length(c) == checksquare(D)
+        @assert length(c) == length(o)
+        return new(c, D, o)
```
schillic:
```suggestion
        return new{VN, MN, VM}(c, D, o)
```
`new` still needs the type parameters.

schillic: Oops, they were completely wrong, sorry :blush:

src/Parallelotope.jl
```diff
@@ -0,0 +1,133 @@
...
+- `directions` -- matrix ``D``; each row is the direction of two corresponding
+                  constraints
+- `offset`     -- vector; each element is the offset of the corresponding constraint
+
+### Notes
+
+Let ``D ∈ \\mathbb{R}^n × n`` be a matrix and let ``o ∈ \\mathbb{R}^n`` be an
+offset vector and ``c ∈ \\mathbb{R}^n`` be a vector corresponding to the
+center. The parallelotope ``P ⊂ \\mathbb{R}^n`` generated by ``D`` and ``o``
+with center in ``c`` is given by the following halfspaces:
```
schillic:
```suggestion
with center in ``c`` is given by the following halfspaces (for ``i = 1, …, n``):
```

src/Parallelotope.jl
```diff
@@ -0,0 +1,133 @@
...
+```math
+    D_{i} . x + c_{i} = o_{i} and
```
schillic:
```suggestion
    D_{i} . x + c_{i} = o_{i} \\text{ and}
```

schillic:
Remaining tasks here:
* get the latest changes from `master`
* add type/methods to docs ([somewhere before here](https://github.com/JuliaReach/LazySets.jl/blob/master/docs/src/lib/representations.md#polygons))
* fix row/column behavior of `genmat`

mforets:
In fact we began to discuss (offline) with @SebastianGuadalupe if it would be interesting to cache the generator representation of the parallelotope. I'm not convinced with the current implementation since it has to compute the genmat for each query of support function / vector.

schillic:
> it has to compute the genmat for each query of support function / vector

Is there no better way to compute them than using the zonotope fallback? Anyway, performance can be improved later.

mforets:
After online discussion with @SebastianGuadalupe, we came up with the idea of defining 3 new types: `HParallelotope`, `GParallelotope`, and `Parallelotope` which combines (through composition) the first two. This way, one can define the parallelotope using the constraints but take advantage of the generator form for sums, etc.

schillic:
> and Parallelotope which combines (through composition) the first two

Like `Polytope` for `HPolytope` and `VPolytope`? We do not even have that :wink: Should we?

It is a good idea, but for simplicity I suggest to separate concerns and start with `HParallelotope` here, then `GParallelotope`, and finally the combination. The `Parallelotope` combination may not be trivial because you may need heuristics when to convert, for which you have to do benchmarks.

Should we have a `togrep` method for `AbstractZonotope`s which returns a `Zonotope`/`GParallelotope`/`Parallelotope` depending on the type? (Actually, all subtypes except for `Zonotope` are parallelotopes.)

mforets:
An illustration in 2D of the functions to convert to generator representation can be found in [this notebook](https://nbviewer.jupyter.org/github/mforets/escritoire/blob/master/sets/HParallelotope.ipynb). The 3D example is also available (but to plot the vertices in the same plot, we need something like #1756 ).

In particular, i point some of the previous items in this PR regarding caching values: it is relatively cheap to make the conversion to G-rep for the `HParallelotope` (much more than general polytopes), which justifies as a design choice to *cache* the `base_vertex` and the `extremal_vertices` in the type, such that for subsequent calls to the support function, or other functions that are easily done with the G-rep can be performed. Since `HParallelotope <: AbstractZonotope`, for instance, the `linear_map` will use the zonopic method which is fast provided that the G-rep is available.

Some other ideas for improvement are also found in that notebook.

schillic:
> The 3D example is also available (but to plot the vertices in the same plot, we need something like #1756 ).

Note that you can plot 3D points with `Plots` just fine.

mforets:
that's true, but i dont think that plays well with `plot3d` in the same plot..

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,306 @@
...
+matrix ``D`` and offset vector ``c``. We denote the *extremal vertices* of ``P``
+with respect to its base vertex ``q`` to be those vertices of ``P`` which have
+an edge in common with ``q``.  The extremal vertices can be computed as the
+solution of the ``n``-dimensional linear systems of equations ``D x = v^i``
+where for each ``i = 1, \\ldots, n``,
+``v = [-c_{n+1}, \\ldots, c_i, \\ldots, -c_{2n}]``.
+
+We refer to [1, Section 3.2.1] for details.
+
+[1] Dreossi, Tommaso, Thao Dang, and Carla Piazza. *Reachability computation for
```
schillic:
```suggestion
[1] Tommaso Dreossi, Thao Dang, and Carla Piazza. *Reachability computation for
```

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,306 @@
...
+for ``i = 1, …, n``. Here ``D_i`` represents the ``i``-th row of ``D`` and ``c_i`` the ``i``-th
+component of ``c``.
+
+For details on the notions given in these notes as well as applications of parallelotopes in
+reachability analysis we refer to [1] and [2]. For conversions between set representations
+we refer to [3].
+
+### References
+
+[1] Dreossi, Tommaso, Thao Dang, and Carla Piazza. *Reachability computation for polynomial dynamical systems.*
```
schillic:
```suggestion
[1] Tommaso Dreossi, Thao Dang, and Carla Piazza. *Reachability computation for polynomial dynamical systems.*
```

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,306 @@
...
+    Formal Methods in System Design 50.1 (2017): 1-38.
+
+[2] Dreossi, Tommaso, Thao Dang, and Carla Piazza. *Parallelotope bundles for polynomial reachability.*
```
schillic:
```suggestion
[2] Tommaso Dreossi, Thao Dang, and Carla Piazza. *Parallelotope bundles for polynomial reachability.*
```

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,306 @@
...
+    Proceedings of the 19th International Conference on Hybrid Systems: Computation and Control. ACM, 2016.
+
+[3] Althoff, Matthias, Olaf Stursberg, and Martin Buss. *Computing reachable sets of hybrid systems using
```
schillic:
```suggestion
[3] Matthias Althoff, Olaf Stursberg, and Martin Buss. *Computing reachable sets of hybrid systems using
```

test/unit_HParallelotope.jl
```diff
@@ -0,0 +1,25 @@
+for N in [Float32, Float64, Rational{Int}]
+    # Example 6 from [1].
+    #
+    # [1] Dreossi, Tommaso, Thao Dang, and Carla Piazza. *Reachability computation for polynomial dynamical systems.*
```
schillic:
```suggestion
    # [1] Tommaso Dreossi, Thao Dang, and Carla Piazza. *Reachability computation for polynomial dynamical systems.*
```

docs/src/lib/representations.md
```diff
@@ -201,6 +201,21 @@ Inherited from [`LazySet`](@ref):
 * [`radius`](@ref radius(::LazySet, ::Real))
 * [`diameter`](@ref diameter(::LazySet, ::Real))
 
+## HParallelotope
+
+```@docs
+HParallelotope
+directions(::HParallelotope{N}) where {N<:Real}
+offset(::HParallelotope{N}) where {N<:Real}
```
schillic: Hm, I am not sure we need these getters because they seem to be type specific. Is there any chance that other set types would have a method with the same name? If not, we can just access the fields.
(Of course we can keep them now that you already added them. See this just as a general comment.)

mforets:
Thanks for fixing the names in the references. I don't know why do they appear disordered; i think they were just copied from google scholar.

docs/src/lib/representations.md
mforets: It's a valid point; these getter functions are type-specific. We may not want to `export` them. If we add another parallelotope type, the getter function may have to do some computation (compare to `base_vertex`, which is for `HParallelotope` some function, but for another set  type, say `GParallelotope` it could be a getter function).

schillic: > If we add another parallelotope type, the getter function may have to do some computation

Great, then they are completely reasonable to have and I take back what I said.

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,306 @@
...
+
+- `P` -- parallelotope in constraint representation
+
+### Output
+
+The base vertex of ``P``.
+
+### Algorithm
+
+Intuitively, the base vertex is the point on which the generators are anchored.
```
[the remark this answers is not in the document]
schillic: I just saw Figure 4 in [1], which explains it. So the generators in [1] and in this documentation are not the same as the generators of our zonotopes. I find this confusing. Can we find an alternative description of the base vertex?

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,306 @@
...
+       constraints_list
+
+"""
+    HParallelotope{N<:Real, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}} <: AbstractZonotope{N}
+
+Type that represents a parallelotope in constraint form.
+
+### Fields
+
+- `directions` -- matrix such that each row is the direction of two parallel constraints
```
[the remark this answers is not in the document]
SebastianGuadalupe: we saw that the regular octagon is a valid zonotope,

```julia
plot(Zonotope([0., 0], [cos(π/8) sin(π/8); cos(3π/8) sin(3π/8); cos(π/8) -sin(π/8); cos(3π/8) -sin(3π/8)]'), ratio=1)
```
image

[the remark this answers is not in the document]
SebastianGuadalupe: in the references that we saw, parallelotopes are particular cases of zonotopes with order 1

src/Sets/HParallelotope.jl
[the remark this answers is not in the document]
SebastianGuadalupe: true, the notion of "generators" in LazySets.Zonotope and the one used in this sentence (line 144)  are not the same. for this reason, we used the terms `base_vertex` and `extremal_vertices`.

however, we agree to change the sentence in line 144 to explain a bit in more detail what does "generator" mean here.

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,307 @@
...
+- `P` -- parallelotope in constraint representation
+
+### Output
+
+The list of constraints of `P`.
+"""
+function constraints_list(P::HParallelotope{N}) where {N<:Real}
+    D, c = P.directions, P.offset
+    n = dim(P)
+    clist = Vector{LinearConstraint{N}}(undef, 2n)
```
mforets:
```suggestion
    clist = Vector{LinearConstraint{N, Vector{N}}}(undef, 2n)
```

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,306 @@
...
+The list of constraints of `P`.
+"""
+function constraints_list(P::HParallelotope{N}) where {N<:Real}
+    D, c = P.directions, P.offset
+    n = size(D, 1)
+    clist = Vector{LinearConstraint{N}}(undef, 2n)
+    @inbounds for i in 1:n
+        clist[i] = LinearConstraint(D[i, :], c[i])
+        clist[i+n] = LinearConstraint(-D[i, :], c[i+n])
+    end
```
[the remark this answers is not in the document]
mforets: I prefer to have the first chunk of n variables corresponding to the positive `D_i ⋅ x ≤ c_{i}` and the second chunk to the negative `-D_i ⋅ x ≤ c_{n+i}`, exactly as in the Notes of the new type.

src/Sets/HParallelotope.jl
```diff
@@ -298,7 +298,7 @@ The list of constraints of `P`.
 function constraints_list(P::HParallelotope{N}) where {N<:Real}
     D, c = P.directions, P.offset
     n = dim(P)
-    clist = Vector{LinearConstraint{N}}(undef, 2n)
+    clist = Vector{LinearConstraint{N, Vector{N}}}(undef, 2n)
```
schillic: @mforets: I think you need to use the vector type of `H` here instead.

mforets: I don't know how to get the vector type of `MN`.

schillic: Ah, I see the problem. Still, it is not nice that a sparse matrix gets converted to dense vectors. Maybe just do `typeof(D[1, :])`?

mforets: Yes, i thought about this in the past but discarded the idea. If you also think it's fine, i'll add it.

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,311 @@
...
+[1] Tommaso Dreossi, Thao Dang, and Carla Piazza. *Reachability computation for polynomial dynamical systems.*
+    Formal Methods in System Design 50.1 (2017): 1-38.
+
+[2] Tommaso Dreossi, Thao Dang, and Carla Piazza. *Parallelotope bundles for polynomial reachability.*
+    Proceedings of the 19th International Conference on Hybrid Systems: Computation and Control. ACM, 2016.
+
+[3] Matthias Althoff, Olaf Stursberg, and Martin Buss. *Computing reachable sets of hybrid systems using
+    a combination of zonotopes and polytopes.* Nonlinear analysis: hybrid systems 4.2 (2010): 233-249.
+"""
+struct HParallelotope{N<:Real, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}} <: AbstractZonotope{N}
```
mforets, line 56: i would suggest to remove the type annotation `N<:Real` here. (we do have this kind of restriction the LazySets but this new `HParallelotope` type should not force using `Real`). @schillic  do you agree?

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,311 @@
...
+The base vertex can be computed as the solution of the ``n``-dimensional linear
+system ``D_i x = c_{n+i}`` for ``i = 1, \\ldots, n``, see [1, Section 3.2.1].
+
+[1] Dreossi, Tommaso, Thao Dang, and Carla Piazza. *Reachability computation for
+    polynomial dynamical systems.* Formal Methods in System Design 50.1 (2017): 1-38.
+"""
+function base_vertex(P::HParallelotope{N}) where {N<:Real}
+    D, c = P.directions, P.offset
+    n = dim(P)
+    v = to_negative_vector(c[n+1:end]) # converts to a dense vector as well
```
mforets, line 158: can we use a view here? `view(c, (n+1):length(c))`

src/Sets/HParallelotope.jl
schillic: Yes, sure.

src/Sets/HParallelotope.jl
SebastianGuadalupe: Yes, it improved the performance a little bit :)

docs/src/lib/sets/HParallelotope.md
```diff
@@ -0,0 +1,44 @@
...
+HParallelotope
+directions(::HParallelotope)
+offset(::HParallelotope)
+dim(::HParallelotope)
+base_vertex(::HParallelotope)
+extremal_vertices(::HParallelotope)
+center(::HParallelotope)
+genmat(::HParallelotope)
+generators(::HParallelotope)
+constraints_list(::HParallelotope)
```
mforets, line 17:
```suggestion
constraints_list(::HParallelotope{N}) where {N}
```

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,311 @@
...
+### Output
+
+An iterator over the generators of `P`.
+"""
+function generators(P::HParallelotope)
+    return generators_fallback(P)
+end
+
+"""
+    constraints_list(P::HParallelotope)
```
mforets, line 286:
```suggestion
    constraints_list(P::HParallelotope{N}) where {N}
```

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,311 @@
...
+
+### Input
+
+- `P` -- parallelotope in constraint representation
+
+### Output
+
+The list of constraints of `P`.
+"""
+function constraints_list(P::HParallelotope{N}) where N
```
mforets, line 298:
```suggestion
function constraints_list(P::HParallelotope{N}) where {N}
```

src/Sets/HParallelotope.jl
```diff
@@ -0,0 +1,311 @@
...
+
+We refer to [1, Section 3.2.1] for details.
+
+[1] Tommaso Dreossi, Thao Dang, and Carla Piazza. *Reachability computation for
+    polynomial dynamical systems.* Formal Methods in System Design 50.1 (2017): 1-38.
+"""
+function extremal_vertices(P::HParallelotope)
+    D, c = P.directions, P.offset
+    n = dim(P)
+    v = to_negative_vector(c[n+1:end])
```
mforets, line 194:
```suggestion
    v = to_negative_vector(view(c, n+1, length(c))
```

docs/src/lib/sets/HParallelotope.md
```diff
@@ -0,0 +1,44 @@
...
+
+# [HParallelotope](@id def_HParallelotope)
+
+```@docs
+HParallelotope
+directions(::HParallelotope)
+offset(::HParallelotope)
+dim(::HParallelotope)
+base_vertex(::HParallelotope)
+extremal_vertices(::HParallelotope)
```
mforets, line 13:
```suggestion
extremal_vertices(::HParallelotope{N}) where {N}
```

PR #1637, opened 2019-08-25
#1636 - Concrete convex hull of polytopes

### LINKED ISSUE #1636: Concrete convex hull of polytopes

schillic:
Closes #1636.

### DIFF
--- a/src/concrete_convex_hull.jl
+++ b/src/concrete_convex_hull.jl
@@ -8,7 +8,7 @@ end
 
 """
     convex_hull(points::Vector{VN};
-                [algorithm]=default_convex_hull_algorithm(points),
+                [algorithm]=nothing,
                 [backend]=nothing,
                 [solver]=nothing
                 )::Vector{VN} where {N<:Real, VN<:AbstractVector{N}}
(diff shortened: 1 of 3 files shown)


src/concrete_convex_hull.jl
```diff
@@ -6,9 +6,58 @@ function default_convex_hull_algorithm(points)
     end
 end
 
+"""
+    function convex_hull(X::LazySet{N}, Y::LazySet{N}; [algorithm]=nothing,
```
mforets:
```suggestion
    convex_hull(X::LazySet{N}, Y::LazySet{N}; [algorithm]=nothing,
```

src/concrete_convex_hull.jl
```diff
@@ -6,9 +6,58 @@ function default_convex_hull_algorithm(points)
...
+function convex_hull(X::LazySet{N}, Y::LazySet{N};
+                     algorithm=nothing,
+                     backend=nothing,
+                     solver=nothing
+                    ) where {N<:Real}
+    n = dim(X)
+    @assert n == dim(Y) "the convex hull requires two sets of the same " *
+                        "dimension, but the sets had dimension $n and $(dim(Y))"
+    if n == 1
+        return overapproximate(ConvexHull(X, Y), Interval)
```
mforets: There exists a special handling of the one-dimensional case. I would rather keep the line `chull` below, and add a case `if n == 1` to return an `Interval`.

schillic: Hm, not sure. For lazy sets I would expect that the support function is much more efficient than `vertices_list`.

mforets: Yes, it will be more efficient in general. But this ultimately depends on the implementation of overapproximate for given types. For example, it may as well return an interval that is strictly bigger than the concrete convex hull, and this is not the expected answer for `convex_hull`, that should be exact.

schillic: Fine, I changed it and created #1638.

PR #1673, opened 2019-09-04
#1661 - Convex hull in 2D removes non-redundant vertices

### LINKED ISSUE #1661: Convex hull in 2D removes non-redundant vertices
The output depends on the ordering of the input vertices.
```julia
julia> V = [[0., 0.], [0., 1.], [0., 2.], [1., 1.]];

julia> convex_hull(V)  # correct
 [1.0, 1.0]
 [0.0, 2.0]
 [0.0, 0.0]

julia> V2 = [[1., 1.], [0., 0.], [0., 1.], [0., 2.]];  # same as V but in different order

julia> convex_hull(V2)  # incorrect
 [0.0, 1.0]
 [1.0, 1.0]
```

SebastianGuadalupe:
Closes #1661.

### DIFF
--- a/src/Utils/comparisons.jl
+++ b/src/Utils/comparisons.jl
@@ -424,7 +424,7 @@ The index of the minimum and maximum of the three inputed numbers.
 ### Examples
 
 ```jldoctest
-julia> arg_minmax(1.4, 52.4, -5.2)
+julia> LazySets.arg_minmax(1.4, 52.4, -5.2)
 (3, 2)
 ```
 """
--- a/src/Utils/comparisons.jl
+++ b/src/Utils/comparisons.jl
@@ -369,3 +369,77 @@ function _leq(x::N, y::N;
               atol::Real=_atol(N)) where {N<:AbstractFloat}
     return x <= y || _isapprox(x, y, rtol=rtol, ztol=ztol, atol=atol)
 end
+
+"""
+    minmax(A, B, C)
+
+Compute the minimum and maximum of three numbers A, B, C.
+
+### Input
+
+- `A` -- first number
+- `B` -- second number
+- `C` -- third number
+
+### Output
+
+The minimum and maximum of the three inputed numbers.
+
+### Examples
+
+```jldoctest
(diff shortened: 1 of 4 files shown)


src/Utils/comparisons.jl
```diff
@@ -369,3 +369,77 @@ function _leq(x::N, y::N;
...
+
+### Input
+
+- `A` -- first number
+- `B` -- second number
+- `C` -- third number
+
+### Output
+
+The minimum and maximum of the three inputed numbers.
```
mforets:
```suggestion
The minimum and maximum of the three given numbers.
```

src/Utils/comparisons.jl
```diff
@@ -369,3 +369,77 @@ function _leq(x::N, y::N;
...
+
+### Input
+
+- `A` -- first number
+- `B` -- second number
+- `C` -- third number
+
+### Output
+
+The index of the minimum and maximum of the three inputed numbers.
```
mforets:
```suggestion
The index of the minimum and maximum of the three given numbers.
```

schillic:
I actually mainly preferred this change, which is still missing:

> and then share the common code, either with another helper function or by restructuring the current function.

It could look like this:

```julia
points[1] = _get_i(min_x, A, B, C)
points[2] = _get_i(max_x, A, B, C)
points[1] = _get_i(min_y, A, B, C)
points[2] = _get_i(max_y, A, B, C)

function _get_i(i, A, B, C)
    return i == 1 ? A : i == 2 ? B : C
end
```

mforets:
fwiw, we kept the new `arg_minmax` approach because it's much faster:

```julia
a = 1.4
b = 52.4
c = -5.2

@btime argmin([$a, $b, $c])
@btime argmin([$a, $b, $c])
  39.240 ns (1 allocation: 112 bytes)
  38.031 ns (1 allocation: 112 bytes)

@btime arg_minmax($a, $b, $c)
  0.031 ns (0 allocations: 0 bytes)

(3, 2)
```

PR #1677, opened 2019-09-08
#1672 - Overapproximation of ResetMap by CartesianProductArray

### LINKED ISSUE #1672: Overapproximation of ResetMap by CartesianProductArray
We only need a reset to zero for now.

schillic:
Closes #1672.

This is a quick implementation. In general it would be preferable to have an implementation for #1113.

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -1118,17 +1118,17 @@ function _overapproximate_lm_cpa!(arr, M, cpa, overapprox_option)
 end
 
 """
-    function overapproximate(rm::ResetMap{N, <:CartesianProductArray{N}},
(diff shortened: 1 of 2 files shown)


mforets:
If the reset map contains non-zeros, can that be accounted by defining an affine map `am` through `get_b`? And then a call to `overapproximate(am, CartesianProductArray, oa)`? (Maybe the latter is not implemented, i haven't checked.)

schillic:
CC @kostakoida

schillic:
> If the reset map contains non-zeros, can that be accounted by defining an affine map `am` through `get_b`? And then a call to `overapproximate(am, CartesianProductArray, oa)`? (Maybe the latter is not implemented, i haven't checked.)

Yes, that can be done (that is why I linked to #1113); no, it is not implemented.

PR #1689, opened 2019-09-14
Add concrete reflection operation

ueliwechsler:
Closes #1559.

Not 100% sure, if overloading `-` operator is a good idea, it could possibly add ambiguities. For example, if we would implement a lazy Minkowski Difference (overapproximation) functionality.

Issue #622, opened 2018-09-12
Bloating set operation


schillic:
Add a lazy set type for bloating in a given norm, ~possibly by a negative coefficient~.
(EDIT: Negative coefficients are theoretically challenging, in particular (but not only) if the set is not big enough; typically we are interested in positive coefficients anyway.)

The implementation of the support vector is quite simple: Just compute the support vector for the original set and add the scaled direction on top (scaled by the bloating coefficient).

PR #1690, opened 2019-09-14
Add interior function

### LINKED ISSUE #126: Optional tolerance argument for membership check

ueliwechsler:
Closes #126.

Functionality that is used in some algorithms to check if a point is contained in the interior of a set.
Often used in the context of algorithms where we have to ensure that the origin is contained in the disturbance set or constraint set.


schillic:
* (This branch is based on #1689. You probably created this branch when you were on the other branch. Please avoid this if not necessary to make reading the changes easier. You can undo it by doing `git rebase master` and then force-pushing.)
* See #126 for the corresponding issue.
* I do not like the name `interior` because it does not sound like a check. Maybe `in_interior`?
* Alternatively we could have a lazy wrapper around a set to enlarge/shrink it (see #622) which, if the coefficient is negative, uses your `interior` implementation for `in`.
* I would make the implementation use a `Ballp` and an option to control the `p`-norm. This automatically creates a `BallInf` if you pass `p=Inf`, but it can also be used with `p=2` for a `Ball2`, which may be more useful in some contexts.

ueliwechsler:
> * (This branch is based on #1689. You probably created this branch when you were on the other branch. Please avoid this if not necessary to make reading the changes easier. You can undo it by doing `git rebase master` and then force-pushing.)

Yep, sorry for that.

> * I do not like the name `interior` because it does not sound like a check. Maybe `in_interior`?

`in_interior` would be fine for me. But maybe `is_interior_point`?

> * I would make the implementation use a `Ballp` and an option to control the `p`-norm. This automatically creates a `BallInf` if you pass `p=Inf`, but it can also be used with `p=2` for a `Ball2`, which may be more useful in some contexts.

👍

> * See #126 for the corresponding issue.
> * Alternatively we could have a lazy wrapper around a set to enlarge/shrink it (see #622) which, if the coefficient is negative, uses your `interior` implementation for `in`.

I don't totally understand what you mean by these two comments.

mforets:
+1 for `is_interior_point`

schillic:
> See #126 for the corresponding issue.

This means that I think this PR closes that issue :wink:

> Alternatively we could have a lazy wrapper around a set to enlarge/shrink it (see #622) [...]

You can safely ignore this point :wink:

src/ConcreteOperations/interior.jl
```diff
@@ -0,0 +1,28 @@
...
+
+
+"""
+    is_interior_point(d::AbstractVector, P::LazySet; p=Inf, ε=_TOL_F64.rtol)
+
+Check if the point `d` is contained in the interior of the convex set `P`.
+
+### Input
+
+- `d`  -- coordinate of point
```
schillic:
```suggestion
- `d`  -- point
```

src/ConcreteOperations/interior.jl
```diff
@@ -0,0 +1,28 @@
...
+- `P`  -- lazy set
```
schillic:
```suggestion
- `P`  -- set
```

src/ConcreteOperations/interior.jl
```diff
@@ -0,0 +1,28 @@
...
+- `p`  -- norm of the Ball used to apply the error tolerance
```
schillic:
```suggestion
- `p`  -- norm of the ball used to apply the error tolerance
```

docs/src/lib/operations.md
```diff
@@ -449,3 +449,9 @@ to_union_of_projections(::Rectification{N}, ::Bool=false) where {N<:Real}
 ```@docs
 RectificationCache
 ```
+
+## Interior
+
+```@docs
+is_interior_point(d::AbstractVector, P::LazySet; ε=_TOL_F64.rtol)
```
schillic:
```suggestion
is_interior_point(::AbstractVector, ::LazySet; p=Inf, ε=_TOL_F64.rtol)
```

src/ConcreteOperations/interior.jl
```diff
@@ -0,0 +1,28 @@
...
+- `ε`  -- error tolerance of check
+
+### Output
+
+Boolean which indicates if point `d` is contained in `P`.
+
+### Algorithm
+
+The implementation checks if a `BallInf` with center `d` and radius `ε` is
```
schillic:
```suggestion
The implementation checks if a `Ballp` of norm `p` with center `d` and radius `ε` is
```

src/ConcreteOperations/interior.jl
```diff
@@ -0,0 +1,28 @@
...
+contained in the set `P` which results in a numerical check for `d ∈ interior(P)`
```
schillic:
```suggestion
contained in the set `P`. This is a numerical check for `d ∈ interior(P)`
```

src/ConcreteOperations/interior.jl
```diff
@@ -0,0 +1,28 @@
...
+with error tolerance `ε` with default value `_TOL_F64.rtol`.
```
schillic:
```suggestion
with error tolerance `ε`.
```
I would not talk about default values in the documentation.

test/unit_interior.jl
```diff
@@ -0,0 +1,6 @@
+P = BallInf(zeros(2), 0.15)
+d = [0.15,0.15]
+@test !is_interior_point(d, P)
+@test !is_interior_point(d, P; p=2.0)
+d = [0.15,0.15] .- LazySets._TOL_F64.rtol
+@test is_interior_point(d, P)
```
schillic: Can you add a positive and a negative test for modifying `p` resp. `ε`?
```julia
@test is_interior_point(d, P; p=2)
d = [0.15, 0.15]
@test !is_interior_point(d, P; ε=1.0)
@test is_interior_point(d, P; ε=0.0)
```

src/ConcreteOperations/interior.jl
```diff
@@ -0,0 +1,28 @@
+export is_interior_point
```
schillic: You can just put this code in [`LazySet.jl`](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Interfaces/LazySet.jl).

src/ConcreteOperations/interior.jl
```diff
@@ -0,0 +1,28 @@
...
+
+Boolean which indicates if point `d` is contained in `P`.
+
+### Algorithm
+
+The implementation checks if a `BallInf` with center `d` and radius `ε` is
+contained in the set `P` which results in a numerical check for `d ∈ interior(P)`
+with error tolerance `ε` with default value `_TOL_F64.rtol`.
+"""
+function is_interior_point(d::AbstractVector, P::LazySet; p=Inf, ε=_TOL_F64.rtol)
```
schillic: I think you should make the default for `ε` depend on the numeric type in `d`/`P`. 
```suggestion
function is_interior_point(d::AbstractVector{N}, P::LazySet{N}; p=Inf, ε=_rtol(N)) where {N<:Real}
```

schillic: I am also not sure whether this parameter should be passed by the user or whether the user should rather modify `_rtol`.

mforets: I would keep this, as both are good IMO: either pass it for single-use or modify it globally for longer scripts or consistency through a given application that uses different functions that rely on `comparisons.jl`.

schillic: (Taking this change of course implies that you adapt the docstring and the reference in `docs/src/lib/operations.md`.)

docs/src/lib/operations.md
```diff
@@ -449,3 +449,9 @@ to_union_of_projections(::Rectification{N}, ::Bool=false) where {N<:Real}
 ```@docs
 RectificationCache
 ```
+
+## Interior
+
+```@docs
+is_interior_point(::AbstractVector, ::LazySet; p=Inf, ε=_TOL_F64.rtol)
```
mforets: To fix the [failed docs build](https://travis-ci.org/JuliaReach/LazySets.jl/jobs/584984815), try without specifying the optional keyword arguments, that is:

```julia
is_interior_point(::AbstractVector, ::LazySet)
```

src/ConcreteOperations/interior.jl
```diff
@@ -0,0 +1,28 @@
+export is_interior_point
+
+
+"""
+    is_interior_point(d::AbstractVector, P::LazySet; p=Inf, ε=_TOL_F64.rtol)
```
mforets:
```suggestion
    is_interior_point(d::AbstractVector, P::LazySet; p=Inf, ε=_rtol(N))
```

docs/src/lib/operations.md
schillic: No, the problem is that by adding the type parameter, the signature has changed (see [my comment here](https://github.com/JuliaReach/LazySets.jl/pull/1690#discussion_r324427969)).

src/ConcreteOperations/interior.jl
schillic: Same thing: copy the whole signature here.

docs/src/lib/operations.md
mforets: Either way is good. AFAIK optional arguments are *not* mandatory in the `@docs`. If this has changed we have to review [this section](https://juliareach.github.io/JuliaReachDevDocs/latest/guidelines/#Extending-the-online-documentation-1).

test/unit_interior.jl
ueliwechsler: `Ballp` does only work if `p` has the same type as the `center` and the `radius`. 
So your first test with `p=2` does not work.
Is there a specific reason for that?

schillic: >Ballp does only work if p has the same type as the center and the radius.

Okay, makes sense.

> Is there a specific reason for that?

No, I just forgot that it needs to be a float.

src/ConcreteOperations/interior.jl
schillic: Ah, sorry, that was ambiguous. I meant the whole function.

test/unit_interior.jl
```diff
@@ -0,0 +1,10 @@
+P = BallInf(zeros(2), 0.15)
```
schillic: I would prefer if we had tests with `Float32` and `Rational{Int}` as well.

schillic:
@ueliwechsler: I could make my requested changes and fix the conflicts if you don't mind.

Issue #1505, opened 2019-07-06
Convert AbstractPolyhedron to Polyhedra.polyhedron


schillic:
Currently we can convert `HPolyhedron`, `HPolytope`, and `VPolytope` to `Polyhedra.polyhedron`. Since we can convert all `AbstractPolyhedron` types to `HPolyhedron` (and any `AbstractPolytope` type to all three of them), we can in principle also convert those types, but there is no elegant way.

#### Proposal
One of two options:
1. Make `polyhedron` a required interface method for `AbstractPolyhedron` types.
2. Since there are only two possible implementations (calling `vertices_list` or `constraints_list`), the only information we need is which of those two implementations is preferred. For example, an `AbstractHyperrectangle` would prefer `constraints_list`. This could be indicated by yet another helper method `is_hrep_preferred`, which might be helpful in other situations as well. Then the implementation is:
```julia
function polyhedron(P::AbstractPolyhedron)
    if is_hrep_preferred(P)
        return polyhedron(HPolyhedron(constraints_list(P)))
    else
        return polyhedron(VPolytope(vertices_list(P)))
    end
end
function is_hrep_preferred(P::AbstractPolyhedron)
    # default implementation
    return true
end

function is_hrep_preferred(P::VPolygon)
    return false
end
...
```

This is one open task in #1078.

mforets:
FWIW the following two work: `polyhedron(convert(HPolytope, P))` or `polyhedron(convert(VPolytope, P))`, and i think this is simple enough to qualify as elegant :)

```julia
julia> using LazySets, Polyhedra
[ Info: Recompiling stale cache file C:\Users\marce\.julia\compiled\v1.2\Polyhedra\17i4E.ji for Polyhedra [67491407-f73d-577b-9b50-8179a7c68029]

julia> P = rand(Hyperrectangle)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([0.9714024861824725, -1.5203513209265647], [1.0954915952635476, 0.7449180042707249])

julia> convert(HPolytope, P) |> polyhedron
Polyhedron DefaultPolyhedron{Float64,MixedMatHRep{Float64,Array{Float64,2}},MixedMatVRep{Float64,Array{Float64,2}}}:
4-element iterator of Polyhedra.HalfSpace{Float64,Array{Float64,1}}:
 HalfSpace([1.0, 0.0], 2.06689408144602)
[...]

PR #1696, opened 2019-09-15
#1110 - Add Chebyshev center

### LINKED ISSUE #1110: Add Chebyshev center
_Originally posted by @schillic in https://github.com/JuliaReach/LazySets.jl/pull/1108#issuecomment-462695764_

schillic:
Closes #1110.

### DIFF
--- a/src/Interfaces/AbstractPolyhedron_functions.jl
+++ b/src/Interfaces/AbstractPolyhedron_functions.jl
@@ -4,7 +4,8 @@ export constrained_dimensions,
        tosimplehrep,
        remove_redundant_constraints,
        remove_redundant_constraints!,
-       linear_map
+       linear_map,
+       chebyshev_center
 
 # default LP solver for floating-point numbers
 function default_lp_solver(N::Type{<:AbstractFloat})
@@ -430,4 +431,53 @@ function plot_recipe(P::AbstractPolyhedron{N}, ε::N=zero(N)) where {N<:Real}
 
         return x, y
     end
+end
+
+"""
+    chebyshev_center(P::AbstractPolyhedron{N};
+                     [get_radius]::Bool=false,
+                     [backend]=default_polyhedra_backend(P, N),
+                     [solver]=JuMP.with_optimizer(GLPK.Optimizer)
+                     ) where {N<:AbstractFloat}
+
+Compute the [Chebyshev center](https://en.wikipedia.org/wiki/Chebyshev_center)
+of a polytope.
+
+### Input
+
+- `P`          -- polytope
+- `get_radius` -- (optional; default: `false`) option to additionally return the
+                  radius of the largest ball enclosed by `P` around the
+                  Chebyshev center
+- `backend`    -- (optional; default: `default_polyhedra_backend(P, N)`) the
+                  backend for polyhedral computations
+- `solver`     -- (optional; default: `JuMP.with_optimizer(GLPK.Optimizer)`) the
+                  LP solver passed to `Polyhedra`
+
+### Output
+
+If `get_radius` is `false`, the result is the Chebyshev center of `P`.
+If `get_radius` is `true`, the result is the pair `(c, r)` where `c` is the
+Chebyshev center of `P` and `r` is the radius of the largest ball with center
+`c` enclosed by `P`.
+
+### Notes
+
+The Chebyshev center is the center of a largest Euclidean ball enclosed by `P`.
+In general, the center of such a ball is not unique (but the radius is).
+"""
+function chebyshev_center(P::AbstractPolyhedron{N};
+                          get_radius::Bool=false,
+                          backend=default_polyhedra_backend(P, N),
+                          solver=JuMP.with_optimizer(GLPK.Optimizer)
+                         ) where {N<:AbstractFloat}
+    require(:Polyhedra; fun_name="chebyshev_center")
+    # convert to HPolyhedron to ensure `polyhedron` is applicable (see #1505)
+    Q = polyhedron(convert(HPolyhedron, P); backend=backend)
+    c, r = Polyhedra.chebyshevcenter(Q, solver)
+
+    if get_radius
+        return c, r
+    end
+    return c
 end
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -123,6 +123,7 @@ This interface defines the following functions:
 ∈(::AbstractVector{N}, ::AbstractPolyhedron{N}) where {N<:Real}
 constrained_dimensions(::AbstractPolyhedron)
 linear_map(::AbstractMatrix{N}, ::AbstractPolyhedron{N}) where {N<:Real}
+chebyshev_center(::AbstractPolyhedron{N}) where {N<:AbstractFloat}
 ```
 
 Plotting (bounded) polyhedra is available, too:
--- a/docs/src/lib/representations.md
+++ b/docs/src/lib/representations.md
@@ -23,6 +23,7 @@ center(::Ball2{N}) where {N<:AbstractFloat}
 rand(::Type{Ball2})
 sample(::Ball2{N}, ::Int) where {N<:AbstractFloat}
 translate(::Ball2{N}, ::AbstractVector{N}) where {N<:AbstractFloat}
+chebyshev_center(::Ball2{N}) where {N<:AbstractFloat}
 ```
 Inherited from [`LazySet`](@ref):
 * [`norm`](@ref norm(::LazySet, ::Real))
--- a/src/Sets/Ball2.jl
+++ b/src/Sets/Ball2.jl
@@ -277,3 +277,40 @@ function sample(B::Ball2{N}, nsamples::Int=1;
     end
     return D
 end
+
+
+# --- Ball2 functions ---
+
+
+"""
+    chebyshev_center(B::Ball2{N}; compute_radius::Bool=false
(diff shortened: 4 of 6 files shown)


src/Interfaces/AbstractPolytope.jl
```diff
@@ -96,6 +97,60 @@ function isempty(P::AbstractPolytope)::Bool
...
+The Chebyshev center is the center of a largest Euclidean ball enclosed by `P`.
+In general, the center of such a ball is not unique (but the radius is).
+
+Formally, letting ``∂P`` denote the boundary of ``P``, the Chebyshev center is
+
+```math
+    \\arg\\min_c \\min_{x ∈ ∂P} ‖x - c‖₂.
+```
+"""
+function chebyshev_center(P::AbstractPolytope{N};
```
schillic: * I wanted to make the implementation available for general polytopes but we need #1505 for that. I think it is still better to have this general method and a nice error message than having two identical methods for `HPolytope` and `VPolytope` and no good error message for other set types (and the generalization is very simple later).
* Mathematically the method could actually be generalized to `AbstractPolyhedron` but I find the concept a bit weird for unbounded sets.

src/Interfaces/AbstractPolytope.jl
```diff
@@ -96,6 +97,60 @@ function isempty(P::AbstractPolytope)::Bool
...
+
+### Notes
+
+The Chebyshev center is the center of a largest Euclidean ball enclosed by `P`.
+In general, the center of such a ball is not unique (but the radius is).
+
+Formally, letting ``∂P`` denote the boundary of ``P``, the Chebyshev center is
+
+```math
+    \\arg\\min_c \\min_{x ∈ ∂P} ‖x - c‖₂.
```
schillic: [Wikipedia](https://en.wikipedia.org/wiki/Chebyshev_center) only lists the definition for the smallest enclosing ball and the book that they refer to also does not give a good definition. Since `Polyhedra` uses the interpretation "largest enclosed ball", I came up with this definition here. Please check it!

src/Interfaces/AbstractPolytope.jl
schillic: > Mathematically the method could actually be generalized to `AbstractPolyhedron` but I find the concept a bit weird for unbounded sets.

Then again, if it does not cost anything...

src/Interfaces/AbstractPolytope.jl
```diff
@@ -96,6 +97,60 @@ function isempty(P::AbstractPolytope)::Bool
...
+```
+"""
+function chebyshev_center(P::AbstractPolytope{N};
+                          backend=default_polyhedra_backend(P, N),
+                          compute_radius::Bool=false
+                         ) where {N<:Real}
```
mforets: [this line](https://github.com/JuliaPolyhedra/Polyhedra.jl/blob/737e49135c5bbfcd1e754398ec449d57e05aec2f/src/center.jl#L10) says that it only works with `Float64` for taking the 2-norm.. i suggest that we let `where {N<:AbstractFloat}`

src/Interfaces/AbstractPolytope.jl
```diff
@@ -96,6 +97,60 @@ function isempty(P::AbstractPolytope)::Bool
...
+    require(:Polyhedra; fun_name="chebyshev_center")
+    require(:JuMP; fun_name="chebyshev_center")
+    require(:GLPK; fun_name="chebyshev_center")
```
mforets: lines 140 and 141 can be removed, as one can do:

```julia

julia> chebyshev_center(convert(HPolytope, rand(HPolygon)))
2-element Array{Float64,1}:
 -2.5065814803100483
  3.4125300460193526
```

src/Interfaces/AbstractPolytope.jl
```diff
@@ -96,6 +97,60 @@ function isempty(P::AbstractPolytope)::Bool
...
+    @assert applicable(polyhedron, P) "try converting the set to HPolytope " *
```
mforets: this seems like an unnecessary / implementation detail restriction. you can assert that the set is polyhedral, such as checking the `constraints_list` is applicable, then let `P = convert(HPolytope, P)` and pass it to Polyhedra's `chebyshevcenter`. a specialized version if the vrep is available is anyways not implemented yet (see [here](https://github.com/JuliaPolyhedra/Polyhedra.jl/blob/737e49135c5bbfcd1e754398ec449d57e05aec2f/src/center.jl#L59))

src/Interfaces/AbstractPolytope.jl
mforets: i dont understand the formula in line 132; shouldn't there be a max? or if i take x = c any point in ∂P then the norm is zero => all points in the border are minimizers.

we can write it as `max r` subject to `B(c, r) ⊆ P,  c ∈ P`. the actual optimization problem implementing this is defined [here](https://github.com/JuliaPolyhedra/Polyhedra.jl/blob/737e49135c5bbfcd1e754398ec449d57e05aec2f/src/center.jl#L10) (which we could also add as math text here in the docstring).

src/Interfaces/AbstractPolytope.jl
schillic: This is what I tried to express in https://github.com/JuliaReach/LazySets.jl/pull/1696#discussion_r324466196.

mforets: Ok. It is not strictly needed as one can first convert `P` to HPolyhedron and then apply `LazySets.polyhedron`. This was the original idea of `polyhedron`, to be a lower-level conversion from a lazy polyhedron to a concrete one from Polyhedra.

schillic: Yes, but it is a bit annoying. In particular, if your set prefers V-rep, this conversion would be bad. But as you pointed out, the V-rep is currently not supported by `Polyhedra`, so for now this seems valid.

mforets: As such, imo it's not particularly interesting to add conversions from different lazysets types to Polyhedra (https://github.com/JuliaReach/LazySets.jl/issues/1505), since they anyways have to pass through HPolytope or HPolyhedron (unless we also convert rays, lines, halfspaces directly -- but again i see no interest in doing so ATM).

schillic: > As such, imo it's not particularly interesting to add conversions from different lazysets types to Polyhedra (#1505), since they anyways have to pass through HPolytope or HPolyhedron

This is specific to this function, right? In general it can be helpful.

src/Interfaces/AbstractPolytope.jl
schillic: This is because [we load those packages with `Polyhedra`](https://github.com/JuliaReach/LazySets.jl/blob/bbda6f23830d5cd90efba23295a0a3ac9cd6a07c/src/Initialization/init_Polyhedra.jl#L6).

src/Interfaces/AbstractPolytope.jl
schillic: I will just remove the definition. The text should be sufficient.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -431,3 +432,50 @@ function plot_recipe(P::AbstractPolyhedron{N}, ε::N=zero(N)) where {N<:Real}
...
+
+### Input
+
+- `P`          -- polytope
+- `get_radius` -- (optional; default: `false`) option to additionally return the
+                  radius of the largest ball enclosed by `P` around the
+                  Chebyshev center
+- `backend`    -- (optional, default: `default_polyhedra_backend(P, N)`) the
+                  backend for polyhedral computations
+
```
mforets: the kwarg `solver` is missing in the inputs description

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -431,3 +432,50 @@ function plot_recipe(P::AbstractPolyhedron{N}, ε::N=zero(N)) where {N<:Real}
...
+### Notes
+
+The Chebyshev center is the center of a largest Euclidean ball enclosed by `P`.
+In general, the center of such a ball is not unique (but the radius is).
+"""
+function chebyshev_center(P::AbstractPolyhedron{N};
+                          get_radius::Bool=false,
+                          backend=default_polyhedra_backend(P, N),
+                          solver=JuMP.with_optimizer(GLPK.Optimizer)
```
mforets: we can use here the [default lp solver](https://github.com/JuliaReach/LazySets.jl/blob/20c26b94f65074f538011a5b1144258a0ea755b2/src/Interfaces/AbstractPolyhedron_functions.jl#L10) -- but only after doing the [change to MOI](https://github.com/JuliaReach/LazySets.jl/pull/1701/files#diff-5e796138f8958d1d7c93fd38b67eb60aR10)

schillic: I know :wink:

schillic: Actually, the answer is "no". `JuMP.with_optimizer(GLPK.Optimizer)` returns an `OptimizerFactory`:
```julia
julia> JuMP.with_optimizer(GLPK.Optimizer)
OptimizerFactory(GLPK.Optimizer, (), Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}())
```

PR #1697, opened 2019-09-15
#1105 - An element for HPolytope is not type stable

### LINKED ISSUE #1105: An element for HPolytope is not type stable

schillic:
Closes #1105.

This is a reimplementation of #1108 based on #1696 (only the last two commits are new).
Note that the same problem also occurs with `Float32`, so I also added a handling for that. It is not very nice that we need to enumerate these cases...

### DIFF
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -144,6 +144,7 @@ This interface defines the following functions:
 isbounded(::AbstractPolytope)
 singleton_list(::AbstractPolytope{N}) where {N<:Real}
 isempty(::AbstractPolytope)
+chebyshev_center(::AbstractPolytope{N}) where {N<:Real}
 ```
 
 #### Polygon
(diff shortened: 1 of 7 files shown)


schillic:
Let us first check the state after #1701 before we discuss this PR.

schillic:
Closed (see #1711).

Issue #1236, opened 2019-03-18
Method to set polyhedra backend globally


mforets:
I suggest that we also offer a method `set_polyhedra_backend(name::String)` or similar. This makes it easier to change the backend, especially from the REPL.

_Originally posted by @schillic in https://github.com/JuliaReach/LazySets.jl/issues/1235#issuecomment-473851072_

PR #1701, opened 2019-09-16
#1493 - Use MathOptInterface

### LINKED ISSUE #1493: Use MathOptInterface
https://github.com/JuliaOpt/MathProgBase.jl is deprecated in favor of https://github.com/JuliaOpt/MathOptInterface.jl

### LINKED ISSUE #1711: Remove tests that use LP solver with non-Float64 inputs
Given that the LP solvers currently expect `Float64` inputs and in particular only give `Float64` outputs, we should remove all tests that call an LP solver with other inputs.

For the same reason we should close #1105 and ignore the second point of #1619.

schillic:
Closes #1493.
~Closes #1711.~

* [x] remove `MathProgInterface` and `GLPKMathProgInterface`; load `MathOptInterface`
* [x] refactor LP creation in `remove_redundant_constraints!` (`AbstractPolyhedron_functions.jl`) (the most complicated place because LPs are created in a loop)
    * [x] create `LPModel` to not pollute the solver with old models
* [x] refactor LP creation in `∈` (`AbstractZonotope.jl`)
* [x] refactor LP creation in `σ_helper` (`HPolyhedron.jl`)
* [x] refactor LP creation in `isempty` (`HPolyhedron.jl`)
* [x] refactor LP creation in `∈` (`VPolytope.jl`)


schillic:
For some reason `MOI` does not seem to like `Rational`s...

UtkarshChemE:
Shouldn't we allow user to select other available solvers like Gurobi or CPLEX?

mforets:
@UtkarshChemE  yes, definitely. note that GLPK is just the default solver, but other solvers *can* be used if the user explicitly passes the `solver` keyword argument.  for examples see:

- https://github.com/JuliaReach/LazySets.jl/pull/1701/files#diff-03fc2d142ee3abd1afcf550eceaa7179R103

- https://github.com/JuliaReach/LazySets.jl/pull/1701/files#diff-03fc2d142ee3abd1afcf550eceaa7179R135


on a related note, we plan to add the possibility for the user to change *globally* the default solver, without having to change LazySets defaults (which are defined [here](https://github.com/JuliaReach/LazySets.jl/blob/20c26b94f65074f538011a5b1144258a0ea755b2/src/Interfaces/AbstractPolyhedron_functions.jl#L10)) once and for all the Julia session, but that would be another PR (similar in spirit to https://github.com/JuliaReach/LazySets.jl/issues/1236 which is also open)

mforets:
As per discussion, i think we should use JuMP instead of MOI.

mforets:
This branch should be rebased with master, since several commits have already been merged.

mforets:
Should we close this as stale? Moreover I think we're using the right library now.

schillic:
I keep the branch in case we want to formulate problems in MOI directly.

PR #1706, opened 2019-09-17
#1705 - Plotting of empty polyhedron crashes

### LINKED ISSUE #1705: Plotting of empty polyhedron crashes
The problem is that `plot_recipe(::EmptySet)` returns `nothing`, but we check `isempty` on the result.

schillic:
Closes #1705.


schillic:
After a discussion with @mforets, we want to try returning `[]` instead of `nothing` instead.

mforets:
> After a discussion with @mforets, we want to try returning [] instead of nothing instead.

from [this function](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Sets/EmptySet.jl#L298)

schillic:
It worked and I think it is the better solution :+1:

PR #1713, opened 2019-09-19
#1500 - Option to apply linear_map in H-rep without invertibility check

### LINKED ISSUE #1500: Option to apply linear_map in H-rep without invertibility check
Currently the `linear_map` implementation for polyhedra looks like this:
https://github.com/JuliaReach/LazySets.jl/blob/2270cc98e2ff69a1c874dde3428b52f030c84ad5/src/AbstractPolyhedron_functions.jl#L300-L317

If you want to apply the map in H-rep and already know that the matrix is invertible, you cannot exploit this fact and you need to keep `check_invertibility=true`.

schillic:
Closes #1500.

CC @ueliwechsler

### DIFF
--- a/src/Interfaces/AbstractPolyhedron_functions.jl
+++ b/src/Interfaces/AbstractPolyhedron_functions.jl
@@ -230,18 +230,19 @@ function remove_redundant_constraints(
 end
 
 """
-    linear_map(M::AbstractMatrix{N},
+    linear_map(M::Union{AbstractMatrix{N}, Nothing},
                P::AbstractPolyhedron{N};
                [algorithm]::String=(issparse(M) ? "division" : "inverse"),
                [check_invertibility]::Bool=true,
                [cond_tol]::Number=DEFAULT_COND_TOL,
+               [inverse]::Union{AbstractMatrix{N}, Nothing}=nothing
                ) where {N<:Real}
 
 Concrete linear map of a polyhedral set.
 
 ### Input
 
-- `M`         -- matrix
+- `M`         -- matrix (or `nothing` if `inverse` is specified; see below)
 - `P`         -- polyhedral set
 - `algorithm` -- (optional; default: `"division"` for sparse `M` and `"inverse"`
                  otherwise) the algorithm to be used; possible choices are:
@@ -257,8 +258,9 @@ Concrete linear map of a polyhedral set.
                            the invertibility check fails, we fall back to the
                            `"vrep"` algorithm; this option is ignored if the
                            `"vrep"` algorithm is used
-- `cond_tol`  -- (optional) tolerance of matrix condition (used to check whether
-                the matrix is invertible)
+- `cond_tol`  -- (optional; default: `DEFAULT_COND_TOL`) tolerance of matrix
+                 condition (used to check whether the matrix is invertible)
+- `inverse`   -- (optional; default: `nothing`) matrix inverse (if known)
 
 ### Output
 
@@ -302,8 +304,9 @@ polyhedron is bounded, which we check.
 
 If the matrix is known to be invertible, the the option `check_invertibility`
 can be used to skip the invertibility test.
-If the matrix inverse is even known, there is also a method that accepts the
-inverse as a keyword argument `inverse`.
+If the matrix inverse is even known, it can be specified with the option
+`inverse`, in which case we ignore the other options and also the original
+matrix `M` (which can also be `nothing` for this reason).
 
 The algorithms `"division"` and `"inverse"` give control - in case `M` is
 invertible - on whether the full matrix inverse is computed or only the left
@@ -325,10 +328,20 @@ New subtypes of the interface should define their own `_linear_map_vrep`
 (resp. `_linear_map_hrep`) for special handling of the linear map; otherwise
 the fallback implementation for `AbstractPolyhedron` is used.
 """
-function linear_map(M::AbstractMatrix{N}, P::AbstractPolyhedron{N};
+function linear_map(M::Union{AbstractMatrix{N}, Nothing},
+                    P::AbstractPolyhedron{N};
                     algorithm::String=(issparse(M) ? "division" : "inverse"),
                     check_invertibility::Bool=true,
-                    cond_tol::Number=DEFAULT_COND_TOL) where {N<:Real}
+                    cond_tol::Number=DEFAULT_COND_TOL,
+                    inverse::Union{AbstractMatrix{N}, Nothing}=nothing) where {N<:Real}
+    # if `inverse` is specified, we ignore other inputs and do not check them
+    if inverse != nothing
+        # we ignore `M` here and pass `inverse` as matrix instead
+        return _linear_map_hrep(inverse, P, true; inverse=inverse)
+    end
+
+    @assert M != nothing "a matrix must be specified unless the inverse is " *
+        "specified as well"
     @assert dim(P) == size(M, 2) "a linear map of size $(size(M)) cannot be " *
         "applied to a set of dimension $(dim(P))"
     supported_algorithms = ["vrep", "inverse", "division"]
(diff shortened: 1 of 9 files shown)


mforets:
The logic of the PR looks good to me. On the other hand, after successive extensions to the concrete linear map interface, i think it has become rather obscure... no?

I suggest to either:

(i) add more examples that illustrate the different use cases, or
(ii) make slight changes of the keyword names.

Thoughts for point (ii):

**Problem:** 

To compute `linear_map(A, X)` where `X` is a polyhedral set. There are two ways:

- vrep
    - slower
    - can always be performed if X is polyhedral
- hrep
    - faster
    - `A` needs to be invertible
        - needs to check invertibility
        - needs to compute inverse

**Keywords:**

- `use_vrep`: if the argument is true, use vrep, otherwise *try to use hrep* or fallback to vrep if one of the preconditions for using the hrep is not fullfilled
- `check_invertibility`: argument to control whether or not the invertibility check is performed. this is useful if you know that the matrix is invertible (but have not computed its inverse).
- `inverse`: argument to pass the matrix inverse, if it is known. in this case, the linear map matrix M can be ommited (`nothing`)

Additional arguments:

- `use_inv`: use `inv` or not
- `cond_tol`: tolerance for invertibility check

**Defaults:**

- `use_vrep` is `false` by default, since we want to try the hrep first (it is relatively fast to check if hrep applies or not)
- `check_invertibility` is `true` by default
- `inverse` is `nothing` by default (i dont know the inverse a priori)

Additional arguments:

- `use_inv`:
- `cond_tol`: use the constant `DEFAULT_COND_TOL`

**Pseudocode:**

```julia
function linear_map(M::Union{AbstractMatrix{N}, Nothing},
                    P::AbstractPolyhedron{N};
                    use_vrep::Bool=false,
                    check_invertibility::Bool=true,
                    inverse::Union{Nothing, AbstractMatrix{N}}=nothing,
                    cond_tol::Number=DEFAULT_COND_TOL,
                    use_inv::Bool=(inverse != nothing || !issparse(M))
                   ) where {N<:Real}
    @assert # check that dimensions match, should work if M is Nothing as well, by checking the dimensions of inverse

    # if use_vrep is activated, the rest of the arguments can be safely ignored
    use_vrep && return _linear_map_vrep(M, P)
    
    is_invertible = true
    if check_invertibility    # check if we can invert
        is_invertible = isinvertible(M; cond_tol=cond_tol)
    end

    if is_invertible
       # if we can invert => use hrep
        return _linear_map_hrep(M, P, use_inv; inverse=inverse)
    else
        # otherwise, use vrep
        return _linear_map_vrep(M, P)
    end
end
```

Comments on the pseudocode:

- there is a possible source of confusion if the user passes `linear_map(A, M, use_vrep=false)` since this does *not necessarily* use hrep: it depends on `A` being invertible. this should be documented.
- the code gives more control on the algorithm than the one in `master`
- if the inverse is known, one can pass `nothing` in the field `M`

schillic:
I also thought about this but did not want to break the current API. But since this seems okay, let me add another proposal. Since passing the inverse is a special use case, I think we can add a new method for this purpose.

```julia
function linear_map(M::Union{AbstractMatrix{N}, Nothing},
                    P::AbstractPolyhedron{N};
                    inverse::AbstractMatrix) where {N<:Real}
    return _linear_map_hrep(M, P, true; inverse=inverse)
end

function linear_map(M::AbstractMatrix{N}, P::AbstractPolyhedron{N};
                    method::Symbol=(issparse(M) ? :division : :inverse),
                    check_invertibility::Bool=true,
                    cond_tol::Number=DEFAULT_COND_TOL) where {N<:Real}
    @assert dim(P) == size(M, 2)
    @assert method in [:vrep, :inverse, :division]

    # check invertibility
    if method != :vrep && check_invertibility && !isinvertible(M; cond_tol=cond_tol)
        method = :vrep
    end

    if method == :vrep
        return _linear_map_vrep(M, P)
    else
        use_inv = method == :inverse
        return _linear_map_hrep(M, P, use_inv)
    end
end
```

schillic:
We agreed to use the version in https://github.com/JuliaReach/LazySets.jl/pull/1713#issuecomment-535833199 with strings instead of symbols.

schillic:
Unfortunately the idea with the two methods does not work. For some reason, julia does not find the method. It only works when passing `nothing` for `M`.
```julia
julia> using LazySets

julia> M = rand(2, 2)
2×2 Array{Float64,2}:
 0.0509302  0.107692
 0.0785203  0.835182

julia> linear_map(nothing, Singleton([1., 1.]); inverse=inv(M))
HPolygon{Float64}(HalfSpace{Float64,VN} where VN<:AbstractArray{Float64,1}[HalfSpace{Float64,Array{Float64,1}}([-2.3040029056097, 1.4944331500210524], 1.0), HalfSpace{Float64,Array{Float64,1}}([-24.50655542281636, 3.1599899249110077], -1.0), HalfSpace{Float64,Array{Float64,1}}([2.3040029056097, -1.4944331500210524], -1.0), HalfSpace{Float64,Array{Float64,1}}([24.50655542281636, -3.1599899249110077], 1.0)])

julia> linear_map(M, Singleton([1., 1.]); inverse=inv(M))
ERROR: MethodError: no method matching linear_map(::Array{Float64,2}, ::Singleton{Float64,Array{Float64,1}}; inverse=[24.50655542281636 -3.1599899249110077; -2.3040029056097 1.4944331500210524])
```
Should we instead assign a new name to the "inverse" function?
For instance `linear_map_inverse(M, P)`

mforets:
ok then i tend to prefer having a single `linear_map` function with several use cases. passing the inverse is just one use case, as there are others. we give the possibility of passing `nothing` if the inverse is known and you don't want to pass the matrix `M` itself (or don't want to compute it). the pseudocode i gave in comment https://github.com/JuliaReach/LazySets.jl/pull/1713#issuecomment-535665908 covers these cases.

schillic:
Finally the build passes.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -285,52 +293,69 @@ that was used:
...
                     cond_tol::Number=DEFAULT_COND_TOL,
-                    use_inv::Bool=!issparse(M)
-                   ) where {N<:Real}
+                    inverse::Union{AbstractMatrix{N}, Nothing}=nothing) where {N<:Real}
+    # if `inverse` is specified, we ignore other inputs and do not check them
+    if inverse != nothing
+        return _linear_map_hrep(M, P, true; inverse=inverse)
+    end
+
+    @assert M != nothing "a matrix must be specified unless the inverse is " *
```
mforets: we can now remove this assertion

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -285,52 +293,69 @@ that was used:
...
+        "specified as well"
     @assert dim(P) == size(M, 2) "a linear map of size $(size(M)) cannot be " *
         "applied to a set of dimension $(dim(P))"
+    supported_algorithms = ["vrep", "inverse", "division"]
```
mforets: i would remove this `supported_algorithms` check and the associated assertion altogether.

otherwise, can we add this as a const outside this function (so we dont have to allocate a `supported_algorithms` vector for each call to this function).

schillic: > i would remove this supported_algorithms definition and the associated assertion altogether.

Hm, okay.

> otherwise, can we add this as a const outside this function (so we dont have to allocate a supported_algorithms vector for each call to this function).

I had assumed that the compiler would do that for me, but I was wrong.

```julia
julia> function foo(x)
    supported_algorithms = ["vrep", "inverse", "division"]
    @assert x in supported_algorithms "supported algorithms: $supported_algorithms, but got $x"
    return true
end
foo (generic function with 1 method)

julia> @code_lowered foo(1)
CodeInfo(
1 ─      supported_algorithms = Base.vect("vrep", "inverse", "division")
│   %2 = x in supported_algorithms
└──      goto #3 if not %2
2 ─      goto #4
3 ─ %5 = Base.getproperty(Base.Main, :Base)
│   %6 = Base.getproperty(%5, :string)
│   %7 = Base.string("supported algorithms: ", supported_algorithms, ", but got ", x)
│   %8 = (%6)(%7)
│   %9 = Base.AssertionError(%8)
└──      Base.throw(%9)
4 ┄      return true
)
```

PR #1716, opened 2019-09-21
#1556 - Make box_approximation an alias

### LINKED ISSUE #1556: Make box_approximation an alias
We have some methods like `overapproximate(Z::Zonotope, ::Type{<:Hyperrectangle})`, but at the same time we have methods like `box_approximation(S::LazySet{N}))`. There should not exist two parallel ways to compute the box approximation because now `box_approximation(Z::Zonotope)` uses the fallback method. Rather, one of the two ways should be an alias for the other, as in:
```julia
box_approximation(X) = overapproximate(X, Type{Hyperrectangle})
```

schillic:
Closes #1556.

~Requires #1214.~

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -84,11 +84,21 @@ Return an approximation of a given set as a hyperrectangle.
 ### Output
 
 A hyperrectangle.
+
+### Algorithm
+
+The center of the hyperrectangle is obtained by averaging the support function
+of the given set in the canonical directions, and the lengths of the sides can
+be recovered from the distance among support functions in the same directions.
 """
 function overapproximate(S::LazySet,
                          ::Type{<:Hyperrectangle};
                         )::Union{Hyperrectangle, EmptySet}
-    return box_approximation(S)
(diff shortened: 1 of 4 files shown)


src/Approximations/overapproximate.jl
```diff
@@ -185,6 +239,49 @@ overapproximate(S::LazySet;
...
+The center and radius of the box are obtained by evaluating the support function
+of the given convex set along the canonical directions.
+"""
+function overapproximate(S::LazySet{N}, ::Type{<:BallInf}) where {N<:Real}
+    n = dim(S)
+    c = Vector{N}(undef, n)
+    r = zero(N)
+    d = zeros(N, n)
+
+    @inbounds for i in 1:n
```
mforets: side comment: we can possibly micro-optimize this loop by using `SingleEntryVector` (related to [this issue](https://github.com/JuliaReach/LazySets.jl/issues/1722) )

mforets: @dfcaporale may be able to have a look at this after he uploads his branch for https://github.com/JuliaReach/LazySets.jl/issues/1722

Issue #991, opened 2019-01-12
Add concrete and abstract traits


mforets:
Proposal: add methods to differentiate between concrete set types and interfaces, and (maybe) operations.

This functionality is useful to apply an operation on any concrete type, for example.

Note that Julia's `Base` has `isconcretetype` and `isabstracttype`, and they behave like this:
 
```julia
julia> isconcretetype(HalfSpace)
false

julia> isconcretetype(AbstractSingleton)
false

julia> isabstracttype(AbstractSingleton)
true

julia> isconcretetype(HalfSpace{Float64})
```

The function's names could be some of `isconcrete` / `isabstract` / `isconcreteset` / `isabstractset`.

schillic:
I like it. What about `isinterface`/`isabstract`, `isconcrete`, and `islazy`? The suffix `set` is okay but not needed.

mforets:
I like the name `isinterface` because it has a clear meaning in LazySets (https://juliareach.github.io/LazySets.jl/latest/lib/interfaces.html). So we would add 3 functions, `isinterface`, `isconcrete` and `islazy`.

schillic:
Note that one cannot create an object of abstract type, so the `isinterface` function seems rather useless.

mforets:
what do you think about adding `isoperation()` that returns `true` for eg. `LinearMap` and `false` for eg. `BallInf`? this would partially solve this issue.

schillic:
`isoperation` sounds good!

schillic:
Given that Julia's `isabstracttype` can be used to check for abstract types and `isoperation(type)` in #1734 can be used to check for lazy and concrete set types (by negation), I think we have covered everything here.

PR #1734, opened 2019-09-26
Add isoperation and isoperationtype

mforets:
See #991.

### DIFF
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -16,7 +16,8 @@ export LazySet,
        isuniversal,
        translate,
        affine_map,
-       is_interior_point
+       is_interior_point,
+       isoperation
 
 """
     LazySet{N}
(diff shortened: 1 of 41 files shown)


src/Interfaces/LazySet.jl
```diff
@@ -588,3 +589,36 @@ function plot_recipe(X::LazySet{N}, ε::N=N(PLOT_PRECISION)) where {N<:Real}
     end
     return plot_recipe(Y, ε)
 end
+
+"""
+    isoperation(::Type{<:LazySet})
+
+Check whether the given lazy set is an operation or not.
```
```suggestion
Check whether the given `LazySet` is an operation or not.
```

src/LazyOperations/ExponentialMap.jl
```diff
@@ -60,6 +60,8 @@ struct SparseMatrixExp{N, MN<:AbstractSparseMatrix{N}} <: AbstractMatrix{N}
     M::MN
 end
 
+isoperation(::Type{SparseMatrixExp}) = true
```
schillic: Was this on purpose? `SparseMatrixExp` is not actually a set (same for `ProjectionSparseMatrixExp`).

mforets:
> UnionSet/UnionSetArray are missing.

i only considered subtype of LazySet, but we can add these too

schillic:
> i only considered subtype of LazySet, but we can add these too

No, you also modified `Complement`.

src/LazyOperations/ExponentialMap.jl
mforets: true, this should be removed (same for the projection)

mforets:
ok, so ill add them as well

mforets:
> I would have expected that you create the method for the instances. I would definitely define the following methods:

or add `isoperation` and `isoperationtype`

src/Sets/Ball1.jl
```diff
@@ -53,6 +53,8 @@ struct Ball1{N<:Real} <: AbstractCentrallySymmetricPolytope{N}
     end
 end
 
+isoperationtype(::Type{Ball1}) = false
```
schillic: Should this not be
```suggestion
isoperationtype(::Type{<:Ball1}) = false
```
(holds for all concrete set types)
```julia
julia> typeof(rand(BallInf))
BallInf{Float64}

julia> typeof(rand(BallInf)) isa Type{BallInf}
false

julia> typeof(rand(BallInf)) isa Type{<:BallInf}
true
```

mforets: yes, this is an omission from the concrete set types (i only added it in the operations)

mforets:
Thanks for the review, im merging this.

PR #1735, opened 2019-09-26
#1729 - Faster support function/vector of CartesianProductArray and sparse direction

### LINKED ISSUE #1729: Faster support function/vector of CartesianProductArray and sparse direction
The loops below only have to be executed for iterations where `d[i0:i1]` is not zero.

https://github.com/JuliaReach/LazySets.jl/blob/d37e8ce4e0338f3bf4dcf120c58e9b48f8e5b32c/src/LazyOperations/CartesianProduct.jl#L419-L428

https://github.com/JuliaReach/LazySets.jl/blob/d37e8ce4e0338f3bf4dcf120c58e9b48f8e5b32c/src/LazyOperations/CartesianProduct.jl#L393-L402

schillic:
Closes #1729.

```julia
N = Float64
n = 100
cpa = CartesianProductArray([BallInf(zeros(N, 1), N(1)) for _ in 1:n])
dd = zeros(N, n); dd[1] = N(1); dd[3] = N(1)
ds = sparsevec([1, 3], N[1, 1], n)

@btime ρ(dd, cpa)
  4.046 μs (101 allocations: 9.39 KiB)

@btime ρ(ds, cpa)  # master
  8.558 μs (301 allocations: 18.83 KiB)

@btime ρ(ds, cpa)  # this branch
  296.284 ns (9 allocations: 656 bytes)
```

Notice that the old implementation was even considerably slower for sparse vectors than for dense vectors due to many allocations of new small sparse vectors.

### DIFF
--- a/src/LazyOperations/CartesianProduct.jl
+++ b/src/LazyOperations/CartesianProduct.jl
@@ -405,6 +405,46 @@ function σ(d::AbstractVector{N}, cpa::CartesianProductArray{N}) where {N<:Real}
     return svec
 end
 
+# faster version for sparse vectors
+function σ(d::AbstractSparseVector{N}, cpa::CartesianProductArray{N}
+          ) where {N<:Real}
+    # idea: We walk through the blocks of `cpa` (i.e., the sets `Xi`) and search
+    # for corresponding non-zero entries in `d` (stored in `indices`).
+    # `next_idx` is the next index of `indices` such that
+    # `next_dim = indices[next_idx]` lies in the next block to consider
+    # (potentially skipping some blocks).
+    svec = similar(d)
+    indices, _ = SparseArrays.findnz(d)
+    if isempty(indices)
+        # direction is the zero vector
+        return an_element(cpa)
+    end
+    next_idx = 1
+    next_dim = indices[next_idx]
+    m = length(indices)
+    i0 = 1
+    for Xi in cpa.array
+        i1 = i0 + dim(Xi) - 1
+        if next_dim <= i1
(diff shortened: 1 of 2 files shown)


mforets:
cc: @kostakoida

schillic:
According to @kostakoida there is a bug here which I have to investigate.

src/LazyOperations/CartesianProduct.jl
```diff
@@ -405,6 +405,41 @@ function σ(d::AbstractVector{N}, cpa::CartesianProductArray{N}) where {N<:Real}
...
+function σ(d::AbstractSparseVector{N}, cpa::CartesianProductArray{N}
+          ) where {N<:Real}
+    svec = similar(d)
+    indices, _ = SparseArrays.findnz(d)
+    if isempty(indices)
+        # direction is the zero vector
+        return an_element(cpa)
+    end
+    next_idx = 1
+    next_dim = indices[next_idx]
```
mforets:
```suggestion
    next_dim = indices[next_idx] # chosen such that the next _dim points to a different block than the current one
```
or some comment along these lines that helps to understand / remember how the algorithm works. this variable does not correspond to a dimension, as the name "next dimension" suggests, since is comparable to the right-end `i1`, and dimensions are `i1 - i0 + 1`, but to an index that has been identified to belong to some block (and `next_dim` are chosen such that they belong to different blocks).

mforets: .. or maybe call the variable `next_blk` for "index in the next block"

PR #1740, opened 2019-09-28
Update LazySet.jl

mforets:
Try to fix the Julia block just after "If we only consider concrete subtypes, then:", see [here](https://juliareach.github.io/LazySets.jl/latest/lib/interfaces/#LazySet-1)

### DIFF
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -73,44 +73,44 @@ julia> length(concrete_subtypes)
 38
 
 julia> println.(concrete_subtypes);
- AffineMap
- Ball1
- Ball2
- BallInf
- Ballp
- CacheMinkowskiSum
- CartesianProduct
- CartesianProductArray
- ConvexHull
- ConvexHullArray
- Ellipsoid
- EmptySet
- ExponentialMap
- ExponentialProjectionMap
- HPolygon
- HPolygonOpt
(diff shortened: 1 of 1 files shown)


mforets:
this change uncovered a failing doctest; the `AffineMap` is missing from the list.

mforets:
i dont understand the error -- is it due to suppressing some values, with the `⋮` .. ?

schillic:
> i dont understand the error -- is it due to suppressing some values, with the `⋮` .. ?

Yes, let us do this workaround:

```julia
julia> for t in subtypes(LazySet, true)
           println(t)
       end
```

mforets:
👍 or  in one line, `julia> println.(subtypes(LazySet, true));`

src/Interfaces/LazySet.jl
```diff
@@ -67,8 +67,13 @@ julia> subtypes(LazySet, false)
...
 ```jldoctest; setup = :(using LazySets: subtypes)
-julia> subtypes(LazySet, true)
-37-element Array{Type,1}:
+julia> concrete_subtypes = subtypes(LazySet, true);
+
+julia> length(concrete_subtypes)
+38
+
+julia> println.(concrete_subtypes);
+ AffineMap
```
```suggestion
AffineMap
```
`println` does not add a leading space, so those need to be removed everywhere.

PR #1742, opened 2019-09-28
Update lazy_intersections.md

mforets:


docs/src/man/lazy_intersections.md
```diff
@@ -24,14 +24,15 @@ using Plots, LazySets, LazySets.Approximations
...
+pell = plot(E₁, aspectratio=1)
+pell = plot!(pell, E₂)
 ```
 
 !!! note
-    If you are wondering about the paremeter `1e-3` passed to `plot`, this
-    parameter controls the accuracy to which the set is plotted (because the set that
-    we actually plot is a polygonal overapproximation of the ellipses!).
+    The accuracy to which this set is plotted can be controlled by passing a numerical argument as in
+    `plot(E₁, 1e-3, ratio=1)`. Here, `1e-3` stands for an upper-bound on the error, measured in terms of
```
schillic:
```suggestion
    `plot(E₁, 1e-3, aspectratio=1)`. Here, `1e-3` stands for an upper-bound on the error, measured in terms of
```

mforets: ok but `ratio` and `aspectratio` are aliases, see [here]https://github.com/JuliaPlots/Plots.jl/blob/4cb3e6a23cc3715691a93882920d6c15e3603d3c/src/args.jl#L584)

schillic: Sure, but I prefer to use the same alias consistently, at least in the same docstring.

mforets:
please don't merge yet -- i'd like to build the docs and check the new plots in my machine previously

PR #1743, opened 2019-09-30
#317 - Handling of non-square SparseMatrixExp

### LINKED ISSUE #317: Handling of non-square SparseMatrixExp
Currently we allow to create a `SparseMatrixExp` that is not square, but most operations, including printing, will crash.
Should we add a dimension check in the constructor?
At least we should document that the matrix must be square, even if that is implicitly clear.

mforets:
Closes #317.

### DIFF
--- a/src/LazyOperations/ExponentialMap.jl
+++ b/src/LazyOperations/ExponentialMap.jl
@@ -65,7 +65,7 @@ struct SparseMatrixExp{N, MN<:AbstractSparseMatrix{N}} <: AbstractMatrix{N}
     M::MN
     function SparseMatrixExp(M::MN) where {N, MN<:AbstractSparseMatrix{N}}
         @assert size(M, 1) == size(M, 2) "the lazy matrix exponential `SparseMatrixExp` " *
-        "requires the given matrix to be square, but it has size $(size(M))"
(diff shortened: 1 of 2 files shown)


src/LazyOperations/ExponentialMap.jl
```diff
@@ -54,10 +57,17 @@ julia> get_columns(E, [10]); # same as get_column(E, 10) but a 100x1 matrix is r
...
-[Expokit](https://github.com/acroy/Expokit.jl) package.
+[Expokit](https://github.com/acroy/Expokit.jl) package. Hence, you will have to
+install and load this optional dependency to have access to the functionality
+of `SparseMatrixExp`.
 """
 struct SparseMatrixExp{N, MN<:AbstractSparseMatrix{N}} <: AbstractMatrix{N}
     M::MN
+    function SparseMatrixExp(M::MN) where {N, MN<:AbstractSparseMatrix{N}}
+        @assert size(M, 1) == size(M, 2) "the lazy matrix exponential `SparseMatrixExp` " *
+        "requires the given matrix to be square, but it has size $(size(M))"
```
schillic:
```suggestion
            "requires the given matrix to be square, but it has size $(size(M))"
```

PR #1750, opened 2019-10-03
Update minkowski_difference.jl

mforets:


src/ConcreteOperations/minkowski_difference.jl
```diff
@@ -27,18 +27,18 @@ This function implements Theorem 2.3 in [1], which we state next.
 
 Suppose ``P`` is a polyhedron
 ```math
-P = \\{z ∈ ℝⁿ: sᵢᵀz ≤ rᵢ, i=1,...,N\\}.
+P = \\{z ∈ \\mathbb{R}^n: sᵢᵀz ≤ rᵢ,~i = 1 , \\ldots, N\\}.
```
schillic:
```suggestion
P = \\{z ∈ \\mathbb{R}^n: sᵢᵀz ≤ rᵢ,~i = 1, …, N\\}.
```
Does the `ℝ` not work?

mforets: In principle it is ok, however i saw that the superscript `n` is not in the right place, see: https://juliareach.github.io/LazySets.jl/latest/lib/binary_functions/#LazySets.minkowski_difference-Union{Tuple{N},%20Tuple{LazySet{N},LazySet{N}}}%20where%20N%3C:Real

mforets: It is too much below the R :laughing:  compare to: https://juliareach.github.io/LazySets.jl/latest/man/polyhedral_approximations/#Preliminaries-1

schillic: But then maybe just do the `^n`.

mforets: I can try if doing the `^n` makes a difference. But how do i get the `\mathbb{R}` on the first place in the Julia's REPL?

schillic: `\bbR<TAB>` (from [here](https://discourse.julialang.org/t/changes-to-latex-shortcuts-in-repl/9110))

mforets: :100:  Now if i do  `\bbR<TAB>\^n<TAB>` it is printed in the correct place in the REPL, however, when i print it here, again the `n` is a bit off, see: `julia> ℝⁿ`.

schillic: I would not consider the markdown rendering here. What does the html output look like? If my impression is right and `\\mathbb{R}` and `ℝ` render as the same symbol, then they should not influence the `n`.

src/ConcreteOperations/minkowski_difference.jl
```diff
@@ -27,23 +27,23 @@ This function implements Theorem 2.3 in [1], which we state next.
 
 Suppose ``P`` is a polyhedron
 ```math
-P = \\{z ∈ ℝⁿ: sᵢᵀz ≤ rᵢ, i=1,...,N\\}.
+P = \\{z ∈ ℝ^n: sᵢᵀz ≤ rᵢ,~i = 1 , …, N\\}.
```
schillic:
```suggestion
P = \\{z ∈ ℝ^n: sᵢᵀz ≤ rᵢ,~i = 1, …, N\\}.
```

PR #1766, opened 2019-10-08
#755 - Refactor Zonotopes' reduce_order

### LINKED ISSUE #755: Refactor Zonotopes' reduce_order
`reduce_order` could be sent to the `Approximations` module as `overapproximate(::Zonotope, ::Zonotope, r)`:

``` 
reduce_order(Z::Zonotope, r)::Zonotope

Reduce the order of a zonotope by overapproximating with a zonotope with less
generators.
```

SebastianGuadalupe:
Closes #755 .

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -1318,7 +1318,7 @@ function overapproximate(cap::Intersection{N,
 end
 
 """
-    overapproximate(Z::Zonotope{N}, ::Type{<:Zonotope}, r) where {N<:Real}
+    overapproximate(Z::Zonotope{N}, ::Type{<:Zonotope}, r::Union{Int, Rational{Int}}) where {N<:Real}
 
 Reduce the order of a zonotope by overapproximating with a zonotope with less
 generators.
@@ -1340,7 +1340,7 @@ This function implements the algorithm described in A. Girard's
 
 If the desired order is smaller than one, the zonotope is *not* reduced.
 """
-function overapproximate(Z::Zonotope{N}, ::Type{<:Zonotope}, r) where {N<:Real}
+function overapproximate(Z::Zonotope{N}, ::Type{<:Zonotope}, r::Union{Int, Rational{Int}}) where {N<:Real}
     c, G = Z.center, Z.generators
     d, p = dim(Z), ngens(Z)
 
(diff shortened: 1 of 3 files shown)


schillic:
What about [this comment](https://github.com/JuliaReach/LazySets.jl/issues/755#issuecomment-428440704)?

mforets:
Here we'd like to not eliminate `reduce_order` from the LazySets namespaces but rather to have it work as an alias the overapproximation with a zonotope with given order.

src/Approximations/overapproximate.jl
```diff
@@ -1316,3 +1316,55 @@ function overapproximate(cap::Intersection{N,
...
+A new zonotope with less generators, if possible.
+
+### Algorithm
+
+This function implements the algorithm described in A. Girard's
+*Reachability of Uncertain Linear Systems Using Zonotopes*, HSCC. Vol. 5. 2005.
+
+If the desired order is smaller than one, the zonotope is *not* reduced.
+"""
+function overapproximate(Z::Zonotope{N}, ::Type{<:Zonotope}, r) where {N<:Real}
```
schillic: I wonder whether we should restrict `r` to type `Number` and/or add another alias method
`overapproximate(Z::Zonotope{N}, r::Number) where {N<:Real}`.

mforets: > I wonder whether we should restrict r to type Number 

👍  here it makes sense for `r` to be either an integer or a rational, if we are going to add the type we can use a small union?

> add another alias method overapproximate(Z::Zonotope{N}, r::Number) where {N<:Real}

👎  the API of `overapproximate` requires the type of the expected set, we export `reduce_order` as a convenience function

schillic: Okay for `r::Union{Integer, Rational}`.

src/Approximations/overapproximate.jl
```diff
@@ -1316,3 +1316,55 @@ function overapproximate(cap::Intersection{N,
...
+A new zonotope with less generators, if possible.
+
+### Algorithm
+
+This function implements the algorithm described in A. Girard's
+*Reachability of Uncertain Linear Systems Using Zonotopes*, HSCC. Vol. 5. 2005.
+
+If the desired order is smaller than one, the zonotope is *not* reduced.
+"""
+function overapproximate(Z::Zonotope{N}, ::Type{<:Zonotope}, r::Union{Int, Rational{Int}}) where {N<:Real}
```
schillic: The restriction to `Int` is a breaking change for no apparent reason. Let us use `Integer` and `Rational` or stick to `Number` instead.

PR #1767, opened 2019-10-11
Update to new Documenter change

schillic:
See the breaking change [here](https://github.com/JuliaDocs/Documenter.jl/blob/master/CHANGELOG.md#version-v0240).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # LazySets.jl
 
 [![Build Status](https://travis-ci.org/JuliaReach/LazySets.jl.svg?branch=master)](https://travis-ci.org/JuliaReach/LazySets.jl)
-[![Docs latest](https://img.shields.io/badge/docs-latest-blue.svg)](http://juliareach.github.io/LazySets.jl/latest/)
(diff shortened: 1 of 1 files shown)


mforets:
should we also bump the Documente dependency?

schillic:
> should we also bump the Documente dependency?

I do not think so. If I understand correctly, it was just a cross-link that was dropped and the URL was already valid before.

mforets:
The question is: if i build LazySets Documentation with v < 0.24.0, does everything work after this PR? I have not tried in my machine, but i can.

mforets:
Actually i haven't built the Docs with the new v0.24 at all. Unfortunately there is no way to preview the built docs online (but i recall seeing some code to do this in another package, but now i forgot).

schillic:
I only changed the links to the final documentation. The building of the documentation is not affected by this PR. It is possible that the links are invalid in some old release of `Documenter`.
If you feel better, we can require the newest version of `Documenter`, but usually it is used automatically anyway.

PR #1777, opened 2019-10-16
#1776 - Fix workflow of linear_map for polyhedra

### LINKED ISSUE #1776: Fix workflow of linear_map for polyhedra
In #1713 we changed the way `linear_map` for polyhedra works. There are two issues to be fixed:
1. We check invertibility also for the `"division"` algorithm, which is not necessary.
2. We fall back to the `"vrep"` algorithm if the `"inverse"` algorithm does not apply. Instead we should throw an error. Note that `"inverse"` is the default value, so we have to allow the input `nothing`, make it the default, and in case of `nothing` we choose the algorithm dynamically as we do now.

schillic:
Closes #1776.

### DIFF
--- a/src/Interfaces/AbstractPolyhedron_functions.jl
+++ b/src/Interfaces/AbstractPolyhedron_functions.jl
@@ -233,11 +233,11 @@ end
 """
     linear_map(M::AbstractMatrix{N},
                P::AbstractPolyhedron{N};
-               [algorithm]::String=(issparse(M) ? "division" : "inverse"),
+               [algorithm]::Union{String, Nothing}=nothing,
                [check_invertibility]::Bool=true,
                [cond_tol]::Number=DEFAULT_COND_TOL,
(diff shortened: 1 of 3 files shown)


src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -343,9 +356,24 @@ function linear_map(M::AbstractMatrix{N},
     @assert dim(P) == size(M, 2) "a linear map of size $(size(M)) cannot be " *
         "applied to a set of dimension $(dim(P))"
 
-    # check invertibility
-    if algorithm != "vrep" && check_invertibility && !isinvertible(M; cond_tol=cond_tol)
-        algorithm = "vrep"
+    if algorithm == nothing
+        if issparse(M) || issquare(M)
```
mforets:
```suggestion
        if issparse(M) || !issquare(M)
```

PR #1778, opened 2019-10-17
#622 - Bloating set operation

### LINKED ISSUE #622: Bloating set operation

schillic:
Closes #622.

```julia
using LazySets, Plots
pl = plot()
for (i, p) in [(1., 2.), (4., 1.), (7., Inf)]
    B = BallInf([i, 0.], 1.);
    X = Bloating(B, 0.1, p);
    plot!(pl, B)
    plot!(pl, X)
end
plot!(; ratio=1)
```
bloating

### DIFF
--- a/src/LazyOperations/Bloating.jl
+++ b/src/LazyOperations/Bloating.jl
@@ -1,0 +1,151 @@
+export Bloating
+
+"""
+    Bloating{N<:AbstractFloat, S<:LazySet{N}} <:LazySet{N}
+
+Type that represents a uniform expansion of a convex set in a given norm (also
+known as *bloating*).
+
+### Fields
+
+- `X` -- convex set
+- `ε` -- (positive) bloating factor
+- `p` -- ``p``-norm (``≥ 1``; default: ``2``)
+
+### Notes
+
+`Bloating(X, ε, p)` is equivalent to the Minkowski sum of `X` and a ball in the
+`p`-norm of radius `ε` centered in the origin `O` (i.e., `X ⊕ Ballp(p, O, ε)`).
+"""
+struct Bloating{N<:AbstractFloat, S<:LazySet{N}} <:LazySet{N}
+    X::S
+    ε::N
+    p::N
+
+    function Bloating(X::S, ε::N, p::N=N(2)) where {N<:AbstractFloat,
+                                                    S<:LazySet{N}}
+        @assert ε > zero(N) "bloating requires a distance > 0, but $ε was given"
+        @assert p >= one(N) "bloating requires a norm >= 1, but $p was given"
+
+        return new{N, S}(X, ε, p)
+    end
+end
+
+isoperationtype(::Type{<:Bloating}) = true
+
+"""
+    dim(B::Bloating)
+
+Return the dimension of a bloated convex set.
+
+### Input
+
+- `B` -- bloated convex set
+
+### Output
+
+The ambient dimension of the bloated set.
+"""
+function dim(B::Bloating)
+    return dim(B.X)
+end
+
+# helper function to compute the bloating ball
+function _bloating_ball(B::Bloating{N}) where {N<:AbstractFloat}
+    return Ballp(B.p, zeros(N, dim(B)), B.ε)
+end
+
+"""
+    σ(d::AbstractVector{N}, B::Bloating{N}) where {N<:AbstractFloat}
+
+Return the support vector of a bloated convex set in a given direction.
+
+### Input
+
+- `d` -- direction
+- `B` -- bloated convex set
+
+### Output
+
+The support vector of the bloated set in the given direction.
+"""
+function σ(d::AbstractVector{N}, B::Bloating{N}) where {N<:AbstractFloat}
+    @assert !iszero(d) "the support vector of the zero direction is undefined"
+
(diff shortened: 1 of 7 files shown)


src/LazyOperations/Bloating.jl
```diff
@@ -0,0 +1,135 @@
...
+
+- `d` -- direction
+- `B` -- bloated convex set
+
+### Output
+
+The support vector of the bloated set in the given direction.
+"""
+function σ(d::AbstractVector{N}, B::Bloating{N}) where {N<:AbstractFloat}
+    return σ(d, B.X) + normalize(d, B.p) * B.ε
```
mforets: here i would assert that the norm of d is not zero, to avoid NaNs in the output.

src/LazyOperations/Bloating.jl
```diff
@@ -0,0 +1,135 @@
...
+
+    function Bloating(X::S, ε::N, p::N=N(2)) where {N<:AbstractFloat,
+                                                    S<:LazySet{N}}
+        @assert ε > zero(N) "bloating requires a distance > 0, but $ε was given"
+        @assert p >= one(N) "bloating requires a norm >= 1, but $p was given"
+
+        return new{N, S}(X, ε, p)
+    end
+end
+
```
mforets: after the type definition we should add 

```julia
isoperationtype(::Type{<:Bloating}) = true
```

src/LazyOperations/Bloating.jl
```diff
@@ -0,0 +1,135 @@
...
+    Bloating{N<:AbstractFloat, S<:LazySet{N}} <:LazySet{N}
+
+Type that represents a (possibly negative) uniform expansion of a convex set in
+a given norm (also known as *bloating*).
+
+### Fields
+
+- `X` -- convex set
+- `ε` -- (positive) bloating factor
+- `p` -- ``p``-norm (``≥ 1``)
```
mforets: let's add that `2` is the default for the inner constructor

src/LazyOperations/Bloating.jl
```diff
@@ -0,0 +1,140 @@
+export Bloating
+
+"""
+    Bloating{N<:AbstractFloat, S<:LazySet{N}} <:LazySet{N}
+
+Type that represents a (possibly negative) uniform expansion of a convex set in
```
mforets: here we have the word "negative" but below it is only 'positive'. as i understand it, if 0 < eps < 1 this is a contraction and otherwise for eps > 1 it is an expansion. i suggest to add this information and / or an example in the constructor's docstring.

src/LazyOperations/Bloating.jl
mforets: :thinking: : the formula doesn't give what i expect for the second term:

```julia
d = rand(3)
p = 2.5
ε = 1e-2

normalize(d, p) * ε
3-element Array{Float64,1}:
 0.0034857477766639185
 0.009630890087713    
 0.002005076494778946

σ(d, Ballp(p, zeros(3), ε))
3-element Array{Float64,1}:
 0.004634748147714765 
 0.009125845625116528 
 0.0032056502022087315
```

src/LazyOperations/Bloating.jl
```diff
@@ -0,0 +1,140 @@
...
+
+- `d` -- direction
+- `B` -- bloated convex set
+
+### Output
+
+The support function of the bloated set in the given direction.
+"""
+function ρ(d::AbstractVector{N}, B::Bloating{N}) where {N<:AbstractFloat}
+    return ρ(d, B.X) + dot(d, normalize(d, B.p) * B.ε)
```
mforets: same comment as for the support vector applies here

src/LazyOperations/Bloating.jl
```diff
@@ -0,0 +1,140 @@
...
+### Output
+
+An element in the bloated convex set.
+
+### Algorithm
+
+The implementation returns the result of `an_element` for the wrapped set.
+"""
+function an_element(B::Bloating)
+    return an_element(B.X)
```
mforets: i think that this only holds in general if the bloating is an expansion. if it is a contraction, an element from B.X may no longer live in B.

src/LazyOperations/Bloating.jl
schillic: Oh, I already forgot that we have this.

src/LazyOperations/Bloating.jl
schillic: Yeah, sorry, this is a left-over of an old version. I only want to allow values > 1.

src/LazyOperations/Bloating.jl
schillic: You are right, my formula was wrong. Now I do it the lazy way and actually compute the result for the ball.

mforets:
The error in the Windows build is unrelated to this PR.

schillic:
I just tried to run the code in the first post and got the same plots. Funny that my wrong implementation still seemed quite correct :smiley:

Issue #1781, opened 2019-10-24
Add upper bounds for dependencies


schillic:
See https://discourse.julialang.org/t/announcement-automatic-merging-for-the-general-registry/29961.

schillic:
In principle I see two ways:
1. Only allow the most recent package versions (i.e., upper bound = lower bound).
2. Have a range of available package versions.

The disadvantage of 1. is that `LazySets` only works with the current environment of packages. If you have to go back to an older version of `LazySets` (e.g., to investigate a regression), you need to reinstall old package versions. And similarly if you need to use an older version of some package, you cannot use the new versions of `LazySets` anymore.
The disadvantage of 2. is that we do not actually test the older versions (Travis always uses the most recent versions that are compatible).

So 1. is too restrictive and 2. is too generous.

I tend to go for 2. For some packages we do not know a lower bound, so we should not specify it. For other packages we do know a lower bound, but it is usually not the most recent version. We can have a Travis build with the lowest dependency versions to alleviate the disadvantage of 2.

mforets:
This was done in [v1.22.0](https://github.com/JuliaReach/LazySets.jl/releases/tag/v1.22.0), implementing option 2.

PR #1782, opened 2019-10-25
Update Project.toml

mforets:
See #1781.


Project.toml
```diff
@@ -19,26 +19,26 @@ SharedArrays = "1a1011a3-84de-559e-8e89-a11a2f7dc383"
...
-Expokit = "≥ 0.1.0"
-GLPK = "≥ 0.10.0"
-GLPKMathProgInterface = "≥ 0.4.0"
-GR = "≥ 0.39.1"
-IntervalArithmetic = "≥ 0.14.0"
-IntervalMatrices = "≥ 0.2.1"
+Expokit = "0.2.0"
+GLPK = "0.11.4"
+GLPKMathProgInterface = "0.4.4"
+GR = "0.42.0"
```
mforets:
```suggestion
GR = "0.41.0"
```

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "LazySets"
 uuid = "b4f0291d-fe17-52bc-9479-3d1a343d9043"
-version = "1.21.0"
+version = "1.22.0"
```
schillic: Can we do this in a separate PR please?

Project.toml
```diff
@@ -21,24 +21,24 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
...
-Makie = "≥ 0.9.4"
-MathProgBase = "≥ 0.7.0"
-Optim = "≥ 0.14.1"
-Plots = "≥ 0.25.1"
-Polyhedra = "0.5.4"
-RecipesBase = "≥ 0.3.0"
-Reexport = "≥ 0.2.0"
-Requires = "≥ 0.4.0"
-TaylorModels = "≥ 0.1.1"
+Expokit = "^0.2"
```
schillic: Any reason why `Expokit` v0.1 is not allowed anymore? The old version allowed it and it seemed to work.
The same holds for `GR`, `IntervalArithmetic`, `RecipesBase`, `Requires`, and `TaylorModels`.
If there is a reason to upgrade one of those packages, I think we should add a test to capture this necessity.

Project.toml
mforets:
```suggestion
version = "1.21.0"
```

Project.toml
mforets: > Any reason why Expokit v0.1 is not allowed anymore? The old version allowed it and it seemed to work.

See: This release establishes compatibility with Julia 0.7/1.0. From now on support for Julia 0.6 will be dropped, https://github.com/acroy/Expokit.jl/releases/tag/v0.2.0

> The same holds for GR, IntervalArithmetic, RecipesBase, Requires, and TaylorModels.
If there is a reason to upgrade one of those packages, I think we should add a test to capture this necessity.

I don't think that we need to add more tests for backwards compatibility  -- if it is not broken, don't fix it :) .

Project.toml
```diff
@@ -21,24 +21,24 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
...
+GLPK = "^0.10, ^0.11"
+GLPKMathProgInterface = "^0.4"
+GR = "^0.41"
```
schillic:
```suggestion
GR = "^0"
```
Your version is already outdated. Since this package is completely optional and I do not expect any breaking changes before v1 because we only use it through `Plots`, I say we do not need to add a bound for the minor version.

Project.toml
schillic: Okay, I checked these packages and I agree to upgrade them.

PR #1795, opened 2019-11-26
#1793 - Add membership for msum of singleton and set

### LINKED ISSUE #1793: Membership between singleton and lazy m-sum
Cf. https://github.com/JuliaReach/LazySets.jl/issues/1792#issue-527564541

mforets:
Closes #1793.

### DIFF
--- a/docs/src/lib/operations.md
+++ b/docs/src/lib/operations.md
@@ -191,7 +191,7 @@ dim(::MinkowskiSum)
 isbounded(::MinkowskiSum)
 isempty(::MinkowskiSum)
 constraints_list(::MinkowskiSum)
-∈(::AbstractVector{N}, ::MinkowskiSum{N, <:AbstractSingleton, <:LazySet}) where {N}
+∈(::AbstractVector{N}, ::MinkowskiSum{N, S1, S2}) where {N, S1<:AbstractSingleton, S2<:LazySet}
 ```
 Inherited from [`LazySet`](@ref):
 * [`norm`](@ref norm(::LazySet, ::Real))
--- a/docs/src/lib/operations.md
+++ b/docs/src/lib/operations.md
@@ -191,7 +191,7 @@ dim(::MinkowskiSum)
 isbounded(::MinkowskiSum)
 isempty(::MinkowskiSum)
 constraints_list(::MinkowskiSum)
-∈(::AbstractVector{N}, ::MinkowskiSum{N, <:AbstractSingleton, <:LazySets}) where {N}
+∈(::AbstractVector{N}, ::MinkowskiSum{N, <:AbstractSingleton, <:LazySet}) where {N}
 ```
 Inherited from [`LazySet`](@ref):
 * [`norm`](@ref norm(::LazySet, ::Real))
--- a/docs/src/lib/operations.md
+++ b/docs/src/lib/operations.md
@@ -191,7 +191,7 @@ dim(::MinkowskiSum)
 isbounded(::MinkowskiSum)
 isempty(::MinkowskiSum)
 constraints_list(::MinkowskiSum)
-∈(::AbstractVector{N}, ::MinkowskiSum{N, <:Singleton, <:LazySets}) where {N}
+∈(::AbstractVector{N}, ::MinkowskiSum{N, <:AbstractSingleton, <:LazySets}) where {N}
 ```
 Inherited from [`LazySet`](@ref):
 * [`norm`](@ref norm(::LazySet, ::Real))
--- a/docs/src/lib/operations.md
+++ b/docs/src/lib/operations.md
@@ -191,6 +191,7 @@ dim(::MinkowskiSum)
 isbounded(::MinkowskiSum)
 isempty(::MinkowskiSum)
 constraints_list(::MinkowskiSum)
+∈(::AbstractVector{N}, ::MinkowskiSum{N, <:Singleton, <:LazySets}) where {N}
 ```
 Inherited from [`LazySet`](@ref):
 * [`norm`](@ref norm(::LazySet, ::Real))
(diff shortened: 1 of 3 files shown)


schillic:
What about the symmetric case?

src/LazyOperations/MinkowskiSum.jl
```diff
@@ -211,6 +211,30 @@ function constraints_list(ms::MinkowskiSum)
     return constraints_list(minkowski_sum(ms.X, ms.Y))
 end
 
+"""
+    ∈(x::AbstractVector{N}, ms::MinkowskiSum{N, <:Singleton, <:LazySets}) where {N}
+
+Check whether a given point is contained in the minkowski sum of a singleton
```
schillic:
```suggestion
Check whether a given point is contained in the Minkowski sum of a singleton
```

src/LazyOperations/MinkowskiSum.jl
```diff
@@ -211,6 +211,30 @@ function constraints_list(ms::MinkowskiSum)
...
+and a lazy set.
```
schillic:
```suggestion
and a convex set.
```

src/LazyOperations/MinkowskiSum.jl
```diff
@@ -211,6 +211,30 @@ function constraints_list(ms::MinkowskiSum)
...
+
+### Input
+
+- `x`  -- point (any vector type)
```
schillic:
```suggestion
- `x`  -- point
```
(just to be consistent with other methods)

src/LazyOperations/MinkowskiSum.jl
```diff
@@ -211,6 +211,30 @@ function constraints_list(ms::MinkowskiSum)
...
+- `ms` -- lazy minkowski sum of a singleton and a lazy set
```
schillic:
```suggestion
- `ms` -- lazy Minkowski sum of a singleton and a convex set
```

src/LazyOperations/MinkowskiSum.jl
```diff
@@ -211,6 +211,30 @@ function constraints_list(ms::MinkowskiSum)
...
+
+### Output
+
+`true` iff ``x ∈ ms``.
+
+### Algorithm
+
+Note that ``x ∈ (S ⊕ P)``, where ``S`` is a singleton set, ``S = \\{s\\}`` and
+``P`` is a lazy set, ``(x-s) ∈ P``.
```
schillic:
```suggestion
``P`` is a convex set, if and only if ``(x-s) ∈ P``.
```

src/LazyOperations/MinkowskiSum.jl
```diff
@@ -211,6 +211,30 @@ function constraints_list(ms::MinkowskiSum)
     return constraints_list(minkowski_sum(ms.X, ms.Y))
 end
 
+"""
+    ∈(x::AbstractVector{N}, ms::MinkowskiSum{N, <:Singleton, <:LazySets}) where {N}
```
schillic:
```suggestion
    ∈(x::AbstractVector{N}, ms::MinkowskiSum{N, <:Singleton, <:LazySet}) where {N}
```

src/LazyOperations/MinkowskiSum.jl
```diff
@@ -211,6 +211,30 @@ function constraints_list(ms::MinkowskiSum)
...
+### Output
+
+`true` iff ``x ∈ ms``.
+
+### Algorithm
+
+Note that ``x ∈ (S ⊕ P)``, where ``S`` is a singleton set, ``S = \\{s\\}`` and
+``P`` is a lazy set, ``(x-s) ∈ P``.
+"""
+function ∈(x::AbstractVector{N}, ms::MinkowskiSum{N, <:Singleton, <:AbstractPolytope}) where {N}
```
schillic:
```suggestion
function ∈(x::AbstractVector{N}, ms::MinkowskiSum{N, <:Singleton, <:LazySet}) where {N}
```
(at least that corresponds to the documentation)

src/LazyOperations/MinkowskiSum.jl
```diff
@@ -211,6 +211,30 @@ function constraints_list(ms::MinkowskiSum)
...
+    return (x - element(ms.X)) ∈ ms.Y
```
schillic: I first thought to rather call the method for a `Translation`, but it is probably cheaper to define it directly.

mforets:
> What about the symmetric case?

I forgot it, thanks for the reminder.

src/LazyOperations/MinkowskiSum.jl
mforets: Ok, if you want to call it "inconsistency" this text in parentheses; i think that it helps, because we don't have point type.

src/LazyOperations/MinkowskiSum.jl
mforets: Yes, i agree that there's no need to wrap into a translation for library code.

mforets:
Moreover, the new method can dispatch on an `AbstractSingleton`, instead of a `Singleton`, to cover the `ZeroSet` as well.

mforets:
About the convex constraint: i think this works with non-convex sets as well.

schillic:
> About the convex constraint: i think this works with non-convex sets as well.

True, then let us just call it "set".

docs/src/lib/operations.md
```diff
@@ -191,6 +191,7 @@ dim(::MinkowskiSum)
 isbounded(::MinkowskiSum)
 isempty(::MinkowskiSum)
 constraints_list(::MinkowskiSum)
+∈(::AbstractVector{N}, ::MinkowskiSum{N, <:AbstractSingleton, <:LazySets}) where {N}
```
mforets:
```suggestion
∈(::AbstractVector{N}, ::MinkowskiSum{N, <:AbstractSingleton, <:LazySet}) where {N}
```

schillic:
I had a look at the error. It seems that `Documenter` does not recognize this construct. I tried all kinds of combinations and only got this one to work:
```julia
function ∈(x::AbstractVector{N}, ms::MinkowskiSum{N, S1, S2}) where {N, S1<:AbstractSingleton, S2<:LazySet}
    ...
end
```

PR #1796, opened 2019-11-26
#1792 - Add algorithm option and set constraints to the default

### LINKED ISSUE #1792: Fallback implementation for inclusion between polytopic sets
```julia
julia> typeof.([Rt, Tg])

2-element Array{DataType,1}:
 Zonotope{Float64}                                                         
 MinkowskiSum{Float64,Singleton{Float64,Array{Float64,1}},BallInf{Float64}}

julia> Rt ⊆ Tg

MethodError: no method matching iterate(::MinkowskiSum{Float64,Singleton{Float64,Array{Float64,1}},BallInf{Float64}})
Closest candidates are:
  iterate(!Matched::Core.SimpleVector) at essentials.jl:604
  iterate(!Matched::Core.SimpleVector, !Matched::Any) at essentials.jl:604
  iterate(!Matched::ExponentialBackOff) at error.jl:214
  ...

Stacktrace:
 [1] in(::Array{Float64,1}, ::MinkowskiSum{Float64,Singleton{Float64,Array{Float64,1}},BallInf{Float64}}) at ./operators.jl:1055
 [2] ∉(::Array{Float64,1}, ::MinkowskiSum{Float64,Singleton{Float64,Array{Float64,1}},BallInf{Float64}}) at ./operators.jl:1068
 [3] issubset at /home/mforets/.julia/dev/LazySets/src/ConcreteOperations/issubset.jl:229 [inlined]
 [4] issubset(::Zonotope{Float64}, ::MinkowskiSum{Float64,Singleton{Float64,Array{Float64,1}},BallInf{Float64}}) at /home/mforets/.julia/dev/LazySets/src/ConcreteOperations/issubset.jl:226
 [5] top-level scope at In[187]:1
```

but

```julia
julia> Tg_poly = HPolytope(constraints_list(Tg));
julia>Rt_poly = HPolytope(constraints_list(Rt));
julia>Rt_poly ⊆ Tg_poly 

true
```

cc: @dfcaporale, @SebastianGuadalupe

mforets:
Closes #1792.

### DIFF
--- a/src/ConcreteOperations/issubset.jl
+++ b/src/ConcreteOperations/issubset.jl
@@ -116,20 +116,7 @@ function ⊆(P::AbstractPolytope{N},
           )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
     @assert dim(P) == dim(H)
 
-    for v in vertices_list(P)
-        if v ∉ H
-            if witness
-                return (false, v)
-            else
-                return false
-            end
-        end
-    end
-    if witness
-        return (true, N[])
-    else
-        return true
-    end
+    return _issusbset_vertices_list(P, H, witness)
 end
 
 
@@ -197,8 +184,8 @@ end
 
 
 """
-    ⊆(P::AbstractPolytope{N}, S::LazySet{N}, [witness]::Bool=false
-     )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+    ⊆(P::AbstractPolytope{N}, S::LazySet{N}, witness::Bool=false;
+      algorithm="constraints")::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
 
 Check whether a polytope is contained in a convex set, and if not, optionally
 compute a witness.
@@ -207,7 +194,15 @@ compute a witness.
 
 - `P` -- inner polytope
 - `S` -- outer convex set
-- `witness` -- (optional, default: `false`) compute a witness if activated
+- `witness`   -- (optional, default: `false`) compute a witness if activated
+- `algorithm` -- (optional, default: `"constraints"`) algorithm for the inclusion
+                 check; available options are:
+
+    * `"constraints"`, using the list of constraints of `P` and support function
+      evaluations of `S`
+
+    * `"vertices"`, using the list of vertices of `P` and membership evaluations
+      of `S`
 
 ### Output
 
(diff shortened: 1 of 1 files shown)


src/ConcreteOperations/issubset.jl
```diff
@@ -116,20 +116,7 @@ function ⊆(P::AbstractPolytope{N},
...
-                return false
-            end
-        end
-    end
-    if witness
-        return (true, N[])
-    else
-        return true
-    end
+    return _issusbset_vertices_list(P, H, witness)
```
schillic:
```suggestion
    return _issubset_vertices_list(P, H, witness)
```

src/ConcreteOperations/issubset.jl
```diff
@@ -197,8 +184,8 @@ end
 
 
 """
-    ⊆(P::AbstractPolytope{N}, S::LazySet{N}, [witness]::Bool=false
-     )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+    ⊆(P::AbstractPolytope{N}, S::LazySet{N}, witness::Bool=false;
```
schillic: Why remove the brackets?

mforets: It was unintentional, if you propose the change i can merge it

schillic:
```suggestion
    ⊆(P::AbstractPolytope{N}, S::LazySet{N}, [witness]::Bool=false;
```

mforets:
The problem with the new default method is that it implicitly assumes that the second set, `S` in the code, is such that the list of constraints is applicable. Since this is not the case in general, we should do something.

I suggest to check if `constraints_list` is applicable, and use `"constraints"`, and otherwise use `"vertices"`.

This is not the first time the issue appears, and a solid approach is to use what is usually called a holy-trait in this case adding functions/types `ispolytopic / PolytopicSet()`. It has zero added cost because for many sets the check is compiled away as it is defined by the type (`AbstractPolytope` would return `ispolytopic` to be true). This would be another PR though.

schillic:
> It has zero added cost

This is only true for non-nested types. For deeply nested lazy sets, this is nontrivial.

src/ConcreteOperations/issubset.jl
```diff
@@ -221,10 +217,29 @@ compute a witness.
...
+        return _issubset_constraints_list(P, S, witness)
+    elseif algorithm == "vertices"
+        return _issubset_vertices_list(P, S, witness)
+    else
+        error("algorithm $algorithm unknown")
+    end
+end
+
+@inline function _default_issubset(P, S)
+    if !applicable(constraints_list, S)
```
schillic: I would typically not check for negated conditions (harder to read) in `if`-`else` blocks.

PR #1799, opened 2019-11-28
#1125 - isuniversal methods for bounded sets

### LINKED ISSUE #1125: Add isuniversal methods for all set types
In #1121 we added the `isuniversal` function. It would be better to have set-type-specific methods.

* [x] generic check for bounded set types (#1121)
* [x] dispatch for existing unbounded set types (#1527)
* [ ] dispatch for existing bounded set types
* [ ] generic witness production for bounded set types: compute the support vector in some direction and add the direction on top

schillic:
Closes #1125.

### DIFF
--- a/src/Interfaces/AbstractPolyhedron_functions.jl
+++ b/src/Interfaces/AbstractPolyhedron_functions.jl
@@ -54,6 +54,41 @@ function ∈(x::AbstractVector{N}, P::AbstractPolyhedron{N})::Bool where {N<:Rea
 end
 
 """
+    isuniversal(P::AbstractPolyhedron{N}, [witness]::Bool=false
+               )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+
+Check whether a polyhedron is universal.
+
+### Input
+
+- `P`       -- polyhedron
+- `witness` -- (optional, default: `false`) compute a witness if activated
+
+### Output
+
+* If `witness` option is deactivated: `true` iff ``P`` is universal
+* If `witness` option is activated:
+  * `(true, [])` iff ``P`` is universal
+  * `(false, v)` iff ``P`` is not universal and ``v ∉ P``
+
+### Algorithm
+
+`P` is universal iff it has no constraints.
+
+A witness is produced using `isuniversal(H)` where `H` is the first linear
+constraint of `P`.
+"""
+function isuniversal(P::AbstractPolyhedron{N}, witness::Bool=false
+                    )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+    constraints = constraints_list(P)
+    if isempty(constraints)
+        return witness ? (true, N[]) : true
+    else
+        return witness ? isuniversal(constraints[1], true) : false
+    end
+end
+
+"""
     constrained_dimensions(P::AbstractPolyhedron)::Vector{Int} where {N<:Real}
 
 Return the indices in which a polyhedron is constrained.
--- a/src/Interfaces/AbstractPolytope.jl
+++ b/src/Interfaces/AbstractPolytope.jl
@@ -121,8 +121,9 @@ function isuniversal(P::AbstractPolytope{N}, witness::Bool=false
                     )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
     if witness
         constraints = constraints_list(P)
-        @assert !isempty(constraints) "polytopes without constraints are not " *
-            "allowed"
+        if isempty(constraints)
+            return (true, N[])  # special case for polytopes without constraints
+        end
         return isuniversal(constraints[1], true)
     else
         return false
--- a/src/Interfaces/AbstractPolytope.jl
+++ b/src/Interfaces/AbstractPolytope.jl
@@ -96,6 +96,39 @@ function isempty(P::AbstractPolytope)::Bool
     return isempty(vertices_list(P))
 end
 
+"""
+    isuniversal(P::AbstractPolytope{N}, [witness]::Bool=false
+               )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+
+Check whether a polyhedron is universal.
+
+### Input
+
+- `P`       -- polyhedron
+- `witness` -- (optional, default: `false`) compute a witness if activated
+
+### Output
+
+* If `witness` option is deactivated: `false`
+* If `witness` option is activated: `(false, v)` where ``v ∉ P``
+
+### Algorithm
+
+A witness is produced using `isuniversal(H)` where `H` is the first linear
+constraint of `P`.
+"""
+function isuniversal(P::AbstractPolytope{N}, witness::Bool=false
+                    )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+    if witness
+        constraints = constraints_list(P)
+        @assert !isempty(constraints) "polytopes without constraints are not " *
+            "allowed"
+        return isuniversal(constraints[1], true)
+    else
+        return false
+    end
+end
+
 # given a polytope P, apply the linear map P to each vertex of P
 # it is assumed that the interface function `vertices_list(P)` is available 
 @inline function _linear_map_vrep(M::AbstractMatrix{N}, P::AbstractPolytope{N}) where {N<:Real}
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -112,6 +112,7 @@ This interface defines the following functions:
 ```@docs
 dim(::AbstractCentrallySymmetric)
 isbounded(::AbstractCentrallySymmetric)
+isuniversal(::AbstractCentrallySymmetric{N}, ::Bool=false) where {N<:Real}
 an_element(::AbstractCentrallySymmetric{N}) where {N<:Real}
 isempty(::AbstractCentrallySymmetric)
 ```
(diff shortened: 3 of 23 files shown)


schillic:
> ok, it should be that it is the full space Rn.

Yes, and this is actually not a new method.

src/Interfaces/AbstractPolytope.jl
```diff
@@ -96,6 +96,32 @@ function isempty(P::AbstractPolytope)::Bool
...
+* If `witness` option is activated: `(false, v)` where ``v ∉ P``
+
+### Algorithm
+
+A witness is produced using `isuniversal(H)` where `H` is the first linear
+constraint of `P`.
+"""
+function isuniversal(P::AbstractPolytope{N}, witness::Bool=false
+                    )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+    return witness ? isuniversal(constraints_list(P)[1], true) : false
```
mforets: The command `constraints_list(P)[1]` will error if `P` has no constraints. As in the case of `HPolyhedron` given above, i would check add the check `if isempty(P.constraints)` here.

mforets: BTW when is a constraint universal? The `a` has to be zero?

schillic: A polytope should always have constraints.

mforets: `HPolytype()` is a valid command..

mforets: At least, can you add a better error message?

src/Interfaces/AbstractPolytope.jl
```diff
@@ -96,6 +96,39 @@ function isempty(P::AbstractPolytope)::Bool
...
+### Algorithm
+
+A witness is produced using `isuniversal(H)` where `H` is the first linear
+constraint of `P`.
+"""
+function isuniversal(P::AbstractPolytope{N}, witness::Bool=false
+                    )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+    if witness
+        constraints = constraints_list(P)
+        @assert !isempty(constraints) "polytopes without constraints are not " *
```
mforets: Why "not allowed"? The following command works:

```julia
julia> using LazySets

julia> HPolytope()
HPolytope{Float64}(HalfSpace{Float64,VN} where VN<:AbstractArray{Float64,1}[])
```

schillic: Yes, the command is allowed, but the instance should only be used after adding bounding constraints. The instance with no constraints is contradictory and erroneous. If we return `true` here (because the polytope does not have any constraints), this adds even more contradictions.
```julia
julia> isbounded(HPolytope())
true

julia> constrained_dimensions(HPolytope())
0-element Array{Int64,1}

julia> ρ([1.], HPolytope())
ERROR: the support function in direction [1.0] is undefined because
the polytope is unbounded

julia> an_element(HPolytope())
ERROR: ArgumentError: An index is out of bound.

julia> [0.] ∈ HPolytope()
ERROR: AssertionError: a 1-dimensional point cannot be an element
of a -1-dimensional set
```

mforets: I think you are thinking it too much. The method `isuniversal` for a `HPolytope()` should return `true`, because it is universal, the same way as you added for `HPolyhedron()`. There is nothing more on that.

schillic: `isuniversal(X) && isbounded(X)` should never hold. This is contradictory. For `AbstractPolytope`, the type implies boundedness and hence non-universality.

mforets: The title of this PR is "isuniversal methods for bounded sets", but i didn't comment it was contradictory / meaningless, although that thought came into my mind as you point out (it should never hold). But here, if `isuniversal(HPolyhedron())` is `true`, there is no good reason to no let `isuniversal(HPolytope())` return `true` as well.. 

> the type implies boundedness and hence non-universality.

Yes, and in that case, `isuniversal` should *not* make checks with the value, and instead return `false` for any `HPolytope`. But that is not what you have written..

schillic: > Yes, and in that case, isuniversal should not make checks with the value, and instead return false for any HPolytope. But that is not what you have written

The method returns `false` unless we produce a witness. For that I use the first constraint. This constraint must exist because otherwise the object does not represent a polytope, which the type implies.

schillic: I see two options:
1. We let `isuniversal(::AbstractPolytope) == false` independent of the concrete object (this is implemented here). Obviously, witness production should not alter that result, so the result should either be `(false, w)` for witness `w` or an error. For `HPolytope`s without any constraints I do not see how to compute `w` - we do not even know the dimension of the set.
2. We let `isuniversal(HPolytope()) == true`, but then we should remove `HPolytope` because obviously there would not be a semantic difference to `HPolyhedron` anymore.

mforets: Eventually we can add `isboundedtype` to refer to the type and `isbounded` to refer to the object. I find that better than the the current implementation, which is

```julia
function isbounded(P::HPolytope, use_type_assumption::Bool=true)::Bool
    if use_type_assumption
        return true
    end
    return isbounded(HPolyhedron(P.constraints))
end
```
Regarding the universal case, i think that the implementation should be:

```julia
function isuniversal(P::HPolytope, use_type_assumption::Bool=true)::Bool
    if use_type_assumption
        return true
    end
    return isuniversal(HPolyhedron(P.constraints))
end
```
In particular, the assertion should be removed, because it is a value, and because it is wrong: polytopes without constraints *are* allowed as one can do `HPolytope()`.

schillic: My concern is **only about witness production**. I only want to throw an error in that case. Your proposal does *not* answer that question. What should `isuniversal(HPolytope(), true; use_type_assumption=true)` return?

```julia
function isuniversal(P::HPolytope, witness::Bool=false; use_type_assumption::Bool=true)::Bool
    if use_type_assumption
        if witness
            # ???
        else
            return false
        end
    end
    ...
end
```

> In particular, the assertion should be removed, because it is a value, and because it is wrong: polytopes without constraints are allowed as one can do `HPolytope()`.

Only because you can do `HPolytope()` does not imply that you are allowed to use the result for anything. I personally do not like that you can do that because it clearly violates type assumptions, and I would argue that we only have this option for historical reasons. I would be happy to remove this constructor and rather require that `HPolytope`s must be constructed from a list of bounding constraints. We have `HPolyhedron` for the general case now and you can always `convert` between the types efficiently. But that discussion is going beyond this issue.

schillic: The following compromise might work for both of us:

* Use a similar implementation for `AbstractPolytope` as for `AbstractPolyhedron` - but with `false` instead of `true`.

```julia
function isuniversal(P::AbstractPolytope{N}, witness::Bool=false
                    )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
    return witness ? isuniversal(constraints_list(P)[1], true) : false
end
```

* But for `HPolytope` we override the definition:

```julia
function isuniversal(P::HPolytope{N}, witness::Bool=false
                    )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
    constraints = constraints_list(P)
    if isempty(constraints)
        return witness ? (true, N[]) : true
    else
        return isuniversal(constraints[1], true)
    end
end
```

Usually I do not like that overriding an interface definition changes the semantics - it should only provide a more efficient implementation. But in this case the result only differs for `P`s that are (IMO) invalid inputs and for those the output is undefined (for me), so *any* output is good for me and I can also live with inconsistencies in these cases.

What do you think?

mforets: > What should isuniversal(HPolytope(), true; use_type_assumption=true) return?

It should return `false`, because the type assumption of polytopes is that they are bounded hence non-universal. The witness production option is by definition ill-defined if we only want to use the type assumption, as producing a witness inevitably needs to consider the value.

schillic: > It should return false

The second argument (`true`) stands for witness production. I disagree that we should not return a witness if a witness was requested.

schillic: > The witness production option is by definition ill-defined if we only want to use the type assumption, as producing a witness inevitably needs to consider the value.

And this is exactly the case when I throw an error. Then I do not understand why you do not like my solution.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -53,6 +53,40 @@ function ∈(x::AbstractVector{N}, P::AbstractPolyhedron{N})::Bool where {N<:Rea
...
+### Algorithm
+
+`P` is universal iff it has no constraints.
+
+A witness is produced using `isuniversal(H)` where `H` is the first linear
+constraint of `P`.
+"""
+function isuniversal(P::AbstractPolyhedron{N}, witness::Bool=false
+                    )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+    if isempty(P.constraints)
```
schillic: Reminder for me: This needs to be changed.

src/Interfaces/AbstractPolytope.jl
schillic: Conclusion: `(true, N[])`

mforets: Ok :+1:  So this will make `isuniversal(HPolytope()) = false` and `isuniversal(HPolytope(), true) = (true, N[])`. I've opened issue https://github.com/JuliaReach/LazySets.jl/issues/1835 to outsource this problem.

mforets:
lgtm once ci passes

PR #1801, opened 2019-11-29
Update Zonotope.jl

mforets:
- The `center` of the Zonotope doesn't need to be a dense vector, but having the output annotated enforces a needless conversion.
- The conversion takes some time.
- Similarly, the overapproximate call in `reduce_order` produces a Zonotope, so there is no need to convert again to a Zonotope.

### DIFF
--- a/src/Sets/Zonotope.jl
+++ b/src/Sets/Zonotope.jl
@@ -113,7 +113,7 @@ Zonotope(center::AbstractVector{N}, generators_list::AbstractVector{VN};
 
 
 """
-    center(Z::Zonotope{N})::Vector{N} where {N<:Real}
+    center(Z::Zonotope{N}) where {N<:Real}
 
 Return the center of a zonotope.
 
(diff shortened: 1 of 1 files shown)


src/Sets/Zonotope.jl
```diff
@@ -275,7 +275,7 @@ A new zonotope with less generators, if possible.
 
 See `overapproximate(Z::Zonotope{N}, ::Type{<:Zonotope}, r::Union{Integer, Rational}) where {N<:Real}` for details.
 """
-function reduce_order(Z::Zonotope{N}, r::Union{Integer, Rational})::Zonotope{N} where {N<:Real}
+function reduce_order(Z::Zonotope{N}, r::Union{Integer, Rational}) where {N<:Real}
```
mforets: I don't know if we want to keep the type annotation in the docstring; i'm fine either way..

schillic: Same, but if in doubt, let us remove it.

PR #1808, opened 2019-12-02
#1446 - Print error message if required external package is missing

### LINKED ISSUE #1446: Print error message if required external package is missing
Check if all methods in [`init.jl`](https://github.com/JuliaReach/LazySets.jl/blob/master/src/init.jl) and [`Approximations/init.jl`](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Approximations/init.jl) have a method that prints an error message outside the `@requires` block.
~(Best wait until resolving #1109.)~

schillic:
Closes #1446.

I defined new methods outside the `Requires` wrappers for better error printing. Alternatively, we could sometimes remove the `Requires` wrappers altogether and instead use our `require` helper function.

The only cases I did not add are functions that are clearly package specific (`default_cddlib_backend`, `default_polyhedra_backend`, `default_lp_solver_polyhedra`, and `polyhedron`).

### DIFF
--- a/src/LazyOperations/ExponentialMap.jl
+++ b/src/LazyOperations/ExponentialMap.jl
@@ -10,6 +10,14 @@ export SparseMatrixExp,
        get_column,
        get_columns
 
+function load_expokit()
+return quote
+
+using .Expokit: expmv
+
+end end  # quote / load_expokit
+
+
 # --- SparseMatrixExp & ExponentialMap ---
 
 """
@@ -84,10 +92,9 @@ function size(spmexp::SparseMatrixExp, ax::Int)::Int
     return size(spmexp.M, ax)
 end
 
-function load_expokit_sparsematrixexp()
-return quote
-
 function get_column(spmexp::SparseMatrixExp{N}, j::Int)::Vector{N} where {N}
+    require(:Expokit; fun_name="get_column")
+
     n = size(spmexp, 1)
     aux = zeros(N, n)
     aux[j] = one(N)
@@ -96,6 +103,7 @@ end
 
 function get_columns(spmexp::SparseMatrixExp{N},
                      J::AbstractArray)::Matrix{N} where {N}
+    require(:Expokit; fun_name="get_columns")
 
     n = size(spmexp, 1)
     aux = zeros(N, n)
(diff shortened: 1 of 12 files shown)


src/LazyOperations/ExponentialMap.jl
```diff
@@ -84,6 +84,24 @@ function size(spmexp::SparseMatrixExp, ax::Int)::Int
     return size(spmexp.M, ax)
 end
 
+function get_column(spmexp::SparseMatrixExp{N}, j::Int) where {N}
```
mforets: I don't understand the purpose of having functions that just call `require`. Do you have a use case? Why do these functions not contain docstrings? Or the idea is that after doing the `using Expokit` this function is re-written with the one that actually contains code? In that case, i think that it is not a good idea. Just put the actual code just after the call to `require`, as we do with Polyhedra functions.

schillic: > Or the idea is that after doing the using Expokit this function is re-written with the one that actually contains code?

Yes, that is the purpose. Currently if you do `get_column(E)` it just says that it does not know about that method. With this PR you get a nice error message that you should load `Expokit`.

> Put the actual code just after the call to require, as we do with Polyhedra functions.

That was the question above. I can try that, but it may not always work. If it works everywhere, we can remove `Requires` and I was not sure whether you want to go there. But fine, I also think this is the better thing to do.

schillic: Done.

src/Sets/VPolytope.jl
```diff
@@ -555,10 +555,6 @@ function load_polyhedra_vpolytope() # function to be loaded by Requires
 return quote
 # see the interface file AbstractPolytope.jl for the imports
 
-export vertices_list,
-       tohrep,
```
mforets: `vertices_list` is redundant withing the file, but `tohrep` and `tovrep` where defined only once in `load_polyhedra_vpolytope`. I wouldn't remove these exports from `VPolytope.jl`.

mforets:
@schillic Have you tried the plot3d function after this change? (I haven't, and since we don't test it..)

schillic:
> Have you tried the plot3d function after this change?

Yes, it still works.

mforets:
Ok, thanks.

src/Sets/VPolytope.jl
schillic: These `export`s were defined deep inside the file (look at the line numbers). They should be at the top of the file. I can move them there if you prefer.

PR #1810, opened 2019-12-03
#1809 - Inconsistent inclusion check

### LINKED ISSUE #1809: Inconsistent inclusion check
```julia
X = BallInf([0.1, 0.2, 0.1], 0.3)
Y = convert(HPolytope, X)
X ⊆ Y    # true
Y ⊆ X    # false
```

Reported by @ueliwechsler in https://github.com/JuliaReach/LazySets.jl/issues/1586#issuecomment-561201760_

mforets:
Closes #1809.

### DIFF
--- a/src/ConcreteOperations/issubset.jl
+++ b/src/ConcreteOperations/issubset.jl
@@ -157,10 +157,8 @@ function ⊆(H1::AbstractHyperrectangle{N},
     @inbounds for i in 1:dim(H1)
         c_dist = c1[i] - c2[i]
         r_dist = radius_hyperrectangle(H1, i) - radius_hyperrectangle(H2, i)
-        # check if c_dist is in the interval [r_dist, -r_dist]
-        left = _leq(r_dist, c_dist)
-        right = _leq(c_dist, -r_dist)
-        if !left || !right
+        # check if c_dist is not in the interval [r_dist, -r_dist]
+        if !_leq(r_dist, c_dist) || !_leq(c_dist, -r_dist)
             if witness
                 # compute a witness 'p' in the difference
                 p = copy(center(H1))
--- a/src/ConcreteOperations/issubset.jl
+++ b/src/ConcreteOperations/issubset.jl
@@ -152,15 +152,20 @@ function ⊆(H1::AbstractHyperrectangle{N},
            witness::Bool=false
           )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
     @assert dim(H1) == dim(H2)
-
-    for i in 1:dim(H1)
-        c_dist = center(H1)[i] - center(H2)[i]
+    zero_N = zero(N)
+    c1 = center(H1)
+    c2 = center(H2)
+    @inbounds for i in 1:dim(H1)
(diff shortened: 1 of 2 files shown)


mforets:
I'd like to add the test, although i'm not sure in which file it should be ..

schillic:
> I'd like to add the test, although i'm not sure in which file it should be ..

Yeah, that's a problem. Eventually I wanted to add a separate test file for binary operations. For now I would add it either to `BallInf` or `HPolytope`.

src/ConcreteOperations/issubset.jl
```diff
@@ -152,15 +152,20 @@ function ⊆(H1::AbstractHyperrectangle{N},
            witness::Bool=false
           )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
     @assert dim(H1) == dim(H2)
-
-    for i in 1:dim(H1)
-        c_dist = center(H1)[i] - center(H2)[i]
+    zero_N = zero(N)
```
schillic: We used to do that in the past, but the compiler will inline `zero(N)` anyway and so I prefer to remove this variable.

src/ConcreteOperations/issubset.jl
```diff
@@ -152,15 +152,20 @@ function ⊆(H1::AbstractHyperrectangle{N},
...
+    c1 = center(H1)
+    c2 = center(H2)
+    @inbounds for i in 1:dim(H1)
+        c_dist = c1[i] - c2[i]
```
schillic: :+1: for the optimization

src/ConcreteOperations/issubset.jl
```diff
@@ -152,15 +152,20 @@ function ⊆(H1::AbstractHyperrectangle{N},
...
         r_dist = radius_hyperrectangle(H1, i) - radius_hyperrectangle(H2, i)
-        if -r_dist < c_dist || c_dist < r_dist
+        # check if c_dist is in the interval [r_dist, -r_dist]
+        left = _leq(r_dist, c_dist)
+        right = _leq(c_dist, -r_dist)
+        if !left || !right
```
schillic: This is a (minor) performance degradation. Previously the condition was short-circuited if the first disjunct was satisfied. Now you evaluate `right` in all cases.
I would prefer to write comments that explain every condition (in particular because `left`/`right` are very ambiguous and do not help me) or write helper functions like `_in_left_bound`.

mforets: Ok i will write the statements in the same line, good point. (`left` and `right` mean left side of the inequality and right side of the inequality respectively).

mforets:
> Can you explain how this closes the issue?

Yes; in the `Y ⊆ X` dispatches on LazySet ⊆ AbstractHyperrectangle, whose implementation is to take the interval hull of `Y` and then use the AbstractHyperrectangle ⊆ AbstractHyperrectangle algorithm.

src/ConcreteOperations/issubset.jl
mforets: Ok :+1:

PR #1812, opened 2019-12-03
#1811 - Volume of a hyperrectangular set

### LINKED ISSUE #1811: Volume of a hyperrectangular set

mforets:
Closes #1811.

- [x] add test


src/ConcreteOperations/volume.jl
```diff
@@ -0,0 +1,34 @@
...
+
+Return the volume of a hyperrectangular set.
+
+### Input
+
+- `H` -- hyperrectangular set
+
+### Output
+
+The (Euclidean) volume of ``H``.
```
schillic: Just out of curiosity: What other types of volume are there?

src/ConcreteOperations/volume.jl
```diff
@@ -0,0 +1,34 @@
...
+
+### Algorithm
+
+The volume of the hyperrectangle ``H`` with vector radius ``r`` is
```
schillic:
```suggestion
The volume of the ``n``-dimensional hyperrectangle ``H`` with vector radius ``r`` is
```
(because `n` is used in the formula)

src/ConcreteOperations/volume.jl
```diff
@@ -0,0 +1,34 @@
...
+    vol = α * prod(r)
+    return vol
+end
+
+# fallback for rational
+function volume(H::AbstractHyperrectangle{N}) where {N}
+    r = radius_hyperrectangle(H)
+    vol = prod(2 .* r)
+    return vol
+end
```
schillic: The following implementations are more efficient because they only create one or zero intermediate vectors.
```julia
function volume2(H::AbstractHyperrectangle{N}) where {N<:Real}
    vol = one(N)
    for ri in radius_hyperrectangle(H)
        vol *= 2 * ri
    end
    return vol
end

function volume3(H::AbstractHyperrectangle{N}) where {N<:Real}
    vol = one(N)
    for i in 1:dim(H)
        vol *= 2 * radius_hyperrectangle(H, i)
    end
    return vol
end

julia> B = BallInf(zeros(Rational{Int}, 1000), 1//2);

julia> @btime volume($B)
  96.468 μs (2 allocations: 31.50 KiB)
1//1

julia> @btime volume2($B)
  94.259 μs (1 allocation: 15.75 KiB)
1//1

julia> @btime volume3($B)
  92.144 μs (0 allocations: 0 bytes)
1//1
```

src/ConcreteOperations/volume.jl
```diff
@@ -0,0 +1,34 @@
...
+The volume of the hyperrectangle ``H`` with vector radius ``r`` is
+``2ⁿ ∏ᵢ rᵢ`` where ``rᵢ`` denotes the ``i``-th component of ``r``.
+"""
+function volume(H::AbstractHyperrectangle{N}) where {N<:AbstractFloat}
+    r = radius_hyperrectangle(H)
+    n = dim(H)
+    α = exp(n*log(2))
+    vol = α * prod(r)
+    return vol
+end
```
schillic: This implementation is faster than your fallback but slightly less precise and slower than my second suggestion in the next comment (but I copied the functions here for convenience).
```julia
function volume2(H::AbstractHyperrectangle{N}) where {N<:Real}
    vol = one(N)
    for ri in radius_hyperrectangle(H)
        vol *= 2 * ri
    end
    return vol
end

function volume3(H::AbstractHyperrectangle{N}) where {N<:Real}
    vol = one(N)
    for i in 1:dim(H)
        vol *= 2 * radius_hyperrectangle(H, i)
    end
    return vol
end

julia> B = BallInf(zeros(100), 0.5);

julia> @btime volume($B)
  120.164 ns (1 allocation: 896 bytes)
1.0000000000000018

julia> @btime volume2($B)
  153.538 ns (1 allocation: 896 bytes)
1.0

julia> @btime volume3($B)
  75.993 ns (0 allocations: 0 bytes)
1.0
```

src/ConcreteOperations/volume.jl
mforets: I meant the volume element in [Euclidean space](https://en.wikipedia.org/wiki/Volume_element#Volume_element_in_Euclidean_space). This function estimates the [Lebesgue measure](https://en.wikipedia.org/wiki/Lebesgue_measure) of the set, but there are many other ways to assign volumes by using a different measure.

src/ConcreteOperations/volume.jl
mforets: Thanks for your comments. I didn't consider optimizing for BallInf, since the application that i had in mind requires to use Hyperrectangles (splitting a BallInf by the longer direction is no longer a box). For hyperrectangles, the functions that i added do not allocate.

After some tests, the implementation that works better for hyperrectangles is the variation `mapreduce(x -> 2x, *, radius_hyperrectangle(H))`. For boxes, your `volume3` is faster for dim approx smaller than 16, eventually there is a crossover with the exponentiation function (`vol = exp(n*log(2r))`) as it has constant cost. For n=100, the exponential is 60% faster than volume3. I would tend to use volume3 however, since n > 20 seems "big" for this type of computation..

schillic: > I didn't consider optimizing for BallInf, since the application that i had in mind requires to use Hyperrectangles (splitting a BallInf by the longer direction is no longer a box). For hyperrectangles, the functions that i added do not allocate.

Oh, true, I forgot that. Whatever you prefer, then.

schillic: > I would tend to use volume3 however, since n > 20 seems "big" for this type of computation.

What about implementing both and choosing the best implementation based on `dim(H)`?

mforets: :ok_hand:  BTW the experiments that i did are in [this notebook](https://nbviewer.jupyter.org/github/mforets/escritoire/blob/master/sets/Volume.ipynb)

src/ConcreteOperations/volume.jl
```diff
@@ -23,16 +23,27 @@ function volume(H::AbstractHyperrectangle{N}) where {N<:Real}
...
     else
         r = B.radius
         vol = exp(n*log(2r))
     end
     return vol
 end
+
+function volume(B::BallInf{N}) where {N<:Real}
+    n = dim(B)
+    vol = _vol_prod(B, n)
```
schillic: Since you optimized for `BallInf`: multiplying `2 * r` `n` times has `2n` multiplications, but you can do with only `n+1`:
```julia
    diameter = 2 * B.radius
    return diameter^dim(B)
```

mforets:
> I would rather distribute these functions to AbstractHyperrectangle.jl and BallInf.jl.

Ok.

src/ConcreteOperations/volume.jl
mforets: I tried it but it didnt show to be faster.. maybe because of the `^` operation?
Feel free to open a follow-up issue.

schillic: Then at least this one should be faster.
```julia
@inline function _vol_prod(B::BallInf{N}, n) where {N<:Real}
    vol = one(N)
    diameter = 2 * radius_hyperrectangle(H, i)
    for i in 1:n
        vol *= diameter
    end
    return vol
end
```

src/Interfaces/AbstractHyperrectangle.jl
```diff
@@ -570,3 +571,34 @@ The `Hyperrectangle` that corresponds to the rectification of `H`.
...
+    return vol
+end
+
+@inline function _vol_prod(H::AbstractHyperrectangle{N}, n) where {N<:Real}
+    vol = one(N)
+    for i in 1:n
+        vol *= 2 * radius_hyperrectangle(H, i)
+    end
+    return vol
+end
```
schillic: This function is now unused, right?

src/Sets/BallInf.jl
```diff
@@ -303,3 +304,29 @@ function translate(B::BallInf{N}, v::AbstractVector{N}) where {N<:Real}
...
+end
+
+function volume(B::BallInf{N}) where {N<:AbstractFloat}
+    n = dim(B)
+    if n < 50
+        vol = _vol_prod(B, n)
+    else
+        r = B.radius
+        vol = exp(n*log(2r))
+    end
```
schillic: The `else` case is untested.

src/ConcreteOperations/volume.jl
mforets: I've added it, although in the tests it doesn't come up as faster than the current one. I wouldn't be surprised if the compiler already does some tricks to optimize the loop.

src/Interfaces/AbstractHyperrectangle.jl
mforets: Yes. Do you want to remove it, or do you suggest to keep it?

src/ConcreteOperations/volume.jl
schillic: Okay, thanks for trying.

src/LazySets.jl
```diff
@@ -124,6 +124,7 @@ include("ConcreteOperations/issubset.jl")
 include("ConcreteOperations/minkowski_difference.jl")
 include("ConcreteOperations/minkowski_sum.jl")
 include("ConcreteOperations/reflection.jl")
+include("ConcreteOperations/volume.jl")
```
schillic:
```suggestion
```

src/Interfaces/AbstractHyperrectangle.jl
mforets: Yes. Do you want to remove it, or do you suggest to keep it?

schillic: I see no point in keeping it if you have decided against using it.

src/Sets/BallInf.jl
mforets: I've added two new tests for BallInf.

mforets:
@schillic  Thanks for your review! 

As usual i will nevertheless wait for CI if something unexpected breaks...

PR #1819, opened 2019-12-04
#1586  Equivalence check for sets

### LINKED ISSUE #1586: Equivalence check for sets
Add `isequivalent(X, Y) = X ⊆ Y && Y ⊆ X`, which is mainly supposed to be a convenience function.

For certain type combinations one can add more efficient methods later.

ueliwechsler:
Close #1586

### DIFF
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -685,18 +685,19 @@ isoperationtype(::LazySet) = error("`isoperationtype` cannot be applied to " *
                                    "a set instance; use `isoperation` instead")
 
 """
-    is_equivalent(X::LazySet, Y::LazySet)
+    isequivalent(X::LazySet, Y::LazySet; eps=0.0; p=Inf )
 
-Return whether two LazySets are equal in the mathematical sense, i.e. equivalent.
+Return whether two LazySets are equal in the mathematical sense, i.e. equivalent. If `eps` is bigger than zero, the equivalence check is relaxed by a `p`-norm ball with radius `eps`.
 
 ### Input
 
 - `X` -- any `LazySet`
 - `Y` -- another `LazySet`
+- `eps` -- (optional, default=0.0) tolerance for equivalence check
+- `p` -- (optional, default=Inf) norm of `Ballp` used for the approximative equivalence check
 
 ### Output
-
-- `true` iff `X` is equivalent to `Y`.
+`true` iff `X` is equivalent to `Y` with tolerance `eps`.
 
 ### Examples
 
@@ -709,9 +710,13 @@ julia> is_equivalent(X, Y)
 true
 ```
 """
-function isequivalent(X::LazySet, Y::LazySet)
+function isequivalent(X::LazySet, Y::LazySet; eps=0.0; p=Inf )
     if X == Y
         return true
     end
-    return X ⊆ Y && Y ⊆ X
+    if eps == 0.0
+        return X ⊆ Y && Y ⊆ X
+    else
+        return X ⊆ Y + Ballp(p, zeros(n),eps) &&
+               Y ⊆ X + Ballp(p, zeros(n),eps)
 end
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -695,11 +695,14 @@ Return whether two LazySets are equal in the mathematical sense, i.e. equivalent
 - `Y` -- another `LazySet`
 
 ### Output
+
 `true` iff `X` is equivalent to `Y`.
 
 ## Algorithm
 
-First, the check `X == Y` is performed which returns `true` if and only if the given sets are of the same type, and have the same values (modulo floating-point tolerance). Otherwise, the double inclusion check `X ⊆ Y && Y ⊆ X` is computed.
+First, the check `X == Y` is performed which returns `true` if and only if the given sets are of the same type,
+and have the same values (modulo floating-point tolerance). Otherwise, the double inclusion check `X ⊆ Y && Y ⊆ X` is
+computed.
 
 ### Examples
 
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -708,7 +708,7 @@ julia> X = BallInf([0.1, 0.2], 0.3)
 julia> Y = convert(HPolytope, X)
 julia> X == Y
 false
-julia> is_equivalent(X, Y)
+julia> isequivalent(X, Y)
 true
 ```
 """
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -719,7 +719,7 @@ true
 ```
 """
 function isequivalent(X::LazySet, Y::LazySet)
-    if X == Y
+    if X ≈ Y
         return true
     end
     return X ⊆ Y && Y ⊆ X
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -685,23 +685,21 @@ isoperationtype(::LazySet) = error("`isoperationtype` cannot be applied to " *
                                    "a set instance; use `isoperation` instead")
 
 """
-    isequivalent(X::LazySet, Y::LazySet; eps=0.0; p=Inf )
+    isequivalent(X::LazySet, Y::LazySet)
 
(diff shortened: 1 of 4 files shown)


src/Interfaces/LazySet.jl
```diff
@@ -682,3 +683,34 @@ end
...
+julia> X = BallInf([0.1, 0.2], 0.3)
+julia> Y = convert(HPolytope, X)
+julia> X == Y
+false
+julia> is_equivalent(X, Y)
+true
+```
+"""
+function isequivalent(X::LazySet, Y::LazySet)
+    if X == Y
```
mforets: Ok; this check is useful if the sets are of the same type, that was the case?

src/Interfaces/LazySet.jl
```diff
@@ -682,3 +683,34 @@ end
...
+
+```
+julia> X = BallInf([0.1, 0.2], 0.3)
+julia> Y = convert(HPolytope, X)
+julia> X == Y
+false
+julia> is_equivalent(X, Y)
+true
+```
+"""
```
mforets: Let's add the algorithm:

```julia
## Algorithm

First, the check `X == Y` is performed which returns `true` if and only if the given sets are of the same type, and have the same values (modulo floating-point tolerance). Otherwise, the double inclusion check `X ⊆ Y && Y ⊆ X` is computed.
```

src/Interfaces/LazySet.jl
ueliwechsler: Yes, I added this because it is not expensive to check `X == Y` and could speed up the check if `X == Y` is already true.

mforets: Ok, makes sense.

ueliwechsler:
Do you guys know, what is wrong with the build travis build in julia 1.2?

mforets:
It says `GitError(Code:ERROR, Class:Net, curl error: Failed to connect to github.com port 443: Connection timed out`, i will try triggering the build again.

schillic:
The v1.2 builds run into timeouts. Not sure why.

src/Interfaces/LazySet.jl
```diff
@@ -682,3 +683,38 @@ end
...
+First, the check `X == Y` is performed which returns `true` if and only if the given sets are of the same type, and have the same values (modulo floating-point tolerance). Otherwise, the double inclusion check `X ⊆ Y && Y ⊆ X` is computed.
+
+### Examples
+
+```
+julia> X = BallInf([0.1, 0.2], 0.3)
+julia> Y = convert(HPolytope, X)
+julia> X == Y
+false
+julia> is_equivalent(X, Y)
```
```suggestion
julia> isequivalent(X, Y)
```

test/unit_util.jl
```diff
@@ -85,4 +86,12 @@ for _dummy_ in 1:1 # avoid global variable warnings
         B = LazySets.delete_zero_columns!(A)
         @test size(B) == (m, n-3)
     end
+
+    # Check if sets are equal
+    n = 3
+    I = Diagonal(ones(n))
+    X = HPolytope([I; -I], ones(2*n))
+    Y = convert(VPolytope, X)
+    @test LazySets.isequivalent(X, Y)
```
schillic:
```suggestion
    @test isequivalent(X, Y)
```

src/Interfaces/LazySet.jl
```diff
@@ -682,3 +683,38 @@ end
...
+### Output
+`true` iff `X` is equivalent to `Y`.
+
+## Algorithm
+
+First, the check `X == Y` is performed which returns `true` if and only if the given sets are of the same type, and have the same values (modulo floating-point tolerance). Otherwise, the double inclusion check `X ⊆ Y && Y ⊆ X` is computed.
+
+### Examples
+
+```
```
```suggestion
```jldoctest
```
This would detect the other error I reported above. However, you need to add blank lines after each command, like so:
```julia
julia> X = BallInf([0.1, 0.2], 0.3);

julia> Y = convert(HPolytope, X);

julia> X == Y
false

julia> isequivalent(X, Y)
true
```

ueliwechsler:
> The v1.2 builds run into timeouts. Not sure why.

Is this a general problem at the moment. Or does it have to do something with my code?

schillic:
> Is this a general problem at the moment.

I think so because it does not even pass the first test.

ueliwechsler:
Do you have any idea, what I could change to make it pass the tests?

schillic:
I actually tried and see the same result in my v1.2 version, so it is not an issue with the build server. I do not have time to debug now, but maybe you want to try running the tests locally in v1.2 and see why they do not pass your tests. If I had to guess, I would say that there is something with `Rational`s for your programs that does not terminate.

ueliwechsler:
Thanks, I will try it later on.

schillic:
Out of curiosity: What was the problem?

ueliwechsler:
> Out of curiosity: What was the problem?

I am not sure. I didn't had the motivation to see why it did not work, once it worked 😆. I changed two main things. First, I added the test at a different location `unit_polytopes.jl` and I removed the use of `Diagonal` and therefore the import of `LinearAlgebra: Diagonal`.

mforets:
The problem was somehow related to polyhedra dependency. For instance, leaving the test in `unit_util.jl` but changing it to

```julia
    # Check if sets are equal by double inclusion
    n = 2
    I = Diagonal(ones(n))
    X = HPolygon([I; -I], ones(2*n))
    Y = convert(VPolygon, X)
    @test isequivalent(X, Y)
```
worked for me. (I just didn't have time to communicate it yesterday!)

mforets:
Merge?

Cc @schillic

src/Interfaces/LazySet.jl
```diff
@@ -682,3 +683,44 @@ end
...
+
+julia> X == Y
+false
+
+julia> isequivalent(X, Y)
+true
+```
+"""
+function isequivalent(X::LazySet, Y::LazySet)
+    if X == Y
```
schillic: Since the inclusion checks are approximate, it could be more consistent to use `X ≈ Y` here. At least this would match the documentation. (So at least the documentation or the code have to be changed here.)
However, `≈` is not controlled by our internal precision variables, so it might also be confusing. Thoughts?

test/unit_Polytope.jl
```diff
@@ -228,6 +228,10 @@ for N in [Float64]
     Y = convert(HPolytope, X)
     @test Y ⊆ X
 
+    # double inclusion check
+    Z = convert(VPolytope, Y)
+    @test isequivalent(Y, Z)
```
schillic: Can you add a negative test as well?

src/Interfaces/LazySet.jl
```diff
@@ -682,3 +683,44 @@ end
...
+
+### Output
+
+`true` iff `X` is equivalent to `Y`.
+
+## Algorithm
+
+First, the check `X == Y` is performed which returns `true` if and only if the given sets are of the same type,
+and have the same values (modulo floating-point tolerance). Otherwise, the double inclusion check `X ⊆ Y && Y ⊆ X` is
+computed.
```
```suggestion
used.
```
I would say that a test is not computed.

src/Interfaces/LazySet.jl
ueliwechsler: I would suggest adding `≈`, then it fits the documentation.

If you want to make it less confusing, we could also just remove this preceding equality check.

test/unit_Polytope.jl
ueliwechsler: Good enough?

src/Interfaces/LazySet.jl
mforets: > I would suggest adding ≈, then it fits the documentation.

I think this is as good idea :thumbsup:

> If you want to make it less confusing, we could also just remove this preceding equality check.

We shouldn't remove this initial comparison: if the sets are of the same type, it is going to be more efficient than the double inclusion check.

> However, ≈ is not controlled by our internal precision variables

Good point, that should be fixed by using `_isapprox` in line 437.

mforets: > Good point, that should be fixed by using _isapprox in line 437.

This is now https://github.com/JuliaReach/LazySets.jl/pull/1824

test/unit_Polytope.jl
schillic: Thanks!

mforets:
Thanks a lot!

PR #1820, opened 2019-12-05
Fix case 0 for linear term in taylor model

mforets:
This PR fixes the conversion to a zonotope for the following vector of taylor models:

```julia
2-element Array{TaylorModels.TaylorModel1{TaylorN{Float64},Float64},1}:
   10.1 + 0.09999999999999964 x₁ + [0, 0]
                             0.0 + [0, 0]
```

In principle one can't construct a taylor model with 0 order, but for whatever reason the zero polynomial, after normalization, `Q`, in [this part of the code](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Approximations/overapproximate.jl#L1027), had zero order and in consequence `linear_polynomial` was failing because the implementation in Taylorseries uses `a[1]`,  but since they start at `a[0]` the term doesn't exist, see [here](https://github.com/JuliaDiff/TaylorSeries.jl/blob/53a865275d678280bfc002548bff0d7a18c99dbd/src/auxiliary.jl#L315).

cc: @lbenet

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -762,9 +762,16 @@ using .TaylorModels: Taylor1, TaylorN, TaylorModelN, TaylorModel1,
                      normalize_taylor, linear_polynomial,
                      constant_term, evaluate, mid
 
-# helper functions
-@inline get_linear_coeffs(p::Taylor1) = linear_polynomial(p).coeffs[2:end]
-@inline get_linear_coeffs(p::TaylorN) = linear_polynomial(p).coeffs[2].coeffs
+@inline function get_linear_coeffs(p::Taylor1, n)
+    return linear_polynomial(p).coeffs[2:end]
+end
+
+@inline function get_linear_coeffs(p::TaylorN, n)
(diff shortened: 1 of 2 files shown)


src/Approximations/overapproximate.jl
```diff
@@ -762,9 +762,16 @@ using .TaylorModels: Taylor1, TaylorN, TaylorModelN, TaylorModel1,
                      normalize_taylor, linear_polynomial,
                      constant_term, evaluate, mid
 
-# helper functions
-@inline get_linear_coeffs(p::Taylor1) = linear_polynomial(p).coeffs[2:end]
-@inline get_linear_coeffs(p::TaylorN) = linear_polynomial(p).coeffs[2].coeffs
+@inline function get_linear_coeffs(p::Taylor1, n)
+    return linear_polynomial(p).coeffs[2:end]
```
schillic: Just to make sure: Here you do not need to check `p.order == 0`?

mforets: Ok, i've added that case as well.

PR #1824, opened 2019-12-10
#1823 - Use Tolerance for getfield in comparison of sets of the same type

### LINKED ISSUE #1823: Use Tolerance for getfield in comparison of sets of the same type
https://github.com/JuliaReach/LazySets.jl/blob/932e804a6ea5590537c7bb53ae5c364403bb772e/src/Interfaces/LazySet.jl#L437

mforets:
Closes #1823.

### DIFF
--- a/src/Utils/comparisons.jl
+++ b/src/Utils/comparisons.jl
@@ -102,7 +102,7 @@ end
 
 default_tolerance(N::Type{<:Number}) = error("default tolerance for numeric type $N is not defined")
 default_tolerance(N::Type{<:Rational}) = Tolerance(zero(N), zero(N), zero(N))
-default_tolerance(N::Type{<:Int}) = Tolerance(zero(N), zero(N), zero(N))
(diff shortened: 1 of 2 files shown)


schillic:
The problem here is this:
```julia
ERROR: MethodError: no method matching _isapprox(::Array{Int64,1}, ::Array{Float64,1})
```

src/Utils/comparisons.jl
```diff
@@ -101,6 +101,8 @@ set_atol(N::Type{NT}, ε::NT) where {NT<:Number} = begin
 end
 
 default_tolerance(N::Type{<:Number}) = error("default tolerance for numeric type $N is not defined")
+default_tolerance(N::Type{<:Rational}) = Tolerance(zero(N), zero(N), zero(N))
+default_tolerance(N::Type{<:Int}) = Tolerance(zero(N), zero(N), zero(N))
```
schillic:
```suggestion
default_tolerance(N::Type{<:Integer}) = Tolerance(zero(N), zero(N), zero(N))
```

PR #1826, opened 2019-12-11
Update AbstractHyperrectangle support function

mforets:
See [this comment](https://github.com/JuliaReach/LazySets.jl/pull/1785#discussion_r356628746).

### DIFF
--- a/src/Interfaces/AbstractHyperrectangle.jl
+++ b/src/Interfaces/AbstractHyperrectangle.jl
@@ -292,11 +292,7 @@ function ρ(d::AbstractVector{N}, H::AbstractHyperrectangle{N}) where {N<:Real}
     c = center(H)
     res = zero(N)
     @inbounds for (i, di) in enumerate(d)
-        if di < zero(N)
-            res += di * (c[i] - radius_hyperrectangle(H, i))
-        elseif di > zero(N)
-            res += di * (c[i] + radius_hyperrectangle(H, i))
-        end
+        res += di * c[i] + abs(di) * radius_hyperrectangle(H, i)
     end
     return res
 end


schillic:
I would not merge this.

mforets:
Why? Do you have an example where this is worse than the version in master?

BTW i wonder how does the computation compare to using directly BLAS functions. I'll have a look.,

schillic:
It should be obvious that adding `n` multiplications is generally slower.

schillic:
Huh, I just tried and this version is actually faster. My world falls down.

schillic:
One advantage of the `master` version is for sparse vectors (the loop skips `0` entries in the vector).

```julia
julia> n = 100; H = Hyperrectangle(ones(n), ones(n)); d = zeros(n); d[1] = 1.;

julia> @btime ρ($d, $H);  # master
  96.579 ns (0 allocations: 0 bytes)

julia> @btime ρ($d, $H);  # this branch
  125.644 ns (0 allocations: 0 bytes)
```

mforets:
That's fair enough. In [this notebook](https://nbviewer.jupyter.org/github/mforets/escritoire/blob/master/sets/Hyperrectangle_supp.ipynb) i added some other benchmarks, that may be useful if we want to review this again in the future.

schillic:
Woops, sorry, I had the wrong numbers in the post [here](https://github.com/JuliaReach/LazySets.jl/pull/1826#issuecomment-564582595). The text was correct. The version here was faster for positive dense vectors.

mforets:
Too late :laughing:  The last figure in [the notebook](https://nbviewer.jupyter.org/github/mforets/escritoire/blob/master/sets/Hyperrectangle_supp.ipynb) shows that `ρ2b` is the fastest in almost all cases, by a factor between 1 and 1.8 with respect to the function in master, `ρ0`.

PR #1833, opened 2019-12-12
#1337 - Handle inexact vectors in ispermutation

### LINKED ISSUE #1337: Handle inexact vectors in ispermutation
If the vectors are floating points, `ispermutation` may fail:

```julia
function ispermutation(u::AbstractVector{T}, v::AbstractVector{T})::Bool where T
    ...
    occurrence_map = Dict{T, Int}()
    has_duplicates = false
    for e ∈ u
        if e ∉ v   
            return false
        end
....
```
It would be nice to add another function that can handle vectors that were originated by different means, in floating point, or add a tolerance keyword argument to `ispermutation`.

mforets:
See #1337.

### DIFF
--- a/src/Utils/comparisons.jl
+++ b/src/Utils/comparisons.jl
@@ -134,7 +134,7 @@ Determine if `x` is smaller than or equal to `y`.
 
 - `x`      -- number
 - `y`      -- another number (of possibly different numeric type than `x`)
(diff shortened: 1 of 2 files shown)


schillic:
I removed the `Closes` above because there is another bug.

mforets:
Ok, thanks.

Issue #1357, opened 2019-05-09
Add star set representation


mforets:
Proposal: add the symbolic set representation of generalized stars.

This set representation has applications in reachability analysis [1].

References:

- [Duggirala, P. S., & Viswanathan, M. 2016, July. Parsimonious, simulation based verification of linear systems. In International Conference on Computer Aided Verification pp. 477-494. Springer, Cham.](http://cs.unc.edu/~psd/files/research/2016/psd-mv-cav-16.pdf)

schillic:
Stars are also used in Hylaa.

mforets:
Ok thanks for the pointer, i thought my ref was the same set of people but it is not, although the intersection is non-empty ([HyLAA ref](http://stanleybak.com/hylaa/)).

mforets:
Star sets used also in [Safety Verification of Cyber-Physical Systems with Reinforcement Learning Control](https://dl.acm.org/doi/abs/10.1145/3358230).

Here is a tentative definition:

```julia
# it is not generally convex but i make it <: LazySet
struct Star{N<:Real, VN<:AbstractVector{N}} <: LazySet{N}
    c::VN # center
    V::Vector{VN} # basis
    P::HPolyhedron{N} # predicate
end
```
In particular, the predicate function is implemented as a conjunction of linear constraints.

PR #1834, opened 2019-12-12
#1838 - Convex type function

### LINKED ISSUE #1838: Add isconvextype
PR:

- #1834

Related:

- #1837

### LINKED ISSUE #1844: Convex type for AbstractCentrallySymmetric sets
Proposal: to let

```julia
isconvextype(::Type{<:AbstractCentrallySymmetric}) = false
```
Currently (after https://github.com/JuliaReach/LazySets.jl/pull/1834), it returns an error.

An example of such sets is the star set, https://github.com/JuliaReach/LazySets.jl/issues/1357.

mforets:
Closes #1838.
Closes #1844.

--- 

To be added in the "Examples" section of the docstring of `isconvextype` (on a different PR)

````
On the other hand, the complement of a set is not convex, hence the intersection of
the lazy complement of a set with a ball in the infinity norm can no longer be
guaranteed to be convex, and `isconvextype` returns `false` in this case:

```jldoctest convex_types
julia> isconvextype(Intersection{Float64, Complement{Float64,BallInf{Float64}}, BallInf{Float64}})
false
```
````

### DIFF
--- a/src/Interfaces/AbstractCentrallySymmetric.jl
+++ b/src/Interfaces/AbstractCentrallySymmetric.jl
@@ -26,6 +26,7 @@ julia> subtypes(AbstractCentrallySymmetric)
 """
 abstract type AbstractCentrallySymmetric{N<:Real} <: LazySet{N} end
 
+isconvextype(::Type{<:AbstractCentrallySymmetric}) = false
 
 """
     dim(S::AbstractCentrallySymmetric)::Int
--- a/src/Interfaces/AbstractCentrallySymmetric.jl
+++ b/src/Interfaces/AbstractCentrallySymmetric.jl
@@ -26,7 +26,7 @@ julia> subtypes(AbstractCentrallySymmetric)
 """
 abstract type AbstractCentrallySymmetric{N<:Real} <: LazySet{N} end
 
-isconvextype(::Type{<:AbstractCentrallySymmetric}) = error("the `isconvextype` function for `AbstractCentrallySymmetric` sets is not implemented")
+isconvextype(::Type{<:AbstractCentrallySymmetric}) = false
 
 """
     dim(S::AbstractCentrallySymmetric)::Int
--- a/src/Interfaces/AbstractCentrallySymmetric.jl
+++ b/src/Interfaces/AbstractCentrallySymmetric.jl
@@ -26,7 +26,7 @@ julia> subtypes(AbstractCentrallySymmetric)
 """
 abstract type AbstractCentrallySymmetric{N<:Real} <: LazySet{N} end
 
-isconvextype(::Type{<:AbstractCentrallySymmetric}) = false
+isconvextype(::Type{<:AbstractCentrallySymmetric}) = error("the `isconvextype` function for `AbstractCentrallySymmetric` sets is not implemented")
 
 """
     dim(S::AbstractCentrallySymmetric)::Int
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -19,7 +19,8 @@ export LazySet,
        is_interior_point,
        isoperation,
        isoperationtype,
-       isequivalent
+       isequivalent,
+       isconvextype
 
 """
     LazySet{N}
@@ -723,4 +724,56 @@ function isequivalent(X::LazySet, Y::LazySet)
         return true
     end
     return X ⊆ Y && Y ⊆ X
+end
+
+"""
+    isconvextype(X::Type{<:LazySet})
+
+Check whether the given `LazySet` type is convex.
+
+### Input
+
+- `X` -- subtype of `LazySet`
+
+### Output
+
+`true` if the given set type is guaranteed to be convex by using only type
+information, and `false` otherwise.
+
+### Notes
+
+Since this operation only acts on types (not on values), it can return false
+negatives, i.e. there may be instances where the type is convex, even though the
+answer of this function is `false`. The examples below illustrate this point.
+
+### Examples
+
+A ball in the infinity norm is always convex, hence we get:
+
+```jldoctest convex_types
+julia> isconvextype(BallInf)
+true
+```
+
+For instance, the union (`UnionSet`) of two sets may in general be either convex
+or not, since convexity cannot be decided by just using type information.
+Hence, `isconvextype` returns `false` if `X` is `Type{<:UnionSet}`.
+
+```jldoctest convex_types
+julia> isconvextype(UnionSet)
+false
+```
+
+However, the type parameters from the set operations allow to decide convexity
+in some cases, by falling back to the convexity of the type of its arguments.
+Consider for instance the lazy intersection. The intersection of two convex sets
+is always convex, hence we can get:
+
+```jldoctest convex_types
+julia> isconvextype(Intersection{Float64, BallInf{Float64}, BallInf{Float64}})
+true
+```
+"""
+function isconvextype(X::Type{<:LazySet})
+    error("`isconvextype` is not implemented for type $X")
 end
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -743,7 +743,7 @@ information, and `false` otherwise.
 ### Notes
 
 Since this operation only acts on types (not on values), it can return false
-negatives, i.e. there may be instances where the type is convex, even though the
+negatives, i.e. there may be instances where the set is convex, even though the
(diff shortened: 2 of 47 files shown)

### CODE REVIEW

File: src/Sets/PolynomialZonotope.jl [hunk 1]
```diff
@@ -88,6 +88,7 @@ struct PolynomialZonotope{N}
 end
 
 isoperationtype(::Type{<:PolynomialZonotope}) = false
+isconvextype(::Type{<:PolynomialZonotope}) = true
```
mforets:
```suggestion
isconvextype(::Type{<:PolynomialZonotope}) = false
```

schillic:
> To be added in the "Examples" section of the docstring of isconvextype (possibly on a different PR)

Is there a reason why you do not want to add that here? I find that useful.

src/LazyOperations/Complement.jl
```diff
@@ -36,6 +38,7 @@ struct Complement{N<:Real, S<:LazySet{N}}
 end
 
 isoperationtype(::Type{<:Complement}) = true
+isconvextype(::Type{<:Complement}) = false
```
schillic: At some point (here or later) we should add that the `Complement` of `HalfSpace`, `EmptySet`, and `Universe` is convex (these are the only cases I can think of).

src/LazyOperations/AffineMap.jl
```diff
@@ -105,6 +105,7 @@ struct AffineMap{N<:Real, S<:LazySet{N}, NM, MAT<:AbstractMatrix{NM},
 end
 
 isoperationtype(::Type{<:AffineMap}) = true
+isconvextype(::Type{AffineMap{N, S, NM, MAT, VN}}) where {N, S, NM, MAT, VN} = isconvextype(S)
```
schillic: You only need to specify the type parameters until you get to `S`. I find that easier to read, but we can keep it now.

src/Interfaces/LazySet.jl
```diff
@@ -724,3 +725,55 @@ function isequivalent(X::LazySet, Y::LazySet)
...
+
+### Output
+
+`true` if the given set type is guaranteed to be convex by using only type
+information, and `false` otherwise.
+
+### Notes
+
+Since this operation only acts on types (not on values), it can return false
+negatives, i.e. there may be instances where the type is convex, even though the
```
```suggestion
negatives, i.e. there may be instances where the set is convex, even though the
```

src/LazyOperations/Complement.jl
mforets: Good point. I've opened https://github.com/JuliaReach/LazySets.jl/issues/1841 to follow up.

src/LazyOperations/AffineMap.jl
mforets: Ok. I'll keep this in mind.

mforets:
> We should add tests for this, but that can be done later.

Feel free to open an issue to follow up on that.

mforets:
> Is there a reason why you do not want to add that here? I find that useful.

The example won't work until `Complement` subtypes `LazySet`:

```julia
julia> Intersection{Float64, Complement{Float64,BallInf{Float64}}, BallInf{Float64}}
ERROR: TypeError: in Intersection, in S1, expected S1<:LazySet{N<:Real}, got Type{Complement{Float64,BallInf{Float64}}}
Stacktrace:
 [1] top-level scope at REPL[6]:1
```

mforets:
Thanks for the review. I'll merge after CI is green (but possibly tomorrow).

Note that i didn't apply `isconvextype` to the set interfaces, such as `AbstractHyperrectangle`. I don't know if it convenient to have that as well. Seems like it would simplify a bit some cases (and we can remove for example the dispatch on `BallInf`, etc.)

schillic:
> The example won't work until Complement subtypes LazySet

I see.

> Note that i didn't apply isconvextype to the set interfaces, such as AbstractHyperrectangle. I don't know if it convenient to have that as well. Seems like it would simplify a bit some cases (and we can remove for example the dispatch on BallInf, etc.)

Right, I also wondered why. Just like for other methods, that would save lines of code (both now and in the future when adding new set types).

src/Interfaces/AbstractCentrallySymmetric.jl
```diff
@@ -26,6 +26,7 @@ julia> subtypes(AbstractCentrallySymmetric)
 """
 abstract type AbstractCentrallySymmetric{N<:Real} <: LazySet{N} end
 
+isconvextype(::Type{<:AbstractCentrallySymmetric}) = false
```
schillic: This requires some discussion. Currently all subtypes of `AbstractCentrallySymmetric` are convex. Clearly, the name does not require convexity. But that is also true for `AbstractPolyhedron` etc., so the question is: Do we want to keep `AbstractPolyhedron` for convex polyhedra? Do we want to rename the interfaces?

mforets: I had in mind the star sets, which are centrally symmetric but are not convex. There is an open issue: https://github.com/JuliaReach/LazySets.jl/issues/1357

I didn't propose to rename the interfaces. But if you feel like this needs further discussion we can follow up next thursday on the telco. In this case, i would let `AbstractCentrallySymmetric` outside this PR and let it for follow up work.

schillic: > In this case, i would let AbstractCentrallySymmetric outside this PR and let it for follow up work.

:+1:

schillic: By the way, the answer `false` for interfaces does not make sense to me. If we do that, then the default method for `LazySet` should return `false` instead.

mforets:
```suggestion
isconvextype(::Type{<:AbstractCentrallySymmetric}) = error("the `isconvextype` function for `AbstractCentrallySymmetric` sets is not implemented")
```

mforets: > By the way, the answer false for interfaces does not make sense to me. 

Why? It means that the interface type does not guarantee that all subtypes are convex.

>  If we do that, then the default method for LazySet should return false instead.

Yes.

mforets: I left `isconvextype` for a general LazySet to gives an error, because it is the higher supertype and it's used as a fallback. But i can also code it to return `false` instead of an error. Thay may be even more consistent as `isconvextype(MinkowsiSum)` would now give `false` instead of giving  the error `ERROR: `isconvextype` is not implemented for type MinkowskiSum`, which is the behavior in this branch.

schillic: > But i can also code it to return false instead of an error. Thay may be even more consistent

I totally agree.

mforets: Ok, done.

schillic: And for `AbstractCentrallySymmetric` you want to keep it?

mforets: No, i propose to do https://github.com/JuliaReach/LazySets.jl/issues/1844. Following this thread we would discuss this in the future, but i can add it here as well.

schillic: Yes, I suggest to do it here. Why introduce something and change it immediately?
Actually, now you can just remove the method.

src/Interfaces/AbstractCentrallySymmetric.jl
```diff
@@ -26,6 +26,7 @@ julia> subtypes(AbstractCentrallySymmetric)
 """
 abstract type AbstractCentrallySymmetric{N<:Real} <: LazySet{N} end
 
+isconvextype(::Type{<:AbstractCentrallySymmetric}) = error("the `isconvextype` function for `AbstractCentrallySymmetric` sets is not implemented")
```
mforets:
```suggestion
isconvextype(::Type{<:AbstractCentrallySymmetric}) = false
```

src/LazyOperations/LinearMap.jl
```diff
@@ -108,6 +108,7 @@ struct LinearMap{N<:Real, S<:LazySet{N},
 end
 
 isoperationtype(::Type{<:LinearMap}) = true
+isconvextype(::Type{LinearMap{N, S}}) where {N, S} = isconvextype(S)
```
mforets: This is wrong: the type argument should either specify *all* the parameters, *or* use `<:` as in `Type{<:LinearMap{N, S}}`.

PR #1847, opened 2019-12-14
#1841 - Convexity of the complement

### LINKED ISSUE #1841: Convexity of the complement
At some point (here or later) we should add that the `Complement` of `HalfSpace`, `EmptySet`, and `Universe` is convex (these are the only cases I can think of).

_Originally posted by @schillic in https://github.com/JuliaReach/LazySets.jl/pull/1834_

mforets:
Closes #1841.

### DIFF
--- a/src/LazyOperations/Complement.jl
+++ b/src/LazyOperations/Complement.jl
@@ -43,9 +43,9 @@ isoperationtype(::Type{<:Complement}) = true
 isconvextype(::Type{<:Complement}) = false
 
 # special cases which are always convex
-isconvextype(::Type{Complement{<:HalfSpace}}) = true
-isconvextype(::Type{Complement{<:EmptySet}}) = true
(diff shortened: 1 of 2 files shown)


src/LazyOperations/Complement.jl
```diff
@@ -38,8 +38,15 @@ struct Complement{N<:Real, S<:LazySet{N}}
 end
 
 isoperationtype(::Type{<:Complement}) = true
+
+# the set complement is not convex in general
 isconvextype(::Type{<:Complement}) = false
 
+# special cases which are always convex
+isconvextype(::Type{Complement{<:HalfSpace}}) = true
```
schillic: This does not work. The first type parameter is the number type. Please add tests to avoid this.

mforets: Good catch, thanks.

PR #1848, opened 2019-12-14
#1845 - Make polynomial zonotope subtype lazyset

### LINKED ISSUE #1845: Make PolynomialZonotope a LazySet

mforets:
Closes #1845.

### DIFF
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -55,7 +55,7 @@ The subtypes of `LazySet` (including abstract interfaces):
 
 ```jldoctest; setup = :(using LazySets: subtypes)
 julia> subtypes(LazySet, false)
-19-element Array{Any,1}:
+20-element Array{Any,1}:
  AbstractCentrallySymmetric
  AbstractPolyhedron
  AffineMap
@@ -73,6 +73,7 @@ julia> subtypes(LazySet, false)
  LinearMap
  MinkowskiSum
  MinkowskiSumArray
+ PolynomialZonotope
  ResetMap
  Translation
 ```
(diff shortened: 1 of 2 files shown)


mforets:
Sorry, can you elaborate on "until we replaced all code parts with the isconvextype trait" ?

schillic:
Algorithms that rely on their argument(s) being convex should dispatch on `isconvextype` instead. For instance, epsilon-close overapproximation of a polynomial zonotope would be incorrect (i.e., not actually epsilon-close).

mforets:
Yes, sure, but that should better be done incrementally, in separate PRs. Think about the `master` branch as the in-development branch. Updating polynomial zonotopes is of independent interest, as in the other cases.

Moreover, note that docstrings already say ".. polygonal approximation of a convex set.". And `isconvextype` of the polynomial zonotope is false. So, it is arguable that something should be done at all. I think it is a good idea to use dispatch in some cases, but in the end, it is a misuse of the function.

schillic:
> Yes, sure, but that should better be done incrementally, in separate PRs.

Agreed, but it should be done in the right order. Breaking things first is not the right order for me.

mforets:
Ok, i think we agree. My point is that this issue *can* be merged independently of making issues that use `isconvextype` because *it doesn't break anything*. There are no tests that will be broken. Calling a function with a non-convex set, while the function needs a convex set, is ultimately a misuse of the function. It will always happen that you can break things. Here we want to use dispatch to not allow such behaviors, which is fine, but it doesn't mean things are broken.

mforets:
> Here we want to use dispatch to not allow such behaviors,

And more interestingly, to selectively apply algorithms based on the property.

mforets:
There are no actionable changes to this PR. It is now blocked by https://github.com/JuliaReach/LazySets.jl/issues/1856 and by https://github.com/JuliaReach/LazySets.jl/issues/1855.

mforets:
Closing as stale; we can merge this and the related PRs in one go..

PR #1850, opened 2019-12-16
support vector particular dispatches for sev

### LINKED ISSUE #1821: Faster hyperrectangle support vector for SingleEntryVector
Follow-up of https://github.com/JuliaReach/LazySets.jl/pull/1785

dfcaporale:
Closes #1821.

### DIFF
--- a/src/Interfaces/AbstractHyperrectangle.jl
+++ b/src/Interfaces/AbstractHyperrectangle.jl
@@ -277,12 +277,12 @@ end
 function _σ_sev_hyperrectangle(d::SingleEntryVector{N}, H::AbstractHyperrectangle{N}) where {N<:Real}
 
     @assert d.n == dim(H) "a $(d.n)-dimensional vector is " *
-                                "incompatible with a $(dim(H))-dimensional set"
+                          "incompatible with a $(dim(H))-dimensional set"
     s = copy(center(H))
     if d.v < zero(N)
-      s[d.i] -= radius_hyperrectangle(H, d.i)
+        s[d.i] -= radius_hyperrectangle(H, d.i)
     else
-      s[d.i] += radius_hyperrectangle(H, d.i)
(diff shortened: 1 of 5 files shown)


src/Interfaces/AbstractHyperrectangle.jl
```diff
@@ -273,6 +273,20 @@ function σ(d::AbstractVector{N}, H::AbstractHyperrectangle{N}) where {N<:Real}
...
 end
 
+# helper function for single entry vector
+function _σ_sev_hyperrectangle(d::SingleEntryVector{N}, H::AbstractHyperrectangle{N}) where {N<:Real}
+
+    @assert d.n == dim(H) "a $(d.n)-dimensional vector is " *
+                                "incompatible with a $(dim(H))-dimensional set"
+    s = copy(center(H))
+    if d.v < zero(N)
+      s[d.i] -= radius_hyperrectangle(H, d.i)
```
mforets: For indentation we use 4 spaces instead of 2.

mforets:
Since only one value of the center is changed, instead of all of them, this PR changes the point which is returned, from a vertex in a corner, to the point aligned with the direction of the ball. I don't consider this breaking though, as we don't guarantee that the point is always the same for different implementations. For instance,

```julia
julia> σ(d, b) # here d is the single entry vector [1.0, 0.0]
2-element Array{Float64,1}:
 1.0
 0.0

julia> σ([1.0, 0.0], b) # same result as in master for d
2-element Array{Float64,1}:
 1.0
 1.0
```

schillic:
> Since only one value of the center is changed, instead of all of them, this PR changes the point which is returned, from a vertex in a corner, to the point aligned with the direction of the ball. I don't consider this breaking though, as we don't guarantee that the point is always the same for different implementations.

Agreed. That is why a test should not rely on those values.

mforets:
> Tests with SEVs would be nice. Example:

I added this test in the last commit.

PR #1859, opened 2019-12-19
#1856 - Add check for convex type in set inclusion of line segment in a convex set

### LINKED ISSUE #1856: Restrict generic inclusion between line segment and lazy sets to convex type
Idea from [this comment](https://github.com/JuliaReach/LazySets.jl/pull/1848#issuecomment-565801073).

mforets:
Closes #1856.

### DIFF
--- a/src/ConcreteOperations/issubset.jl
+++ b/src/ConcreteOperations/issubset.jl
@@ -552,8 +552,7 @@ end
 
 
 """
-    ⊆(L::LineSegment{N}, S::LazySet{N}, [witness]::Bool=false
-     )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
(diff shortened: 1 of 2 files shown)


src/ConcreteOperations/issubset.jl
```diff
@@ -576,8 +575,13 @@ optionally compute a witness.
 Since ``S`` is convex, ``L ⊆ S`` iff ``p ∈ S`` and ``q ∈ S``, where ``p, q`` are
 the end points of ``L``.
 """
-function ⊆(L::LineSegment{N}, S::LazySet{N}, witness::Bool=false
-          )::Union{Bool, Tuple{Bool, Vector{N}}} where {N<:Real}
+function ⊆(L::LineSegment{N}, S::ST, witness::Bool=false) where {N<:Real, ST<:LazySet{N}}
+
+    if !isconvextype(ST)
```
mforets, line 580: it would be preferable to use a trait an obtain a NotImplementedError here

PR #1861, opened 2019-12-19
#1855 - Restrict epsilon-close approx to convex sets

### LINKED ISSUE #1855: Restrict epsilon-close approx to convex sets
Idea from [this comment](https://github.com/JuliaReach/LazySets.jl/pull/1848#issuecomment-565801073).

mforets:
Closes #1855.

This PR implements a restriction on the argument type such that epsilon-close overapproximation of a set which is not guaranteed to be convex by its type is forbidden.

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -28,19 +28,25 @@ end
                     ::Type{<:HPolygon},
                     [ε]::Real=Inf) where {N<:Real}
 
-Return an approximation of a given 2D convex set.
-If no error tolerance is given, or is `Inf`, the result is a box-shaped polygon.
-Otherwise the result is an ε-close approximation as a polygon.
+Return an approximation of a given 2D convex set using iterative refinement.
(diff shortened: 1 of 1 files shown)


mforets:
Ok, i agree to revert the commit and update the docstring instead.

src/Approximations/overapproximate.jl
```diff
@@ -26,26 +26,31 @@ end
 """
     overapproximate(S::LazySet{N},
                     ::Type{<:HPolygon},
-                    [ε]::Real=Inf)::HPolygon where {N<:Real}
+                    [ε]::Real=Inf) where {N<:Real}
 
-Return an approximation of a given 2D convex set.
-If no error tolerance is given, or is `Inf`, the result is a box-shaped polygon.
-Otherwise the result is an ε-close approximation as a polygon.
+Return an approximation of a given 2D convex set using iterative refinement.
```
```suggestion
Return an approximation of a given 2D set using iterative refinement.
```

PR #1862, opened 2019-12-20
#1605 - Split page for set types in documentation

### LINKED ISSUE #1605: Split page for set types in documentation
We have three page for set types: interfaces, concrete set types, and lazy set types.
I suggest that we split them into single pages for each set type to make it easier to find something.

schillic:
Closes #1605.

As a comment for the near future: I noticed that there are some sentences about non-convex sets not being part of the `LazySet` interface. These will need to be changed.


mforets:
I think this change is in the good direction but requires more work:

- Rename the interface sections to their real names, eg. `AbstractHyperrectangle` instead of `Hyperrectangle`, or `AbstractPolygon` instead of `Polygon`.

- Allow top-level browsing for the (concrete) set representations. They are now not visible from the search panel.

schillic:
> Rename the interface sections to their real names, eg. AbstractHyperrectangle instead of Hyperrectangle, or AbstractPolygon instead of Polygon.

Okay, these were the old names (just to say).

> Allow top-level browsing for the (concrete) set representations. They are now not visible from the search panel.

I agree that a quick access is neat, but I think there are too many set types for the top level (you would not find anything else anymore). We could have a section page "Quick links" on the first page. Can you explain how you imagine the panel?

schillic:
I added a new version addressing the comments.

mforets:
> Can you explain how you imagine the panel?

This discussion has things in common with the one for [adding src files into subfolders](https://github.com/JuliaReach/LazySets.jl/issues/15790) that we had a while ago.

I find that the panel names `Interfaces`, `Sets`, `Lazy Operations`, `Concrete Operations` is satisfactory. I have some preference of that choice over using `Set Interfaces`, `Concrete set representations`, `Lazy set representations`, and `Binary Functions on Sets`. One reason is that the former choice is sort of less ambiguous; "concrete sets" are also lazy in several ways.

schillic:
> I find that the panel names Interfaces, Sets, Lazy Operations, Concrete Operations is satisfactory.

Alright, I changed them (and moved the "Concrete Operations" up).

docs/make.jl
```diff
@@ -12,31 +12,70 @@ makedocs(
...
-        "Iterative Refinement" => "man/iterative_refinement.md",
-        "Interval Hulls" => "man/interval_hulls.md",
-        "Convex Hulls" => "man/convex_hulls.md",
-        "Operations on Sets" => "man/set_operations.md",
-        "A Reachability Algorithm" => "man/reach_zonotopes.md",
-        "A Hybrid Reachability Algorithm" => "man/reach_zonotopes_hybrid.md",
-        "Concrete Polyhedra" => "man/concrete_polyhedra.md",
-        "Parallel Approximations" => "man/parallel_approximations.md",
-        "Lazy Intersections" => "man/lazy_intersections.md"
+            "Manual" => Any[
```
mforets:
```suggestion
        "Manual" => Any[
```

schillic:
These commits were supposed to stay separated...

PR #1863, opened 2019-12-21
#1682 - Add element type function

### LINKED ISSUE #1682: Function to obtain the numeric type of a set
It is sometimes helpful to obtain the numeric type (`N`) of a set without having to do dispatch, in particular for results of function calls (e.g., in unit tests).
I think `eltype` would be a good function name.
Since we use `N` as the first type parameter for sets, the implementation is simple:
```julia
@inline function eltype(::Type{<:LazySet{N}}) where {N}
    return N
end

@inline function eltype(::LazySet{N}) where {N}
    return N
end
```

mforets:
Closes #1682.


mforets:
Thanks for the review !

Issue #1839, opened 2019-12-12
Refactor array-set types to their own files


mforets:
From https://github.com/JuliaReach/LazySets.jl/issues/215

- [x] Minkowski sum array
- [x] UnionSetArray
- [x] ConvexHullArray
- [x] CartesianProductArray
- [x] IntersectionArray

schillic:
`IntersectionArray` is missing.

PR #1865, opened 2019-12-21
#1839 - Move CPA to its own file

mforets:
See #1839.


src/LazyOperations/CartesianProductArray.jl
```diff
@@ -0,0 +1,584 @@
+import Base: *, ∈, isempty
+
+export CartesianProductArray,
+       array,
+       swap,
```
schillic:
```suggestion
```

src/LazyOperations/CartesianProductArray.jl
```diff
@@ -0,0 +1,584 @@
...
+```jldoctest
+julia> using LazySets: block_structure
+
+julia> cpa = CartesianProductArray([BallInf(zeros(n), 1.0) for n in [3, 1, 2]]);
+
+julia> block_structure(cpa)
+3-element Array{UnitRange{Int64},1}:
+1:3
+4:4
+5:6
```
schillic: There needs to be a leading space in these lines.

mforets:
Thanks for the review!

PR #1866, opened 2019-12-22
Fix docs

schillic:
~In #1863 there were no type restrictions, but Julia/`Documenter` do not like that. The build in that PR probably passed due to nondeterminism.~

That again apparently only worked on my side due to nondeterminism.


mforets:
Thanks!

schillic:
The problem still persists...

PR #1868, opened 2019-12-22
#1619 - Add an_element for polyhedron

### LINKED ISSUE #1619: Problems with an_element for rational HPolyhedron
```julia
julia> an_element(HPolyhedron{Rational{Int}}())  # empty list of constraints crashes
ERROR: ArgumentError: An index is out of bound.

julia> an_element(HPolyhedron([HalfSpace([1//1], 0//1)]))  # numeric type changes
1-element Array{Float64,1}:
 0.0
```

See also #1105.
The second point should be ignored (see #1711).

mforets:
Closes #1619.

### DIFF
--- a/src/Interfaces/AbstractPolyhedron_functions.jl
+++ b/src/Interfaces/AbstractPolyhedron_functions.jl
@@ -5,7 +5,8 @@ export constrained_dimensions,
        remove_redundant_constraints,
        remove_redundant_constraints!,
        linear_map,
-       chebyshev_center
+       chebyshev_center,
+       an_element
 
 # default LP solver for floating-point numbers
 function default_lp_solver(N::Type{<:AbstractFloat})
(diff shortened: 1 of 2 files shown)


src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -578,3 +579,30 @@ function chebyshev_center(P::AbstractPolyhedron{N};
...
+
+An element of a polyhedron.
+
+### Algorithm
+
+An element of the polyhedron is obtained by evaluating its support vector along
+direction ``[1, 0, …, 0]``.
+"""
+function an_element(P::AbstractPolyhedron{N}) where {N<:Real}
+    if dim(P) == -1
```
schillic: Maybe store `dim(P)`?

mforets: ok

mforets: (i think there is no difference in performance, just for readability)

mforets:
> Why did you not do #1867 here?

It is unrelated to #1619.

schillic:
> It is unrelated to #1619.

I do not see why you are not allowed to implement an *efficient* fix instead of just any fix.

mforets:
Keeping `sparsevec` in this branch is the current behavior for all polyhedra. So i take less risks :p . In #1619 i will propose to do both changes as once (this and the generic `an_element`).

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -578,3 +579,31 @@ function chebyshev_center(P::AbstractPolyhedron{N};
...
 end
+
+"""
+    an_element(P::AbstractPolyhedron{N}) where {N<:Real}
+
+Return some element of a convex set.
+
+### Input
+
+- `S` -- convex set
```
```suggestion
- `P` -- polyhedron
```

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -578,3 +579,31 @@ function chebyshev_center(P::AbstractPolyhedron{N};
...
+### Output
+
+An element of a polyhedron.
+
+### Algorithm
+
+An element of the polyhedron is obtained by evaluating its support vector along
+direction ``[1, 0, …, 0]``.
+"""
+function an_element(P::AbstractPolyhedron{N}) where {N<:Real}
```
schillic: This requires a new entry in the docs.

```
┌ Warning: 1 docstring potentially missing:
│ 
│     LazySets.an_element :: Union{Tuple{AbstractPolyhedron{N}}, Tuple{N}} where N<:Real
```

PR #1872, opened 2019-12-27
#1871 - Documentation nondeterministically breaks due to missing eltype function

### LINKED ISSUE #1871: Documentation nondeterministically breaks due to missing eltype function
See #1866, which did not actually solve the problem.

```
[ Info: CheckDocument: running document checks.
┌ Warning: 1 docstring potentially missing:
│ 
│     Base.eltype :: Union{Tuple{Type{#s21} where #s21<:LazySet{N}}, Tuple{N}} where {N}
└ @ Documenter.DocChecks ~/.julia/packages/Documenter/IGqbY/src/DocChecks.jl:63

```

schillic:
Closes #1871.

### DIFF
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -89,8 +89,7 @@ plot3d!
 ==(::LazySet, ::LazySet)
 ≈(::LazySet, ::LazySet)
 copy(::LazySet)
-Base.eltype(::Type{<:LazySet{N}}) where {N}
-Base.eltype(::LazySet{N}) where {N}
+eltype
 ```
 
 ### Aliases for set types


mforets:
Did you see if the change show the docs for all the 59+2 `eltype` methods, or only the 2 methods defined in LazySets? I wonder what is distinct of this function wrt `copy` or other overloaded `Base` functions.

In other cases a workaround for similar (?) problems was to explicitly write the type parameter, as in

```julia
eltype(::Type{T}) where {N, T<:LazySet{N}}

schillic:
It shows only the 2 methods and fixes the issue (I tested several times). The reason why just writing the method name is sufficient is that we restrict the docs to our package:

https://github.com/JuliaReach/LazySets.jl/blob/d7b526f2f5ebf2fc0422df890b37adc274930bf7/docs/make.jl#L9

mforets:
Ok, sounds excellent!

PR #1873, opened 2019-12-27
Move reflect() to LazySet.jl

schillic:


mforets:
I prefer to keep concrete operations on sets in the folder `ConcreteOperations`. In other words, concrete operations are those that have (or are planned to have) a lazy counterpart. And keep the stuff in `src/Interfaces/LazySets.jl` to comparisons, or the generic functions on types. It seems that you want to go the other way round. It's fine, but i find this an unnecessary change.

schillic:
> I prefer to keep concrete operations on sets in the folder ConcreteOperations

True, but this only holds for *binary* operations.
We define all other *unary* operations together with the set types (or interfaces). Look at the files in [`Interfaces`](https://github.com/JuliaReach/LazySets.jl/tree/master/src/Interfaces)/[`Sets`](https://github.com/JuliaReach/LazySets.jl/tree/master/src/Sets)/[`LazyOperations`](https://github.com/JuliaReach/LazySets.jl/tree/master/src/LazyOperations) and you find concrete operations; example:

https://github.com/JuliaReach/LazySets.jl/blob/e2b6731f51c4c95459ffa2faf795950a0ba0d38d/src/Interfaces/LazySet.jl#L347-L349

> i find this an unnecessary change

This PR restores the above-mentioned organization of the library, so I find this semi-important.

mforets:
Ok, sounds perfect.

Issue #1279, opened 2019-04-09
Add Introduction to Convex Sets to the LazySets docs 


mforets:
See [these slides](https://schillic.github.io/resources/supplement/slides_ConvexSets2019.pdf).

* [x] basic sets ( #1876)
* [ ] set operations
* [ ] advanced sets
* [ ] support function

mforets:
Draft: https://github.com/JuliaReach/LazySets.jl/wiki/EP:-Documentation

PR #1849, opened 2019-12-15
#1846 - Make unions a LazySet

mforets:
Closes #1846.

Issue #1318, opened 2019-04-30
Plot recipe for set union


mforets:
The types `UnionSet` and `UnionSetArray` do not yet have plotting functionality.

PR #1876, opened 2019-12-27
#1279 - Introduction to convex sets: basic sets

schillic:
This is part 1 of #1279.


mforets:
> add LaTeXStrings requirement for docs

Shouldn't it go to `docs/Project.toml`?

docs/src/man/convex_sets.md
```diff
@@ -0,0 +1,258 @@
...
+
+function plot_hyperplane!(x, y, linecolor)
+    plot!(x, y, linecolor=linecolor, arrow=:arrow, width=8)
+    plot!([x[2], x[1]], [y[2], y[1]], linecolor=linecolor, arrow=:arrow, width=8)
+end
+```
+
+## Basic convex sets
+
+A set ``X ⊆ ℝⁿ`` is convex if for any two points ``x``, ``y`` in ``X`` the whole
```
mforets: We discussed this elsewhere, i know, but reading the ``ℝⁿ`` symbol in the generated doc looks like something is wrong with it the symbol. Since this is not docstring, we can use ``\\mathbb{R}^n`` which properly aligned the super-index ``n``.

docs/src/man/convex_sets.md
```diff
@@ -0,0 +1,258 @@
...
+    X = \{λ · ~ x + (1 − λ) · y \mid x, y ∈ X , λ ∈ [0, 1] ⊆ ℝ\}
+```
+
+Below we show an example of a convex set and a non-convex set.
+
+```@example convex_sets
+pentagon = VPolygon([[1., 2.], [2., 1.], [1.5, 0.], [0.5, 0.], [0., 1.]])
+
+plot_raw()  # hide
+plot!(xlims=(0., 2.), ylims=(0., 2.), xticks=[0., 1., 2.],  # hide
```
mforets: The tick markers are huge (at least when i visualize the generated docs). Can we make them smaller? Perhaps this can be tuned with the option `guidefontsize`, or `tickfontsize`, see the tables [here](https://docs.juliaplots.org/latest/attributes/).

mforets: .. or can we just use the usual plot recipes?, eg. show the command `plot(pentagon)` and the resulting plot.

docs/src/man/convex_sets.md
```diff
@@ -0,0 +1,258 @@
...
+      yticks=[1., 2.], top_margin=3mm)  # hide
+plot1(pentagon, true)  # hide
+```
+
+```@example convex_sets
+hourglass = [VPolygon([[0., 2.], [1., 0.9], [2, 2.]]),
```
mforets: This set can be defined as a `UnionSet` of two triangles, but it requires  https://github.com/JuliaReach/LazySets.jl/pull/1849 and https://github.com/JuliaReach/LazySets.jl/issues/1318.

docs/src/man/convex_sets.md
```diff
@@ -0,0 +1,258 @@
...
+      yticks=[1., 2.], top_margin=3mm)  # hide
+plot3(hourglass, true; linecolor=c3)  # hide
+```
+
+### Empty set
+
+The simplest convex set is the empty set.
+
+```@example convex_sets
+E = EmptySet()
```
mforets: There's nothing interesting about the empty set, i would remove the section and the plot. Or merge the section in another one.

schillic: This manual should eventually cover all convex sets in the library. It does not hurt to have one sentence about the empty set IMO.

docs/src/man/convex_sets.md
schillic: The reason was `tickfont=font(30, "Times"), guidefontsize=45`, which I used for the slides, but I agree that they were too big for a browser :+1:
I replaced that by `tickfont=font(15, "Times")` because the default font size is too small for my taste.

docs/src/man/convex_sets.md
schillic: I added the `UnionSet` now that we have #1318.

docs/src/man/convex_sets.md
```diff
@@ -0,0 +1,269 @@
...
+A set ``X ⊆ \mathbb{R}^n`` is convex if for any two points ``x``, ``y`` in ``X``
+the whole line segment connecting ``x`` and ``y`` is also contained in ``X``.
+Formally:
+
+```math
+    X = \{λ · x + (1 − λ) · y \mid x, y ∈ X, λ ∈ [0, 1] ⊆ \mathbb{R}\}
+```
+
+Below we show an example of a convex set and a non-convex set.
+
```
mforets, line 78: i would say here that `VPolygon` represents the convex hull of the given vertices, and add a sentence of about `HPolygon` too (in both cases with links to the types' docs / wikipedia)

docs/src/man/convex_sets.md
```diff
@@ -0,0 +1,269 @@
...
+hourglass = UnionSet(VPolygon([[0., 2.], [1., 0.9], [2, 2.]]),
+                     VPolygon([[0., 0.], [1., 1.1], [2, 0.]]))
+
+plot_raw()  # hide
+plot!(xlims=(0., 2.), ylims=(0., 2.), xticks=[0., 1., 2.],  # hide
+      yticks=[1., 2.], top_margin=3mm)  # hide
+plot3(hourglass, true; linecolor=c3)  # hide
+```
+
+### Empty set
```
mforets, line 98: i wouldn't have a whole subsection just this.. same below for singletons ¯\_(ツ)_/¯  .. it's like there are many other important sets which are missing here, so having these subsections makes me think they are all the types. otherwise , you can  add more text saying there are many other types and this is just for illustration.

docs/src/man/convex_sets.md
```diff
@@ -0,0 +1,269 @@
...
+![../assets/ball_two_thirds.png](../assets/ball_two_thirds.png)
+
+[Source: [Wikimedia](https://commons.wikimedia.org/wiki/File:Astroid.svg)]
+
+### Unbounded sets
+
+#### Hyperplane
+
+A hyperplane is the generalization of the line (in 2D) and the plane (in 3D).
+Formally, a hyperplane is parametric in a normal vector ``d`` and a constant
```
mforets, line 217: (i would use letters `a` and `b` because we use that in other parts of the documentation and the code)

docs/src/man/convex_sets.md
schillic: I added the `ZeroSet` that was missing. All other set types are supposed to be added later (lazy operations and composite types for which you need to know the operations).

PR #1880, opened 2019-12-28
#1878 #1879 - genmat for AbstractHyperrectangle, generalize Zonotope conversion

### LINKED ISSUE #1878: Define genmat for AbstractHyperrectangle
Currently we use the fallback implementation

https://github.com/JuliaReach/LazySets.jl/blob/85e4b30770a2d888a1b10951ed687e1ac49accbd/src/Interfaces/AbstractHyperrectangle.jl#L59-L61

which is

https://github.com/JuliaReach/LazySets.jl/blob/85e4b30770a2d888a1b10951ed687e1ac49accbd/src/Interfaces/AbstractZonotope.jl#L68-L74

However, in another place we use

https://github.com/JuliaReach/LazySets.jl/blob/85e4b30770a2d888a1b10951ed687e1ac49accbd/src/convert.jl#L258-L260

which seems more efficient.

### LINKED ISSUE #1879: Generalize conversion from AbstractHyperrectangle to Zonotope
The `convert` implementation

https://github.com/JuliaReach/LazySets.jl/blob/85e4b30770a2d888a1b10951ed687e1ac49accbd/src/convert.jl#L258-L260

should be generalized from `AbstractHyperrectangle` to `AbstractZonotope` and rely on `genmat` instead of a custom implementation (after doing #1878 this will be equivalent).

schillic:
Closes #1878.
Closes #1879.

### DIFF
--- a/src/Interfaces/AbstractHyperrectangle.jl
+++ b/src/Interfaces/AbstractHyperrectangle.jl
@@ -58,18 +58,15 @@ A matrix where each column represents one generator of `H`.
 """
 function genmat(H::AbstractHyperrectangle{N}) where {N<:Real}
     r = radius_hyperrectangle(H)
-    if any(iszero, r)
-        I = Int[]
-        vals = N[]
-        for (i, ri) in enumerate(r)
-            if !iszero(ri)
-                push!(I, i)
-                push!(vals, ri)
-            end
+    I = Int[]
+    vals = N[]
+    for (i, ri) in enumerate(r)
+        if !iszero(ri)
+            push!(I, i)
+            push!(vals, ri)
         end
-        return sparse(I, 1:length(I), vals, length(r), length(I))
     end
-    return Diagonal(r)
+    return sparse(I, 1:length(I), vals, length(r), length(I))
 end
 
 # iterator that wraps the generator matrix
--- a/src/Interfaces/AbstractHyperrectangle.jl
+++ b/src/Interfaces/AbstractHyperrectangle.jl
@@ -44,7 +44,7 @@ isconvextype(::Type{<:AbstractHyperrectangle}) = true
 
 
 """
-   genmat(H::AbstractHyperrectangle)
+   genmat(H::AbstractHyperrectangle{N}) where {N<:Real}
 
 Return the generator matrix of a hyperrectangular set.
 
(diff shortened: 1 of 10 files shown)


src/Interfaces/AbstractHyperrectangle.jl
```diff
@@ -56,8 +56,20 @@ Return the generator matrix of a hyperrectangular set.
 
 A matrix where each column represents one generator of `H`.
 """
-function genmat(H::AbstractHyperrectangle)
-    return genmat_fallback(H)
+function genmat(H::AbstractHyperrectangle{N}) where {N<:Real}
+    r = radius_hyperrectangle(H)
+    if any(iszero, r)
```
mforets: I would return a `Diagonal(sparse(r))` in general here. Other code can take advantage knowning that this matrix is `Diagonal`. And returning two different matrix types may introduce a type-instability.

schillic: This was my first implementation, but then a test fails. We assume that the generator matrix has no zero columns.

mforets: Do you remember which test fails? I don't see any reason why `Diagonal(r)` would be wrong; it probably uncovers another bug..?

schillic: https://github.com/JuliaReach/LazySets.jl/blob/fd488b412a0643e94f6e18a800b02bbaf3c2aaa3/test/unit_Hyperrectangle.jl#L26-L30

schillic: I changed it to always returning a `SparseMatrix`. Problem solved :+1:

mforets:
@schillic Do you remember why some tests needed to be changed / downgraded to Float64?

schillic:
> @schillic Do you remember why some tests needed to be changed / downgraded to Float64?

Yes, see the build errors:

```
  MethodError: no method matching qr(::SparseMatrixCSC{Float32,Int64}, ::Type{Val{true}}; tol=7.1525574f-6)
```

src/Interfaces/AbstractHyperrectangle.jl
```diff
@@ -56,8 +56,17 @@ Return the generator matrix of a hyperrectangular set.
 
 A matrix where each column represents one generator of `H`.
 """
-function genmat(H::AbstractHyperrectangle)
-    return genmat_fallback(H)
+function genmat(H::AbstractHyperrectangle{N}) where {N<:Real}
```
mforets: Consider the following implementation:

```julia
function genmat(H::AbstractHyperrectangle)
    r = radius_hyperrectangle(H)
    return dropzeros!(sparse(Diagonal(r)))
end
```

mforets:
> Yes, see the build errors:

For the conversions or the inclusion tests?

mforets:
I tried to make a comparative table and this is what i got (feel free to edit it):

||genmat|edge cases|breaks|adequacy|
|--------------|-------------|-----|----|-----|
|Diagonal|immediate (5ns, 16bytes)|contains redundant gens for the *flat* case||ops with diagonal matrices are very specialized/fast|
|sparse|not immediate (283ns,  2.67kb) with [this function](https://github.com/JuliaReach/LazySets.jl/pull/1880#discussion_r362442736)|no redundant gens for the flat case|tests with inclusion/conversion in different numeric fields ok (see [here](https://github.com/JuliaReach/LazySets.jl/pull/1880#issuecomment-570125037))|

In sum, i'm yet not 100% sure to return a sparse matrix instead of a diagonal in the generator matrix. I guess it is hard to evaluate without actual use cases..

schillic:
I tried `Diagonal` and tests were failing. I do not see how a `Diagonal` can be non-square. Feel free to open an alternative solution.

PR #1883, opened 2019-12-28
Refactor ConvexHullArray

mforets:
See https://github.com/JuliaReach/LazySets.jl/issues/1839.


src/LazyOperations/ConvexHull.jl
```diff
@@ -3,9 +3,7 @@ import Base.isempty
 export ConvexHull, CH,
        convex_hull,
        convex_hull!,
-       ConvexHullArray, CHArray,
        ConvexHull!,
-       array,
        swap
```
schillic: This is a method of the `CHArray` (but it does not matter where we write the `export`s).

mforets: > swap is a method of the CHArray

Are you sure?

schillic: Nah, that was incorrect.

PR #1884, opened 2019-12-28
#1686 - Add togrep methods for AbstractZonotopes

### LINKED ISSUE #1686: Add togrep methods for AbstractZonotopes
> Should we have a togrep method for AbstractZonotopes which returns a Zonotope/GParallelotope/Parallelotope depending on the type? (Actually, all subtypes except for Zonotope are parallelotopes.)

Ok, these conversions can be added as well.

_Originally posted by @mforets in https://github.com/JuliaReach/LazySets.jl/pull/1632#issuecomment-530735813_

schillic:
Closes #1686.

### DIFF
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -131,8 +131,8 @@ Return a generator representation of a zonotopic set.
 ### Output
 
 The same set in generator representation.
-This fallback implementation returns a `Zonotope`; in general, other generator
-representations may be returned.
+This fallback implementation returns a `Zonotope`; however, more specific
+implementations may return other generator representations.
 """
 function togrep(Z::AbstractZonotope)
     return convert(Zonotope, Z)
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -4,7 +4,8 @@ export AbstractZonotope,
        genmat,
        generators,
        ngens,
-       order
+       order,
+       togrep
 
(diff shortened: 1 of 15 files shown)


src/Interfaces/AbstractZonotope.jl
```diff
@@ -118,6 +119,25 @@ function generators_fallback(Z::AbstractZonotope{N}) where {N<:Real}
...
+
+### Input
+
+- `Z` -- zonotopic set
+
+### Output
+
+The same set in generator representation.
+This fallback implementation returns a `Zonotope`; in general, other generator
+representations may be returned.
```
mforets: > ; in general, other generator representations may be returned.

I don't find the part after the semicolon very clear.. can we delete it? Or say that to use other representation, try using `convert` method directly.

schillic: I wanted to say here: do not rely on the fact that `togrep` *always* returns a `Zonotope`. This is not about `convert`ing.

mforets: Ah, i see what you mean. The source of my confusion was the part ".. may be returned." (..by whom? Answer: *not* by this function; by other, specific implementations of `togrep` ).

schillic: Okay, would the following help?
```
This fallback implementation returns a `Zonotope`; however, more specific
implementations may return other generator representations.
```

mforets: yes

PR #1886, opened 2019-12-29
Refactor MSA

mforets:
See https://github.com/JuliaReach/LazySets.jl/issues/1839


src/LazyOperations/MinkowskiSumArray.jl
```diff
@@ -0,0 +1,165 @@
+export MinkowskiSumArray,
+       array,
```
mforets:
```suggestion
       array
```

Issue #1874, opened 2019-12-27
Iterative refinement fails for affine map with static vector


mforets:
```julia
julia> dump(Q)
AffineMap{Float64,Hyperrectangle{Float64,StaticArrays.SArray{Tuple{2},Float64,1,2},StaticArrays.SArray{Tuple{2},Float64,1,2}},Float64,Array{Float64,2},StaticArrays.SArray{Tuple{2},Float64,1,2}}
  M: Array{Float64}((2, 2)) [0.13999999999999999 0.33999999999999997; 0.2 0.1]
  X: Hyperrectangle{Float64,StaticArrays.SArray{Tuple{2},Float64,1,2},StaticArrays.SArray{Tuple{2},Float64,1,2}}
    center: StaticArrays.SArray{Tuple{2},Float64,1,2}
      data: Tuple{Float64,Float64}
        1: Float64 5.551115123125783e-17
        2: Float64 0.0
    radius: StaticArrays.SArray{Tuple{2},Float64,1,2}
      data: Tuple{Float64,Float64}
        1: Float64 0.8150157120277686
        2: Float64 0.6297654635197406
  v: StaticArrays.SArray{Tuple{2},Float64,1,2}
    data: Tuple{Float64,Float64}
      1: Float64 0.09998194208717585
      2: Float64 2.659605357965139e-5

julia> plot(Q)
MethodError: no method matching addapproximation!(::LazySets.Approximations.PolygonalOverapproximation{Float64}, ::StaticArrays.SArray{Tuple{2},Float64,1,2}, ::Array{Float64,1}, ::StaticArrays.SArray{Tuple{2},Float64,1,2}, ::Array{Float64,1})
Closest candidates are:
  addapproximation!(::LazySets.Approximations.PolygonalOverapproximation, !Matched::Array{N<:Real,1}, ::Array{N<:Real,1}, !Matched::Array{N<:Real,1}, ::Array{N<:Real,1}) where N<:Real at /home/mforets/.julia/dev/LazySets/src/Approximations/iterative_refinement.jl:125

Stacktrace:
[...]

PR #1892, opened 2020-01-02
#1875 -  Iterative refinement with abstract vectors

### LINKED ISSUE #1875: Iterative refinement with abstract vectors
The fields in a `LocalApproximation` and `addapproximation!` can be trivially generalized to any abstract vector. With possibly a performance boost if statically sized vectors are used. This change ~~would~~ may also fix https://github.com/JuliaReach/LazySets.jl/issues/1874.

mforets:
Closes #1875.

- [x] fix tests
- [x] update docstrings

### DIFF
--- a/docs/src/lib/approximations.md
+++ b/docs/src/lib/approximations.md
@@ -60,7 +60,7 @@ PolygonalOverapproximation
 new_approx(S::LazySet, p1::VN, d1::VN,
            p2::VN, d2::VN) where {N<:AbstractFloat, VN<:AbstractVector{N}}
 addapproximation!(Ω::PolygonalOverapproximation, p1::VN, d1::VN, p2::VN, d2::VN) where {N<:Real, VN<:AbstractVector{N}}
-refine(approx::LocalApproximation, S::LazySet)
+refine(::LocalApproximation, ::LazySet)
 tohrep(::PolygonalOverapproximation)
 approximate(S::LazySet{N}, ε::N) where {N<:AbstractFloat}
 constraint(::LocalApproximation)
--- a/docs/src/lib/approximations.md
+++ b/docs/src/lib/approximations.md
@@ -57,11 +57,12 @@ box_approximation_helper
 ```@docs
 LocalApproximation
 PolygonalOverapproximation
-new_approx(::LazySet, ::Vector{Float64}, ::Vector{Float64}, ::Vector{Float64}, ::Vector{Float64})
-addapproximation!(::PolygonalOverapproximation, ::Vector{Float64}, ::Vector{Float64}, ::Vector{Float64}, ::Vector{Float64})
-refine(::LocalApproximation, ::LazySet)
+new_approx(S::LazySet, p1::VN, d1::VN,
+           p2::VN, d2::VN) where {N<:AbstractFloat, VN<:AbstractVector{N}}
+addapproximation!(Ω::PolygonalOverapproximation, p1::VN, d1::VN, p2::VN, d2::VN) where {N<:Real, VN<:AbstractVector{N}}
+refine(approx::LocalApproximation, S::LazySet)
 tohrep(::PolygonalOverapproximation)
-approximate(::LazySet{Float64}, ::Float64)
+approximate(S::LazySet{N}, ε::N) where {N<:AbstractFloat}
 constraint(::LocalApproximation)
 ```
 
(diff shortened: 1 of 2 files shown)


src/Approximations/iterative_refinement.jl
```diff
@@ -17,12 +17,12 @@ Type that represents a local approximation in 2D.
...
-    d1::Vector{N}
-    p2::Vector{N}
-    d2::Vector{N}
-    q::Vector{N}
+struct LocalApproximation{N<:Real, VN<:AbstractVector{N}}
+    p1::VN
+    d1::VN
+    p2::VN
+    d2::VN
+    q::VN
```
schillic: Are you sure that all vectors should have the same type? For instance, `q` comes from
`q = element(intersection(Line(d1, dot(d1, p1)), Line(d2, dot(d2, p2))))`.

mforets: No, i'm not sure if we need to add more types to fix https://github.com/JuliaReach/LazySets.jl/issues/1874, but this is a first step in that direction. This change shouldn't break anything, because in  `master`, all vectors also have the same type. In this branch, all vectors also have the same type. `VN` is just a parameter so we can use static vectors.

schillic: I agree that it is still an improvement over the current state :+1:

mforets:
The branch is ready for review.

src/Approximations/iterative_refinement.jl
```diff
@@ -56,14 +56,16 @@ Type that represents the polygonal approximation of a convex set.
...
+    constraints::Vector{LinearConstraint{N, VN}}
+end
 
-    function PolygonalOverapproximation(S::LazySet{N}) where {N<:Real}
-        return new{N}(S, Vector{LocalApproximation{N}}(), Vector{LinearConstraint{N,Vector{N}}}())
-    end
+function PolygonalOverapproximation(S::SN) where {N<:Real, SN<:LazySet{N}}
+    empty_local_approx = Vector{LocalApproximation{N, Vector{N}}}()
+    empty_constraints = Vector{LinearConstraint{N,Vector{N}}}()
+    return PolygonalOverapproximation(S, empty_local_approx, empty_constraints)
```
schillic: The parameter `SN` is not used. Is that correct?

docs/src/lib/approximations.md
```diff
@@ -57,11 +57,12 @@ box_approximation_helper
 ```@docs
 LocalApproximation
 PolygonalOverapproximation
-new_approx(::LazySet, ::Vector{Float64}, ::Vector{Float64}, ::Vector{Float64}, ::Vector{Float64})
-addapproximation!(::PolygonalOverapproximation, ::Vector{Float64}, ::Vector{Float64}, ::Vector{Float64}, ::Vector{Float64})
-refine(::LocalApproximation, ::LazySet)
+new_approx(S::LazySet, p1::VN, d1::VN,
+           p2::VN, d2::VN) where {N<:AbstractFloat, VN<:AbstractVector{N}}
+addapproximation!(Ω::PolygonalOverapproximation, p1::VN, d1::VN, p2::VN, d2::VN) where {N<:Real, VN<:AbstractVector{N}}
+refine(approx::LocalApproximation, S::LazySet)
```
schillic:
```suggestion
refine(::LocalApproximation, ::LazySet)
```
This line did not actually change.

src/Approximations/iterative_refinement.jl
```diff
@@ -56,19 +56,21 @@ Type that represents the polygonal approximation of a convex set.
...
```
schillic: The parameter `SN` is not used. If that is correct, then I do not understand why you added it.

mforets: Sorry, i didn't understand your comment.

mforets: Oh, do you mean that you'd like to see `PolygonalOverapproximation(S::LazySet{N}) where {N<:Real}` as before instead of `PolygonalOverapproximation(S::SN) where {N<:Real, SN<:LazySet{N}}` ? I added it as a reflex, and these forms are equivalent as far as i know.

schillic: > I added it as a reflex

Okay. I mentioned it because often a change with no effect is suspicious and indicates that something was forgotten.

> these forms are equivalent as far as i know

Yes, they are equivalent. I find the new version slightly harder to read (because I have to look to the right in order to find out the dispatch) but we can keep it.

mforets: Ok. If you make the suggestion i would approve it, otherwise, either version is fine to me.

PR #1894, opened 2020-01-03
Revision of Ball1.jl

schillic:
This PR may seem a bit strange. I am not particularly interested in `Ball1`. My original plan was to *quickly* go through all files and do some clean-ups. But I got lost in the details and only made it through the first file :disappointed:

Here are the commit descriptions. There is a **breaking change** in the third commit that is debatable.

1. several changes:
   * removed output-type parameters
   * "faster" `vertices_list` implementation by removing `push!`es and look-ups of vector entries (the "faster" is in quotes because the benchmark below shows that there is no dfference for large inputs and even a minor slowdown for 2D inputs...)
   * added assertion message
   * used `enumerate`
   * removed redundant `LazySets.` prefix
2. added fail-fast option to membership test (see small benchmark below; in the best case the implementation takes constant time but in the worst case that option is considerably slower (~30%))
3. removed constructor with type parameter (IMO it has no use case)
4. added type parameter for `center` vector

```julia
julia> d = zeros(2); d2 = 10*ones(2); B = Ball1(zeros(2), 1.);
julia> d3 = zeros(10000); d4 = 10*ones(10000); B2 = Ball1(zeros(10000), 1.);

# --- positive answer ---

julia> @btime ∈($d, $B)  # before change
  8.211 ns (0 allocations: 0 bytes)
true

julia> @btime ∈($d, $B, true)
  8.588 ns (0 allocations: 0 bytes)
true

julia> @btime ∈($d, $B, false)
  8.580 ns (0 allocations: 0 bytes)
true


julia> @btime ∈($d3, $B2)  # before change
  11.188 μs (0 allocations: 0 bytes)
true

julia> @btime ∈($d3, $B2, true)
  14.897 μs (0 allocations: 0 bytes)
true

julia> @btime ∈($d3, $B2, false)
  11.181 μs (0 allocations: 0 bytes)
true

# --- negative answer ---

julia> @btime ∈($d2, $B)  # before change
  8.209 ns (0 allocations: 0 bytes)
false

julia> @btime ∈($d2, $B, true)
  7.464 ns (0 allocations: 0 bytes)
false

julia> @btime ∈($d2, $B, false)
  8.577 ns (0 allocations: 0 bytes)
false


julia> @btime ∈($d4, $B2)  # before change
  11.181 μs (0 allocations: 0 bytes)
false

julia> @btime ∈($d4, $B2, true)
  7.466 ns (0 allocations: 0 bytes)
false

julia> @btime ∈($d4, $B2, false)
  11.187 μs (0 allocations: 0 bytes)
false
```

### DIFF
--- a/src/Sets/Ball1.jl
+++ b/src/Sets/Ball1.jl
@@ -144,7 +144,7 @@ function σ(d::AbstractVector{N}, B::Ball1{N}) where {N<:Real}
 end
 
 """
-    ∈(x::AbstractVector{N}, B::Ball1{N}) where {N<:Real}
+    ∈(x::AbstractVector{N}, B::Ball1{N}, [failfast]::Bool=false) where {N<:Real}
 
 Check whether a given point is contained in a ball in the 1-norm.
 
@@ -152,6 +152,7 @@ Check whether a given point is contained in a ball in the 1-norm.
 
 - `x` -- point/vector
 - `B` -- ball in the 1-norm
+- `failfast` -- (optional, default: `false`) optimization for negative answer
 
 ### Output
 
@@ -159,10 +160,11 @@ Check whether a given point is contained in a ball in the 1-norm.
 
 ### Notes
 
-This implementation is worst-case optimized, i.e., it is optimistic and first
-computes (see below) the whole sum before comparing to the radius.
-In applications where the point is typically far away from the ball, a fail-fast
-implementation with interleaved comparisons could be more efficient.
+The default behavior (`failfast == false`) is worst-case optimized, i.e., the
+implementation is optimistic and first computes (see below) the whole sum before
+comparing to the radius.
+In applications where the point is typically far away from the ball, the option
+`failfast == true` terminates faster.
 
 ### Algorithm
 
@@ -182,12 +184,15 @@ julia> [.5, 1.5] ∈ B
 true
 ```
 """
-function ∈(x::AbstractVector{N}, B::Ball1{N}) where {N<:Real}
+function ∈(x::AbstractVector{N}, B::Ball1{N}, failfast::Bool=false) where {N<:Real}
     @assert length(x) == dim(B) "a $(length(x))-dimensional vector is " *
         "incompatible with a $(dim(B))-dimensional set"
     sum = zero(N)
     for (i, xi) in enumerate(x)
         sum += abs(B.center[i] - xi)
+        if failfast && sum > B.radius
+            return false
+        end
     end
     return sum <= B.radius
 end
--- a/src/Sets/Ball1.jl
+++ b/src/Sets/Ball1.jl
@@ -4,7 +4,7 @@ import Base: rand,
 export Ball1
 
 """
-    Ball1{N<:Real} <: AbstractCentrallySymmetricPolytope{N}
+    Ball1{N<:Real, VN<:AbstractVector{N}} <: AbstractCentrallySymmetricPolytope{N}
 
 Type that represents a ball in the 1-norm (also known as the Manhattan norm).
 The ball is also known as a
@@ -28,7 +28,7 @@ Unit ball in the 1-norm in the plane:
 
 ```jldoctest ball1_constructor
 julia> B = Ball1(zeros(2), 1.)
-Ball1{Float64}([0.0, 0.0], 1.0)
+Ball1{Float64,Array{Float64,1}}([0.0, 0.0], 1.0)
 julia> dim(B)
 2
 ```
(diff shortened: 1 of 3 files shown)


src/Sets/Ball1.jl
```diff
@@ -179,19 +181,23 @@ julia> [.5, 1.5] ∈ B
...
 """
-function ∈(x::AbstractVector{N}, B::Ball1{N})::Bool where {N<:Real}
-    @assert length(x) == dim(B)
+function ∈(x::AbstractVector{N}, B::Ball1{N}, failfast::Bool=false) where {N<:Real}
+    @assert length(x) == dim(B) "a $(length(x))-dimensional vector is " *
+        "incompatible with a $(dim(B))-dimensional set"
     sum = zero(N)
-    for i in eachindex(x)
-        sum += abs(B.center[i] - x[i])
+    for (i, xi) in enumerate(x)
```
mforets:
```suggestion
    @inbounds for (i, xi) in enumerate(x)
```

src/Sets/Ball1.jl
```diff
@@ -179,19 +181,23 @@ julia> [.5, 1.5] ∈ B
...
+        sum += abs(B.center[i] - xi)
+        if failfast && sum > B.radius
```
mforets: Here, `failfast` is a loop invariant and can be extracted out of the loop. I'm curious if Julia automatically detects this.

schillic: I always thought that `Bool` arguments are compiled away.

I tried an example and first was shocked: the `goto #8 if not failfast` instruction below looks like this is not the case.
But then I searched a bit and found [this](https://discourse.julialang.org/t/how-does-constant-propagation-work/22735). So I am still quite convinced that it works, but I do not see a quick way to prove it.

```julia
julia> @code_lowered ∈([1.], Ball1([1.], 1.), false)
CodeInfo(
1 ──       Core.NewvarNode(:(sum))
│          Core.NewvarNode(:(@_6))
│    %3  = Base.getproperty(LazySets, :are_assertions_enabled)
│    %4  = (%3)()
└───       goto #5 if not %4
2 ── %6  = LazySets.length(x)
│    %7  = LazySets.dim(B)
│    %8  = %6 == %7
└───       goto #4 if not %8
3 ──       goto #5
4 ── %11 = Base.getproperty(Base.Main, :Base)
│    %12 = Base.getproperty(%11, :string)
│    %13 = LazySets.length(x)
│    %14 = Base.string("a ", %13, "-dimensional vector is ")
│    %15 = LazySets.dim(B)
│    %16 = Base.string("incompatible with a ", %15, "-dimensional set")
│    %17 = %14 * %16
│    %18 = (%12)(%17)
│    %19 = Base.AssertionError(%18)
└───       Base.throw(%19)
5 ┄─       sum = LazySets.zero($(Expr(:static_parameter, 1)))
│    %22 = LazySets.enumerate(x)
│          @_6 = Base.iterate(%22)
│    %24 = @_6 === nothing
│    %25 = Base.not_int(%24)
└───       goto #13 if not %25
6 ┄─ %27 = @_6
│    %28 = Core.getfield(%27, 1)
│    %29 = Base.indexed_iterate(%28, 1)
│          i = Core.getfield(%29, 1)
│          @_9 = Core.getfield(%29, 2)
│    %32 = Base.indexed_iterate(%28, 2, @_9)
│          xi = Core.getfield(%32, 1)
│    %34 = Core.getfield(%27, 2)
│    %35 = sum
│    %36 = Base.getproperty(B, :center)
│    %37 = Base.getindex(%36, i)
│    %38 = %37 - xi
│    %39 = LazySets.abs(%38)
│          sum = %35 + %39
└───       goto #8 if not failfast
7 ── %42 = sum
│    %43 = Base.getproperty(B, :radius)
│          @_10 = %42 > %43
└───       goto #9
8 ──       @_10 = false
9 ┄─       goto #11 if not @_10
10 ─       return false
11 ─       @_6 = Base.iterate(%22, %34)
│    %50 = @_6 === nothing
│    %51 = Base.not_int(%50)
└───       goto #13 if not %51
12 ─       goto #6
13 ┄ %54 = sum
│    %55 = Base.getproperty(B, :radius)
│    %56 = %54 <= %55
└───       return %56
)
```

src/Sets/Ball1.jl
```diff
@@ -179,19 +181,23 @@ julia> [.5, 1.5] ∈ B
...
-    @assert length(x) == dim(B)
+function ∈(x::AbstractVector{N}, B::Ball1{N}, failfast::Bool=false) where {N<:Real}
+    @assert length(x) == dim(B) "a $(length(x))-dimensional vector is " *
+        "incompatible with a $(dim(B))-dimensional set"
     sum = zero(N)
-    for i in eachindex(x)
-        sum += abs(B.center[i] - x[i])
+    @inbounds for (i, xi) in enumerate(x)
+        sum += abs(B.center[i] - xi)
+        if failfast && sum > B.radius
```
mforets:
```suggestion
        if failfast && !_leq(sum, B.radius)
```

src/Sets/Ball1.jl
```diff
@@ -179,19 +181,23 @@ julia> [.5, 1.5] ∈ B
...
+            return false
+        end
     end
     return sum <= B.radius
```
mforets:
```suggestion
    return _leq(sum, B.radius)
```

mforets:
Let's revisit the membership function in another issue, i think the PR is good to go.

mforets:
> "faster" vertices_list implementation by removing push!es and look-ups of vector entries (the "faster" is in quotes because the benchmark below shows that there is no dfference for large inputs and even a minor slowdown for 2D inputs...)

Yes, im not sure that removing `push!` gives performance boost in general.

PR #1896, opened 2020-01-05
#1201 - Add a dimension to EmptySet

### LINKED ISSUE #1201: Add a dimension to EmptySet

schillic:
Closes #1201.

The only non-trivial change is:
Since `EmptySet` now requires the dimension as input, I changed
`∅ = EmptySet{Float64}()`
to
`∅ = EmptySet{Float64}`
(i.e., you need to say `∅(k)` etc. to create an `EmptySet` of dimension `k`).


src/Sets/EmptySet.jl
```diff
@@ -142,7 +143,7 @@ The output is always `false`.
 ### Examples
 
 ```jldoctest
-julia> [1.0, 0.0] ∈ ∅
+julia> [1.0, 0.0] ∈ EmptySet(2)
```
mforets:
```suggestion
julia> [1.0, 0.0] ∈ ∅(2)
```

PR #1907, opened 2020-01-08
#1902 - Concrete linear map for interval

### LINKED ISSUE #1902: Concrete linear map for interval

mforets:
Closes #1902.

### DIFF
--- a/src/Sets/Interval.jl
+++ b/src/Sets/Interval.jl
@@ -558,7 +558,7 @@ number of rows) of `M`:
 - If `size(M, 1) > 1`, the output is a zonotope whose center is `M * center(x)`
   and it has the single generator, `M * g`, where `g = (high(x)-low(x))/2`.
 """
-function linear_map3(M::AbstractMatrix{N}, x::Interval{N}) where {N<:Real}
+function linear_map(M::AbstractMatrix{N}, x::Interval{N}) where {N<:Real}
     @assert size(M, 2) == 1 "a linear map of size $(size(M)) " *
                             "cannot be applied to an interval"
     nout = size(M, 1)
--- a/src/Sets/Interval.jl
+++ b/src/Sets/Interval.jl
@@ -563,19 +563,28 @@ function linear_map(M::AbstractMatrix{N}, x::Interval{N}) where {N<:Real}
                             "cannot be applied to an interval"
     nout = size(M, 1)
     if nout == 1
-        α = M[1, 1]
-        return Interval(α * x.dat)
+        return _linear_map_interval(M, x)
     else
-        cx = IntervalArithmetic.mid(x.dat)
-        gx = cx - min(x)
-        c = Vector{N}(undef, nout)
-        gen = Matrix{N}(undef, nout, 1)
-        @inbounds for i in 1:nout
-            c[i] = M[i, 1] * cx
-            gen[i] = M[i, 1] * gx
-        end
-        return Zonotope(c, gen, remove_zero_generators=false)
+        return _linear_map_zonotope(M, x)
     end
+end
+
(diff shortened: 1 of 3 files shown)


mforets:
Let

```julia
julia> x = rand(Interval)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([-0.371826, 0.0751367])

julia> M = hcat(0.2...)
1×1 Array{Float64,2}:
 0.2
```

In `master`:

```julia
@btime linear_map($M, $x)
  689.939 ns (14 allocations: 1008 bytes)
Zonotope{Float64}([-0.029668852409503867], [0.04469617356262765])
```

In this branch:

```julia
@btime linear_map($M, $x)

  12.958 ns (0 allocations: 0 bytes)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([-0.0743651, 0.0150274])
```

Computing the interval directly is much faster and it is also the expected set representation for the concrete linear map of an interval. Changing the set representation is a breaking change too, but i didn't have to update tests because `linear_map` of an interval was not tested.

src/Sets/Interval.jl
```diff
@@ -514,3 +515,24 @@ We consider the interval as a line segment with y coordinate equal to zero.
...
+
+- `M` -- matrix
+- `x` -- interval
+
+### Output
+
+An interval obtained by scaling `x` by the matrix `M`.
+"""
+function linear_map(M::AbstractMatrix{N}, x::Interval{N}) where {N<:Real}
+    @assert size(M) == (1, 1) "a linear map of size $(size(M)) " *
```
schillic: Do we always require that `M` is square? I think you should be able to have a different output dimension in principle.

mforets: > Do we always require that M is square?

No, because there can be projections. An interval has dimension 1 so it cannot be projected over a set of dimension smaller than 1 -- and it is not clear what does `M_{2x1} * Interval` represent. Anyhow, I can change the assertion to the weaker `size(M, 2) == 1` if you prefer.

schillic: > and it is not clear what does M_{2x1} * Interval represent

In `master` it works just fine:
```julia
julia> M = hcat([1., 2., 3.])
3×1 Array{Float64,2}:
 1.0
 2.0
 3.0

julia> X = LazySets.Interval(0., 1.);

julia> linear_map(M, X)
Zonotope{Float64}([0.5, 1.0, 1.5], [0.5; 1.0; 1.5])

julia> P = convert(HPolyhedron, X)
HPolyhedron{Float64}(HalfSpace{Float64,VN} where VN<:AbstractArray{Float64,1}[HalfSpace{Float64,LazySets.Arrays.SingleEntryVector{Float64}}([1.0], 1.0), HalfSpace{Float64,LazySets.Arrays.SingleEntryVector{Float64}}([-1.0], -0.0)])

julia> linear_map(M, P)
VPolytope{Float64}(Array{Float64,1}[[1.0, 2.0, 3.0], [0.0, 0.0, 0.0]])
```
So I suggest to (a) weaken the constraint and (b) make it work for `size(M, 1) > 1` as well. We can still return an `Interval` if `size(M, 1) == 1` of course.

mforets: Ok, i agree with your idea. The drawback is that `linear_map(::AbstractMatrix, ::Interval)` is now type unstable, since the dimension is not available at runtime.

schillic: > The drawback is that linear_map(::AbstractMatrix, ::Interval) is now type unstable, since the dimension is not available at compile time.

Yes, so maybe you want to oursource the `Interval` case to a new function `linear_map_1d` which you could also call directly instead.

Issue #576, opened 2018-08-20
Plots for unbounded sets


schillic:
We have the undounded set types `HalfSpace`, `Hyperplane`, `Line`, `HPolyhedron`, and `Universe`.
~I think we should print an error message when trying to plot such sets. Currently the plotting crashes when trying to overapproximate the sets.~ EDIT: We show an error now.

I also thought about adding a common interface/abstract supertype like `UnboundedSet`. However, I do not see any other application at the moment. And it would have the dual problem to what we had with `ConvexSet`, namely that set operations involving these set types would not be `UnboundedSet`s.

mforets:
it would be nice to plot such sets in 2D. but one needs to pass xmin/xmax values for the plots.

schillic:
Good point!

schillic:
This function plots a half-space by first obtaining the x- and y-limits and then computing the two y-coordinates at the left and right border. (One caveat: If the half-space has coefficient 0 for dimension y, the code does not work.) Finally it plots the given line segment and fills the area on the correct side of the line segment.
```julia
function plot_halfspace(H)
    pa = p.series_list[1].plotattributes
    x, y = pa[:x], pa[:y]
    y = [(H.a[1] * z - H.b) / (-H.a[2]) for z in x]
    fillrange = minimum(y)
    plot(x, y, fillrange=fillrange

)
end
```
![half-space](https://user-images.githubusercontent.com/9656686/55196820-e2a3f280-51b0-11e9-822a-3dec9e9da774.png)

Note that if the axis ranges are extended later, the filling is not extended, so in the example you get a triangle.

schillic:
See `#1422` in [`Plots.jl`](https://github.com/JuliaPlots/Plots.jl) and the referenced issues. They have `hspan` and `vspan`, which could be a start.

ueliwechsler:
I think, something like this would work 

```julia
function bounded_HPolytope_from_unbounded(P::HPolyhedron{N}, limit=1000) where {N}
    A, a = tosimplehrep(P)
    n = LazySets.dim(P)
    B = [Matrix{N}(I,n,n); -Matrix{N}(I,n,n)]
    b = limit*ones(N,2*n)
    return HPolytope([A;B], [a;b])
end
```
[...]

PR #1909, opened 2020-01-09
#576 - Plots for unbounded sets

schillic:
This was a first proposal for #576 but got superseded by #1910.

### DIFF
--- a/src/Plotting/plot_recipes.jl
+++ b/src/Plotting/plot_recipes.jl
@@ -82,6 +82,15 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
             if Xi isa Intersection
                 res = plot_recipe(Xi, ε, Nφ)
             else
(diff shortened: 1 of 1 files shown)


schillic:
This approach requires the user to specify the plot limits explicitly. There might be a way to obtain them from the current plot window, though (would require more research).

```julia
julia> using LazySets, Plots

julia> B = BallInf(zeros(2), 2.);

julia> P = HalfSpace([1., 1.], 1.);

julia> plot(B)
```



```julia
julia> plot!(P, xlims=[-1., 1.], ylims=[-1.,1.])
```



If we change the plot limits, the old half-space looks odd. This can be mitigated by increasing the value of `ε`, but you can always find a counterexample. (The value is necessary because the plot window always shows slightly more than the defined plot limits.)

```julia
julia> plot!(P, xlims=[-1., 2.], ylims=[-3.,1.])
```

PR #1910, opened 2020-01-10
#576 - Plots for unbounded sets

### LINKED ISSUE #576: Plots for unbounded sets

ueliwechsler:
Closes #576.

Things to be addressed in follow-up issues:
* Add a macro for common code in all plot recipes for extending the plot limits (difficult because that code itself is already inside a macro) ([link to discussion](https://github.com/JuliaReach/LazySets.jl/pull/1910#discussion_r365823637), #1917).
* Support plotting of nested `Intersection`s of unbounded sets ([link to comment](https://github.com/JuliaReach/LazySets.jl/pull/1910#issuecomment-574183717), #1913).

### DIFF
--- a/src/Plotting/plot_recipes.jl
+++ b/src/Plotting/plot_recipes.jl
@@ -82,6 +82,15 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
             if Xi isa Intersection
                 res = plot_recipe(Xi, ε, Nφ)
             else
+                if !isbounded(Xi)
+                    # bound unbounded set with the plotting range
+                    p = plot!()
+                    xlim = xlims(p)
+                    ylim = ylims(p)
+                    low = [xlim[1], ylim[1]]
+                    high = [xlim[2], ylim[2]]
+                    Xi = intersection(Xi, Hyperrectangle(low=low, high=high))
+                end
                 # hard-code overapproximation here to avoid individual
                 # compilations for mixed sets
                 Pi = overapproximate(Xi, ε)
@@ -161,6 +170,15 @@ julia> plot(B, 1e-2)  # faster but less accurate than the previous call
         seriesalpha --> DEFAULT_ALPHA
         seriescolor --> DEFAULT_COLOR
 
+        if !isbounded(X)
+            # bound unbounded set with the plotting range
+            pa = plotattributes
+            xlims, ylims = pa[:xlims], pa[:ylims]
+            ε = 1.0
+            low = [xlims[1] - ε, ylims[1] - ε]
+            high = [xlims[2] + ε, ylims[2] + ε]
+            X = intersection(X, Hyperrectangle(low=low, high=high))
+        end
         res = plot_recipe(X, ε)
         if isempty(res)
             res
--- a/src/Plotting/plot_recipes.jl
+++ b/src/Plotting/plot_recipes.jl
@@ -49,15 +49,17 @@ function _update_plot_limits!(lims, X::LazySet)
     box_max = high(box)
     for (idx,symbol) in enumerate([:x, :y])
         if lims[symbol] != :auto
-            # scaling factor for obtaining offset from window width
-            ϵ = 0.05
-            # width of the plotting window in the direction `symbol`
+            # width of the plotting window including the new set in the `symbol` direction
             width = max(lims[symbol][2], box_max[idx]) -
                     min(lims[symbol][1], box_min[idx])
 
+            # scaling factor for beautification
+            ϵ = 0.05
+            offset = width*ϵ
+
             # extend the current plot limits if the new set (plus a small offset) falls outside
-            lims[symbol] = (min(lims[symbol][1], box_min[idx] - width*ϵ),
-                            max(lims[symbol][2], box_max[idx] + width*ϵ))
+            lims[symbol] = (min(lims[symbol][1], box_min[idx] - offset),
+                            max(lims[symbol][2], box_max[idx] + offset))
         end
     end
     nothing
--- a/src/Plotting/plot_recipes.jl
+++ b/src/Plotting/plot_recipes.jl
@@ -77,6 +77,31 @@ function _update_plot_limits!(lims, X::LazySet)
     nothing
 end
 
+function _set_auto_limits_to_extrema!(lims, extr)
+    # if the limit is :auto, set the limits to the current extrema
+    for symbol in [:x, :y]
+        if lims[symbol] == :auto
+            emin, emax = extr[symbol]
+            # if the extrema are (-Inf,Inf), i.e. an empty plot,
+            # set it to (-1.,1.)
+            lmin = isinf(emin) ? -1.0 : emin
+            lmax = isinf(emax) ? 1.0 : emax
+            lims[symbol] = (lmin, lmax)
+        end
+    end
+    # otherwise keep the old limits
+    nothing
+end
+
+function bounding_hyperrectangle(lims)
+    low_lim = [lims[:x][1] - DEFAULT_PLOT_LIMIT, lims[:y][1] - DEFAULT_PLOT_LIMIT]
+    high_lim = [lims[:x][2] + DEFAULT_PLOT_LIMIT, lims[:y][2] + DEFAULT_PLOT_LIMIT]
+    return Hyperrectangle(low=low_lim, high=high_lim)
+end
+
+
+
+
 """
     plot_list(list::AbstractVector{VN}, [ε]::N=N(PLOT_PRECISION),
               [Nφ]::Int=PLOT_POLAR_DIRECTIONS, [fast]::Bool=false; ...)
@@ -232,23 +257,8 @@ julia> plot(B, 1e-2)  # faster but less accurate than the previous call
         extr = _extract_extrema(p)
 
         if !isbounded(X)
-            # if the limit is :auto, set the limits to the current extrema
-            for symbol in [:x, :y]
-                if lims[symbol] == :auto
-                    emin, emax = extr[symbol]
-                    # if the extrema are (-Inf,Inf), i.e. an empty plot,
-                    # set it to (-1.,1.)
-                    lmin = isinf(emin) ? -1.0 : emin
-                    lmax = isinf(emax) ? 1.0 : emax
-                    lims[symbol] = (lmin, lmax)
-                end
-            end
-            # otherwise keep the old limits
-
-            # bound unbounded set
-            low_lim = [lims[:x][1] - DEFAULT_PLOT_LIMIT, lims[:y][1] - DEFAULT_PLOT_LIMIT]
-            high_lim = [lims[:x][2] + DEFAULT_PLOT_LIMIT, lims[:y][2] + DEFAULT_PLOT_LIMIT]
-            X = intersection(X, Hyperrectangle(low=low_lim, high=high_lim))
+            _set_auto_limits_to_extrema!(lims, extr)
+            X = intersection(X,  bounding_hyperrectangle(lims))
 
         # if there is already a plotted set and the limits are fixed,
         # automatically adjust the axis limits (e.g. after plotting a unbounded set)
@@ -458,14 +468,29 @@ julia> plot(X, -1., 100)  # equivalent to the above line
     seriescolor --> DEFAULT_COLOR
     seriestype := :shape
 
-    # update fixed plot limits if necessary
+    # extract limits and extrema of already plotted sets
     p = plotattributes[:plot_object]
-    if length(p) > 0
-        lims = _extract_limits(p)
+    lims = _extract_limits(p)
+    extr = _extract_extrema(p)
+
+    if !isbounded(cap)
+        _set_auto_limits_to_extrema!(lims, extr)
+        bounding_box = bounding_hyperrectangle(lims)
+        if !isbounded(cap.X)
+            bounded_X = Intersection(bounding_box, cap.X)
+            cap = Intersection(bounded_X, cap.Y)
+        else
+            cap = Intersection(bounding_box, cap.Y)
+        end
+
+    # if there is already a plotted set and the limits are fixed,
+    # automatically adjust the axis limits (e.g. after plotting a unbounded set)
+    elseif length(p) > 0
         _update_plot_limits!(lims, cap)
-        xlims --> lims[:x]
-        ylims --> lims[:y]
     end
+
+    xlims --> lims[:x]
+    ylims --> lims[:y]
 
     plot_recipe(cap, ε)
 end
--- a/src/Plotting/plot_recipes.jl
+++ b/src/Plotting/plot_recipes.jl
@@ -83,15 +83,6 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
             if Xi isa Intersection
                 res = plot_recipe(Xi, ε, Nφ)
             else
-                if !isbounded(Xi)
-                    # bound unbounded set with the plotting range
-                    p = plot!()
-                    xlim = xlims(p)
-                    ylim = ylims(p)
-                    low = [xlim[1], ylim[1]]
-                    high = [xlim[2], ylim[2]]
-                    Xi = intersection(Xi, Hyperrectangle(low=low, high=high))
-                end
                 # hard-code overapproximation here to avoid individual
                 # compilations for mixed sets
                 Pi = overapproximate(Xi, ε)
--- a/src/Plotting/plot_recipes.jl
+++ b/src/Plotting/plot_recipes.jl
@@ -181,7 +181,7 @@ julia> plot(B, 1e-2)  # faster but less accurate than the previous call
                     emin = extr[symbol].emin
                     emax = extr[symbol].emax
                     # if the extrema are (-Inf,Inf), i.e. an empty plot,
-                    # set it to (0.,1.)
+                    # set it to (-1.,1.)
                     lmin = isinf(emin) ? -1.0 : emin
                     lmax = isinf(emax) ? 1.0 : emax
                     lims[symbol] = (lmin, lmax)
@@ -198,7 +198,7 @@ julia> plot(B, 1e-2)  # faster but less accurate than the previous call
 
         # if there is already a plotted shape and the limits are fixed,
         # automatically adjust the axis limits (e.g. after plotting a unbounded set)
-        elseif  length(p) > 0 && lims[:x] != :auto && lims[:y] != :auto
+        elseif length(p) > 0 && lims[:x] != :auto && lims[:y] != :auto
             # update limits such that the new set is contained in the plot windows
             box = box_approximation(X)
             min_box = low(box)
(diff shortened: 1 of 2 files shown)


src/Plotting/plot_recipes.jl
```diff
@@ -82,6 +83,15 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
...
             else
+                if !isbounded(Xi)
+                    # bound unbounded set with the plotting range
+                    p = plot!()
+                    xlim = xlims(p)
+                    ylim = ylims(p)
+                    low = [xlim[1], ylim[1]]
+                    high = [xlim[2], ylim[2]]
+                    Xi = intersection(Xi, Hyperrectangle(low=low, high=high))
+                end
```
schillic: Just a note: This part comes from my attempt but does not work (used for plotting lists of sets, which I never tested in the end).

ueliwechsler:
```julia 
using LazySets, Plots

B = BallInf(zeros(2), 5.);
B2 = BallInf(15*ones(2), 3.);

P1 = HalfSpace([1., 1.], 1.);
P2 = HalfSpace([-1., 1.], 1.);
p = plot(B)
plot!(P1)
```
Plotting of bounded set `B` and unbouned set `P1`. The set that is plotted instead of `P1` is 
```julia
DEFAULT_PLOT_LIMIT = 1000
low_lim = [lims[:x][1] - DEFAULT_PLOT_LIMIT, lims[:y][1] - DEFAULT_PLOT_LIMIT]
high_lim = [lims[:x][2] + DEFAULT_PLOT_LIMIT, lims[:y][2] + DEFAULT_PLOT_LIMIT]
X = intersection(X, Hyperrectangle(low=low_lim, high=high_lim))
```
where the `lims[x]` and `lims[y]` is extractet from the plotattributes and corresponds to `xlims` and `ylims` . Caveat: In Atom with `gr` backend, this plot is not expressed correctly)
LazySets2
If we zoom out, we see the entire set
``` julia
xlims!((-1300,1300)); ylims!((-1300,1300))
```
LazySets6

Adding another Halfspace is done in the same way.
```julia
xlims!((-20,20)); ylims!((-20,10))
plot!(P2)
```
LazySets4

Now, if we want to add another set. In the current implementation, the set limits of the plotting windows would not be adjsuted since `xlims` and `ylims` is set to a value and not `:auto`.
I added a functionality, such that the limits are adjusted if the new set is not contained in the old limits (this would change for all sets). As a result, we have
```julia
plot!(B2)
```

LazySets5

src/Plotting/plot_recipes.jl
ueliwechsler: Thanks, I will remove it.

schillic: You might have to add some other code here instead (as I said, I never tested it).

ueliwechsler:
As an additional note:
For a empty plot, the extrema is `(Inf,-Inf)`, we set the limit to `(0,1)` as default in this case, as a result:
```julia
p = plot()
plot!(P1)
```
In Atom with `gr()` this is plotted as (this is a screenshot instead of a `savefig`)

image
Adjusting the limits fixes this (again a screenshot)
```julia
xlims!((-20,20))
ylims!((-20,10))
plot!(Singleton([2000,0]))
```
However, the recipe for a  `Singleton` was not extended yet therefore, the limits are also not extended
image

src/Plotting/plot_recipes.jl
```diff
@@ -161,6 +162,56 @@ julia> plot(B, 1e-2)  # faster but less accurate than the previous call
...
+
+        if !isbounded(X)
+            # if the limit is :auto, set the limits to the current extrema
+            for symbol in [:x, :y]
+                if lims[symbol] == :auto
+                    emin = extr[symbol].emin
+                    emax = extr[symbol].emax
+                    # if the extrema is (-Inf,Inf), i.e. an empty plot,
+                    # set it to (0.,1.)
+                    lmin = isinf(emin) ? 0.0 : emin
```
schillic:
```suggestion
                    lmin = isinf(emin) ? -1.0 : emin
```
A default window centered in the origin feels slightly better (requires to update the comment above as well).

src/Plotting/plot_recipes.jl
```diff
@@ -161,6 +162,56 @@ julia> plot(B, 1e-2)  # faster but less accurate than the previous call
         seriesalpha --> DEFAULT_ALPHA
         seriescolor --> DEFAULT_COLOR
 
+        # extract limit and extrema of plotted shapes if existings
```
schillic:
```suggestion
        # extract limits and extrema of already plotted sets
```

src/Plotting/plot_recipes.jl
```diff
@@ -161,6 +162,56 @@ julia> plot(B, 1e-2)  # faster but less accurate than the previous call
...
+            end
+        end
+
+        if !isbounded(X)
+            # if the limit is :auto, set the limits to the current extrema
+            for symbol in [:x, :y]
+                if lims[symbol] == :auto
+                    emin = extr[symbol].emin
+                    emax = extr[symbol].emax
+                    # if the extrema is (-Inf,Inf), i.e. an empty plot,
```
schillic:
```suggestion
                    # if the extrema are (-Inf,Inf), i.e. an empty plot,
```

src/Plotting/plot_recipes.jl
```diff
@@ -161,6 +162,56 @@ julia> plot(B, 1e-2)  # faster but less accurate than the previous call
...
+            for symbol in [:x, :y]
+                lims[symbol] = subplot[Symbol(symbol,:axis)][:lims]
+                extr[symbol] = subplot[Symbol(symbol,:axis)][:extrema]
+            end
+        end
+
+        if !isbounded(X)
+            # if the limit is :auto, set the limits to the current extrema
+            for symbol in [:x, :y]
+                if lims[symbol] == :auto
```
schillic: You need to do either of these:
* Add an `else` case to the first `if` and fill the dictionaries with default values there.
* Check below whether the dictionaries contain the keys before you access them.

(Or maybe you want a mixture: default values for `extr` and a check for `lims`?)

Otherwise there is an error (in a fresh session):
```julia
julia> using LazySets, Plots

julia> plot(HalfSpace([1., 1.], 1.))
ERROR: KeyError: key :x not found
```

src/Plotting/plot_recipes.jl
```diff
@@ -161,6 +162,56 @@ julia> plot(B, 1e-2)  # faster but less accurate than the previous call
...
+            end
+            # otherwise keep the old limits
+
+            # bound unbounded set
+            low_lim = [lims[:x][1] - DEFAULT_PLOT_LIMIT, lims[:y][1] - DEFAULT_PLOT_LIMIT]
+            high_lim = [lims[:x][2] + DEFAULT_PLOT_LIMIT, lims[:y][2] + DEFAULT_PLOT_LIMIT]
+            X = intersection(X, Hyperrectangle(low=low_lim, high=high_lim))
+
+        # if there is already a plotted shape and the limits are fixed,
+        # automatically adjust the  axis limits (e.g. after plotting a unbounded set)
```
schillic:
```suggestion
        # automatically adjust the axis limits (e.g. after plotting a unbounded set)
```

src/Plotting/plot_recipes.jl
```diff
@@ -161,6 +162,56 @@ julia> plot(B, 1e-2)  # faster but less accurate than the previous call
...
+        elseif  length(p) > 0 && lims[:x] != :auto && lims[:y] != :auto
+            # update limits such that the new set is contained in the plot windows
+            box = box_approximation(X)
+            min_box = low(box)
+            max_box = high(box)
+            for (idx,symbol) in enumerate([:x, :y])
+                # if the new set is not contained in limits, extend the limit
+                ϵ = 0.05
+                min_lim = min(lims[symbol][1], min_box[idx]*(1 - ϵ))
+                max_lim = max(lims[symbol][2], max_box[idx]*(1 + ϵ))
```
schillic: 1. The multiplication only makes sense for sets in the positive quadrant.
2. A relative scaling means that the same sets shifted by a large scalar will result in different plots.
3. For checking whether the current bounds should be extended, this additional offset should not be considered.

In the interest of simplicity, I would just ignore this beautification.

schillic:
> However, the recipe for a Singleton was not extended yet therefore, the limits are also not extended

Good catch. Can you outsource your auto-update code to a helper function (something like `_plot_update!`)? Then we can just call it from the other plot recipes.

src/Plotting/plot_recipes.jl
schillic: > requires to update the comment above as well

src/Plotting/plot_recipes.jl
ueliwechsler: Thanks I added the default cases in the outsourced methods `_extract_limits` and `_extract_extrema` and used
```julia
xlims --> lims[:x]
ylims --> lims[:y]
```
to update the limits of the plot independent of `plotattributes[:plot_object]`

src/Plotting/plot_recipes.jl
ueliwechsler: Sure, thanks. Was just not pushed already

src/Plotting/plot_recipes.jl
ueliwechsler: > 1. The multiplication only makes sense for sets in the positive quadrant.

Fixed it by 
```
            ϵ = 0.05
            min_offset = abs(min_box[idx])*ϵ
            max_offset = abs(max_box[idx])*ϵ
            min_lim = min(lims[symbol][1], min_box[idx] - min_offset)
            max_lim = max(lims[symbol][2], max_box[idx] + max_offset)
```

> 2. A relative scaling means that the same sets shifted by a large scalar will result in different plots.
> 3. For checking whether the current bounds should be extended, this additional offset should not be considered.

 I disagree with these points. I am under the opinion that the plot offset should be dependent on the plot size and not the set size.  I really liked the resulting extension of the plot limits by this implementation. However, I did not test the plotting behaviour entierly, so if there are reasons why my implementation doesn't make sense I gladly change it.

ueliwechsler:
I did refactor the code quite a bit.
I outsourced ` _extract_limits` and `_extract_extrema` additionally I outsourced `_update_plot_limits!` as suggested in the comment above. 

I extended `plot_singleton`, `plot_linesegment` and `plot_intersection` with a `_update_plot_limits!` parts.

As a result, the following code snippet 
```julia
using LazySets, Plots
# unbounded set
p = plot(HalfSpace([1., 0.], 0.5))
# bounded set
X = BallInf(zeros(2), 4.)
plot!(X)
# bounded Intersection
X = Ball2([1.,3], 3.) ∩ Ball2([2.,3],2.)
plot!(X)
# LineSegment
L = LineSegment([-5., 0.], [5., 1.]);
plot!(L)
plot!(Singleton([15.,-6.]))
plot!(Singleton([-15.,6.]))
# unbounded intersection
p = plot!(HalfSpace([-1., 1.], 1.) ∩ HalfSpace([-5., -3.], 1.)) #fails
p = plot!(HPolyhedron([-1. 1.; -5. -3.], [-6.,-20]))
```
results in

image3773

As you see, the plotting of an unbounded `Intersection` still fails. Should I just extend the `plot_intersection` recipe with the same code as `plot_recipe`, i.e. outsource the functionality.

And finally, if the interesting features of the unbounded plot are not visible in the resulting plot window, the user has to find them manually, but as discussed for this PR this should be okay.

src/Plotting/plot_recipes.jl
schillic: Okay, does that not destroy the `:auto` for bounded sets now?

src/Plotting/plot_recipes.jl
schillic: Take the values `min_box[idx] == 0` and `max_box[idx] == 1000`. Then `min_offset == 0` and `max_offset == 50`. I doubt that this is what you want.

schillic:
> As you see, the plotting of an unbounded Intersection still fails. Should I just extend the plot_intersection recipe with the same code as plot_recipe, i.e. outsource the functionality.

Is there a problem with `Intersection`s? I know that they are special, but if you can apply the same fix, sure!

src/Plotting/plot_recipes.jl
```diff
@@ -203,6 +285,15 @@ julia> plot(Singleton([0.5, 1.0]))
...
     seriestype := :scatter
 
+    # update fixed plot limits if necessary
+    p = plotattributes[:plot_object]
+    if length(p) > 0
+        lims = _extract_limits(p)
+        _update_plot_limits!(lims, S)
+        xlims --> lims[:x]
+        ylims --> lims[:y]
+    end
```
schillic: It would be nice to have a macro:
```julia
macro update_plot_limits!(S)
    p = plotattributes[:plot_object]
    if length(p) > 0
        lims = _extract_limits(p)
        _update_plot_limits!(lims, S)
        xlims --> lims[:x]
        ylims --> lims[:y]
    end
end
```

src/Plotting/plot_recipes.jl
ueliwechsler: Okay, I see you are totally right, I didn't think this through. But I think we are on the right track. I will just keep trying for the moment and keep your examples as a test case.

src/Plotting/plot_recipes.jl
ueliwechsler: I can work on it.

src/Plotting/plot_recipes.jl
schillic: I think you want to have the *width* (max - min) and scale that.

src/Plotting/plot_recipes.jl
schillic: Hopefully you can just copy-paste what I wrote (maybe with some additional `quote`).

schillic: There might be problems with a "macro call within a macro". If it does not work immediately, we can defer it to a new issue. So do not waste time on it.

schillic:
> I can fix the errors in the unit tests if you want. They were a reminder that we should test plotting "when it's done".

I checked and one just needs to remove the error tests:

https://github.com/JuliaReach/LazySets.jl/blob/9c0853f96d8602a70879922561b4a865d1430543/test/unit_plot.jl#L108-L111

replace by

```julia
plot(hs)
plot(hp)
plot(l)
plot(uni)
```

I leave that small change to you :wink:

src/Plotting/plot_recipes.jl
ueliwechsler: No, because the default case in `_extract_limits` is `lims[:x] = (:auto); lims[:y] = (:auto)`

schillic: If I understand correctly, that default is only used if there is no plot window yet.

src/Plotting/plot_recipes.jl
ueliwechsler: I am not sure anymore, what I really want 😅
It's close to what you proposed. But I think I will use the width of the entire plotting windows and not only the set.

schillic: That's fine :+1:

ueliwechsler:
> Is there a problem with `Intersection`s? I know that they are special, but if you can apply the same fix, sure!

Yes, there is a problem. In `overapproximate` in the `plot_recipe` of the `Intersection` type, i get the error 
```
AssertionError: the first set in the intersection must be bounded
```
I am trying to figure out, how to do a workaround.

schillic:
> Yes, there is a problem. In `overapproximate` in the `plot_recipe` of the `Intersection` type, i get the error
> 
> ```
> AssertionError: the first set in the intersection must be bounded
> ```

Maybe I misunderstand. I do not think that you need to extend the capability to plot unbounded `Intersection`s here. Just plotting a *bounded* `Intersection` should continue working as expected.

ueliwechsler:
> ust plotting a _bounded_ `Intersection` should continue working as expected.

This does still work.

I made an adjustment to the `plot_intersection` recipe that handles the lazy intersection of two unbounded sets. However, the recursion of multiple lazy intersections of unbounded sets is a bit trickier to handle. We can open an issue for that.

src/Plotting/plot_recipes.jl
ueliwechsler: I am not sure why, but I didn't work immediately, so let's open a new issue.

src/Plotting/plot_recipes.jl
```diff
@@ -11,6 +11,105 @@ DEFAULT_GRID = true
...
+
+            # extend the current lower limit of `symbol`, if the new set minus
+            # beautification offset is outside of the current plotting limits
+            if box_min[idx] - offset*ϵ ≤ lims[symbol][1]
+                lim_min = box_min[idx] - offset*ϵ
+            else
+                lim_min = lims[symbol][1]
+            end
+
+            lims[symbol] = (lim_min, lim_max)
```
schillic: Shorter:
```julia
# extend the current plot limits if the new set (plus a small offset) falls outside
lims[symbol][1] = min(lims[symbol][1], box_min[idx] - offset*ϵ)
lims[symbol][2] = max(lims[symbol][2], box_max[idx] + offset*ϵ)
```

src/Plotting/plot_recipes.jl
```diff
@@ -11,6 +11,105 @@ DEFAULT_GRID = true
...
+function _update_plot_limits!(lims, X::LazySet)
+    box = box_approximation(X)
+    box_min = low(box)
+    box_max = high(box)
+    for (idx,symbol) in enumerate([:x, :y])
+        if lims[symbol] != :auto
+            # scaling factor for beautification
+            ϵ = 0.05
+            # compute the size of the plotting windows in the direction `symbol`
+            offset = max(lims[symbol][2], box_max[idx]) -
```
schillic: I think `offset` is not the right term. You compute the interval approximation of two intervals (namely the current plotting range and the box approximation `X` in dimension `idx`). Ignoring the `ϵ` factor, this interval will be the new plotting range in dimension `idx`. So maybe `new_range` or `width`?

src/Plotting/plot_recipes.jl
schillic: Agreed. I added a list of follow-up issues to the OP.

src/Plotting/plot_recipes.jl
schillic: > If I understand correctly, that default is only used if there is no plot window yet.

I guess I did not understand correctly :smiley:

PR #1931, opened 2020-02-03
#1925 - Refactor boundnedness check for HPolygon

### LINKED ISSUE #1925: Refactor boundnedness check for HPolygon
https://github.com/JuliaReach/LazySets.jl/blob/ffe0320ab37a2cc7e06eaf461f04b8743554c4aa/src/Interfaces/AbstractHPolygon.jl#L575 converts to `HPolyhedron`, but `isbounded_unit_dimensions` can be called directly and this makes `isbounded(::HPolygon, false)` ~15x faster.

mforets:
Closes #1925.

### DIFF
--- a/src/Interfaces/AbstractHPolygon.jl
+++ b/src/Interfaces/AbstractHPolygon.jl
@@ -576,6 +576,9 @@ function isbounded(P::AbstractHPolygon, use_type_assumption::Bool=true)
     if use_type_assumption
         return true
     end
+    if isempty(constraints_list(P))
+        return false
+    end
     return isbounded_unit_dimensions(P)
 end
 
--- a/src/Interfaces/AbstractHPolygon.jl
+++ b/src/Interfaces/AbstractHPolygon.jl
@@ -576,7 +576,7 @@ function isbounded(P::AbstractHPolygon, use_type_assumption::Bool=true)
     if use_type_assumption
         return true
     end
-    return isbounded(HPolyhedron(P.constraints))
+    return isbounded_unit_dimensions(P)
 end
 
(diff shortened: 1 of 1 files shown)


mforets:
```julia
using Revise, LazySets, BenchmarkTools

P = rand(HPolygon);

@btime isbounded($P, false) # master
  52.567 μs (310 allocations: 20.11 KiB)

@btime LazySets.isbounded_unit_dimensions($P) # master
  3.140 μs (76 allocations: 3.63 KiB)

@btime isbounded($P, false) # this branch
  3.123 μs (76 allocations: 3.63 KiB)
```

schillic:
Apparently the special case of a polygon with no constraints needs to be treated differently:
```julia
LazySets.Polygon: Error During Test at test\unit_Polygon.jl:87
  Test threw exception
  Expression: !(isbounded(HPolygon{N}(), false)) && !(isbounded(HPolygonOpt{N}(), false))
  AssertionError: the polygon has no constraints
```

schillic:
Do you think that this implementation is also more efficient in higher dimensions? Then we should rather change `isbounded(::Vector{LinearConstraint})`.

mforets:
The "gain" in this PR comes from the fact that evaluating the support function of the given `HPolygon` directly is cheaper than converting to a polyhedron and evaluating its support function (using LPs). 

> Then we should rather change isbounded(::Vector{LinearConstraint}).

Where is that function defined?

schillic:
> Where is that function defined?

Sorry, I missed the conversion to `HPolyhedron`.

mforets:
Thanks for the review.

PR #1938, opened 2020-02-04
#1937 - Extend matrix by orthogonalization

mforets:
Closes #1937.

Issue #1934, opened 2020-02-03
Revise "division" algorithm for linear map 


mforets:
- Forbid calling it for non-invertible matrices.
- Rename it to "inverse_left_division" (because "division" and "inverse' is rather random and confusing -- both methods use the matrix inverse)
- Add in the docs that it "inverse" is preferable in general, but "inverse_left_division" has shown better speed
    - specially for small dimensions (2,3,4 basically), where "inverse_left_division" is ~30% faster
    - at 5-6 there is a crossover then "inverse" has better scaling

- use `"lift"` as default when appropriate

xref: https://github.com/JuliaReach/LazySets.jl/issues/1928

Issue #1802, opened 2019-11-29
Concretely type linear constraint arrays


mforets:
The type parameter of the normal direction in the half-space type was not added from the beginning. As a consequence there are methods which return non-concretely-typed containers where they shouldn't. For instance, 

```julia
julia> x = rand(Hyperrectangle)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([0.6356749195435474, -0.7138079011876447], [0.6689691491156268, 0.32291592190199064])

julia> constraints_list(x)
4-element Array{HalfSpace{Float64,VN} where VN<:AbstractArray{Float64,1},1}:
 HalfSpace{Float64,LazySets.Arrays.SingleEntryVector{Float64}}([1.0, 0.0], 1.304644068659174)   
 HalfSpace{Float64,LazySets.Arrays.SingleEntryVector{Float64}}([0.0, 1.0], -0.39089197928565406)
 HalfSpace{Float64,LazySets.Arrays.SingleEntryVector{Float64}}([-1.0, 0.0], 0.03329422957207939)
 HalfSpace{Float64,LazySets.Arrays.SingleEntryVector{Float64}}([0.0, -1.0], 1.0367238230896354) 
```

Searching for `LinearConstraint{N}` reveals other places.

For ref: https://docs.julialang.org/en/v1/manual/performance-tips/index.html#Avoid-containers-with-abstract-type-parameters-1

mforets:
```julia
h = rand(Hyperrectangle)
m = minkowski_difference(h, h)

HPolytope{Float64}(HalfSpace{Float64,VN} where VN<:AbstractArray{Float64,1}[HalfSpace{Float64,Array{Float64,1}}([1.0, 0.0], 0.0), HalfSpace{Float64,Array{Float64,1}}([0.0, 1.0], 0.0), HalfSpace{Float64,Array{Float64,1}}([-1.0, 0.0], 0.0), HalfSpace{Float64,Array{Float64,1}}([0.0, -1.0], 0.0)])

```

mforets:
```julia
function overapproximate(X::LazySet{N}, dir::AbstractDirections{N}) where {N}
    halfspaces = Vector{LinearConstraint{N}}() # ??
    sizehint!(halfspaces, length(dir))
    T = isbounding(dir) ? HPolytope : HPolyhedron
    H = T(halfspaces)
    for d in dir
        addconstraint!(H, LinearConstraint(d, ρ(d, X)))
[...]

PR #1939, opened 2020-02-04
#1936 - Concrete linear map using variable lifting

### LINKED ISSUE #1936: Concrete linear map using variable lifting
xref: https://github.com/JuliaReach/LazySets.jl/issues/1928

mforets:
Closes #1936.

Requires: ~~https://github.com/JuliaReach/LazySets.jl/pull/1938~~

Note that this function does not change the default. I think this should be done, but we can do it in a follow-up PR after addressing https://github.com/JuliaReach/LazySets.jl/issues/1934.

### DIFF
--- a/src/Interfaces/AbstractPolyhedron_functions.jl
+++ b/src/Interfaces/AbstractPolyhedron_functions.jl
@@ -525,7 +525,7 @@ function _linear_map_hrep_lift(M::AbstractMatrix{N}, P::AbstractPolyhedron{N}) w
 
     Pext = HPolytope(cext)
 
-    # now Mext is invertible and we can apply the linear with inverse function
+    # now Mext is invertible and we can apply the inverse algorithm
     return _linear_map_hrep(Mext, Pext, true, inverse=inv_Mext)
 end
 
--- a/src/Interfaces/AbstractPolyhedron_functions.jl
+++ b/src/Interfaces/AbstractPolyhedron_functions.jl
@@ -288,6 +288,10 @@ Concrete linear map of a polyhedral set.
   - `"inverse"`  -- compute the matrix inverse and apply it to each constraint
                     of `P`
   - `"division"` -- divide each constraint of `P` by `M` from the left
+  - `"lift"`     -- for rectangular non-invertible matrices, extend `M` to an
+                    invertible matrix and use the H-representation (only applies
+                    if the number of rows of `M` is bigger than the number of
+                    columns, and if `M` is full rank)
 - `check_invertibility` -- (optional, deault: `true`) check if the linear map is
                            invertible, in which case this function uses the
                            matrix inverse; if this flag is set to `false`, we
@@ -336,9 +340,21 @@ If `M` is dense and square and invertible (either assumed or checked; see option
 `check_invertibility`), we use `"inverse"`.
 Otherwise, we use `"vrep"`.
 
+Internally, this function operates on the level of the `AbstractPolyhedron`
+interface, but the actual algorithm uses dispatch on the concrete type of `P`,
+depending on the algorithm that is used:
+
+- `_linear_map_vrep(M, P)` if the vertex approach is used
+- `_linear_map_hrep(M, P, use_inv)` if the invertibility criterion is used,
+  where `use_inv` is determined by the algorithm
+
+New subtypes of the interface should define their own `_linear_map_vrep`
+(resp. `_linear_map_hrep`) for special handling of the linear map; otherwise
+the fallback implementation for `AbstractPolyhedron` is used.
+
 ### Algorithm
 
-This function implements two algorithms for the linear map:
+This function mainly implements two approaches for the linear map:
 
 - If the matrix ``M`` is invertible (which we check via a sufficient condition),
   then ``y = M x`` implies ``x = \\text{inv}(M) y`` and we transform the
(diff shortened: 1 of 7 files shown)


src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -414,16 +427,27 @@ function linear_map(M::AbstractMatrix{N},
...
+
+    elseif algorithm == "inverse"
+        return _linear_map_hrep(M, P, true)
+
+    elseif algorithm == "division"
+        return _linear_map_hrep(M, P, false)
+
+    elseif algorithm == "lift"
+        # check preconditions for this algorithm
+        !(m > n) && throw(ArgumentError("this function requires that the number " *
```
```suggestion
        m <= n && throw(ArgumentError("this function requires that the number " *
```

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -414,16 +427,27 @@ function linear_map(M::AbstractMatrix{N},
...
+        "of rows of the linear map is greater than the number of columns, but " *
+        "they are of size $m and $n respectively"))
+
+        r = rank(M)
+        !(r == n) && throw(ArgumentError("the rank of the given matrix is " *
```
```suggestion
        r != n && throw(ArgumentError("the rank of the given matrix is " *
```

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -414,16 +427,27 @@ function linear_map(M::AbstractMatrix{N},
...
+        "$r, but this function assumes that it is $n"))
```
```suggestion
        "$r, but this function requires that it is $n"))
```

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -483,6 +507,28 @@ function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::AbstractPolyhedron{N},
...
+# preconditions have been checked in the caller function
+function _linear_map_hrep_lift(M::AbstractMatrix{N}, P::AbstractPolyhedron{N}) where {N<:Real}
+    m, n = size(M)
+
+    # we extend M to an invertible m x m matrix by appending m-n columns
+    # orthogonal to the column space of M
+    Mext, inv_Mext = extend(M, check_rank=false)
+
+    # append zeros to the existing constraints, in the last m-n coordinates
+    cext = [HalfSpace(vcat(c.a, zeros(N, m-n)), c.b) for c in constraints_list(P)]
```
schillic: I wanted to suggest to outsource `vcat(c.a, zeros(N, m-n))` to a function call `extend(c.a, m-n)` and define that function elsewhere. But maybe that is overkill. As you prefer.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -483,6 +507,28 @@ function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::AbstractPolyhedron{N},
...
+
+    # now fix the last m-n coordinates to zero
+    id_out = Matrix(one(N)*I, m-n, m-n)
+    cext = vcat(cext, [HalfSpace(vcat(zeros(N, n), id_out[i, :]), zero(N)) for i in 1:(m-n)],
```
schillic: Is `vcat(zeros(N, n), id_out[i, :])` not just `SingleEntryVector(n+i, m, one(N))`?

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -483,6 +507,28 @@ function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::AbstractPolyhedron{N},
...
+                      [HalfSpace(vcat(zeros(N, n), -id_out[i, :]), zero(N)) for i in 1:(m-n)])
+
+    Pext = HPolytope(cext)
+
+    # now Mext is invertible and we can apply the linear with inverse function
```
```suggestion
    # now Mext is invertible and we can apply the inverse algorithm
```

src/Interfaces/AbstractPolyhedron_functions.jl
mforets: I think it is readable with a `vcat` (but there's always a personal bias on such things, of course). While we are at it, it is true that if we preallocate the `cext` and write a for loop this has some chance of giving some "free" efficiency.

mforets:
@cqql if you have some further comments/thoughts with respect to this function don't hesitate to write ;)

src/Interfaces/AbstractPolyhedron_functions.jl
mforets: That is correct, but note that `c.a` is not in general a `SingleEntryVector`. However, we are allowed to make the change you propose because `HPolytope` does not still carry the array type information of it's constraints list. But that is planned  (https://github.com/JuliaReach/LazySets.jl/issues/1802 and related issues) and then using mixed types is not a good idea (or this function will simply break). I'm not sure what to do..

mforets: I propose to leave it like it is now and open an issue for later with reference to [this idea](https://github.com/JuliaReach/LazySets.jl/pull/1939/files#r375331312) as well.

schillic: I do not think that a concrete `Vector` (that you use here) is always correct for the same reason. So I suggest to change it :smile:

schillic: `vcat` does the conversion to a common type for you:
```julia
julia> v = [1, 2]
2-element Array{Int64,1}:
 1
 2

julia> vcat(v, [3.0, 4.0])
4-element Array{Float64,1}:
 1.0
 2.0
 3.0
 4.0
```
So I think we are good anyway.

mforets: True, but it is not recursive,

```julia
julia> x = SingleEntryVector(1, 2, 1.0)
2-element SingleEntryVector{Float64}:
 1.0
 0.0

julia> y = rand(2)
2-element Array{Float64,1}:
 0.11055345051539978
 0.8948212121880166 

julia> vcat([HalfSpace(x, 1.0)], [HalfSpace(y, 2.0)])
2-element Array{HalfSpace{Float64,VN} where VN<:AbstractArray{Float64,1},1}:
 HalfSpace{Float64,SingleEntryVector{Float64}}([1.0, 0.0], 1.0)                     
 HalfSpace{Float64,Array{Float64,1}}([0.11055345051539978, 0.8948212121880166], 2.0)
```

schillic: Okay, as you prefer.

cqql:
Looks fine to me. The only remaining thing is just that the broken `"division"` algorithm is still included and should be replaced by an `"elimination"` algorithm that falls back to `"lift"` if possible.

mforets:
Exactly, that is now PR https://github.com/JuliaReach/LazySets.jl/pull/1950.

PR #1947, opened 2020-02-08
Add basetype function

schillic:
This implements the idea(s) from [here](https://discourse.julialang.org/t/extract-type-name-only-from-parametric-type/14188).


mforets:
Can we add these examples in the docstring?

```julia
julia> z = rand(Zonotope);

julia> basetype(z)
Zonotope

julia> basetype(z + z)
MinkowskiSum

julia> basetype(LinearMap(rand(2, 2), z+z))
LinearMap

```

Issue #1207, opened 2019-03-07
Quantifier elimination for projections of unbounded sets


mforets:
Currently, the list of constraints of an unbounded set cannot be obtained in some cases, e.g. the projection of a halfspace:

```julia
julia> S = Approximations.project(HalfSpace([1.0, 1.0, 0.0], 1.0), [1], LinearMap)
LinearMap{Float64,HalfSpace{Float64},Float64,SparseArrays.SparseMatrixCSC{Float64,Int64}}(
  [1, 1]  =  1.0, HalfSpace{Float64}([1.0, 1.0, 0.0], 1.0))

julia> constraints_list(S)
ERROR: ArgumentError: the linear map in vertex representation for an unbounded set is not defined
Stacktrace:
 [1] _linear_map_vrep(::SparseArrays.SparseMatrixCSC{Float64,Int64}, ::HalfSpace{Float64}) at /Users/forets/.julia/dev/LazySets/src/AbstractPolyhedron_functions.jl:335
 [2] #linear_map#18(::Bool, ::Float64, ::Bool, ::Function, ::SparseArrays.SparseMatrixCSC{Float64,Int64}, ::HalfSpace{Float64}) at /Users/forets/.julia/dev/LazySets/src/AbstractPolyhedron_functions.jl:313
 [3] linear_map at /Users/forets/.julia/dev/LazySets/src/AbstractPolyhedron_functions.jl:309 [inlined]
 [4] constraints_list(::LinearMap{Float64,HalfSpace{Float64},Float64,SparseArrays.SparseMatrixCSC{Float64,Int64}}) at /Users/forets/.julia/dev/LazySets/src/LinearMap.jl:377
 [5] top-level scope at none:0
```

A robust approach would be to use [quantifier elimination](https://en.wikipedia.org/wiki/Quantifier_elimination) to find the list of constraints above (in this case, the universal set). In this direction i've seen that [Reduce](https://chakravala.github.io/Reduce.jl/latest/man/11-substitution.html#.2.5-Substitutions-for-General-Expressions-1) offers quantifier elimination, but i haven't tried it.

schillic:
I have not thought about this thoroughly, but I have the feeling that the answer can be computed with a simpler algorithm. Here are some preliminary ideas to investigate (they may or may not work):
1. `constraints_list(project(X::AbstractPolyhedron)) == project(constraints_list(X))`
[...]

Issue #998, opened 2019-01-13
Add vertices_list for HPolyhedron


schillic:
Currently we throw an error. However, the implementation for `HPolytope` works for unbounded polytopes and does what you would expect.
I think we have decided to not support vertices of polyhedra, but since it is so easy, should we support it?
```julia
julia> using LazySets, Polyhedra

julia> constraints = [LazySets.HalfSpace([1., 0.], 0.), LazySets.HalfSpace([0., 1.], 0.)];

julia> vertices_list(HPolyhedron(constraints))
ERROR: ArgumentError: the list of vertices of a (possibly unbounded) polyhedron is not
 defined; if the polyhedron is bounded, try converting to `HPolytope` first
Stacktrace:
 [1] vertices_list(::HPolyhedron{Float64}) at LazySets/src/HPolyhedron.jl:631
 [2] top-level scope at none:0

julia> vertices_list(HPolytope(constraints))
1-element Array{Array{Float64,1},1}:
 [0.0, 0.0]
```

mforets:
I agree to let `vertices_list(::HPolyhedron)` work too.

schillic:
We can use this implementation in the (currently) undefined case of [`linear_map(::HPolyhedron)`](https://github.com/JuliaReach/LazySets.jl/pull/980/files#diff-795b585c4704738125c4e81ad4bff018R527).

schillic:
> We can use this implementation in the (currently) undefined case of linear_map(::HPolyhedron).

However, adding `vertices_list` is not a good idea because we would usually assume that the original set is the convex hull of the vertices (for instance the mentioned `linear_map`).
I changed my mind and vote for choosing a different name for the `vertices_list(::HPolyhedron)` method or for adding another method for polytopes that is identical to `vertices_list` but does not exist for `HPolyhedron` (which we then have to use internally).

schillic:
Usually the `vertices_list` method is defined and more efficient than going to H-rep first.
We should now of course define `vertices_list` in `AbstractPolyhedron_functions.jl`, exactly as you defined it here.
```julia
function vertices_list(P::AbstractPolyhedron{N}; check_boundedness::Bool=true) where {N<:Real}
    if check_boundedness && !isbounded(P)
        throw ArgumentError("polyhedra need to be bounded to obtain their vertices")
    end
    return vertices_list(HPolytope(constraints_list(P), check_boundedness=false)))
end
```

[...]

PR #1950, opened 2020-02-10
#1928 - Major update of concrete linear map

### LINKED ISSUE #1928: Inconsistency in linear_map between vrep and division algorithm
Hey,
check out the following example. It defines a polytope `P` that is restricted to 0 in its last two coordinates. It is bounded and non-empty. Therefore, applying a linear map should clearly not change its emptiness. And yet, the two algorithms disagree. `"vrep"` simply maps the vertices and finds the correct answer that `M * P` is also non-empty. But `"division"` produces an empty polytope.

I have looked at the implementation of `_linear_map_hrep` and it solves a system of linear equations to update the constraints. I am no expert in geometry but wouldn't a matrix with less rows than columns, i.e. a projection into a lower-dimensional space, require a proper variable elimination procedure like Fourier-Motzkin, even if the matrix has full row rank?

```julia
import Polyhedra
using LazySets

B = [
    -0.0 -0.0 -1.0 -0.0;
    -0.0 -0.0 -0.0 -1.0;
    0.0 0.0 1.0 0.0;
    0.0 0.0 0.0 1.0;
    2.9947065529531898 -1.1028295687569036 -0.0 -0.0;
    3.04356887693959 1.0 -0.0 -0.0;
    -3.043568876939177 -1.0 -0.0 -0.0;
    -2.7154753896639945 1.0 -0.0 -0.0;
    -3.1209566358111682 -1.0 -0.0 -0.0;
    -3.117487135897941 1.0 -0.0 -0.0;
    3.1209566358106873 1.0 -0.0 -0.0;
    4.400608112720001 -1.4115882186157707 -0.0 -0.0
]
b = [
    0.0,
    0.0,
    -0.0,
    -0.0,
    -1.0,
    3.98729727200628,
    -3.3782603325110476,
    1.803258851263224,
    -3.4045375136630747,
    1.6766140435881927,
    4.0235794003261,
    -1.0
]
P = HPolytope(B, b)
@assert !isempty(P)
@assert isbounded(P, false)

M = [
    -0.36650336 1.6418293 -0.6497878 -1.5073241;
    -0.3861024 -1.0513428 -1.083672 -0.6236442;
    0.41887435 -0.30923694 1.6104311 1.7603866
]

@assert !isempty(linear_map(M, P; algorithm="vrep"))
@assert !isempty(linear_map(M, P; algorithm="division"))
```

### LINKED ISSUE #1934: Revise "division" algorithm for linear map

### LINKED ISSUE #1935: Concrete linear map using block elimination
xref: https://github.com/JuliaReach/LazySets.jl/issues/1928

mforets:
Closes #1928.
Closes #1934.
Closes #1935.

----

Some comments on the user-facing changes:

- Now `linear_map(M, X)` never passes to the vrep by default; this is achieved by using the new `"elimination"` algorithm. However, `algorithm="vrep"` is still available if `vertices_list` is applicable *and* the set is not an `HPolyhedron` (but see the comments in the code, we should update this eventually).
- The defaults have changed, see the new function `_default_linear_map_algorithm`.
- Tests that used to throw an error now work, because the new algorithm `elimination` works for non-invertible matrices in H-representation, even for possibly unbounded polyhedra.

Some comments about the internal changes:

- We still use dispatch on two stages: basically, first on the type and then on the algorithm. This simplifies a bit the logic for what does each set type want to give as output type.
- Along the same lines, note that `_linear_map_hrep` functions do not return sets; they return lists of constraints.
- Originally the `_linear_map_hrep` function was meant to be used for `inverse`/`inverse_right` algorithms only. That was fine but we now have 2 more algorithms that work in the hrep, `lift` and `elimination`. Instead of passing around lots of possible kwargs (in particular, their possible combinations are meaningless), it was cleaner to dispatch on singleton types / types that carry the key data such as the matrix inverse or the polyhedra backend.
- Swapped `_linear_map_hrep_helper <--> _linear_map_hrep` names because i think it is more appropriate to call helper functions the intermediate function that dispatches on the set type, and reserve the name `_linear_map_hrep` for the actual computation.
- Added some checking functions `_check_algorithm_applies` which dispatch on each linear map algorithm. The motivation to outsource these checks is that a priori we have to make them twice:
    - If no algorithm is passed (`algorithm == nothing`), we want to decide which algorithm is applicable according to the defaults.
    - If some algorithm is passed, we still want to check if it applies (unless explicitly said not to do it, eg. with the `check_invertibility` flag), and raise an error message.


mforets:
@schillic sorry for the "patch bomb". As an excuse, my initial attempt was to do #1934 and #1935 separately, but that didn't make a lot of sense since elimination plays a central role and i would have to make the same work (especially docs) twice, since elimination is now chosen as one of the defaults.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -330,124 +444,187 @@ that was used:
...
+on heuristics on the types and values of `M` and `P`:
+
+- If the `"inverse"` algorithm applies, it is used. This algorithm is applied
+  unconditionally if the `inverse` matrix is passed, or if `check_invertibility`
+  is set to `true`.
+- Otherwise, if the `"lift"` algorithm applies, it is used.
+- Otherwise, the `"elimination"` algorithm is used.
+
+Note that `"inverse"` does not require the external library `Polyhedra`, and so
+does `"lift"`. However, the fallback method `"lift"` does require `Polyhedra`
```
mforets:
```suggestion
does `"inverse_right"`. However, the fallback method `"lift"` does require `Polyhedra`
```

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -330,124 +444,187 @@ that was used:
...
+on heuristics on the types and values of `M` and `P`:
+
+- If the `"inverse"` algorithm applies, it is used. This algorithm is applied
+  unconditionally if the `inverse` matrix is passed, or if `check_invertibility`
+  is set to `true`.
+- Otherwise, if the `"lift"` algorithm applies, it is used.
+- Otherwise, the `"elimination"` algorithm is used.
+
+Note that `"inverse"` does not require the external library `Polyhedra`, and so
+does `"inverse_right"`. However, the fallback method `"lift"` does require `Polyhedra`
```
mforets:
```suggestion
does `"inverse_right"`. However, the fallback method `"elimination"` does require `Polyhedra`
```

cqql:
I think the default algorithm should be `vrep` if `P` is a `VPolytope`. Everything else looks fine to me.

mforets:
Yes. This is handled by dispatch on the top-level `linear_map`, see [this function](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Sets/VPolytope.jl#L261).

schillic:
@mforets: I added a "Closes #1207" in the OP. Can you rebase and add a test for `project`, please?

src/Interfaces/AbstractPolytope.jl
```diff
@@ -144,13 +144,12 @@ end
...
-                                  inverse::Union{Nothing, AbstractMatrix{N}}=nothing
-                                 ) where {N<:Real}
-    constraints = _linear_map_hrep_helper(M, P, use_inv; inverse=inverse)
+function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::AbstractPolytope{N},
+                                 algo::AbstractLinearMapAlgorithm) where {N<:Real}
+    constraints = _linear_map_hrep(M, P, algo)
     m = size(M, 1) # output dimension
     if m == 1
+        # TODO: create interval directly ?
         return convert(Interval, HPolygon(constraints))
```
schillic: Good point. We can solve that later.
How would it look like? Do we know that `_linear_map_hrep` returns exactly two constraints?

src/Sets/HalfSpace.jl
```diff
@@ -484,3 +477,14 @@ function translate(hs::HalfSpace{N}, v::AbstractVector{N}; share::Bool=false
     b = hs.b + dot(hs.a, v)
     return HalfSpace(a, b)
 end
+
+function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::HalfSpace{N},
+                          algo::AbstractLinearMapAlgorithm) where {N<:Real}
+    constraints = _linear_map_hrep(M, P, algo)
+    if length(constraints) == 1
+        c = first(constraints)
+        return HalfSpace(c.a, c.b)
```
schillic: Should `c` not be a `HalfSpace` already? :smile:

src/Sets/Hyperplane.jl
```diff
@@ -432,11 +432,13 @@ function _constraints_list_hyperplane(a::AbstractVector{N}, b::N
...
-                         ) where {N<:Real}
-    constraint = _linear_map_hrep_helper(M, P, use_inv, inverse=inverse)[1]
-    return Hyperplane(constraint.a, constraint.b)
+function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::Hyperplane{N},
+                                 algo::AbstractLinearMapAlgorithm) where {N<:Real}
+    constraints = _linear_map_hrep(M, P, algo)
+    @assert length(constraints) == 2
+    # TODO: assert that the constraints define a hyperplane?
+    c = first(constraints)
+    return Hyperplane(c.a, c.b)
```
schillic: I think a unit test should be enough.

src/Sets/Line.jl
```diff
@@ -323,11 +323,17 @@ function constrained_dimensions(L::Line{N}) where {N<:Real}
...
+                                 algo::AbstractLinearMapAlgorithm) where {N<:Real}
+    constraints = _linear_map_hrep(M, P, algo)
+    if length(constraints) == 1
+        c = first(constraints)
+        return Line(c.a, c.b)
+    else
+        # TODO: check and transform 2 constraints to a line? ie. add constructor
+        # from constraints? (see also the corresponding function for HalfSpace)
+        return HPolyhedron(constraints)
+    end
```
schillic: I do not understand why this code is not the same as for `Hyperplane` since the `Line` is a `Hyperplane` in 2D. In particular, since the original constraints are identical up to the sign, it should not happen that one of them gets removed. There can probably be a common helper function to handle `Line` and `Hyperplane` together.

However, I agree that the old code was wrong in two respects:
1. It did not check whether the output dimension is 2.
2. It did not check whether the the map is a projection, which can destroy the property of being a hyperplane/line. Take the 2D example `x + y = 0` and project to the x axis, which gives you the 1D universe.

test/unit_Hyperplane.jl
```diff
@@ -64,10 +64,15 @@ for N in [Float64, Rational{Int}, Float32]
...
-    # test concrete linear map of a hyerplane
+    # test concrete linear map of a hyperplane
     H = Hyperplane(N[1, -1], N(0)) # x = y
     M = N[1 0; 0 0] # non-invertible matrix
-    @test_throws ArgumentError linear_map(M, H)
+    if N == Float32 || N == Float64
+        @test linear_map(M, H) isa Hyperplane{Float64}
+    elseif N == Rational{Int}
+        @test linear_map(M, H) isa Hyperplane{Rational{BigInt}}
+    end
```
schillic: I would test for the concrete result, not just the type.

test/unit_Line.jl
```diff
@@ -72,9 +72,13 @@ for N in [Float64, Rational{Int}, Float32]
...
     L = Line(N[1, -1], N(0)) # x = y
     M = N[1 0; 0 0] # non-invertible matrix
-    @test_throws ArgumentError linear_map(M, L)
-    M = N[2 2; 0 1] # invertible matrix
-    @test linear_map(M, L) == Line(N[0.5, -2.0], N(0.0))
+    if N == Rational{Int}
+        @test linear_map(M, L) isa HPolyhedron{Rational{BigInt}}
+    elseif N == Float32 || N == Float64
+        @test linear_map(M, L) isa HPolyhedron{Float64}
+    end
```
schillic: Same (test for the concrete result, not just the type).

test/unit_Line.jl
```diff
@@ -72,9 +72,13 @@ for N in [Float64, Rational{Int}, Float32]
...
+    #M = N[2 2; 0 1] # invertible matrix
+    #@test linear_map(M, L) == Line(N[0.5, -2.0], N(0.0)) #TODO: see _linear_map_hrep_helper in Line.jl
```
schillic: Revisit this after fixing the `Line` case. I do not see why this would not work.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -526,7 +708,31 @@ function _linear_map_hrep_lift(M::AbstractMatrix{N}, P::AbstractPolyhedron{N}) w
...
+    id_m = Matrix(one(N)*I, m, m)
+    backend = algo.backend
+    method = algo.method
+
+    # extend the polytope storing the y variables first
+    Ax_leq_b = [Polyhedra.HalfSpace(vcat(zeros(N, m), c.a), c.b) for c in constraints_list(P)]
+    y_eq_Mx = [Polyhedra.HyperPlane(vcat(-id_m[i, :], M[i, :]), zero(N)) for i in 1:m]
+
+    Phrep = Polyhedra.hrep(y_eq_Mx, Ax_leq_b)
+    Phrep_cdd = polyhedron(Phrep, backend)
```
schillic: I would not use `cdd` as name. The code is general.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -526,7 +708,31 @@ function _linear_map_hrep_lift(M::AbstractMatrix{N}, P::AbstractPolyhedron{N}) w
...
+    Peli_block = Polyhedra.eliminate(Phrep_cdd, (m+1):(m+n), method)
+    Peli_block = Polyhedra.removeduplicates(Polyhedra.hrep(Peli_block))
+
+    return constraints_list(HPolytope(Peli_block)) # TODO: take constraints directly?
```
```suggestion
    return constraints_list(HPolyhedron(Peli_block)) # TODO: take constraints directly?
```

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -476,39 +655,42 @@ function _linear_map_vrep(M::AbstractMatrix{N}, P::AbstractPolyhedron{N}) where
...
 end
 
-function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::AbstractPolyhedron{N},
-                                 use_inv::Bool;
-                                 inverse::Union{Nothing, AbstractMatrix{N}}=nothing
-                                ) where {N<:Real}
+# preconditions should have been checked in the callinear_map_hrep(M, P, algo)ler function
+function _linear_map_hrep(M::AbstractMatrix{N}, P::AbstractPolyhedron{N},
+                          algo::LinearMapInverse) where {N}
+    inverse = isempty(algo.inverse) ? inv(M) : algo.inverse
```
schillic: What does `isempty(::Matrix)` mean? I think you want to remove the type constraint that `algo.inverse` has to be a matrix and instead assign the default value `nothing`.
Maybe even define a function `hasinverse(algo::LinearMapInverse)` for clarity.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -266,42 +266,156 @@ function remove_redundant_constraints(
...
+                            "invertible matrix"))
+        return false
+    end
+
+    dense_condition = !issparse(M)
+    if !dense_condition
+        throw_error && throw(ArgumentError("the inverse of a sparse matrix can " *
+        "often be dense and can cause the computer to run out of memory. If you " *
+        "are sure you have enough memory, please convert your matrix to a dense matrix; " *
+        "try to pass `Matrix(M)`"))
```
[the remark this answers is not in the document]
mforets: yes, i copied over the message from julia,

```julia
julia> using LinearAlgebra, SparseArrays

julia> inv(sprandn(2, 2, 1.))
ERROR: The inverse of a sparse matrix can often be dense and can cause the computer to run out of memory. If you are sure you have enough memory, please convert your
 matrix to a dense matrix.
Stacktrace:
 [2] inv(::SparseMatrixCSC{Float64,Int64}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/SparseArrays/src/linalg.jl:1263
 [3] top-level scope at REPL[2]:1
```
i agree to shorten it and suggest to use `"inverse_right"` instead.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -266,42 +266,156 @@ function remove_redundant_constraints(
...
+        throw_error && throw(ArgumentError("algorithm \"lift\" requires that the number " *
+            "of rows of the linear map is greater than the number of columns, but " *
+            "they are of size $m and $n respectively"))
+        return false
+    end
+
+    rank_condition = rank(M) == n
+    if !rank_condition
+        throw_error && throw(ArgumentError("the rank of the given matrix is " *
+            "$r, but algorithm \"lift\" requires that it is $n"))
```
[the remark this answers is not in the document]
mforets: oops, good catch!

src/Interfaces/AbstractPolyhedron_functions.jl
mforets: The reason was that the empty matrix is used as a check that the inverse has not been assigned yet; this is needed [here](https://github.com/JuliaReach/LazySets.jl/pull/1950/files#diff-5e796138f8958d1d7c93fd38b67eb60aR599, because the matrix is computed in the `_linear_map_hrep_helper`.

ou're right that it is a bit odd to use the struct that way. After a second thought, here is my new proposal:

- Observe that the `_check_algorithm_applies` functions do not actually need an instances of the algorithm, so we can dispatch on its type instead.
- Remove the line `inverse = isempty(algo.inverse) ? inv(M) : algo.inverse`, and compute it before hand.

I think that the logic is better in this way :+1:

schillic: >     * Observe that the `_check_algorithm_applies` functions do not actually need an instance of the algorithm, so we can dispatch on its type instead.

I think that is what I meant: allow the type `Nothing` for `algo.inverse` and then you can dispatch on that.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -330,124 +444,187 @@ that was used:
...
-Otherwise, we use `"vrep"`.
-
-Internally, this function operates on the level of the `AbstractPolyhedron`
-interface, but the actual algorithm uses dispatch on the concrete type of `P`,
-depending on the algorithm that is used:
-
-- `_linear_map_vrep(M, P)` if the vertex approach is used
-- `_linear_map_hrep(M, P, use_inv)` if the invertibility criterion is used,
-  where `use_inv` is determined by the algorithm
+### Algorithm
```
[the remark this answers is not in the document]
mforets: ok, i swapped them

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -266,42 +266,156 @@ function remove_redundant_constraints(
...
+    if _check_algorithm_applies(M, P, LinearMapInverse(), cond_tol=cond_tol)
+        algo = LinearMapInverse(inv(M))
+    elseif _check_algorithm_applies(M, P, LinearMapLift())
+        algo = LinearMapLift()
+    else
+        require(:Polyhedra; fun_name="linear_map with elimination")
+        require(:CDDLib; fun_name="linear_map with elimination")
+        if backend == nothing
+            backend = default_cddlib_backend(N)
+        end
```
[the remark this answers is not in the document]
mforets: I think it's fine to keep the `require` inside the clause. Or i could move these require to the actual algorithm `_linear_map_hrep` (as it is the case for the `vrep` algorithm).

The fact of requiring `CDDLib` is that it is the only library (afaik) that does variable elimination.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -330,124 +444,187 @@ that was used:
...
+and variable lifting. Depending on the properties of `M` and `P`, one algorithm
+may be preferable over the other. Details on the algorithms are given in the
+following subsections.
+
+If the algorithm argument is not specified, a default option is chosen based
+on heuristics on the types and values of `M` and `P`:
+
+- If the `"inverse"` algorithm applies, it is used. This algorithm is applied
+  unconditionally if the `inverse` matrix is passed, or if `check_invertibility`
+  is set to `true`.
```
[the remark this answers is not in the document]
mforets: In commit `improve check_invertibility case` i fixed this and also improved the logic for check invertibility, such that if the user passes `check_invertibility=false`, either `inverse` or `inverse_right` is used (or the algorithm passed), *without* checking for invertibility.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -330,124 +444,187 @@ that was used:
...
-        "$r, but this function requires that it is $n"))
+        algo = LinearMapLift()
+        _check_algorithm_applies(M, P, algo, throw_error=true)
+        return _linear_map_hrep_helper(M, P, algo)
 
-        return _linear_map_hrep_lift(M, P)
     else
         throw(ArgumentError("got unknown algorithm \"$algorithm\"; " *
-            "available choices: \"vrep\", \"inverse\", \"division\", \"lift\""))
+            "available choices: \"vrep\", \"inverse\", \"inverse_right\", \"lift\", \"elimination\""))
```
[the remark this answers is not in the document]
mforets: alright, i reordered them (as they are chosen in by default)

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -474,16 +474,25 @@ following subsections.
 If the algorithm argument is not specified, a default option is chosen based
 on heuristics on the types and values of `M` and `P`:
 
-- If the `"inverse"` algorithm applies, it is used. This algorithm is applied
-  unconditionally if the `inverse` matrix is passed, or if `check_invertibility`
-  is set to `true`.
+- If the `"inverse"` algorithm applies, it is used.
+- If the `"inverse_right"` algorithm applies, it is used.
```
```suggestion
- Otherwise, if the `"inverse_right"` algorithm applies, it is used.
```

src/Interfaces/AbstractPolyhedron_functions.jl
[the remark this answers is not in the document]
schillic: > The fact of requiring `CDDLib` is that it is the only library (afaik) that does variable elimination.

Fine, but it should still make sense to move it inside the `if` because at some point there may be a different package.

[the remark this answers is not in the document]
mforets: Ok, i see what you mean now :+1: 

```julia
        require(:Polyhedra; fun_name="linear_map with elimination")
        if backend == nothing
            require(:CDDLib; fun_name="linear_map with elimination")
            backend = default_cddlib_backend(N)
        end
```

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -330,124 +444,187 @@ that was used:
...
+`_linear_map_vrep`) are invoked, which dispatch on the set type. Then, each helper
+function calls the concrete implementation of `_linear_map_hrep`, which dispatches
+on the algorithm, and returns a list of constraints.
+
+To simplify working with different algorithms different sets of options, the
+types `<: AbstractLinearMapAlgorithm` are used. These types are singleton type
+or types that carry only the key data for the given algorithm, such as the matrix
+inverse or the polyhedra backend.
+
+New subtypes of the `Polyhedra` interface may define their own helper functions
```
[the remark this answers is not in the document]
mforets: Yes, i changed this to 

> New subtypes of the `AbstractPolyhedron` interface

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -330,124 +444,187 @@ that was used:
...
-    elseif algorithm == "inverse"
-        return _linear_map_hrep(M, P, true)
+    elseif algorithm == "vrep"
+        return _linear_map_vrep(M, P)
 
-    elseif algorithm == "division"
-        return _linear_map_hrep(M, P, false)
+    elseif got_inv
+        algo = LinearMapInverse()
+        check_invertibility && _check_algorithm_applies(M, P, algo; cond_tol=cond_tol, throw_error=true)
```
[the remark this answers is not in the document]
mforets: I would leave this as is, because inside this clause (`got_inv`), the user has forced `algorithm="inverse"`.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -330,124 +444,187 @@ that was used:
...
+    elseif algorithm == "elimination" || algorithm == "elim"
+        require(:Polyhedra; fun_name="linear_map with elimination")
+        require(:CDDLib; fun_name="linear_map with elimination")
+        if backend == nothing
+            backend = default_cddlib_backend(N)
+        end
+        if elimination_method == nothing
+            elimination_method = Polyhedra.BlockElimination()
+        end
+        algo = LinearMapElimination(backend, elimination_method)
```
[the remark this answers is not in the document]
mforets: Good, i refactored that part into a new function `_get_elimination_instance(N, backend, elimination_method)`.

src/Interfaces/AbstractPolyhedron_functions.jl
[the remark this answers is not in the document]
schillic: I do not see the logic in checking invertibility but not density if the algorithm is fixed.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -330,124 +444,187 @@ that was used:
...
+is good as a workaround when `M` is sparse (since the `inv` function is not available
+for sparse matrices).
+
+### Vertex representation
+
+This algorithm is invoked with the keyword algorithm `algorithm="vrep"`.
+The idea is to convert the polyhedron to its vertex representation and apply the
+linear map to each vertex of `P`.
+
+The returned set is a polyhedron in vertex representation. Note that conversion of
```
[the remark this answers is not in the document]
mforets: True. At least, until we have https://github.com/JuliaReach/LazySets.jl/issues/998

src/Interfaces/AbstractPolyhedron_functions.jl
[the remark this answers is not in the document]
mforets: To give priority to the algorithm choice. (= if you are not sure, use the default).

src/Interfaces/AbstractPolytope.jl
mforets: > How would it look like?
> Do we know that _linear_map_hrep returns exactly two constraints?

I can't guarantee it always returns 2 constraints.. as we're not explicitly removing redundant constraints..

src/Sets/HalfSpace.jl
mforets: i did this for clarity (and to ensure returning a new HalfSpace instance) but i think we can just return `fist(constraints)` directly :+1:

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -266,42 +266,156 @@ function remove_redundant_constraints(
...
+
+    return true
+end
+
+function _check_algorithm_applies(M::AbstractMatrix{N}, P::AbstractPolyhedron{N},
+                                   ::LinearMapVRep; throw_error=false) where {N}
+
+    # TODO: the second check should be !isbounded(P) but this may be expensive;
+    # see also #998
+    if !applicable(vertices_list, P) || (P isa HPolyhedron)
```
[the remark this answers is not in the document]
mforets: ok, i changed this check to

```julia
    is_polytopic = P isa AbstractPolytope
    if !is_polytopic

File: src/Sets/Line.jl (hunk 6 above)
Comment (mforets) [reply src/Sets/Line.jl]: ok, i have removed the TODO and kept 

```julia
function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::Line{N},
                                 algo::AbstractLinearMapAlgorithm) where {N<:Real}
     constraints = _linear_map_hrep(M, P, algo)
     @assert length(constraints) == 2
     c = first(constraints)
     return Line(c.a, c.b)
end
```
in particular, what i meant is that we are not checking that the 2 constraints are of the form `ax <= b` and `ax >= b`, which implies `ax = b`.

src/Interfaces/AbstractPolytope.jl
schillic: Now that I see this again, `convert(Interval, HPolygon)` looks really wrong to me. Why does this even work? :smile:

src/Sets/Line.jl
schillic: What about my second concern? I think that is really a problem:
> It did not check whether the map is a projection, which can destroy the property of being a hyperplane/line. Take the 2D example x + y = 0 and project to the x axis, which gives you the 1D universe.

test/unit_Line.jl
mforets: Good, the test can be uncommented as it works now.

src/Sets/Line.jl
mforets: Not sure that i follow, do you mind opening a follow up issue? (no regression in this branch if you say it wasn't checked before..)

src/Interfaces/AbstractPolytope.jl
mforets: :smile:  it was silly to convert from a polygon to an interval, it doesn't work (and it shouldn't):

```julia
julia> convert(Interval, rand(HPolygon))
ERROR: AssertionError: can only convert a one-dimensional HPolygon{Float64} to `Int
erval`
Stacktrace:
 [1] macro expansion at /home/mforets/.julia/dev/LazySets/src/Assertions/Assertions
.jl:23 [inlined]
 [2] convert(::Type{Interval}, ::HPolygon{Float64}) at /home/mforets/.julia/dev/Laz
ySets/src/convert.jl:513
 [3] top-level scope at REPL[6]:1
```
So i propose to change that line to `convert(Interval, HPolytope(constraints))`.

src/Sets/Line.jl
schillic: #1989

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -526,7 +708,31 @@ function _linear_map_hrep_lift(M::AbstractMatrix{N}, P::AbstractPolyhedron{N}) w
...
+function _linear_map_hrep(M::AbstractMatrix{N}, P::AbstractPolyhedron{N},
+                          algo::LinearMapElimination) where {N<:Real}
+    m, n = size(M)
+    id_m = Matrix(one(N)*I, m, m)
+    backend = algo.backend
+    method = algo.method
+
+    # extend the polytope storing the y variables first
+    Ax_leq_b = [Polyhedra.HalfSpace(vcat(zeros(N, m), c.a), c.b) for c in constraints_list(P)]
+    y_eq_Mx = [Polyhedra.HyperPlane(vcat(-id_m[i, :], M[i, :]), zero(N)) for i in 1:m]
```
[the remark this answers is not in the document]
schillic: Now you do the `-` twice :wink:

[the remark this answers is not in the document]
mforets: Line number?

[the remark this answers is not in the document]
schillic: > `₋Id_m` is a symbol

:roll_eyes:
Then it is a bad name :smiley:

mforets:
Thanks for your thorough review :+1:  I'm glad we could get this new functionality into the codebase :checkered_flag: 

I'll merge when CI has finished..

Issue #1429, opened 2019-06-11
Outsource project from Approximations


schillic:
Reason: There is no approximation involved and we can use this method in `LazySets` as well.

https://github.com/JuliaReach/LazySets.jl/blob/0aa98258aa428bfd5527973665d50c51750f9a19/src/Approximations/decompositions.jl#L577-L584

This will also remove the dependency of `LazySets` on `Approximations` (see https://github.com/JuliaReach/LazySets.jl/pull/1421#discussion_r292359894).

Furthermore, we could add an optional argument to skip the `constrained_dimensions` check. We can also rename to `project_unconstrained` to make this clear.

For the concrete projection (using `linear_map`) we should also be able to pass the `algorithm` argument.

Issue #1331, opened 2019-05-02
Add default value in project


mforets:
I think it would be nice to have a default output set type in `LazySets.Approximations.project`.

```julia
juila> Z = rand(Zonotope);

julia>  project(Z, [1, 2])
MethodError: no method matching project(::Zonotope{Float64}, ::Array{Int64,1})

julia>  project(Z, [1, 2], LinearMap)
LazySets.LinearMap{Float64,Zonotope{Float64},Float64,SparseArrays.SparseMatrixCSC{Float64,Int64}}(
  [1, 1]  =  1.0
  [2, 2]  =  1.0, Zonotope{Float64}([0.25482, -1.88379], [1.17086 1.00467 -0.526207 -0.958695; -0.338528 0.680383 0.687293 1.76753]))
```

Using a lazy linear map by default sounds good to me.

CC: @kostakoida

schillic:
I am fine with that. The implementation probably requires to add a new method.

mforets:
Actually i think this deserves further discussion. I would argue to make some (breaking) changes:

1- `project` should return the "exact" projection, i.e. a concrete set.
2- `Projection(::LazySet, ::AbstractVector)` should be a helper constructor for the linear map of a projection where you pass the variables on which you want to project
3- Other `project` functions can be removed, e.g. those that are a composition of a lazy projection (with a `LinearMap`) and with an overapproximation.

Note: some `project` methods do already behave like 2, eg. `project(P::HPolyhedron{N}, block::AbstractVector{Int})` and `project(H::HalfSpace{N}, block::AbstractVector{Int})`.

schillic:
`project` is originally just a helper function for `decompose`. So I suggest to rename this helper method in a first step.

mforets:
> 1- project should return the "exact" projection, i.e. a concrete set.

This is now the case in the current development version. There is still room for improvement and specialization. For instance,

```julia
julia> project(Z, [1, 2])
HPolyhedron{Float64}(HalfSpace{Float64,VN} where VN<:AbstractArray{Float64,1}[HalfSpace{Float64,Array{Float64,1}}([-0.27016058730099896, -0.962815276711571
2], 3.234200957777295), HalfSpace{Float64,Array{Float64,1}}([0.27016058730099896, 0.9628152767115712], -0.012263323795623116), HalfSpace{Float64,Array{Floa
[...]

PR #1956, opened 2020-02-13
#1726 - Concrete decomposition using linear_map

### LINKED ISSUE #1726: Concrete decomposition using linear_map
Offer the option to do `decompose` not with the lazy `LinearMap` but with the concrete `linear_map`.

schillic:
Closes #1726.

We use dispatch to choose the projection algorithm. I first thought about creating a dummy type for this purpose but I think passing `nothing` is acceptable since the concrete projection should actually be the standard.

### DIFF
--- a/src/Approximations/decompositions.jl
+++ b/src/Approximations/decompositions.jl
@@ -259,7 +259,7 @@ end
 """
     project(S::LazySet{N},
             block::AbstractVector{Int},
-            ::Nothing,
+            [::Nothing=nothing],
             [n]::Int=dim(S)
            ) where {N<:Real}
 
@@ -269,7 +269,7 @@ Project a high-dimensional set to a given block by using a concrete linear map.
 
 - `S`       -- set
 - `block`   -- block structure - a vector with the dimensions of interest
-- `nothing` -- used for dispatch
+- `nothing` -- (default: `nothing`) used for dispatch
 - `n`       -- (optional, default: `dim(S)`) ambient dimension of the set `S`
 
 ### Output
@@ -282,7 +282,7 @@ We apply the function `linear_map`.
 """
 @inline function project(S::LazySet{N},
                          block::AbstractVector{Int},
-                         ::Nothing,
+                         ::Nothing=nothing,
                          n::Int=dim(S)
                         ) where {N<:Real}
     m = length(block)
--- a/src/Approximations/decompositions.jl
+++ b/src/Approximations/decompositions.jl
@@ -237,7 +237,8 @@ function decompose(S::LazySet{N},
                    block_options::Union{Type{<:LazySet},
                                         Pair{<:UnionAll, <:Real},
                                         Real,
-                                        Type{<:AbstractDirections}
+                                        Type{<:AbstractDirections},
+                                        Nothing
                                        }
                   ) where {N<:Real}
     n = dim(S)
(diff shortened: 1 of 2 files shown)


src/Approximations/decompositions.jl
```diff
@@ -220,7 +220,7 @@ julia> typeof(res[1]), typeof(res[2])
 """
 function decompose(S::LazySet{N},
                    partition::AbstractVector{<:AbstractVector{Int}},
-                   block_options
+                   block_options::AbstractVector
```
schillic: This change is unrelated but I found it safer to restrict the type because we iterate over this object.

src/Approximations/decompositions.jl
```diff
@@ -255,6 +256,40 @@ function decompose(S::LazySet, block_options; block_size::Int=1)
...
+
+### Output
+
+A set representing the projection of the set `S` to block `block`.
+
+### Algorithm
+
+We apply the function `linear_map`.
+"""
+@inline function project(S::LazySet{N},
```
mforets: I'm not sure about inlining.. is there some general rule that applies to using `@inline`?

src/Approximations/decompositions.jl
```diff
@@ -255,6 +256,40 @@ function decompose(S::LazySet, block_options; block_size::Int=1)
...
+                         block::AbstractVector{Int},
+                         ::Nothing,
+                         n::Int=dim(S)
```
mforets: Here we can add the `algorithm` option that is passed to the linear map. But i agree to keep the signature of this new project function short at least for now.

src/Approximations/decompositions.jl
```diff
@@ -255,6 +256,40 @@ function decompose(S::LazySet, block_options; block_size::Int=1)
...
+
+A set representing the projection of the set `S` to block `block`.
+
+### Algorithm
+
+We apply the function `linear_map`.
+"""
+@inline function project(S::LazySet{N},
+                         block::AbstractVector{Int},
+                         ::Nothing,
```
mforets: The call `decompose(b, partition, nothing)` with a `nothing` is strange.. can we make this the default, so we don't have to enter the dummy argument?

src/Approximations/decompositions.jl
schillic: I copied this from the other `project` methods.

src/Approximations/decompositions.jl
schillic: The issue #1726 that this PR closes wants to offer a concrete `decompose`. The `project` methods are called from `decompose` by just passing its argument, and it is not possible to have no argument in `decompose`.

https://github.com/JuliaReach/LazySets.jl/blob/da8213f0ba11c9b1513e5f8100378db0ea3af444/src/Approximations/decompositions.jl#L221-L229

We wanted to add the default in #1331. But I agree that the change is simple and I can do it here already.

src/Approximations/decompositions.jl
schillic: I suggest to add the algorithm when we outsource the `project` methods in #1429.

src/Approximations/decompositions.jl
mforets: Ok, thanks.

src/Approximations/decompositions.jl
mforets: Ok. We can review the interface of decompose later. Also, i remember that it was type unstsble for a reachability use case.

About the default, here or later as you prefer.

src/Approximations/decompositions.jl
mforets: Alright 👍

src/Approximations/decompositions.jl
schillic: It was a good idea, I did not see that.
```julia
julia> B = rand(BallInf)
BallInf{Float64}(t[-0.41391500435046535, 0.9008017376015398], 0.25252464855799645)                                             
                                                                                                                                     
julia> project(B, [1])
Zonotope{Float64}([-0.41391500435046535], [0.25252464855799645])                                                                     
```

PR #1959, opened 2020-02-14
#1207 - Concrete projection of unbounded sets

### LINKED ISSUE #1207: Quantifier elimination for projections of unbounded sets

mforets:
Closes #1207.


src/Approximations/decompositions.jl
```diff
@@ -636,7 +641,8 @@ function project(P::AbstractPolyhedron{N}, block::AbstractVector{Int}) where {N}
     if constrained_dimensions(P) ⊆ block
         return HPolyhedron([HalfSpace(c.a[block], c.b) for c in constraints_list(P)])
     else
-        error("the concrete projection of a polyhedron " *
-              "for a general block structure is not implemented yet")
```
schillic: The documentation still says that this case is not supported and should be updated.

mforets: Good catch, i have updated the Algorithm section.

I've also hard-coded the return type, because this is what the function's "Output" documentation says.

mforets: (Note that the concrete linear map, although it will operate in hrep, it may return other set types eg. HPolygon's iirc).

schillic: > I've also hard-coded the return type, because this is what the function's "Output" documentation says.

You know that you can also change the documentation instead :wink:

mforets: Yes, but i would rather move that (breaking) change for consideration in a follow-up issue. In general, i think it's a good idea that the returned type is "closer" to the actual set. But for consistency we should also apply the same logic for existing case; in particular, note that the existing `constrained_dimensions(P) ⊆ block` case *always* returns an `HPolyhedron`, irrespective of the set dimension.

mforets: Should i resolve this conversation? I forgot what we wanted to do here...

schillic: > Yes, but i would rather move that (breaking) change for consideration in a follow-up issue. In general, i think it's a good idea that the returned type is "closer" to the actual set.

This is what you suggested. I leave it to you to decide what you want :wink:

mforets: Ok, ill open an issue. The proposal here is to return an interval / halfspace / line / hpolygon / hpolytope / hpolyhedron depending on the result of the projection..

PR #1962, opened 2020-02-14
#1960 - Pass constructor argument in concrete m-sum  

### LINKED ISSUE #1960: Pass constructor argument to concrete m-sum of zonotopic sets
The concrete Minkowski sum of zonotopic sets does not receive the `remove_zero_generators` keyword argument to the zonotope constructor. This should be added.

A possible caveat is that the `AbstractZonotope` interface does not require `remove_zero_generators` to be defined (and i think that.. it shouldn't :) ). We should rather dispatch on `Zonotopes`. After all, there is no logic for the return type in that fallback method (it is hardcoded to return a `Zonotope`).

mforets:
Closes #1960.

### DIFF
--- a/src/ConcreteOperations/minkowski_sum.jl
+++ b/src/ConcreteOperations/minkowski_sum.jl
@@ -218,8 +218,8 @@ function minkowski_sum(H1::AbstractHyperrectangle{N},
 end
 
 """
-    minkowski_sum(Z1::AbstractZonotope{N}, Z2::AbstractZonotope{N})
-        where {N<:Real}
+    minkowski_sum(Z1::AbstractZonotope{N}, Z2::AbstractZonotope{N};
+                  [remove_zero_generators::Bool]) where {N<:Real}
 
 Concrete Minkowski sum of a pair of zonotopic sets.
 
@@ -227,6 +227,8 @@ Concrete Minkowski sum of a pair of zonotopic sets.
 
 - `Z1` -- zonotopic set
 - `Z2` -- zonotopic set
+- `remove_zero_generators` -- (optional, default: `true`) if `true`, remove the
+                              zero generators from the resulting zonotope
 
 ### Output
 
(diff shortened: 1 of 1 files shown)


schillic:
I disagree with this approach. There is no reason to check for zero generators. I do not want to propagate sanity checks to all functions.

mforets:
I would suggest to outsource move your concerns to the default value to a separate issue. This function does not change the behavior in `master`; it just allows users to pass the option.

src/ConcreteOperations/minkowski_sum.jl
```diff
@@ -235,9 +237,13 @@ A `Zonotope` corresponding to the concrete Minkowski sum of `Z1` and `Z2`.
...
+equal to zero) of `Z1 ⊕ Z2` are removed.
 """
-function minkowski_sum(Z1::AbstractZonotope{N},
-                       Z2::AbstractZonotope{N}) where {N<:Real}
-    return Zonotope(center(Z1) + center(Z2), [genmat(Z1) genmat(Z2)])
+function minkowski_sum(Z1::AbstractZonotope{N}, Z2::AbstractZonotope{N};
+                       remove_zero_generators::Bool=true) where {N<:Real}
+    cnew = center(Z1) + center(Z2)
+    Gnew = hcat(genmat(Z1), genmat(Z2))
+    return Zonotope(cnew, Gnew, remove_zero_generators=remove_zero_generators)
```
schillic:
```suggestion
    return Zonotope(cnew, Gnew, remove_zero_generators=false)
```

mforets: Ok.. i disagree here. We may prefer to change this in the future, i know, but for now and as i said in the other issue, i prefer to use the constructor's defaults ( = keep the behavior in master).

schillic: Well, then we seem to disagree.

mforets: True, but we agree that the option should be added. Should we merge this PR?

schillic: > True, but we agree that the option should be added.

No, I do not see a reason to add the option.

PR #1963, opened 2020-02-14
Update Project.toml

mforets:


mforets:
I restarted the CI build because the julia v1.1 was broken with a segfault.

Issue #996, opened 2019-01-13
Add common polytopic methods to lazy operations


schillic:
For instance we have `vertices_list`, `constraints_list`, and `linear_map` for concrete-type polytopes but often not for lazy operations. The idea is to add these methods and just fall back to the definitions on the wrapped sets. See [this comment](https://github.com/JuliaReach/LazySets.jl/issues/941#issuecomment-448195948) for a list of `vertices_list` implementations. Sometimes the implementation is not trivial, e.g., see #675 for how to implement it for `MinkowskiSum`.

mforets:
Would this definition make sense? `ispolytopic(X) = applicable(vertices_list, X) && isbounded(X)`

Then we can write generic fallbacks, eg. 

```julia
function convert(VPolygon, X)
    if ispolytopic(X) && dim(X) == 2
        return VPolygon(vertices_list(X))
    else
         throw(ArgumentError(...))
    end
end
```

schillic:
> Would this definition make sense?

Yes, but that was not the point of this issue and is rather something that is enabled by it.

mforets:
True, thanks. We have a couple of related issues about fallbacks; perhaps my comment was how i would like to implement https://github.com/JuliaReach/LazySets.jl/issues/1387.

mforets:
For linear map, we could add the implementation that takes the constraints list of the cartesian product (https://github.com/JuliaReach/LazySets.jl/pull/2194/files#diff-ace1e54fabcf477ffe72b37d1fb1baeeR277)

PR #1965, opened 2020-02-14
#1113 - Add AbstractAffineMap interface

### LINKED ISSUE #1113: Add AbstractAffineMap interface

schillic:
Closes #1113.

I was not sure about `ExponentialProjectionMap`, ~so I left that one out.~ but now it is part as well.

The third commit is unrelated. I found it when printing `subtypes(AbstractAffineMap)` did not produce `ExponentialMap` but `ExponentialMap{N, S} where ...`.


mforets:
`get_A` and `get_b` are arguably arbitrary name choices. Let's use `matrix(..)` and `vector(..)`? Or `matrix` and `affine_term`.

mforets:
> I was not sure about ExponentialProjectionMap, so I left that one out.

i think this also qualitifes as `<:AbstractAffineMap`, since

> Type that represents the projection of a sparse matrix exponential, i.e.,
``L⋅\\exp(M)⋅R`` for a given sparse matrix ``M``.

Hence the matrix is `L⋅\\exp(M)` and the affine term is zero.

schillic:
> `get_A` and `get_b` are arguably arbitrary name choices. Let's use `matrix(..)` and `vector(..)`? Or `matrix` and `affine_term`. For the getter function of the wrapped set we can use `set(..)`.

Done.

> > I was not sure about ExponentialProjectionMap, so I left that one out.
> 
> I think this also qualitifes as `<:AbstractAffineMap`, since
> 
> > Type that represents the projection of a sparse matrix exponential, i.e.,
> > `L⋅\\exp(M)⋅R` for a given sparse matrix `M`.
> 
> Hence the matrix is `L⋅\\exp(M)⋅R` and the affine term is zero.

Thanks, done.

src/LazyOperations/ResetMap.jl
```diff
@@ -314,7 +303,7 @@ function constraints_list(rm::ResetMap{N}) where {N<:Real}
         return res
     end
 
-    constraints = copy(constraints_list(LinearMap(get_A(rm), rm.X)))
+    constraints = copy(constraints_list(LinearMap(matrix(rm), rm.X)))
```
mforets:
```suggestion
    constraints = copy(constraints_list(LinearMap(matrix(rm), set(rm))))
```

src/LazyOperations/ResetMap.jl
```diff
@@ -142,18 +144,22 @@ reset map.
...
-function _get_b_from_dictionary(dict::Dict{Int, N}, n::Int) where {N<:Real}
+function _vector_from_dictionary(dict::Dict{Int, N}, n::Int) where {N<:Real}
     b = sparsevec(Int[], N[], n)
     for (i, val) in dict
         b[i] = val
     end
     return b
 end
 
+function set(rm::ResetMap)
```
mforets: should we `@inline` this?

src/Interfaces/AbstractAffineMap.jl
```diff
@@ -0,0 +1,303 @@
...
+the affine map, i.e. `x ↦ M*x + v` for each vertex `x` of `X`. By default, the
+convex hull operation is taken before returning this list. For dimensions three
+or higher, this operation relies on the functionality through the concrete
+polyhedra library `Polyhedra.jl`.
+
+If you are not interested in taking the convex hull of the resulting vertices
+under the affine map, pass `apply_convex_hull=false` as a keyword argument.
+
+Note that we assume that the underlying set `X` is polyhedral, either concretely
+or lazily, i.e. there the function `vertices_list` should be applicable.
```
mforets: can we check that `vertices_list` is applicable and send an error message otherwise? or define and use `ispolytopic(X)` from [this comment](https://github.com/JuliaReach/LazySets.jl/issues/996#issuecomment-586589969)

src/Interfaces/AbstractAffineMap.jl
```diff
@@ -0,0 +1,303 @@
...
+### Notes
+
+We assume that the underlying set `X` is polyhedral, i.e., offers a method
+`constraints_list(X)`.
+
+### Algorithm
+
+Falls back to the list of constraints of the translation of a lazy linear map.
+"""
+function constraints_list(am::AbstractAffineMap{N}) where {N<:Real}
```
mforets: Similar comment to the `vertices_list`.

src/LazyOperations/ResetMap.jl
schillic: `rm.X` is old code that I did not touch. I will change it now. But while it is better to never do that, I think it is fine to use the field access instead of the getter function in the main file that defines the type. We do that all over this library.

src/LazyOperations/ResetMap.jl
schillic: AFAIK Julia does this automatically for such obvious getter functions.

src/Interfaces/AbstractAffineMap.jl
schillic: This function just calls `_constraints_list_translation` which already has such a check.

src/Interfaces/AbstractAffineMap.jl
```diff
@@ -0,0 +1,303 @@
...
+### Input
+
+- `am` -- affine map
+
+### Output
+
+`true` iff the wrapped set is empty and the affine vector is empty.
+"""
+function isempty(am::AbstractAffineMap)
+    return isempty(set(am))
```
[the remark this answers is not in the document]
schillic: I just copied this from `AffineMap.jl`. I do not even understand that comment. I will remove "and the affine vector is empty."

src/LazyOperations/ResetMap.jl
mforets: > I think it is fine to use the field access instead of the getter function in the main file that defines the type. 

Yes, certainly. I commented because I expected to see `set(..)` as there is `matrix(..)` in the same line.

src/Interfaces/AbstractAffineMap.jl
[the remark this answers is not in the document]
mforets: you're right; taking the minkowski sum with an empty set doesn't add anything. 

```julia
julia> ∅(2) + rand(2)
Translation{Float64,Array{Float64,1},EmptySet{Float64}}(EmptySet{Float64}(2), [0.1618462193818917, 0.5400420114536386])

julia> isempty(ans)
true
```

PR #1967, opened 2020-02-15
Lazy projection constructor

mforets:
See https://github.com/JuliaReach/LazySets.jl/issues/1331.

Requires ~https://github.com/JuliaReach/LazySets.jl/pull/1959.~

Remains:

- [x] Add test.
- [x] Add new function to the documentation.


src/LazyOperations/LinearMap.jl
```diff
@@ -451,3 +451,9 @@ The polytope representing the linear map of the lazy linear map of a set.
 function linear_map(M::AbstractMatrix{N}, lm::LinearMap{N}) where {N<:Real}
      return linear_map(M * lm.M, lm.X)
 end
+
+# lazy projection
+function Projection(X::LazySet{N}, vars::AbstractVector{Int}) where {N<:Real}
+    M = projection_matrix(N, dim(X), vars)
```
schillic: Can we use `variables` or `dimensions`?

src/Approximations/decompositions.jl
```diff
@@ -321,7 +314,7 @@ A lazy `LinearMap` representing the projection of the set `S` to block `block`.
                          set_type::Type{<:LinearMap},
                          n::Int=dim(S)
                         ) where {N<:Real}
-    M = _projection_matrix(N, n, block)
+    M = projection_matrix(block, n, N)
     return M * S
```
schillic: Here you could immediately do `return Projection(S, block)`.

src/Arrays/matrix_operations.jl
```diff
@@ -240,3 +241,38 @@ function extend(M::AbstractMatrix; check_rank=true)
...
+
+### Output
+
+
+A sparse matrix that corresponds to the projection onto the variables in `block`.
+
+### Examples
+
+```jldoctest projection_matrix
+julia> projection_matrix([1, 3], 4)
```
mforets: Apparently we need `julia> LazySets.Arrays.projection_matrix([1, 3], 4)` here, or load `LazySets.Arrays` in these docs. @schillic  do you suggest something in particular? Otherwise i would just do `LazySets.Arrays.projection_matrix([1, 3], 4)

schillic: I would do it as in a similar situation:

https://github.com/JuliaReach/LazySets.jl/blob/ef38a632e370dd745aa31dd32abd4bcc544d6c97/src/Arrays/vector_operations.jl#L79-L83

So:

````julia
```jldoctest projection_matrix
julia> using LazySets: projection_matrix

julia> projection_matrix([1, 3], 4)

[...]
````

PR #1972, opened 2020-02-16
Bump compat for RecipesBase to v0.8

schillic:
This is #1955 with the required adaptation of the docs.


mforets:
We should probably tag a new patch release now. (cf. https://stackoverflow.com/questions/60150101/cannot-install-differentialequations-package-to-julia)

Issue #1914, opened 2020-01-15
Clarification on the behaviour of isbounded for "unbounded" HPolygon and HPolytope 


ueliwechsler:
I stumbled over this behaviour for an unbounded polyhedral set
```julia
X1 = HPolyhedron([0. 1.], [2.]); isbounded(X1) # false
X2 = HPolygon([0. 1.], [2.]); isbounded(X2) # true
X3 = HPolytope([0. 1.], [2.]); isbounded(X3) # true
```
From the documentation for `HPolygon` we have that:
>  Another type assumption is that the polygon is bounded. The option check_boundedness can  
  be used to assert this. This option is deactivated by default 
... 
>  The function isbounded can be used to manually assert boundedness.

Shouldn't the method `isbounded` return false for the `HPolygon` and `HPolytope`?

schillic:
This is a shortcoming in the documentation. It should say:

> The function `isbounded(P, false)` can be used to manually assert boundedness.

The `false` here ignores the type assumption and actually performs a check.

ueliwechsler:
Thank you! But then, we should also use `isbounded(P, false)` instead of `isbounded(P)` in  #576,  right?

schillic:
No, we do not have to guarantee correct behavior if you use a bounded type to represent an unbounded set. This is a user error.

mforets:
> Shouldn't the method isbounded return false for the HPolygon and HPolytope?

There is always this tension between using only the type information or the value for a `isxyz` function. And I agree that it's unexpected to get `true` in the cases above; after all, someone can argue what is the purpose of `isbounded` if it doesn't check the actual it receives, i.e., its value? There is a kwarg for the "expected" behavior, but still.. A reasonable compromise that I like, and we started to adopt for some newer functions, is to define two functions, `isboundedtype` (uses the value), and `isbounded` (doesn't use the value).

ueliwechsler:
> This is a user error.

I agree, but we could help the user not make this mistake of using a bounded type for an unbounded set. The problem is that the error messages tend to be quite ugly and hard to figure out for these cases since they are not handled as exceptions.

What about warning the user or throwing an error if he tries to create "a bounded unbounded set" such as `HPolytope([0;1.],2.)`? 


> `isbounded` (uses the value), and `isboundedtype` (only uses type information if it is available).

Nice, this would make it more intuitive for me.

schillic:
[...]

PR #1979, opened 2020-02-17
#1926 - Add isboundedtype

### LINKED ISSUE #1926: Add isboundedtype
See [this comment](https://github.com/JuliaReach/LazySets.jl/issues/1914#issuecomment-574742790).

schillic:
Closes #1926.

This version returns `false` for lazy set types. A more sophisticated approach (to be added later) would require to define a function `isboundedtype(::Type{<:LazySet})` based on set types that can then be applied to type parameters like so:

```julia
function isboundedtype(cp::CartesianProduct{N, S1, S2}) where {N, S1, S2}
    return isboundedtype(S1) && isboundedtype(S2)
end
```

### DIFF
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -268,7 +268,7 @@ Determine whether a set is bounded only based on type information.
 
 `true` if the set is bounded based on type information.
 Note that some sets may still represent an unbounded set even though their type
-actually (example: [`HPolytope`](@ref)).
+actually does not (example: [`HPolytope`](@ref)).
 
 ### Notes
 
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -256,6 +256,30 @@ Alias for the support vector σ.
 const support_vector = σ
 
 """
+    isboundedtype(::Type{<:LazySet})
+
+Determine whether a set type only represents bounded sets.
+
+### Input
+
+- `LazySet` -- set type for dispatch
+
+### Output
+
+`true` if the set type only represents bounded sets.
+Note that some sets may still represent an unbounded set even though their type
+actually does not (example: [`HPolytope`](@ref)).
+
+### Notes
+
+By default this function returns `false`.
+All set types that can determine boundedness should override this behavior.
+"""
+function isboundedtype(::Type{<:LazySet})
+    return false
+end
+
+"""
     isboundedtype(S::LazySet)
 
 Determine whether a set is bounded only based on type information.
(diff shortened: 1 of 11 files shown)


src/Interfaces/LazySet.jl
```diff
@@ -251,6 +252,30 @@ Alias for the support vector σ.
...
+`true` if the set is bounded based on type information.
+Note that some sets may still represent an unbounded set even though their type
+actually (example: [`HPolytope`](@ref)).
+
+### Notes
+
+By default this function returns `false`.
+All set types that can determine boundedness should override this behavior.
+"""
+function isboundedtype(S::LazySet)
```
mforets: i would expect that `isboundedype` acts on types. we could, *maybe*, allow it to receive instances since there is (almost) no room for confusion, but in the first place, it should work types.

mforets: (correspondingly, `isbounded` should work on instances in the first place..)

schillic: This is not what you proposed in the issue, which links [here](https://github.com/JuliaReach/LazySets.jl/issues/1914#issuecomment-574742790). The motivation was that the result of `P = HPolytope()` depends on either the type (a polytope *is* bounded) or on the instance (a set with no constraints *is not* bounded). This is what `isbounded(P)` and `isboundedtype(P)` was supposed to differentiate.

src/Interfaces/LazySet.jl
```diff
@@ -251,6 +252,30 @@ Alias for the support vector σ.
 """
 const support_vector = σ
 
+"""
+    isboundedtype(S::LazySet)
```
mforets, line 256:
```suggestion
    isboundedtype(::LazySet)
```

src/Interfaces/LazySet.jl
```diff
@@ -251,6 +252,30 @@ Alias for the support vector σ.
...
+
+### Input
+
+- `S` -- set
+
+### Output
+
+`true` if the set is bounded based on type information.
+Note that some sets may still represent an unbounded set even though their type
+actually (example: [`HPolytope`](@ref)).
```
mforets, line 268:
```suggestion
actually not (example: [`HPolytope`](@ref)).
```

mforets:
> can we have isboundedtype(::Type{<:LazySet}) in this PR as well?

note that defining `isboundedtype` on types on the first place lets one define `isboundedtype` just once by resorting to the type of the argument.

src/Interfaces/LazySet.jl
schillic: `S` is referred to in the docstring.

schillic: With the new implementation this is actually needed.

PR #1981, opened 2020-02-17
#1911 - Type instability in genmat_fallback

### LINKED ISSUE #1911: Type instability in genmat_fallback
```julia
x = rand(Interval)

@code_warntype genmat(x)

Variables
  #self#::Core.Compiler.Const(LazySets.genmat, false)
  H::Interval{Float64,IntervalArithmetic.Interval{Float64}}

Body::Any
1 ─ %1 = LazySets.genmat_fallback(H)::Any
└──      return %1
```

schillic:
Closes #1911.

```julia
# master
julia> @btime genmat(rand(Interval))
  423.462 ns (7 allocations: 464 bytes)

julia> @btime genmat(rand(BallInf, dim=100))
  48.197 μs (315 allocations: 98.44 KiB)

# this branch
julia> @btime genmat(rand(Interval))
  175.353 ns (4 allocations: 272 bytes)

julia> @btime genmat(rand(BallInf, dim=100));
  21.260 μs (7 allocations: 80.02 KiB)
```

```julia
julia> @code_warntype genmat(rand(Interval))
Variables
  #self#::Core.Compiler.Const(LazySets.genmat, false)
  H::Interval{Float64,IntervalArithmetic.Interval{Float64}}

Body::Array{Float64,2}
1 ─ %1 = LazySets.genmat_fallback(H)::Array{Float64,2}
└──      return %1
```

### DIFF
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -69,7 +69,11 @@ function genmat_fallback(Z::AbstractZonotope{N}) where {N<:Real}
     if isempty(gens)
         return Matrix{N}(undef, dim(Z), 0)
     end
-    return hcat(gens...)
+    G = Matrix{N}(undef, dim(Z), length(gens))
+    @inbounds for (i, g) in enumerate(gens)
+        G[:, i] = g
+    end
+    return G
 end
 
 # iterator that wraps the generator matrix
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -58,18 +58,47 @@ Fallback definition of `genmat` for zonotopic sets.
 
 ### Input
 
-- `Z` -- zonotopic set
+- `Z`     -- zonotopic set
+- `gens`  -- (optional; default: `generators(Z)`) iterator over generators
+- `ngens` -- (optional; default: `nothing`) number of generators or `nothing` if
+             unknown
 
 ### Output
 
 A matrix where each column represents one generator of `Z`.
+
+### Notes
+
+Passing the number of generators is much more efficient as otherwise the
+generators have to be obtained from the iterator (`gens`) and stored in an
+intermediate vector until the final result matrix can be allocated.
 """
-function genmat_fallback(Z::AbstractZonotope{N}) where {N<:Real}
-    gens = generators(Z)
+function genmat_fallback(Z::AbstractZonotope{N};
+                         gens=generators(Z),
+                         ngens=nothing) where {N<:Real}
     if isempty(gens)
         return Matrix{N}(undef, dim(Z), 0)
+    elseif ngens == nothing
+        return _genmat_fallback_generic(Z, gens)
+    else
+        return _genmat_fallback_ngens(Z, gens, ngens)
     end
-    G = Matrix{N}(undef, dim(Z), length(gens))
+end
+
+function _genmat_fallback_generic(Z::AbstractZonotope{N}, gens) where {N<:Real}
+    Gv = Vector{Vector{N}}()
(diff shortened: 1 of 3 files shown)


src/Interfaces/AbstractZonotope.jl
```diff
@@ -69,7 +69,11 @@ function genmat_fallback(Z::AbstractZonotope{N}) where {N<:Real}
     if isempty(gens)
         return Matrix{N}(undef, dim(Z), 0)
     end
-    return hcat(gens...)
+    G = Matrix{N}(undef, dim(Z), length(gens))
+    @inbounds for (i, g) in enumerate(gens)
```
schillic: I realized only now that `length(gens)` is not good because that creates the generators just for computing the length and then creates them another time in the loop. I have to look into `push!` for matrices.

mforets: > that creates the generators just for computing the length

Why? Isn't `Base.length(it::FallbackGeneratorIterator) = it.n_plus_one - 1` what is used in the generic fallback?

schillic: True, but there can be other implementations apart from `FallbackGeneratorIterator` that dynamically compute the generators. The `HyperrectangleGeneratorIterator` goes a hybrid way by precomputing the `length` and storing it, but I am not sure that this is smart. We might change this in the future.

https://github.com/JuliaReach/LazySets.jl/blob/c61beeacc7ad7eae8634a03f8e2fb7aef65bb8c9/src/Interfaces/AbstractHyperrectangle.jl#L64-L98

mforets: Ok, i see. Then you could be looking to `push!` to a vector and then [reshape](https://docs.julialang.org/en/v1/base/arrays/#Base.reshape) it.

schillic: After benchmarking the generic implementation, I decided to have both versions. The generic fallback with an additional `Vector` is used by default but currently all our types can efficiently determine the number of generators and so I let them call the more efficient version.
```julia
julia> @btime genmat(rand(Interval));  # knowing the generators
  179.210 ns (4 allocations: 272 bytes)

julia> @btime genmat(rand(Interval));  # not knowing the generators
  309.431 ns (7 allocations: 496 bytes)

julia> @btime genmat(rand(BallInf, dim=100));  # knowing the generators
  25.119 μs (7 allocations: 80.02 KiB)

julia> @btime genmat(rand(BallInf, dim=100));  # not knowing the generators
  26.991 μs (114 allocations: 169.72 KiB)
```
For `Interval`s I think we can do even better by implementing a custom `genmat` method (but not here).

src/Sets/LineSegment.jl
```diff
@@ -210,7 +210,8 @@ Return the generator matrix of a line segment.
 A matrix with a single column representing the generator of `L`.
 """
 function genmat(L::LineSegment)
-    return genmat_fallback(L)
+    ngens = L.p == L.q ? 0 : 1
```
mforets:
```suggestion
    ngens = _isapprox(L.p, L.q) ? 0 : 1
```

schillic: The main code checks equality and I just copy the behavior here. I would prefer to add functional changes later.

mforets: > The main code 

Where is it?

> I would prefer to add functional changes later.

This function introduces a bug, unless we restrict the numeric type of the line segment to an exact ring such as rationals. I agree to defer fixing bugs for later, but do you mind opening a follow-up issue? (if it already exists i didn't notice)

schillic: #2001

PR #1982, opened 2020-02-17
#1779 - Support function for empty sets

### LINKED ISSUE #1779: isbounded of empty polyhedron crashes
```julia
julia> isbounded(HPolyhedron([HalfSpace([1.], 0.), HalfSpace([-1.], -1.)]))
ERROR: the support vector is undefined because the polyhedron is empty
Stacktrace:
 [2] σ_helper(::LazySets.Arrays.SingleEntryVector{Float64}, ::HPolyhedron{Float64}, ::GLPKMathProgInterface.GLPKInterfaceLP.GLPKSolverLP) at .julia/dev/LazySets/src/Sets/HPolyhedron.jl:183
 [3] #ρ#86(::GLPKMathProgInterface.GLPKInterfaceLP.GLPKSolverLP, ::typeof(ρ), ::LazySets.Arrays.SingleEntryVector{Float64}, ::HPolyhedron{Float64}) at .julia/dev/LazySets/src/Sets/HPolyhedron.jl:105
 [4] ρ at .julia/dev/LazySets/src/Sets/HPolyhedron.jl:105 [inlined]
 [5] isbounded_unit_dimensions(::HPolyhedron{Float64}) at .julia/dev/LazySets/src/Interfaces/LazySet.jl:210
 [6] isbounded(::HPolyhedron{Float64}) at .julia/dev/LazySets/src/Sets/HPolyhedron.jl:215
 [7] top-level scope at REPL[66]:1
```

schillic:
Closes #1779.

The first commit refactors the code a bit without functional changes (other than printing a better error message in the support function).

The second commit is a proposal to define the support function for empty sets as `-Inf`.

### DIFF
--- a/src/Sets/EmptySet.jl
+++ b/src/Sets/EmptySet.jl
@@ -79,10 +79,10 @@ Evaluate the support function of an empty set in a given direction.
 
 ### Output
 
-An error.
+`-Inf`.
 """
 function ρ(d::AbstractVector{N}, ∅::EmptySet{N}) where {N<:Real}
-    error("the support function of an empty set does not exist")
+    return N(-Inf)
 end
 
 """
--- a/src/Sets/HPolyhedron.jl
+++ b/src/Sets/HPolyhedron.jl
@@ -98,6 +98,7 @@ direction.
 ### Output
 
 The support function of the polyhedron.
+If the polyhedron is empty, the result is `-Inf`.
 If a polytope is unbounded in the given direction, we throw an error.
 If a polyhedron is unbounded in the given direction, the result is `Inf`.
 """
@@ -105,8 +106,7 @@ function ρ(d::AbstractVector{N}, P::HPoly{N};
            solver=default_lp_solver(N)) where {N<:Real}
     lp, unbounded, empty = σ_helper(d, P, solver)
     if empty
-        error("the support function in direction $(d) is undefined because " *
-              "the polytope is empty")
+        return N(-Inf)
     elseif unbounded
         if P isa HPolytope
             error("the support function in direction $(d) is undefined " *
--- a/src/Sets/HPolyhedron.jl
+++ b/src/Sets/HPolyhedron.jl
@@ -103,8 +103,11 @@ If a polyhedron is unbounded in the given direction, the result is `Inf`.
 """
 function ρ(d::AbstractVector{N}, P::HPoly{N};
            solver=default_lp_solver(N)) where {N<:Real}
-    lp, unbounded = σ_helper(d, P, solver)
-    if unbounded
+    lp, unbounded, empty = σ_helper(d, P, solver)
+    if empty
+        error("the support function in direction $(d) is undefined because " *
+              "the polytope is empty")
+    elseif unbounded
         if P isa HPolytope
             error("the support function in direction $(d) is undefined " *
                   "because the polytope is unbounded")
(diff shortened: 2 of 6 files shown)


mforets:
> The second commit is a proposal to define the support function for empty sets as -Inf.

It is not clear to me why would getting infinities be better than a nice error message. Shouldn't we at least print a warning?

schillic:
> It is not clear to me why would getting infinities be better than a nice error message. Shouldn't we at least print a warning?

The motivation was that many functions rely on `support_function` and then just crash (see the example in the issue this PR tries to close). Making the support function work gives a better experience (e.g., you should be able to call functions on sets such as empty `HPolytope`s without getting an error). The only alternative I saw was to add additional checks to each function that can potentially call the support function of empty sets, which I found too complicated.

mforets:
I agree if we add a warning (through `Base.@warn`) that it's computing over an empty set. Otherwise, i think that returning an arbitrary real is worse than a crash (which would potentially help to uncover an issue or a bug).

mforets:
>  (see the example in the issue this PR tries to close).

i think the behavior in that example is fine: the polyhedron *is* empty and the message says precisely that; i think that is better than `true` or `false` since none of them applies. mathematically, the supp func is defined for non-empty sets (https://en.wikipedia.org/wiki/Support_function)

schillic:
>  it could otherwise return `true` as the empty set is bounded

It *should* return `true`. Why should it throw an error? That was the whole point of the issue.

schillic:
> mathematically, the supp func is defined for non-empty sets (https://en.wikipedia.org/wiki/Support_function)

I know, but it is also defined for closed sets and yet we return `Inf` for unbounded directions. The reason was the same: we wanted to avoid a crash at the support-function level and handle this case in the functions that called the support function instead.

mforets:
> Why should it throw an error?

i agree that it should return true. however, i think it is better to raise warnings when computing support functions over empty sets. in real applications, i think that is most helpful than getting infinities.

schillic:
> better to raise warnings when computing support functions over empty sets

Then for consistency we should also raise a warning when computing the support function over unbounded sets.

mforets:
well that's a different animal.. an unbounded set can have a finite support function over a given direction.. but yes, i remember that we have (had?) such error messages for support functions in the unbounded case.

schillic:
> well that's a different animal.. an unbounded set can have a finite support function over a given direction.

Yes, I meant "in a direction where the set is unbounded."

> but yes, i remember that we have (had?) such error messages for support functions in the unbounded case.

Exactly. To me this has the same taste. I am okay with adding a warning in those cases instead. But I think these cases should be treated in the same way.
But just to say it again, this will lead to additional warnings in places where you would not expect them (for example in `isbounded` where we query the support function as a sub-procedure).

src/Sets/EmptySet.jl
```diff
@@ -79,10 +79,10 @@ Evaluate the support function of an empty set in a given direction.
 
 ### Output
 
-An error.
+`-Inf`.
 """
 function ρ(d::AbstractVector{N}, ∅::EmptySet{N}) where {N<:Real}
-    error("the support function of an empty set does not exist")
+    return N(-Inf)
```
mforets: As discussed in the thread, i suggest to add a warning here.

src/Sets/HPolyhedron.jl
```diff
@@ -98,13 +98,16 @@ direction.
...
 If a polytope is unbounded in the given direction, we throw an error.
 If a polyhedron is unbounded in the given direction, the result is `Inf`.
 """
 function ρ(d::AbstractVector{N}, P::HPoly{N};
            solver=default_lp_solver(N)) where {N<:Real}
-    lp, unbounded = σ_helper(d, P, solver)
-    if unbounded
+    lp, unbounded, empty = σ_helper(d, P, solver)
+    if empty
+        return N(-Inf)
```
mforets: Also here..?

src/Sets/EmptySet.jl
```diff
@@ -79,10 +79,10 @@ Evaluate the support function of an empty set in a given direction.
 
 ### Output
 
-An error.
+`-Inf`.
```
mforets, line 82: obs: \rho([-1.0], x <= 1) is -Inf is *correct*

PR #1986, opened 2020-02-18
Use faster 2d directions algorithm by default

mforets:
The version of directions comparison using quadrants is both faster and more general than the version using `jump2pi`. For these reasons, we could remove the implementation altogether.

```julia
julia> p = rand(HPolygon);

julia> d = rand(2);

julia> @btime σ($d, $p); # master
  737.704 ns (12 allocations: 688 bytes)

julia> @btime σ($d, $p); # this branch
  632.355 ns (12 allocations: 688 bytes)
```

### DIFF
--- a/src/Interfaces/AbstractPolygon.jl
+++ b/src/Interfaces/AbstractPolygon.jl
@@ -140,6 +140,10 @@ directions using the right-hand rule (see [`is_right_turn`](@ref)).
 In particular, this method does not use the arctangent.
 """
 function <=(u::AbstractVector{N}, v::AbstractVector{N}) where {N<:Real}
+    return _leq_quadrant(u, v)
+end
+
+function _leq_quadrant(u::AbstractVector{N}, v::AbstractVector{N}) where {N<:Real}
     qu, qv = quadrant(u), quadrant(v)
     if qu == qv
         # same quadrant, check right-turn with center 0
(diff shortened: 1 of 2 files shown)


schillic:
Did you try polygons with dozens of constraints? I think a `rand` polygon has only a few constraints.

And if we remove the implementation, we should also remove `jump2pi` because this is the only place where it was used.

mforets:
> Did you try polygons with dozens of constraints?

I knew since long ago that tested individually, the version with quadrant is faster, so I had planned to add the algorithms as an option (as in https://github.com/JuliaReach/LazySets.jl/pull/864). But now I don't see the point of keeping both algorithms given this situation (less general + slower). I'm also fine to keep the trig algorithm as a separate unexported function.

```julia
julia> p = rand(VPolygon, num_vertices=1000) |> tohrep;

julia> length(p.constraints)
999

julia> @btime σ($dir, $p) # master
  1.137 μs (12 allocations: 688 bytes)
2-element Array{Float64,1}:
 5.059188018707831
 5.959855447966407

julia> @btime σ($dir, $p) # this branch
  796.222 ns (12 allocations: 688 bytes)
2-element Array{Float64,1}:
 5.059188018707831
 5.959855447966407
```
>  we should also remove jump2pi 

Yes.

schillic:
I am a bit confused that you first used `HPolygon` and then used `VPolygon`. Do they both use this function?

I am fine with removing something that has no use case :+1:

mforets:
>I am a bit confused that you first used HPolygon and then used VPolygon.

I forgot that there is also a kwarg for HPolygons, i could have used `rand(HPolygon, num_constraints=100)`.

>  Do they both use this function?

Only `HPolygon` uses it, in [this line](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Sets/HPolygon.jl#L147).

schillic:
Oh sorry, I overlooked the `tohrep`. Never mind then.

mforets:
I decided to refactor the code instead of removing the function.

src/Interfaces/AbstractPolygon.jl
```diff
@@ -140,6 +140,10 @@ directions using the right-hand rule (see [`is_right_turn`](@ref)).
 In particular, this method does not use the arctangent.
 """
 function <=(u::AbstractVector{N}, v::AbstractVector{N}) where {N<:Real}
+    return _leq_quadrant(u, v)
```
schillic: I think we should `assert` that the vectors have length 2 here (I know we did not do this before). `<=` may easily be used in wrong places.

src/Interfaces/AbstractPolygon.jl
```diff
@@ -140,6 +140,14 @@ directions using the right-hand rule (see [`is_right_turn`](@ref)).
 In particular, this method does not use the arctangent.
 """
 function <=(u::AbstractVector{N}, v::AbstractVector{N}) where {N<:Real}
+    @assert length(u) == length(v) == 2 "comparison of vectors `u` and `v` " *
+           "by their direction requires they are are of length 2, " *
```
schillic:
```suggestion
           "by their direction requires they are of length 2, " *
```

PR #1998, opened 2020-02-21
Run CompatHelper only once a day

schillic:


schillic:
For some reason `CompatHelper` does not seem to accept custom settings.

PR #2005, opened 2020-02-26
#2002 - VPolytope constructor from matrix columns

### LINKED ISSUE #2002: VPolytope constructor from matrix columns
Extend https://github.com/JuliaReach/LazySets.jl/issues/1987 to higher dimensions, ie. such that the following works: 

```julia
julia> using StaticArrays; VPolytope(@SMatrix(rand(4, 10)))
```

mforets:
Closes #2002.

### DIFF
--- a/test/unit_Polytope.jl
+++ b/test/unit_Polytope.jl
@@ -246,6 +246,10 @@ for N in [Float64]
     q = VPolytope([N[0, 1], N[0, 2]])
     @test N[0, 1//2] ∉ q
 
+    # creation from a matrix (each column is a vertex)
+    pmat = VPolytope(copy(N[0 0; 1 0; 0 1]'))
+    @test p == pmat
+
     # inclusion (see #1809)
     X = BallInf(N[0.1, 0.2, 0.1], N(0.3))
     Y = convert(HPolytope, X)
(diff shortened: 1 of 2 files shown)


test/unit_Polytope.jl
```diff
@@ -246,6 +246,10 @@ for N in [Float64]
     q = VPolytope([N[0, 1], N[0, 2]])
     @test N[0, 1//2] ∉ q
 
+    # creation from a matrix (each column is a vertex)
+    pmat = VPolytope(copy(N[0 0; 1 0; 0 1]'))
```
schillic: Why the `copy` and the transpose?

mforets: Each vertex is a column (hence the transpose); the `copy` is to materialize the lazy transpose, this is standard.

schillic: Okay, I thought that you could as well write the final matrix directly, but if it helps in reading... (for me it does not :wink:)

mforets: This is what comes natural to me, i don't know .. :) I have to think about where to put the commas, if i have to write the final matrix directly.

mforets: You can add a new example in the docstrings if you'd prefer, examples never hurt..

PR #2007, opened 2020-02-26
#2003 - Make isposdef check optional in Ellipsoid constructor

### LINKED ISSUE #2003: Make isposdef check optional in Ellipsoid constructor
Use: `Ellipsoid(c, M, check_positive_definite=false)`. This came up recently when using for `M` an `MArray` (`isposdef` is not defined yet for such array type).

mforets:
Closes #2003.

### DIFF
--- a/src/Sets/Ellipsoid.jl
+++ b/src/Sets/Ellipsoid.jl
@@ -24,21 +24,39 @@ of an ellipse.
 - `shape_matrix` -- real positive definite matrix, i.e. it is equal to its transpose
                     and ``x^\\mathrm{T}Qx > 0`` for all nonzero ``x``
 
+## Notes
+
+By default, the inner constructor checks that the given shape matrix is positive
+definite. Use the flag `check_posdef=false` to disable this check.
+
 ### Examples
 
-If the center is not specified, it is assumed that the center is the origin.
-For instance, a 3D ellipsoid with center at the origin and the shape matrix being
+An ellipsoid can be created passing its center and shape matrix (which should be
+positive definite).
+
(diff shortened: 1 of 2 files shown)


src/Sets/Ellipsoid.jl
```diff
@@ -56,39 +73,34 @@ julia> σ(ones(3), E)
...
 isoperationtype(::Type{<:Ellipsoid}) = false
 isconvextype(::Type{<:Ellipsoid}) = true
 
-# convenience constructor without type parameter
-Ellipsoid(c::AbstractVector{N}, Q::AbstractMatrix{N}) where {N<:AbstractFloat} =
-    Ellipsoid{N}(c, Q)
-
 # convenience constructor for ellipsoid centered in the origin
-Ellipsoid(Q::AbstractMatrix{N}) where {N<:AbstractFloat} =
-    Ellipsoid(zeros(N, size(Q, 1)), Q)
```
schillic: This change is unrelated and has been done in #1946 already. It is not clear to me whether we want to have that. Can you comment in #1946?

mforets: #1946 has other changes too; i didn't have time to consider them yet..

mforets: But you're right; i shouldn't have removed that constructor. It's just weird to have `Ellipsoid{N}` in line 82  and then again `new{N}(...)`  below.

schillic: > #1946 has other changes too

No, there are just some consequences. That is why I think we should discuss these changes before removing stuff. You can undo the changes here or we wait until we reached an agreement.

Issue #2010, opened 2020-02-26
Use point type of the VRep in conversion


mforets:
```julia
function VPolytope(P::VRep{N}) where {N}
    # TODO: use point type of the VRep
    vertices = Vector{Vector{N}}()
    for vi in Polyhedra.points(P)
        push!(vertices, vi)
    end
    return VPolytope(vertices)
end
```

Issue #2009, opened 2020-02-26
Remove constructor of VPolytope from VPolygon


mforets:
IIRC we've been removing creating of types from other types, users should do instead `convert(target_type, type)` (or `overapproximate`). The following still remains in `VPolytope.jl`:

```julia
# constructor from a polygon in V-representation
function VPolytope(P::VPolygon, share::Bool=false)
    v = vertices_list(P)
    if !share
        v = copy(v)
    end
    return VPolytope(v)
end
```
Should we remove it?

schillic:
> Should we remove it?

I agree, but the sharing is an important option.

PR #2012, opened 2020-02-26
#2004 - Generalize vertices type in VPolytope

### LINKED ISSUE #2004: Generalize vertices type in VPolytope
It should be:

```julia
struct VPolytope{N, VT<:AbstractVector{N}} <: AbstractPolytope{N}
    vertices::Vector{VT}
end
```

mforets:
Closes #2004.

### DIFF
--- a/src/Sets/VPolytope.jl
+++ b/src/Sets/VPolytope.jl
@@ -12,7 +12,7 @@ export VPolytope,
        tovrep
 
 """
-    VPolytope{N<:Real} <: AbstractPolytope{N}
+    VPolytope{N<:Real, VN<:AbstractVector{N}} <: AbstractPolytope{N}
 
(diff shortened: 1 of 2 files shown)


mforets:
tests pass locally. i opened some related-follow up PRs (#2009, #2010, #2011).

mforets:
Bump :trumpet:

schillic:
I already commented in #2004 that the container should also be parameterized.

mforets:
oh boy, i think we are good like this. generalizing for generalizing is not good.

mforets:
Thanks for the review!

> I will add the type parameter for the container in a follow-up PR.

As i said there, i think that the outer vector is the correct level of abstraction. For a specific case, a new set type may be better. Or we can reconsider generalizing the outer type, but for now, i think we can leave this as is..

PR #2018, opened 2020-02-28
Make bounds check in hyperrectangle constructor optional

### LINKED ISSUE #2015: Make bounds check in hyperrectangle constructor optional
I think it should be "undefined" (can crash, use at your own risk, etc.). Compare to `IntervalArithmetic.interval` which checks that the interval is well formed, and `IntervalArithmetic.Interval` which doesn't peform the check.

```julia
julia> IntervalArithmetic.interval(2, 1)
ERROR: ArgumentError: `[2, 1]` is not a valid interval. Need `a ≤ b` to construct `interval(a, b)`.
Stacktrace:
 [1] interval(::Int64, ::Int64) at /home/mforets/.julia/packages/IntervalArithmetic/lyDPa/src/intervals/intervals.jl:111
 [2] top-level scope at REPL[5]:1

julia> IntervalArithmetic.Interval(2, 1)
[2, 1]

julia> diam(ans)
-1.0
```
This option would be more for debugging purposes, or speedup the creation of a very large number of rectangles (which you know are well defined).

----

Ref: [this comment](https://github.com/sisl/NeuralVerification.jl/pull/88#issuecomment-591409318).

### LINKED ISSUE #2017: Bounds check in Hyperrectangle are done twice

mforets:
Closes #2015.
Closes #2017.

### DIFF
--- a/src/Sets/Hyperrectangle.jl
+++ b/src/Sets/Hyperrectangle.jl
@@ -75,7 +75,7 @@ Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([-1.0, 1.0], [2.0, 1.0
 By default, the constructor checks that that radius of the hyperrecatangle
 is nonnegative. To supress this check, use the `check_bounds` optional flag
 in the constructor. Note that if `check_bounds` is set to `false`, the behavior
-the set may be wrong/undefined.
+of a set with contradictory bounds is undefined.
 """
 struct Hyperrectangle{N<:Real, VNC<:AbstractVector{N}, VNR<:AbstractVector{N}
                      } <: AbstractHyperrectangle{N}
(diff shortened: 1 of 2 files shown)


src/Sets/Hyperrectangle.jl
```diff
@@ -71,18 +71,25 @@ julia> h = [1.0, 2.0];
 julia> Hyperrectangle(low=l, high=h)
 Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([-1.0, 1.0], [2.0, 1.0])
 ```
+
+By default, the constructor checks that that radius of the hyperrecatangle
```
schillic:
```suggestion
By default, the constructor checks that that radius of the hyperrectangle
```

src/Sets/Hyperrectangle.jl
```diff
@@ -71,18 +71,25 @@ julia> h = [1.0, 2.0];
...
+is nonnegative. To supress this check, use the `check_bounds` optional flag
+in the constructor. Note that if `check_bounds` is set to `false`, the behavior
+the set may be wrong/undefined.
```
schillic:
```suggestion
of a set with contradictory bounds is undefined.
```

src/Sets/Hyperrectangle.jl
mforets: > hyperrecatangle

sounds funny :laughing:  we should define such type

mforets:
Thanks for the review. As you've seen, i also removed the double check as it was completely unnecessary since it is always done in the inner constructor..

PR #2021, opened 2020-02-28
#2014 - Specialize symmetric interval hull for intervals

### LINKED ISSUE #2014: Specialize symmetric interval hull for intervals
In master:

```julia
julia> r = rand(Interval)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([-1.07068, -0.688518])

julia> @btime symmetric_interval_hull($r)
  123.561 ns (7 allocations: 544 bytes)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([0.0], [1.0706737823323762])
```

mforets:
Closes #2014.

### DIFF
--- a/src/Approximations/box_approximations.jl
+++ b/src/Approximations/box_approximations.jl
@@ -56,11 +56,25 @@ end
 box_approximation_symmetric(∅::EmptySet) = ∅
 
 """
-    symmetric_interval_hull
+    symmetric_interval_hull(X::LazySet)
 
-Alias for `box_approximation_symmetric`.
+Return a set that tightly overapproximates `X` with a hyperrectangular set symmetric
+respect to the origin.
 """
-symmetric_interval_hull = box_approximation_symmetric
+function symmetric_interval_hull(X::LazySet)
+    return box_approximation_symmetric(X)
+end
+
+# interval specialization
+function symmetric_interval_hull(x::Interval)
+    abs_inf = abs(min(x))
+    abs_sup = abs(max(x))
+    if abs_sup > abs_inf
+        return Interval(-abs_sup, abs_sup)
+    else
+        return Interval(-abs_inf, abs_inf)
+    end
+end
 
 """
     box_approximation_helper(S::LazySet{N}) where {N<:Real}
--- a/src/Approximations/box_approximations.jl
+++ b/src/Approximations/box_approximations.jl
@@ -69,11 +69,8 @@ end
 function symmetric_interval_hull(x::Interval)
     abs_inf = abs(min(x))
     abs_sup = abs(max(x))
-    if abs_sup > abs_inf
-        return Interval(-abs_sup, abs_sup)
-    else
-        return Interval(-abs_inf, abs_inf)
-    end
+    bound = max(abs_sup, abs_inf)
+    return Interval(-bound, bound)
 end
 
 """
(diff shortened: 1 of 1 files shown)


mforets:
```julia
julia> x = rand(Interval)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([-1.03243, -0.0673332])

@btime symmetric_interval_hull($x)
  0.016 ns (0 allocations: 0 bytes)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([-1.03243, 1.03243])

@btime box_approximation_symmetric($x) # falback (master)
  117.774 ns (6 allocations: 512 bytes)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([0.0], [1.0324268415018834])
```

src/Approximations/box_approximations.jl
```diff
@@ -56,11 +56,25 @@ end
 box_approximation_symmetric(∅::EmptySet) = ∅
 
 """
-    symmetric_interval_hull
+    symmetric_interval_hull(X::LazySet)
 
-Alias for `box_approximation_symmetric`.
+Return a set that tightly overapproximates `X` with a hyperrectangular set symmetric
+respect to the origin.
```
schillic:
```suggestion
with respect to the origin.
```

src/Approximations/box_approximations.jl
```diff
@@ -56,11 +56,25 @@ end
...
 """
-symmetric_interval_hull = box_approximation_symmetric
+function symmetric_interval_hull(X::LazySet)
+    return box_approximation_symmetric(X)
+end
```
schillic: Here `symmetric_interval_hull` was defined as an alias of `box_approximation_symmetric`. Hence you need to override the definition of `box_approximation_symmetric` instead. Otherwise that function will not make use of the new method.

src/Approximations/box_approximations.jl
```diff
@@ -56,11 +56,25 @@ end
...
+
+# interval specialization
+function symmetric_interval_hull(x::Interval)
+    abs_inf = abs(min(x))
+    abs_sup = abs(max(x))
+    if abs_sup > abs_inf
+        return Interval(-abs_sup, abs_sup)
+    else
+        return Interval(-abs_inf, abs_inf)
+    end
```
schillic: I prefer this simpler code:
```suggestion
    bound = max(abs_sup, abs_inf)
    return Interval(-bound, bound)
```

src/Approximations/box_approximations.jl
mforets: the point in this PR was to specialize `symmetric_interval_hull` for intervals. i didnt change `box_approximation_symmetric` in purpose: it stil remains a *box* (rectangle).

src/Approximations/box_approximations.jl
mforets: i like the suggestion :+1:

src/Approximations/box_approximations.jl
mforets: i should probably add a test with the new behavior..

schillic: I do not think that this is useful. They have just been aliases before and we have never cared for a difference.

mforets: :man_shrugging:  do you care if i now care for a difference?

schillic: Yes, this is a semantic change and I disagree.

mforets: Fair enough, this change is untelated to the purpose of this pr anyways

src/Approximations/box_approximations.jl
```diff
@@ -57,11 +57,18 @@ box_approximation_symmetric(∅::EmptySet) = ∅
 
 """
     symmetric_interval_hull
-
 Alias for `box_approximation_symmetric`.
```
schillic:
```suggestion

Alias for `box_approximation_symmetric`.
```

PR #2024, opened 2020-02-28
#1999 - Add vector parameter for Ball2

### LINKED ISSUE #1999: Add vector parameter for Ball2
See https://github.com/JuliaReach/LazySets.jl/issues/481.

mforets:
Closes #1999.

### DIFF
--- a/src/Sets/Ball2.jl
+++ b/src/Sets/Ball2.jl
@@ -6,7 +6,7 @@ export Ball2,
        volume
 
 """
-    Ball2{N<:AbstractFloat} <: AbstractCentrallySymmetric{N}
+    Ball2{N<:AbstractFloat, VN<:AbstractVector{N}} <: AbstractCentrallySymmetric{N}
 
 Type that represents a ball in the 2-norm.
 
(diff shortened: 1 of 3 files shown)


src/Sets/Ball2.jl
```diff
@@ -259,12 +262,12 @@ Random sampling with uniform distribution in `B` is computed using Muller's meth
...
 """
-function sample(B::Ball2{N}, nsamples::Int=1;
+function sample(B::Ball2{N, VN}, nsamples::Int=1;
                 rng::AbstractRNG=GLOBAL_RNG,
-                seed::Union{Int, Nothing}=nothing) where {N<:AbstractFloat}
+                seed::Union{Int, Nothing}=nothing) where {N<:AbstractFloat, VN<:AbstractVector{N}}
     require(:Distributions; fun_name="sample")
     n = dim(B)
-    D = Vector{Vector{N}}(undef, nsamples) # preallocate output
+    D = Vector{VN}(undef, nsamples) # preallocate output
```
schillic: You cannot know that the samples will have the same vector type as the set.

mforets: That's weird.. but possible. So you suggest to hardcode to vector?

mforets: i checked and https://github.com/JuliaReach/LazySets.jl/blob/master/src/Utils/samples.jl#L187 is only defined for cache D with Vector. this means that `D = Vector{VN}(undef, nsamples)` will not work if `VN` is different from vector. ill open a separate issue, and revert my change here.

mforets:
```suggestion
    D = Vector{Vector{N}}(undef, nsamples) # preallocate output
```

mforets:
Thx for the review!

PR #2025, opened 2020-02-28
#2022 - Faster symmetric interval hull for 2D hyperrectangles

### LINKED ISSUE #2022: Faster symmetric interval hull for 2D hyperrectangles
some "micro-optimization" in the 2D case:

```julia
function sih3(H::Hyperrectangle{N}) where {N}
    # @assert dim(H) == 2 ... (from outside)
    @inbounds begin
        a1p = abs(H.center[1] + H.radius[1])
        a1m = abs(H.center[1] - H.radius[1])
        a2p = abs(H.center[2] + H.radius[2])
        a2m = abs(H.center[2] - H.radius[2])
    end  
    r1 = a1p > a1m ? a1p : a1m
    r2 = a2p > a2m ? a2p : a2m
    return Hyperrectangle(zeros(N, 2), [r1, r2])
end
```

```julia
julia> H = rand(Hyperrectangle, dim=2)

julia> @btime box_approximation_symmetric($H)
  140.574 ns (6 allocations: 512 bytes)

julia> @btime sih3($H)
  55.775 ns (3 allocations: 224 bytes)
```

It is worth noting that we haven't yet used the SingleEntryVector to speedup the generic function.

mforets:
Closes #2022.

Requires: ~~https://github.com/JuliaReach/LazySets.jl/pull/2021~~.

### DIFF
--- a/src/Approximations/box_approximations.jl
+++ b/src/Approximations/box_approximations.jl
@@ -56,11 +56,25 @@ end
 box_approximation_symmetric(∅::EmptySet) = ∅
 
 """
-    symmetric_interval_hull
+    symmetric_interval_hull(X::LazySet)
 
-Alias for `box_approximation_symmetric`.
+Return a set that tightly overapproximates `X` with a hyperrectangular set symmetric
+respect to the origin.
 """
-symmetric_interval_hull = box_approximation_symmetric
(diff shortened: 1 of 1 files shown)


mforets:
Superseded by https://github.com/JuliaReach/LazySets.jl/pull/2030.

PR #2026, opened 2020-02-28
#2019 - Special case concrete Minkowski sum for intervals

### LINKED ISSUE #2019: Special case concrete Minkowski sum for intervals
Currently, the concrete minkowski sum between a pair of intervals falls back to the [AbstractHyperrectangle](https://github.com/JuliaReach/LazySets.jl/blob/a4c2db2cb07f7b66f441074303cbfdba4bc276e8/src/ConcreteOperations/minkowski_sum.jl#L216). It should rather use the concrete `+`.

mforets:
Closes #2019.

### DIFF
--- a/src/ConcreteOperations/minkowski_sum.jl
+++ b/src/ConcreteOperations/minkowski_sum.jl
@@ -248,6 +248,25 @@ function minkowski_sum(Z1::AbstractZonotope{N}, Z2::AbstractZonotope{N};
     return Zonotope(cnew, Gnew, remove_zero_generators=remove_zero_generators)
 end
 
+"""
+    minkowski_sum(x::Interval, y::Interval)
+
+Concrete Minkowski sum of a pair of intervals.
+
+### Input
+
+- `x` -- hyperrectangular set
+- `y` -- hyperrectangular set
+
+### Output
+
+An `Interval` corresponding to the concrete Minkowski sum of `x` and `y`.
+
+### Algorithm
+
+The function takes the sum of `x` and `y` following the rules of interval
+arithmetic.
+"""
 function minkowski_sum(x::Interval, y::Interval)
     return Interval(x.dat + y.dat)
 end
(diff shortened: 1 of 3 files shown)


mforets:
In this branch,

```julia
julia> x, y = rand(Interval), rand(Interval)

julia> @btime minkowski_sum($x, $y)
  0.017 ns (0 allocations: 0 bytes)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([0.0864657, 2.39401])
```
In master,

```julia
julia> @btime minkowski_sum($x, $y)
  154.339 ns (7 allocations: 608 bytes)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([-0.46852105974655833], [0.6100585759329054])
```

test/unit_Interval.jl
```diff
@@ -56,6 +56,11 @@ for N in Ns
     v = vertices_list(m)
     @test N[1.5] in v && N[-2] in v
 
+    # the concrete Minkowski sum of intervals returns an interval
+    sum_x_y = minkowski_sum(x, y)
+    @test sum_x_y isa Interval
+    @test min(sum_x_y) == N[-2] && max(sum_x_y) == N[1.5]
```
```suggestion
    @test minkowski_sum(x, y) == Interval(N(-2), N(1.5))
```

src/ConcreteOperations/minkowski_sum.jl
```diff
@@ -247,3 +247,7 @@ function minkowski_sum(Z1::AbstractZonotope{N}, Z2::AbstractZonotope{N};
     Gnew = hcat(genmat(Z1), genmat(Z2))
     return Zonotope(cnew, Gnew, remove_zero_generators=remove_zero_generators)
 end
+
+function minkowski_sum(x::Interval, y::Interval)
+    return Interval(x.dat + y.dat)
+end
```
schillic: The documentation would still point to `minkowski_sum(::AbstractHyperrectangle, ::AbstractHyperrectangle)`, which may suggest that there is no specialized implementation. Just saying...

test/unit_Interval.jl
mforets:
```suggestion
    @test min(sum_x_y) == N(-2) && max(sum_x_y) == N(1.5)
```

src/ConcreteOperations/minkowski_sum.jl
mforets: ok. im fine adding the docstring for this specialized method, i think that is what we do in LazySets always

PR #2030, opened 2020-02-29
#2022 - Faster symmetric interval hull for 2D hyperrectangles

### LINKED ISSUE #2022: Faster symmetric interval hull for 2D hyperrectangles

mforets:
Closes #2022.

### DIFF
--- a/src/Approximations/box_approximations.jl
+++ b/src/Approximations/box_approximations.jl
@@ -45,11 +45,7 @@ The center of the box is the origin, and the radius is obtained by computing the
 maximum value of the support function evaluated at the canonical directions.
 """
 function box_approximation_symmetric(S::LazySet{N}) where {N<:Real}
-    _box_approximation_symmetric_fallback(S)
-end
-
-# fallback returns a hyperrectangular set
-function _box_approximation_symmetric_fallback(S::LazySet{N}) where {N<:Real}
+    # fallback returns a hyperrectangular set
     (c, r) = box_approximation_helper(S)
     if r[1] < 0
         return EmptySet{N}(dim(S))
--- a/src/Approximations/box_approximations.jl
+++ b/src/Approximations/box_approximations.jl
@@ -49,7 +49,7 @@ function box_approximation_symmetric(S::LazySet{N}) where {N<:Real}
 end
 
 # fallback returns a hyperrectangular set
-function _box_approximation_symmetric_fallback(S::LazySet)
+function _box_approximation_symmetric_fallback(S::LazySet{N}) where {N<:Real}
     (c, r) = box_approximation_helper(S)
     if r[1] < 0
         return EmptySet{N}(dim(S))
--- a/src/Approximations/box_approximations.jl
+++ b/src/Approximations/box_approximations.jl
@@ -45,6 +45,11 @@ The center of the box is the origin, and the radius is obtained by computing the
 maximum value of the support function evaluated at the canonical directions.
 """
 function box_approximation_symmetric(S::LazySet{N}) where {N<:Real}
+    _box_approximation_symmetric_fallback(S)
+end
+
+# fallback returns a hyperrectangular set
+function _box_approximation_symmetric_fallback(S::LazySet)
     (c, r) = box_approximation_helper(S)
     if r[1] < 0
         return EmptySet{N}(dim(S))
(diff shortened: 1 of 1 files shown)


src/Approximations/box_approximations.jl
```diff
@@ -70,6 +75,29 @@ function box_approximation_symmetric(x::Interval)
...
+        # the hyperrectangle is seen as the cartesian product [v₁⁻, v₁⁺] x [v₂⁻, v₂⁺]
+        v₁⁺ = abs(H.center[1] + H.radius[1])
+        v₁⁻ = abs(H.center[1] - H.radius[1])
+        v₂⁺ = abs(H.center[2] + H.radius[2])
+        v₂⁻ = abs(H.center[2] - H.radius[2])
+    end
+    r₁ = max(v₁⁺, v₁⁻)
+    r₂ = max(v₂⁺, v₂⁻)
+    return Hyperrectangle(zeros(N, 2), [r₁, r₂])
+end
```
schillic: This code does not make use of two dimensions. It can easily be generalized to n dimensions.
```julia
function _box_approximation_symmetric(H::Hyperrectangle{N}) where {N}
    n = dim(H)
    r = Vector{N}(undef, n)
    @inbounds for i in 1:n
        v⁺ = abs(H.center[i] + H.radius[i])
        v⁻ = abs(H.center[i] - H.radius[i])
        r[i] = max(v⁺, v₁)
    end
    return Hyperrectangle(zeros(N, n), r)
end
```

mforets: yes. however, i haven't compared this version with the fallback especially for higher dimensions. i expect that if we use single entry vector in the fallback the results should be similar.

schillic: I expect that the generalized version is equivalent to your version in 2D. So there should be a gain.

src/Approximations/box_approximations.jl
```diff
@@ -45,6 +45,11 @@ The center of the box is the origin, and the radius is obtained by computing the
 maximum value of the support function evaluated at the canonical directions.
 """
 function box_approximation_symmetric(S::LazySet{N}) where {N<:Real}
+    _box_approximation_symmetric_fallback(S)
+end
+
+# fallback returns a hyperrectangular set
+function _box_approximation_symmetric_fallback(S::LazySet)
```
mforets:
```suggestion
function _box_approximation_symmetric_fallback(S::LazySet{N}) where {N<:Real}
```

src/Approximations/box_approximations.jl
mforets: > I expect that the generalized version is equivalent to your version in 2D.

I confirm; the overhead of the generalized version (ie. using the `for` loop instead of writing the 4 operations explicitly) is only ~2ns, so i won't make a particular case.

>  i haven't compared this version with the fallback especially for higher dimensions.

now i have compared, and the new version scales much better, here are some numbers:

```julia
H100 = Hr(100)
@btime symmetric_interval_hull($H100)
@btime _box_approximation_symmetric($H100)

  17.692 μs (6 allocations: 4.41 KiB)
  183.716 ns (3 allocations: 1.78 KiB)

H50 = Hr(50)
@btime symmetric_interval_hull($H50)
@btime _box_approximation_symmetric($H50)

  5.612 μs (6 allocations: 2.45 KiB)
  115.675 ns (3 allocations: 1.00 KiB)

H20 = Hr(20)
@btime symmetric_interval_hull($H20)
@btime _box_approximation_symmetric($H20)

  1.191 μs (6 allocations: 1.20 KiB)
  80.315 ns (3 allocations: 512 bytes)

H10 = Hr(10)
@btime symmetric_interval_hull($H10)
@btime _box_approximation_symmetric($H10)

  336.631 ns (6 allocations: 832 bytes)
  68.873 ns (3 allocations: 352 bytes)
```

mforets: while we are at it, this version is ~10..20% faster for dimensions >= 20:

```julia
@inline function _absmax_interval(c::N, r::N) where {N}
    if c >= zero(N)
        m = c + r
    else
        m = -c + r
    end
    return m
end

function _box_approximation_symmetric_opt(H::Hyperrectangle{N}) where {N<:Real}
    n = dim(H)
    r = Vector{N}(undef, n)
    @inbounds for i in 1:n
        r[i] = _absmax_interval(H.center[i], H.radius[i])
    end
    return Hyperrectangle(zeros(N, n), r)
end
```

src/Approximations/box_approximations.jl
```diff
@@ -45,6 +45,11 @@ The center of the box is the origin, and the radius is obtained by computing the
 maximum value of the support function evaluated at the canonical directions.
 """
 function box_approximation_symmetric(S::LazySet{N}) where {N<:Real}
+    _box_approximation_symmetric_fallback(S)
+end
+
+# fallback returns a hyperrectangular set
+function _box_approximation_symmetric_fallback(S::LazySet{N}) where {N<:Real}
```
schillic: I think this can be removed again.

PR #2038, opened 2020-03-01
#2020 - Make + operator lazy for intervals

### LINKED ISSUE #2020: Make `+` lazy for Intervals
When we wrote the `Interval` type we left `+` to be "particular" as taking the concrete m-sum, with the idea that it is fast to do for intevals.. I changed my mind, and i think that `+` should use the lazy, as with the other set types. (but note #2019).

mforets:
Closes #2020.

### DIFF
--- a/src/Sets/Interval.jl
+++ b/src/Sets/Interval.jl
@@ -181,24 +181,6 @@ function center(x::Interval{N}) where {N<:Real}
 end
 
 """
-    +(x::Interval{N}, y::Interval{N}) where {N<:Real}
-
-Return the sum of the intervals.
-
-### Input
-
-- `x` -- interval
-- `y` -- interval
-
-### Output
-
-The sum of the intervals as a new `Interval` set.
-"""
-function +(x::Interval{N}, y::Interval{N}) where {N<:Real}
-    return Interval(x.dat + y.dat)
-end
-
-"""
     -(x::Interval{N}, y::Interval{N}) where {N<:Real}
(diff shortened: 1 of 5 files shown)


src/Sets/Interval.jl
```diff
@@ -180,24 +180,6 @@ function center(x::Interval{N}) where {N<:Real}
...
-- `x` -- interval
-- `y` -- interval
-
-### Output
-
-The sum of the intervals as a new `Interval` set.
-"""
-function +(x::Interval{N}, y::Interval{N}) where {N<:Real}
-    return Interval(x.dat + y.dat)
-end
```
schillic: The docs need to be removed.

src/convert.jl
```diff
@@ -930,3 +930,21 @@ function convert(::Type{MinkowskiSumArray},
...
+
+"""
+    convert(::Type{Interval}, x::MinkowskiSum{N, IT, IT}) where {N, IT<:Interval{N}}
+
+Converts the Minkowski sum of two intervals into an interval.
+
+### Input
+
+- `Interval` -- type used for dispatch
+- `X`        -- Minkowski sum of a pair of intervals
```
```suggestion
- `x`        -- Minkowski sum of a pair of intervals
```
Although I would prefer to use capital letters for sets...

src/Sets/Interval.jl
mforets: aha! good catch

PR #2039, opened 2020-03-01
#2032 - Add corresponding matrix type and vector type

### LINKED ISSUE #2032: Add corresponding vector type and matrix type
`Vector` <--> `Matrix`
`SparseVector` <--> `SparseMatrixCSC`
etc., and using regular arrays as fallback
cf. implementation in Polyhedra.jl: https://github.com/JuliaPolyhedra/Polyhedra.jl/blob/master/src/Polyhedra.jl#L48

mforets:
Closes #2032.

### DIFF
--- a/src/Arrays/matrix_vector_operations.jl
+++ b/src/Arrays/matrix_vector_operations.jl
@@ -28,7 +28,7 @@ Return a corresponding vector type with respect to type `T`.
 
 ### Input
 
-- `T` -- vector type
+- `T` -- vector or matrix type
 
 ### Output
 
--- a/src/Arrays/matrix_vector_operations.jl
+++ b/src/Arrays/matrix_vector_operations.jl
@@ -20,3 +20,57 @@ function inner(x::AbstractVector{N}, A::AbstractMatrix{N}, y::AbstractVector{N}
               ) where {N}
     dot(x, A * y)
 end
+
+"""
+    _vector_type(T)
+
+Return a corresponding vector type with respect to type `T`.
+
+### Input
+
+- `T` -- vector type
+
+### Output
+
+A vector type that corresponds in some sense (see Notes below) to `T`.
+
+### Notes
+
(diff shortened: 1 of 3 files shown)


src/Arrays/matrix_vector_operations.jl
```diff
@@ -20,3 +20,57 @@ function inner(x::AbstractVector{N}, A::AbstractMatrix{N}, y::AbstractVector{N}
...
 end
+
+"""
+    _vector_type(T)
+
+Return a corresponding vector type with respect to type `T`.
+
+### Input
+
+- `T` -- vector type
```
schillic:
```suggestion
- `T` -- matrix type
```

mforets: Actually, this should be either vector or matrix type.

mforets:
```suggestion
- `T` -- vector or matrix type
```

src/Arrays/matrix_vector_operations.jl
```diff
@@ -20,3 +20,57 @@ function inner(x::AbstractVector{N}, A::AbstractMatrix{N}, y::AbstractVector{N}
...
+function _vector_type end
+
+"""
+    _matrix_type(T)
+
+Return a corresponding matrix type with respect to type `T`.
+
+### Input
+
+- `T` -- vector type
```
mforets:
```suggestion
- `T` -- vector or matrix type
```

mforets:
> Maybe add to the documentation that the Otherwise case may change in the future.

one comment here: we do have to add more dispatches eg for `SArray` and `MArrays` types, but the "otherwise" case is still going to be the fallback. Also, these functions can be easily extended by users code, so we have set types and set operations (once we use these functions, were appropriate) that "just work" with such custom vector/array type pairs in a predictable way.

PR #2062, opened 2020-03-12
#2055 - Area of 2D set

### LINKED ISSUE #2055: Area of 2D set
- https://es.wikipedia.org/wiki/F%C3%B3rmula_del_%C3%A1rea_de_Gauss
- https://en.wikipedia.org/wiki/Shoelace_formula

mforets:
Closes #2055.

### DIFF
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -23,7 +23,8 @@ export LazySet,
        isoperationtype,
        isequivalent,
        isconvextype,
-       area
+       area,
+       surface
 
 """
     LazySet{N}
@@ -881,9 +882,9 @@ true
 isconvextype(X::Type{<:LazySet}) = false
 
 """
-    area(X::LazySet{N}) where {N}
+    surface(X::LazySet{N}) where {N}
 
-Compute the area of a set.
+Compute the surface area of a set.
 
 ### Input
 
@@ -891,19 +892,19 @@ Compute the area of a set.
 
 ### Output
 
-A number representing the area of `X`.
+A number representing the surface area of `X`.
 """
-function area(X::LazySet{N}) where {N}
+function surface(X::LazySet{N}) where {N}
     if dim(X) == 2
-        return area_2D(X)
+        return area(X)
     else
-        throw(ArgumentError("the area function only applies to two-dimensional sets, " *
-                            "but the given set is $(dim(X))-dimensional"))
+        throw(ArgumentError("the surface function is only implemented for " *
+                    "two-dimensional sets, but the given set is $(dim(X))-dimensional"))
     end
 end
 
 """
-    area_2D(X::LazySet{N}) where {N}
+    area(X::LazySet{N}) where {N}
 
 Compute the area of a two-dimensional polytopic set using the Shoelace formula.
 
@@ -915,19 +916,24 @@ Compute the area of a two-dimensional polytopic set using the Shoelace formula.
 
 A number representing the area of `X`.
 
+### Notes
+
+This algorithm is applicable to any lazy set `X` such that its list of vertices,
+`vertices_list`, can be computed.
+
 ### Algorithm
 
-Let `m` be the number of vertices of `X`. The following instances are special-cased:
+Let `m` be the number of vertices of `X`. The following instances are considered:
 
 - `m = 0, 1, 2`: the output is zero.
 - `m = 3`: the triangle case is computed using the Shoelace formula with 3 points.
 - `m = 4`: the quadrilateral case is obtained by the factored version of the Shoelace
            formula with 4 points.
 
-Otherwise, the general Shoelace formula is used; for detals see the wikipedia article
-[Shoelace formula](https://en.wikipedia.org/wiki/Shoelace_formula).
+Otherwise, the general Shoelace formula is used; for detals see the wikipedia
+article [Shoelace formula](https://en.wikipedia.org/wiki/Shoelace_formula).
 """
-function area_2D(X::LazySet{N}) where {N}
+function area(X::LazySet{N}) where {N}
     @assert dim(X) == 2 "this function only applies to two-dimensional sets, " *
     "but the given set is $(dim(X))-dimensional"
 
@@ -939,28 +945,36 @@ function area_2D(X::LazySet{N}) where {N}
     end
 
     if m == 3 # triangle
-        A = vlist[1]
-        B = vlist[2]
-        C = vlist[3]
-        res = A[1] * (B[2] - C[2]) + B[1] * (C[2] - A[2]) + C[1] * (A[2] - B[2])
-        return abs(res/2)
+        res = _area_triangle(vlist)
 
     elseif m == 4 # quadrilateral
-        A = vlist[1]
-        B = vlist[2]
-        C = vlist[3]
-        D = vlist[4]
-        res = A[1] * (B[2] - D[2]) + B[1] * (C[2] - A[2]) + C[1] * (D[2] - B[2]) + D[1] * (A[2] - C[2])
-        return abs(res/2)
+        res = _area_quadrilateral(vlist)
 
     else # general case
-        return _area_2D(vlist)
+        res = _area_polygon(vlist)
     end
 
+    return res
+end
+
+function _area_triangle(v::Vector{VN}) where {N, VN<:AbstractVector{N}}
+    A = v[1]
+    B = v[2]
+    C = v[3]
+    res = A[1] * (B[2] - C[2]) + B[1] * (C[2] - A[2]) + C[1] * (A[2] - B[2])
     return abs(res/2)
 end
 
-function _area_2D(v::Vector{VN}) where {N, VN<:AbstractVector{N}}
+function _area_quadrilateral(v::Vector{VN}) where {N, VN<:AbstractVector{N}}
+    A = v[1]
+    B = v[2]
+    C = v[3]
+    D = v[4]
+    res = A[1] * (B[2] - D[2]) + B[1] * (C[2] - A[2]) + C[1] * (D[2] - B[2]) + D[1] * (A[2] - C[2])
+    return abs(res/2)
+end
+
+function _area_polygon(v::Vector{VN}) where {N, VN<:AbstractVector{N}}
     m = length(v)
     res = zero(N)
     for i in 1:m-1
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -897,8 +897,8 @@ function area(X::LazySet{N}) where {N}
     if dim(X) == 2
         return area_2D(X)
     else
-        throw(ArgumentError("the area of a $(dim(X))-dimensional set is not " *
-                            "implemented yet"))
+        throw(ArgumentError("the area function only applies to two-dimensional sets, " *
+                            "but the given set is $(dim(X))-dimensional"))
     end
 end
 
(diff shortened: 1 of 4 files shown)


test/unit_area.jl
```diff
@@ -0,0 +1,21 @@
...
+p = Interval(0.0, 1.0) × Interval(0.0, 0.0)
+@test area(p) ≈ 0.0
+
+# triangle
+p = VPolygon([[13, 14.], [16, 30.], [50, 10.]])
+@test area(p) ≈ 302.0
+
+# quadrilateral
+vlist = [[3, 4.], [5, 11.], [12, 8.], [9, 5.], [5, 6.]]
+p = VPolygon(vlist) # convex hull => 5 points
```
mforets:
```suggestion
p = VPolygon(vlist) # convex hull => 4 points
```

src/Interfaces/LazySet.jl
```diff
@@ -878,3 +879,93 @@ true
...
+### Output
+
+A number representing the area of `X`.
+"""
+function area(X::LazySet{N}) where {N}
+    if dim(X) == 2
+        return area_2D(X)
+    else
+        throw(ArgumentError("the area of a $(dim(X))-dimensional set is not " *
+                            "implemented yet"))
```
schillic: Can you add some definition of the area of a non-2D set? I would just define it for 2D sets. Otherwise maybe it should be called `volume` or `surface` if it generalizes.

mforets: Yes, i had in mind the (surface) area.

mforets: What change do you propose? The area *is* defined only for 2D sets in this PR. Or you suggest to change the error message to something else, like "the area function only applies to two-dimensional sets, but the given set is $(dim(X))-dimensional"?

schillic: Yes, I would change the error message to the one you suggested. We can still have this proxy method `area` but then if we want to add more methods, we have to call them `area2D` (just saying).

mforets: ah, i see. we can rename `area -> surface` (any dimension, in particular uses `area` for 2D) and `area_2D -> area` (only 2D). but this wouldn't be breaking, as `surface` is not available yet, we can do this at a later moment.

src/Interfaces/LazySet.jl
```diff
@@ -878,3 +879,93 @@ true
...
+- `m = 4`: the quadrilateral case is obtained by the factored version of the Shoelace
+           formula with 4 points.
+
+Otherwise, the general Shoelace formula is used; for detals see the wikipedia article
+[Shoelace formula](https://en.wikipedia.org/wiki/Shoelace_formula).
+"""
+function area_2D(X::LazySet{N}) where {N}
+    @assert dim(X) == 2 "this function only applies to two-dimensional sets, " *
+    "but the given set is $(dim(X))-dimensional"
+
```
schillic: I would not check this again. Or remove the other method completely.

src/Interfaces/LazySet.jl
```diff
@@ -878,3 +879,93 @@ true
...
+    vlist = vertices_list(X)
+    m = length(vlist)
+
+    if 0 <= m <= 2
```
```suggestion
    if m <= 2
```

src/Interfaces/LazySet.jl
```diff
@@ -878,3 +879,93 @@ true
...
+        return zero(N)
+    end
+
+    if m == 3 # triangle
+        A = vlist[1]
+        B = vlist[2]
+        C = vlist[3]
+        res = A[1] * (B[2] - C[2]) + B[1] * (C[2] - A[2]) + C[1] * (A[2] - B[2])
+        return abs(res/2)
```
schillic: It is probably beneficial if you outsource these cases to methods as well because then it can dispatch on the vector type.

test/unit_area.jl
```diff
@@ -0,0 +1,21 @@
+# not implemented for dimension >= 3
```
schillic:
```suggestion
# not implemented for dimension != 2
```

test/unit_area.jl
```diff
@@ -0,0 +1,21 @@
...
+p = VPolygon([[13, 14.], [16, 30.], [50, 10.]])
+@test area(p) ≈ 302.0
+
+# quadrilateral
+vlist = [[3, 4.], [5, 11.], [12, 8.], [9, 5.], [5, 6.]]
+p = VPolygon(vlist) # convex hull => 4 points
+@test area(p) ≈ 35.0
+
+# general case (non-convex)
+@test LazySets._area_2D(vlist) ≈ 30.0
```
schillic: What does "non-convex" mean?
I would rather test the `area` method for the case `m > 4` instead of the helper method.

src/Interfaces/LazySet.jl
mforets: ok

test/unit_area.jl
mforets: the area function applies to non-convex polygons too, and this example here is the one in wikipedia: see "More Complex Example" in https://en.wikipedia.org/wiki/Shoelace_formula

mforets: but if you have another polygon example to test i can do add it, sure.

mforets: >  instead of the helper method.

note that i tested the helper alone because `VPolygon(..)` would take the convex hull (unless i ask it not to do so..)

src/Interfaces/LazySet.jl
mforets: the change you propose is connected to the previous thread, https://github.com/JuliaReach/LazySets.jl/pull/2062#discussion_r391646940 . i can make the proposed change now, about adding a `surface` function. the top-level `area` should also check that `X` is two-dimensional -- so it's computed twice if the user calls `surface(p)`. but this shouldn't be a problem, since that operation is super cheap.

mforets:
this will be ready for review once CI passes.

test/unit_area.jl
schillic: Okay, I understand now. Thanks for the explanation.

src/Interfaces/LazySet.jl
schillic: The `surface` function did not exist back then (I think). With the new structure I agree to test it twice.

mforets: ok!

test/unit_area.jl
```diff
@@ -0,0 +1,25 @@
...
+# quadrilateral
+vlist = [[3, 4.], [5, 11.], [12, 8.], [9, 5.], [5, 6.]]
+p = VPolygon(vlist) # convex hull => 4 points
+@test area(p) ≈ 35.0
+
+# test surface function, which falls back to the area function for 2D sets
+@test surface(p) ≈ 35.0
+
+# general case (non-convex)
+@test LazySets._area_polygon(vlist) ≈ 30.0
```
schillic:
```suggestion
for N in [Float64, Float32, Rational{Int}]
    # not implemented for dimension != 2
    p = BallInf(zeros(N, 3), N(1))
    @test_throws ArgumentError surface(p)
    @test_throws AssertionError area(p)

    # sets with zero area
    p = Singleton([0, 1])
    @test area(p) ≈ N(0)
    p = Interval(N(0), N(1)) × Interval(N(0), N(0))
    @test area(p) ≈ N(0)

    # triangle
    p = VPolygon([N[13, 14], N[16, 30], N[50, 10]])
    @test area(p) ≈ N(302)

    # quadrilateral
    vlist = [N[3, 4], N[5, 11], N[12, 8], N[9, 5], N[5, 6]]
    p = VPolygon(vlist) # convex hull => 4 points
    @test area(p) ≈ N(35)

    # more than four vertices
    p2 = VPolygon([N[1, 0], N[0, 0], N[0, 1], N[1/2, 2], [1, 1]])
    @test area(p2) == N[3/2]

    # non-convex polygon (via list of vertices)
    @test LazySets._area_polygon(vlist) ≈ N(30)

    # test surface function, which falls back to the area function for 2D sets
    @test surface(p) ≈ N(35)
end
```

test/unit_area.jl
```diff
@@ -0,0 +1,31 @@
+for N in [Float64, Float32, Rational{Int}]
+    # not implemented for dimension != 2
+    p = BallInf(zeros(N, 3), N(1))
+    @test_throws ArgumentError surface(p)
+    @test_throws AssertionError area(p)
+
+    # sets with zero area
+    p = Singleton([0, 1])
```
mforets:
```suggestion
    p = Singleton(N[0, 1])
```

src/Interfaces/LazySet.jl
```diff
@@ -878,3 +880,106 @@ true
...
+    return abs(res/2)
+end
+
+function _area_polygon(v::Vector{VN}) where {N, VN<:AbstractVector{N}}
+    m = length(v)
+    res = zero(N)
+    for i in 1:m-1
+        @inbounds res += v[i][1] * v[i+1][2] - v[i+1][1] * v[i][2]
+    end
+    @inbounds res += v[m][1] * v[1][2] - v[1][1] * v[m][2]
```
```suggestion
    @inbounds res = v[m][1] * v[1][2] - v[1][1] * v[m][2]
    for i in 1:m-1
        @inbounds res += v[i][1] * v[i+1][2] - v[i+1][1] * v[i][2]
    end
```
Saves one line and one addition :muscle:

schillic: I expect that the formula with only one multiplication per iteration is much more efficient. Is there a reason why you chose this implementation?

test/unit_area.jl
```diff
@@ -0,0 +1,31 @@
...
+    @test area(p) ≈ N(302)
+
+    # quadrilateral
+    vlist = [N[3, 4], N[5, 11], N[12, 8], N[9, 5], N[5, 6]]
+    p = VPolygon(vlist) # convex hull => 4 points
+    @test area(p) ≈ N(35)
+
+    # more than four vertices
+    p2 = VPolygon([N[1, 0], N[0, 0], N[0, 1], N[1/2, 2], [1, 1]])
+    @test area(p2) == N[3/2]
```
schillic:
```suggestion
    @test area(p2) == N(3/2)
```

src/Interfaces/LazySet.jl
mforets: :) i won't stop you to try and pull request a more efficient implementation, but for this PR the branch is fairly good i think.

> Is there a reason why you chose this implementation?

no.

mforets:
Thanks for your review @schillic, let me know if you would like to consider further https://github.com/JuliaReach/LazySets.jl/pull/2062#discussion_r391851484 or it is ready to merge from your side.

src/Interfaces/LazySet.jl
schillic: What about the code suggestion at the top?

mforets: if you insist :)

mforets: for the record i tried this version today and it is actually *slower*, 129ns vs 100ns for 100 for a random polygon with 100 vertices.

```julia
function _area_polygon2(v::Vector{VN}) where {N, VN<:AbstractVector{N}}
    m = length(v)
    @inbounds res = (v[1][1] + v[m][1]) * (v[1][2] - v[m][2])
    for i in 1:m-1
        @inbounds res += (v[i+1][1] + v[i][1]) * (v[i+1][2] - v[i][2])
    end
    return abs(res/2)
end
```

PR #2068, opened 2020-03-13
#1901 - Revisit delete_zero_columns!

### LINKED ISSUE #1901: Revisit delete_zero_columns!
- The name `delete_zero_columns!` is a misnomer, because it does not remove columns inplace; it just returns a view over the nonzero columns.
- Rename the function to `nonzero_columns`. 
- Add a better implementation for sparse matrices.
- Update Zonotope code. This makes creation of zonotopes faster by a large margin (esp. in high dimensions), see https://github.com/JuliaReach/LazySets.jl/issues/1975#issuecomment-587000133

mforets:
Closes #1901.

See the description of changes in https://github.com/JuliaReach/LazySets.jl/issues/1901#issue-545510461.

### DIFF
--- a/src/Arrays/matrix_operations.jl
+++ b/src/Arrays/matrix_operations.jl
@@ -5,7 +5,7 @@ export _At_mul_B,
        issquare,
        isinvertible,
        cross_product,
-       delete_zero_columns!,
+       nonzero_columns,
        extend,
        projection_matrix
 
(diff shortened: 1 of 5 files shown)


src/Arrays/matrix_operations.jl
```diff
@@ -147,46 +147,32 @@ cross_product(M::AbstractSparseMatrix) = cross_product(Matrix(M))
...
+        if !iszero(view(A, :, i))
+            push!(nzcol, i)
         end
     end
+    return nzcol
+end
+
+function nonzero_columns(A::SparseMatrixCSC)
+    return collect(i for i in 1:A.n if A.colptr[i] < A.colptr[i+1])
 end
```
schillic: Can you add a test for sparse matrices?

mforets:
I added a commit that fixes that Singleton case (and improves it a bit, since it is now specialized for this type). On the other hand, it still breaks to use a Diagonal with all zeros. This whole thing of removing zero generators *in the inner constructor* doesn't look great, because it can potentially change the matrix type.

src/Sets/Zonotope.jl
```diff
@@ -107,11 +107,15 @@ struct Zonotope{N<:Real, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}} <: Abstra
...
         if remove_zero_generators
-            generators = delete_zero_columns!(generators)
+            nzcol = nonzero_columns(generators)
+            if length(nzcol) < size(generators, 2)
+                G = view(generators, :, nzcol)
+            end
         end
         MT = typeof(generators)
-        new{N, VN, MT}(center, generators)
+        new{N, VN, MT}(center, G)
```
schillic: Instead you can do (I tested it):
```julia
        return _Zonotope(center, G)
    end

    # constructor with fixed types
    function _Zonotope(center::VN, generators::MN) where {N<:Real,
                                                          VN<:AbstractVector{N},
                                                          MN<:AbstractMatrix{N}}
        return new{N, VN, MN}(center, generators)
    end
```

mforets: ok i did that change, thanks.

---

as commented above, i still consider that it's worth reviewing this default check, if we want to keep it or not. it imposes a high slowdown for the creation of zonotopes, easily 10x up to 50x from tests that i did. recall that the motivation for this constructor check (https://github.com/JuliaReach/LazySets.jl/pull/1246) was the conversion from flat hyperrectangles to zonotopes. however this can be done done like so:

```julia
function convert(::Type{Zonotope}, H::AbstractHyperrectangle)
    G = genmat(H)
    if isflat(H)
         # get nonzero columns
         # new G wih removed generators
    end
    return Zonotope(center(H), G)
end
```

mforets: i added a proof of this principle in https://github.com/JuliaReach/LazySets.jl/pull/2072

schillic: > it imposes a high slowdown for the creation of zonotopes

In internal code we should deactivate the check as long as we can guarantee that nothing bad happens. Your flat hyperrectangle conversion is a good example. But for the general API I would keep the checks to avoid confusion about unexpected behavior. For performance-critical code there is an option to deactivate the check.

> easily 10x up to 50x

A factor is not a fair comparison because the rest of the constructor is constant.

mforets: > A factor is not a fair comparison because the rest of the constructor is constant.

yes, but the application that i have in mind is reachability analysis where you create thousands or hundreds of thousands of zonotopes.

schillic: > For performance-critical code there is an option to deactivate the check.

mforets: yes, [that(https://github.com/JuliaReach/LazySets.jl/pull/2068#discussion_r392668546 is something you proposed already and now for a 2nd time :)  i agree about the premise of deactivating this check internally so long as the operation does not create zero generators. but thinking this through, i don't know of any meaningful operation / use that would create zero generators, so i was thinking if we should rather remove this option altogether (in the constructor). the flat hyperrectangle case can be handled separately as stated above.

schillic: > i don't know of any meaningful operation / use that would create zero generators

The user interface for creating a `Zonotope` lets you pass an arbitrary matrix.

mforets:
This PR is superseded by https://github.com/JuliaReach/LazySets.jl/pull/2072.

Issue #1975, opened 2020-02-16
Improve performance of  minkowski_sum of two Zonotopes


ueliwechsler:
As discussed on gitter, the performance of the `minkowsk_sum` for two `Zonotope` can be improved.

As far as I understand it, this can be done in two ways.
1. set `remove_zero_generators` to `false`
```julia
function minkowski_sum_v2(Z1::Zonotope{N}, Z2::Zonotope{N}) where {N}
    G = hcat(Z1.generators, Z2.generators)
    return Zonotope(Z1.center + Z2.center, G, remove_zero_generators=false)
end
```
2. improving the performance of the `delete_zero_columns!` method
https://github.com/JuliaReach/LazySets.jl/blob/054f1d2aac78e074595855b25fd35c3b789c7ff2/src/Sets/Zonotope.jl#L96

Here some benchmark, just for fun ;)
```julia
using LazySets
using BenchmarkTools

function minkowski_sum_v2(Z1::Zonotope{N}, Z2::Zonotope{N}) where {N}
    G = hcat(Z1.generators, Z2.generators)
    return Zonotope(Z1.center + Z2.center, G, remove_zero_generators=false)
end

n = 100
P = convert(Zonotope, BallInf(zeros(n), 1.0))
Q = convert(Zonotope, BallInf(zeros(n), 2.0))
@btime minkowski_sum($P, $Q)
@btime minkowski_sum_v2($P, $Q)

## n = 10
# 2.511 μs (94 allocations: 8.41 KiB)
#881.836 ns (33 allocations: 3.80 KiB)
## n = 20
# 4.671 μs (154 allocations: 14.61 KiB)
# 916.667 ns (33 allocations: 5.47 KiB)
## n = 100
# 35.600 μs (634 allocations: 64.47 KiB)
# 1.867 μs (33 allocations: 19.06 KiB)
```

schillic:
Option 1 has been implemented in #1962.

ueliwechsler:
Ah cool, that's great. My bad.

mforets:
Thanks for the benchmarks @ueliwechsler, some comments:

- In the release [v1.28.0](https://github.com/JuliaReach/LazySets.jl/releases/tag/v1.28.0) we added the option `remove_zero_generators` but set to `true` by default: this is the same default of the Zonotope constructor, so the performance does not really change :) unless you compute `minkowski_sum(Z1, Z2, remove_zero_generators=false)`.
[...]

PR #2072, opened 2020-03-15
Remove zero gens check in zonotope constructor

mforets:
Closes https://github.com/JuliaReach/LazySets.jl/issues/1901.
Closes https://github.com/JuliaReach/LazySets.jl/issues/1975

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -1033,9 +1033,6 @@ function overapproximate(vTM::Vector{TaylorModel1{T, S}},
 
     # compute overapproximation
     return _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
-
-    # remove generators which are zero
-    return remove_zero_generators(Z)
 end
 
 """
@@ -1123,13 +1120,11 @@ function overapproximate(vTM::Vector{TaylorModelN{N, T, S}},
     gen_rem = Vector{T}(undef, m) # generators for the remainder
 
     # compute overapproximation
-    Z = _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
-
-    # remove generators which are zero
-    return remove_zero_generators(Z)
+    return _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
 end
 
 function _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
+    nz_gen_rem = 0 # counter for non-zero generators of the remainder
     @inbounds for (i, x) in enumerate(vTM)
         xpol, xdom = polynomial(x), domain(x)
 
@@ -1148,9 +1143,22 @@ function _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
         α = mid(rem_nonlin)
         c[i] = constant_term(Q) + α  # constant terms
         gen_lin[i, :] = get_linear_coeffs(Q) # linear terms
-        gen_rem[i] = abs(rem_nonlin.hi - α)
+        aux = abs(rem_nonlin.hi - α)
+        gen_rem[i] = aux
+        if !isapproxzero(aux)
+            nz_gen_rem += 1
+        end
     end
-    return Zonotope(c, hcat(gen_lin, Diagonal(gen_rem)))
+    if nz_gen_rem > 0
+        D = Diagonal(gen_rem)
+        if nz_gen_rem < size(D, 2)
+            D = D[:, 1:nz_gen_rem]
+        end
+        Z = Zonotope(c, hcat(gen_lin, D))
+    else
+        Z = Zonotope(c, gen_lin)
+    end
+    return Z
 end
 
 end # quote
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -432,7 +432,8 @@ function _overapproximate_convex_hull_zonotope_G05(
                  (G1[:, 1:ngens(Z2)] .- G2) / N(2),
                  G1[:, ngens(Z2)+1:end])
     end
-    return Zonotope(c, G)
+    Z = Zonotope(c, G)
+    return remove_zero_generators(Z)
 end
 
 function _overapproximate_convex_hull_zonotope_GGP09(
@@ -1032,6 +1033,9 @@ function overapproximate(vTM::Vector{TaylorModel1{T, S}},
 
     # compute overapproximation
     return _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
+
+    # remove generators which are zero
+    return remove_zero_generators(Z)
 end
 
 """
@@ -1119,7 +1123,10 @@ function overapproximate(vTM::Vector{TaylorModelN{N, T, S}},
     gen_rem = Vector{T}(undef, m) # generators for the remainder
 
     # compute overapproximation
-    return _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
+    Z = _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
+
+    # remove generators which are zero
+    return remove_zero_generators(Z)
 end
 
 function _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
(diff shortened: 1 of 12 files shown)


src/Sets/Zonotope.jl
```diff
@@ -100,28 +100,39 @@ struct Zonotope{N<:Real, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}} <: Abstra
...
 end
 
 isoperationtype(::Type{<:Zonotope}) = false
 isconvextype(::Type{<:Zonotope}) = true
 
 # constructor from center and list of generators
-function Zonotope(center::VN, generators_list::AbstractVector{VN};
-                  remove_zero_generators::Bool=true) where {N<:Real, VN<:AbstractVector{N}}
-    return Zonotope(center, hcat(generators_list...); remove_zero_generators=remove_zero_generators)
+using LazySets.Arrays: _vector_type, _matrix_type
```
schillic, line 117: Put this line to the top?

src/Sets/Zonotope.jl
```diff
@@ -100,28 +100,39 @@ struct Zonotope{N<:Real, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}} <: Abstra
...
+function Zonotope(center::VN, generators_list::AbstractVector{VN}) where {N<:Real, VN<:AbstractVector{N}}
+    MT = _matrix_type(VN)
+    G = MT(undef, length(center), length(generators_list)) # TODO: generic undef creator?
+    for gi in generators_list
+        @inbounds G[:, i] = gi
```
schillic, line 122:
```suggestion
    for (j, gj) in enumerate(generators_list)
        @inbounds G[:, j] = gj
```

src/Sets/Zonotope.jl
mforets: done

src/Sets/Zonotope.jl
```diff
@@ -100,28 +100,39 @@ struct Zonotope{N<:Real, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}} <: Abstra
...
+end
+
+function remove_zero_generators(Z::Zonotope{N, VN, MN}) where {N<:Real,
+                                                               VN<:AbstractVector{N},
+                                                               MN<:AbstractMatrix{N}}
+    generators = genmat(Z)
+    nzcol = nonzero_columns(generators)
+    p = length(nzcol)
+    pmax = size(generators, 2)
+    G = (p == pmax) ? generators : view(generators, :, nzcol)
```
mforets, line 134: i'll change `view(generators, :, nzcol)` to not return the view here.

mforets:
> Looks good so far!

alright, thx. from my side, it remains to update the tests.

src/convert.jl
```diff
@@ -248,15 +248,40 @@ Converts a hyperrectangular set to a zonotope.
...
+- `Zonotope` -- type, used for dispatch
+- `H`        -- hyperrectangular set
 
 ### Output
 
 A zonotope.
 """
-function convert(::Type{Zonotope}, H::AbstractHyperrectangle)
-    return Zonotope(center(H), Diagonal(radius_hyperrectangle(H)))
+function convert(::Type{Zonotope}, H::AbstractHyperrectangle{N}) where {N}
```
mforets, line 258: the implementation was wrong; i added a new one. not sure though if we should keep this `if isflat(H)` branch, since, if it's not flat, the code in lines 260-276 is also ok.

src/Sets/Zonotope.jl
```diff
@@ -5,6 +5,9 @@ export Zonotope,
        scale,
        reduce_order
 
+# constructor from center and list of generators
+using LazySets.Arrays: _vector_type, _matrix_type
```
schillic, line 10: I think the comment should have stayed below :wink:

mforets: pfff

mforets:
there is a failing test, apparently the algorithm converting a TaylorModel to a zonotope does not remove zero columns,

```julia
    vTM = [TaylorModels.TaylorModelN(pi, I, x0, D) for pi in [p₁, p₂]]
    Z1 = overapproximate(vTM, Zonotope)
    @test center(Z1) == N[3, -2.5]
```

as a hot-fix i'll just remove zeros in the overapproximation.

src/convert.jl
```diff
@@ -248,15 +248,40 @@ Converts a hyperrectangular set to a zonotope.
...
+        sizehint!(Gnz, n * n)
+        @inbounds for (i, ri) in enumerate(r)
+            if ri != zero(N)
+                col = zeros(N, n)
+                col[i] = ri
+                append!(Gnz, col)
+                nzgen += 1
+            end
+        end
+        G = reshape(Gnz, n, nzgen)
```
schillic, line 274: It is probably more efficient to initialize `Gnz = zeros(N, n*n)` (or even `Gnz = zeros(N, n, n)`) and then only edit the entry that is not zero. But we can keep this code as well.

src/convert.jl
```diff
@@ -248,15 +248,40 @@ Converts a hyperrectangular set to a zonotope.
...
+function convert(::Type{Zonotope}, H::AbstractHyperrectangle{N}) where {N}
+    if isflat(H)
+        r = radius_hyperrectangle(H)
+        n = length(r)
+
+        nzgen = 0
+        Gnz = Vector{N}()
+        sizehint!(Gnz, n * n)
+        @inbounds for (i, ri) in enumerate(r)
+            if ri != zero(N)
```
schillic, line 267: Do you want to do an approximate zero check here?

src/Approximations/overapproximate.jl
```diff
@@ -1032,6 +1033,9 @@ function overapproximate(vTM::Vector{TaylorModel1{T, S}},
 
     # compute overapproximation
     return _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
+
+    # remove generators which are zero
+    return remove_zero_generators(Z)
```
schillic, line 1038:
```suggestion
    Z = _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)

    # remove generators which are zero
    return remove_zero_generators(Z)
```

src/convert.jl
mforets: maybe, i have t think about it. not in this PR.

src/Approximations/overapproximate.jl
mforets: this part, is still broken i have to think about how to fix it..

src/convert.jl
mforets: it is not clear to me why that alternative would be more efficient; in general, push / append are fast. but yes, those are possible alternatives and we can possibly micro-optimize this as well. the change is nevertheless much faster than the version in master (See tests in the minkowski_sum issue https://github.com/JuliaReach/LazySets.jl/issues/1975#issuecomment-587000133)

src/Approximations/overapproximate.jl
mforets: finally, i fixed the zero gens issue in the main loop. i dont want to close https://github.com/JuliaReach/LazySets.jl/issues/2106 yet, to give it a second thought later

src/Approximations/overapproximate.jl
```diff
@@ -1148,9 +1143,22 @@ function _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
...
+        aux = abs(rem_nonlin.hi - α)
+        gen_rem[i] = aux
+        if !isapproxzero(aux)
+            nz_gen_rem += 1
+        end
+    end
+    if nz_gen_rem > 0
+        D = Diagonal(gen_rem)
+        if nz_gen_rem < size(D, 2)
+            D = D[:, 1:nz_gen_rem]
```
schillic, line 1155: This does not make sense to me. I would expect that you have to remove those entries that are zero (and not just the last ones)

mforets: you're right, thanks. ill fix it in the loop

mforets: oops, i think that this is still wrong because `gen_rem[nz_gen_rem] = aux` may not place the element in the correct *row*

mforets: fixed -- i reverted these changes, and now drop the zeros in the `nonzero_columns` for sparse matrices

PR #2075, opened 2020-03-18
Add delaunay triangulation through MiniQhull / Qhull

### LINKED ISSUE #2074: Add Delaunay triangulation through MiniQhull (optional dependency)
https://github.com/gridap/MiniQhull.jl

mforets:
Closes #2074.

---

Use:

```julia
julia> using LazySets, MiniQhull

julia> V = rand(VPolygon, num_vertices=10);

julia> D = delaunay(V);

julia> typeof(D)
UnionSetArray{Float64,VPolytope{Float64,Array{Float64,1}}}

julia> using Plots

julia> plot(V); plot!(D)
```

Screenshot from 2020-03-18 17-16-57


src/Initialization/init_MiniQhull.jl
```diff
@@ -0,0 +1,36 @@
...
+
+- `X` -- set
+
+### Output
+
+A union of polytopes in vertex representation.
+
+### Notes
+
+This functions requires that you have properly installed
```
schillic:
```suggestion
This function requires that you have properly installed
```

mforets:
thanks for the review. yes, a test would be great but im not sure that is trivial, because MiniQhull requires the C library (Qhull) to be available in the user's machine. on the other hand if you see here https://github.com/gridap/MiniQhull.jl/blob/master/.travis.yml#L14 they install in travis such library. so i can try to add a test, preferable on another pr.

Issue #1175, opened 2019-02-28
Zonotope overapproximation of zonotope-polyhedron intersection


schillic:

mforets:
somehow related reference: http://www-ljk.imag.fr/membres/Antoine.Girard/Publications/hscc2008b.pdf

PR #2081, opened 2020-03-23
 Zonotope overapproximation of polytope given generator directions 

schillic:
See also #1175.

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -1512,3 +1512,126 @@ function overapproximate(am::AbstractAffineMap{N, <:AbstractHyperrectangle{N}},
     radius_MX = abs.(M) * radius_hyperrectangle(X)
     return Hyperrectangle(center_MXv, radius_MX)
 end
+
+"""
+    overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
+                    dir::AbstractDirections{N};
+                    solver=default_lp_solver(N)) where {N<:Real}
+
+Overapproximate a polytopic set with a zonotope of minimal total generator sum
+using only generators in the given directions.
+
+### Input
+
+- `X`        -- polytopic set
+- `Zonotope` -- type for dispatch
+- `dir`      -- directions used for the generators
+- `solver`   -- (optional, default: `default_lp_solver(N)`) the backend used to
+                solve the linear program
+
+### Output
+
+A zonotope that overapproximates `X` and uses at most the directions provided in
+`dir` (redundant directions will be ignored).
+Under this constraint, the zonotope has the minimal sum of generator vectors.
+
+### Notes
+
+The algorithm only requires one representative of each generator direction and
+their additive inverse (e.g. only one of `[1, 0]` and `[-1, 0]`) and assumes
+that the directions are normalized.
+We preprocess the directions in that respect.
+
+### Algorithm
+
+We solve a linear program parametric in the vertices ``v_j`` of `X` and the
+directions ``d_k`` in `dir` presented in [1, Section 4.2] (adapting the notation
+to the one used in this library).
+
+```math
+    \\min \\sum_{k=1}^l α_k \\
+    s.t. \\
+    c + \\sum_{k=1}^l b_{kj} * d_k = v_j \\quad \\forall j \\
+    -α_k ≤ b_{kj} ≤ α_k \\quad \\forall k, j \\
+    α_k ≥ 0 \\quad \\forall k
+```
+
+The resulting zonotope has center `c` and generators `α_k · d_k`.
+
+Note that the first type of side constraints is vector-based and that the
+nonnegativity constraints (last type) are not stated explicitly in [1].
+
+[1] Zonotopes as bounding volumes, L. J. Guibas et al, Proc. of Symposium on
+    Discrete Algorithms, pp. 803-812.
+"""
+function overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
+                         dir::AbstractDirections{N};
+                         solver=default_lp_solver(N)) where {N<:Real}
+    # TODO "normalization" here involves two steps: removing opposite directions
+    # and normalizing the direction vector
+    # for the latter we can use the normalization information from dispatch on
+    # the directions (some typical such directions are normalized by definition)
+    function normalize_directions(dir::AD) where {AD}
+        dirs = Vector{Vector{N}}()
+        for d in dir
+            d_normalized = normalize(d)
+            if (d_normalized ∉ dirs) && (-d_normalized ∉ dirs)
+                push!(dirs, d_normalized)
+            end
+        end
+        return dirs
+    end
+
+    n = dim(X)
+    dirs = normalize_directions(dir)
+    l = length(dirs)
+    V = vertices_list(X)
+    m = length(V)
+    nvariables = l + n + l * m  # l 'α_k' + n 'p[i]' + lm 'b_lj'
+    nconstraints = n * m + 2 * l * m  # nm 'p_j' + 2lm 'b_kj'
+
+    obj = vcat(ones(l), zeros(nvariables - l))
+    A = spzeros(N, nconstraints, nvariables)
+    sense = vcat(fill('=', n * m), fill('<', 2 * l * m))
+    b = zeros(N, nconstraints)
+    r = 1
+    # constraints p[i] + \sum_k v_j*b_kj = p_j
+    col_offset_p = l
+    col_offset_b = l + n + 1
+    for (j, vj) in enumerate(V)
+        for i in 1:n
+            A[r, col_offset_p + i] = N(1)  # p[i]
+            for (k, dk) in enumerate(dirs)
+                A[r, col_offset_b + (k - 1) * m] = dk[i]
+            end
+            b[r] = vj[i]
+            r += 1
+        end
+        col_offset_b += 1
+    end
+    # constraints -α_k ± b_kj <= 0
+    col_offset_b = l + n
+    for k in 1:l
+        for j in 1:m
+            for sign_b in N[1, -1]
+                A[r, k] = N(-1)  # - α_k
+                A[r, col_offset_b + j] = sign_b  # ± b_kj
+                r += 1
+            end
+        end
+        col_offset_b += m
+    end
+    lower_bounds = vcat(zeros(N, l), fill(N(-Inf), nvariables - l))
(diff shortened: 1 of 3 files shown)


src/Approximations/overapproximate.jl
```diff
@@ -1510,3 +1510,121 @@ function overapproximate(am::AbstractAffineMap{N, <:AbstractHyperrectangle{N}},
...
+nonnegativity constraints (last type) are not stated explicitly in [1].
+
+[1] Zonotopes as bounding volumes, L. J. Guibas et al, Proc. of Symposium on
+    Discrete Algorithms, pp. 803-812.
+"""
+# note: this set can generally be reduced
+function overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
+                         dir::AbstractDirections{N};
+                         solver=default_lp_solver(N)) where {N<:Real}
+    function normalize_directions(dir::AD) where {AD}
```
mforets: seems like abstract directions are not concretely typed. although that is somehow unrelated to this PR, it would be nice to have them `AbstractDirections{N, VN}`  so we can dispatch on `VN`.

src/Approximations/overapproximate.jl
```diff
@@ -1510,3 +1510,121 @@ function overapproximate(am::AbstractAffineMap{N, <:AbstractHyperrectangle{N}},
...
+            d_normalized = normalize(d)
+            if (d_normalized ∉ dirs) && (-d_normalized ∉ dirs)
+                push!(dirs, d_normalized)
+            end
+        end
+        return dirs
+    end
+
+    n = dim(X)
+    dirs = normalize_directions(dir)
```
mforets: this normalization should be optional, thinking that as use case for this function is to fix a template, normalize it (outside the function) and call `overapproximate` many times.

src/Approximations/overapproximate.jl
```diff
@@ -1510,3 +1510,121 @@ function overapproximate(am::AbstractAffineMap{N, <:AbstractHyperrectangle{N}},
...
+### Algorithm
+
+We solve a linear program parametric in the vertices ``v_j`` of `X` and the
+directions ``d_k`` in `dir` presented in [1, Section 4.2] (adapting the notation
+to the one used in this library).
+
+```math
+    \\min \\sum_{k=1}^l c_k \\
+    s.t. \\
+    \\text{center} + \\sum_{k=1}^l b_{kj} * d_k = v_j \\quad \\forall j \\
```
mforets: can you use `c` for `\\text{center}` , in either case  define it below ?

src/Approximations/overapproximate.jl
```diff
@@ -1510,3 +1510,121 @@ function overapproximate(am::AbstractAffineMap{N, <:AbstractHyperrectangle{N}},
...
+        end
+        return dirs
+    end
+
+    n = dim(X)
+    dirs = normalize_directions(dir)
+    l = length(dirs)
+    V = vertices_list(X)
+    m = length(V)
+    nvariables = l + n + l * m  # l 'c_k' + n 'p[i]' + lm 'b_lj'
```
mforets: what is `p[i]`? the coordinates of the new center?

src/LazySets.jl
```diff
@@ -25,6 +25,40 @@ import .Assertions: activate_assertions, deactivate_assertions
...
+
+If the package is loaded, this function has no effect.
+Otherwise it prints an error message.
+
+### Algorithm
+
+This function uses `@assert` and hence loses its ability to print an error
+message if assertions are deactivated.
+"""
+function require(package::Symbol; fun_name::String="", explanation::String="")
```
mforets: is this change related to the PR? also, this check can also be done with a macro, an implementation is https://github.com/mforets/ReachabilityAnalysis.jl/blob/master/src/Initialization/init.jl#L62

src/Approximations/overapproximate.jl
```diff
@@ -1510,3 +1510,121 @@ function overapproximate(am::AbstractAffineMap{N, <:AbstractHyperrectangle{N}},
...
+We solve a linear program parametric in the vertices ``v_j`` of `X` and the
+directions ``d_k`` in `dir` presented in [1, Section 4.2] (adapting the notation
+to the one used in this library).
+
+```math
+    \\min \\sum_{k=1}^l c_k \\
+    s.t. \\
+    \\text{center} + \\sum_{k=1}^l b_{kj} * d_k = v_j \\quad \\forall j \\
+    -c_k ≤ b_{kj} ≤ c_k \\quad \\forall k, j \\
+    c_k ≥ 0 \\quad \\forall k, j \\
```
mforets: not sure that i understand the notation in lines 1553 and 1554, because line 1554 says "for all j" but `c_k` has only 1 index (k)

src/Approximations/overapproximate.jl
mforets: since some of `AbstractDirections` are normalized by construction, we can create an `IsNormalized` trait and use it.

schillic: Normalization here implies that you do not have the opposite direction. All our `AbstractDirections` contain opposite direction.

src/Approximations/overapproximate.jl
schillic: Yes. This is the notation from the paper.

src/LazySets.jl
schillic: In the end it was not needed but if we want to use `require` in other modules then this is necessary, so I kept it.

src/Approximations/overapproximate.jl
schillic: Line 1554 is a copy-paste error.

src/Approximations/overapproximate.jl
schillic: `c` conflicts with `c_k`.

src/Approximations/overapproximate.jl
schillic: Currently there is no implementation satisfying the normalization. We can do that later.

src/Approximations/overapproximate.jl
schillic: I renamed `c_k` to `α_k`.

src/Approximations/overapproximate.jl
mforets: yes; by construction we know which are opposite, we can use that info.

src/Approximations/overapproximate.jl
mforets: you can at least save the step `d_normalized = normalize(d)`

src/LazySets.jl
mforets: do you mean other submodules of LazySets? can't you just import the function in submodules?

src/Approximations/overapproximate.jl
```diff
@@ -1510,3 +1510,123 @@ function overapproximate(am::AbstractAffineMap{N, <:AbstractHyperrectangle{N}},
...
+            for sign_b in N[1, -1]
+                A[r, k] = N(-1)  # - α_k
+                A[r, col_offset_b + j] = sign_b  # ± b_kj
+                r += 1
+            end
+        end
+        col_offset_b += m
+    end
+    lower_bounds = vcat(zeros(N, l), fill(N(-Inf), nvariables - l))
+    lp = linprog(obj, A, sense, b, lower_bounds, Inf, solver)
```
mforets: for the problem formulation, i would prefer that we use the JuMP interface which is much less verbose (dependency which is anyways loaded)

src/Approximations/overapproximate.jl
schillic: But a `IsNormalized` trait would still return `false` in all cases.

src/Approximations/overapproximate.jl
schillic: I tried that first and spent quite some time but could not get it to work. All other LPs in the library also do not use `JuMP` so I do not see why this is necessary now.

src/Approximations/overapproximate.jl
schillic: Only for box directions and this is a cheap operation so I would prefer to not complicate the code with conditions for one type.

src/LazySets.jl
schillic: I did, but the function must be exported before it can be imported. But we load `Approximations.jl` before the function was defined in `init.jl`.

https://github.com/JuliaReach/LazySets.jl/blob/894f77bd2309a80e02d5be50acfb6dbc7c617947/src/LazySets.jl#L137-L157

src/Approximations/overapproximate.jl
mforets: > Only for box directions

why? polar and spherical directions also have (2-norm) one. and custom directions may have it if one defines them.

> complicate the code with conditions for one type

hmm? we can dispatch on the directions type / the trait.

src/LazySets.jl
mforets: > I did, but the function must be exported before it can be imported.

not sure that i understood. why do you want to `import`? but we can do `using LazySets: foo` inside the `Approximations` module as for other functions (eg. `_rtol`).

schillic: By *imported* I meant `using` of course.

> but we can do using LazySets: foo inside the Approximations module

Yes, but only after moving the function upward.

src/Approximations/overapproximate.jl
```diff
@@ -1510,3 +1510,123 @@ function overapproximate(am::AbstractAffineMap{N, <:AbstractHyperrectangle{N}},
...
+
+The resulting zonotope has center `c` and generators `α_k · d_k`.
+
+Note that the first type of side constraints is vector-based and that the
+nonnegativity constraints (last type) are not stated explicitly in [1].
+
+[1] Zonotopes as bounding volumes, L. J. Guibas et al, Proc. of Symposium on
+    Discrete Algorithms, pp. 803-812.
+"""
+# note: this set can generally be reduced
```
mforets, line 1565: do you mean that it can be reduced without (or with a small) error?

schillic: Huh, I do not understand this comment and it does not make any sense at the top of the function. I will remove it :sweat_smile:

src/Approximations/overapproximate.jl
schillic: I added a TODO in the code.

src/LazySets.jl
schillic: I removed this change now as it was not needed.

schillic:
I had to remove zero generators in the test. Should we do that by default in the function?

PR #2084, opened 2020-03-25
Add Julia v1.4 to CI

mforets:


mforets:
@schillic  maybe this is a good moment to add v1.4 to Travis CI ?

schillic:
Sure, that is what this PR is trying to do, right? If it works I give you :+1:

.appveyor.yml
```diff
@@ -1,6 +1,6 @@
 environment:
   matrix:
-  - julia_version: 1.3
+  - julia_version: 1.4
```
mforets, line 3: @schillic this change is arbitrary; if you prefer we can keep both v1.3 and v.1.4

schillic: I would always build on the newest version. So I agree.

.travis.yml
```diff
@@ -9,6 +9,7 @@ os:
 julia:
   - 1.0
   - 1.3
```
schillic, line 11:
```suggestion
```
to save build resources for the whole organization...
But we would then also need to change the docs build to v1.4

schillic:
The doctests fail. Apparently the types of `Vector`s are not printed anymore. Note that we run doctests in the normal test suite and hence we cannot generally support different Julia versions. I think the best way to resolve this is to not print these types in doctests at all.

mforets:
true, now i remember having seen this kind of error when i built docs locally with julia v1.4 here i would just revert and build docs with v1.3

schillic:
This change will not be enough. You have to adapt the doctests.

mforets:
i thought that with this change the doctests will still be done in the julia version in which the docs are deployed (which is still v1.3). if that is not the case, i think that is what makes sense..

schillic:
Right, we run doctests in the normal test suite.

mforets:
i agree that if we insist on doctests as a fundamental part of the test suite then we should test with all julia versions, but then we have to adapt many doctests for changes that are "non-breaking" for julia base such as changing in display format. this happened in the 0.6 to 1.0 transition too. we can otherwise make the doctests to not be required to each build.

schillic:
> if we insist on doctests as a fundamental part of the test suite

We test several features only in doctests. I would prefer to get informed about broken features. Also, the reported code coverage would decrease if we excluded doctests.

Given the rather fundamental change (types of `Vector`s), I agree that we need to find a feasible solution because we often want to print `Vector`s as results. I can try to remove the outputs and if it turns out to be too much we can run the doctests only in the build that we use for creating the docs. But then we should henceforth make new doctests redundant by adding similar tests in the normal test suite. This is anyway a good idea.

PR #2088, opened 2020-03-30
#2087 - Updates to template directions

### LINKED ISSUE #2051: Narrow element type in abstract directions
```julia
julia> typeof(first(BoxDirections(2)))
LazySets.Arrays.SingleEntryVector{Float64}

julia> eltype(BoxDirections(2)) # should be SingleEntryVector ! 
AbstractArray{Float64,1}
```
Same applies for other template directions types.

### LINKED ISSUE #2087: Updates to AbstractDirections
- Add `isnormalized`
- Add type parameter to `AbstractDirections{N}` -> `AbstractDirections{N, VN}`
-  ...

mforets:
Closes #2087.
Closes #2051.

This change is breaking.

---

Remains:

- [x] fix doctests
- [x] fix tests
- [x] fix/add some examples in the template directions docstrings

### DIFF
--- a/src/Approximations/template_directions.jl
+++ b/src/Approximations/template_directions.jl
@@ -251,7 +251,7 @@ By default, each direction is represented in this iterator as a sparse vector:
 julia> eltype(dirs)
 SparseArrays.SparseVector{Float64,Int64}
 ```
-In two dimensions, the directions defined by `BoxDiagDirections` are normal to
+In two dimensions, the directions defined by `OctDirections` are normal to
 the facets of an octagon.
 
 ```jldoctest dirs_Oct
--- a/src/Approximations/template_directions.jl
+++ b/src/Approximations/template_directions.jl
@@ -76,7 +76,9 @@ Returns whether the given template directions is normalized with respect to the
 
 `true` if the 2-norm of each element in `ad` is one and `false` otherwise.
 """
-function isnormalized(ad::AbstractDirections) end
+function isnormalized(ad::AbstractDirections)
+    return false
+end
 
 # ==================================================
 # Box directions
--- a/src/Approximations/template_directions.jl
+++ b/src/Approximations/template_directions.jl
@@ -2,7 +2,7 @@ import LazySets: dim, inner
 using LazySets: isapproxzero
 
 """
-    AbstractDirections{N}
+    AbstractDirections{N, VN}
 
 Abstract type for template direction representations.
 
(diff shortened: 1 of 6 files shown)


src/Approximations/template_directions.jl
```diff
@@ -59,45 +62,97 @@ function isbounding(ad::AbstractDirections)
...
+
+### Input
+
+- `ad` -- template directions
+
+### Output
+
+`true` if the 2-norm of each element in `ad` is one and `false` otherwise.
+"""
+function isnormalized(ad::AbstractDirections) end
```
schillic, line 79:
```suggestion
function isnormalized(ad::AbstractDirections)
    return false
end
```
This defines the function for all subtypes.

mforets: Yes, that sets an arbitrary default value.

schillic: It is not arbitrary. It is the pessimistic case. In applications you hope for the positive case.

mforets:
this issue also closes https://github.com/JuliaReach/LazySets.jl/issues/2051

src/Approximations/template_directions.jl
```diff
@@ -268,45 +500,45 @@ Then the Cartesian components of each direction are obtained with
...
+        throw(ArgumentError("Nφ = $Nφ is invalid; it shoud be at least 1"))
+    end
+    stack = Vector{Vector{N}}(undef, Nφ)
+    φ = range(N(0), N(2*pi), length=Nφ+1)  # discretization of the polar angle
+
+    @inbounds for i in 1:Nφ  # skip last (repeated) angle
+        stack[i] = N[cos(φ[i]), sin(φ[i])]
     end
+    return PolarDirections{N, Vector{N}}(Nφ, stack)
 end
```
schillic, line 530: Is there a reason for making this an outer constructor?

src/Approximations/template_directions.jl
```diff
@@ -431,34 +669,49 @@ User-defined template directions.
...
-                ) where {N, VN<:AbstractVector{N}} =
-    CustomDirections{N, VN}(directions; n=n, isbounding=isbounding)
+function CustomDirections(directions::Vector{VN},
+                          n::Int=_determine_dimension(directions);
+                          check_boundedness::Bool=true,
+                          check_normalization::Bool=true) where {N, VN<:AbstractVector{N}}
+    bounded = check_boundedness ? _isbounding(directions) : false
+    normalized = check_normalization ? _isnormalized(directions) : false
+    return CustomDirections(directions, n, bounded, normalized)
+end
```
schillic, line 712: How do you create a template where you know that the directions are bounding?

src/Approximations/template_directions.jl
mforets: so it would be better to use StaticArrays for this template, since their size is known from the type (same for SphericalDirections). we can add one `PolarDirections{N, SVector{...}` in the future.

src/Approximations/template_directions.jl
mforets: i would in that case just use the type's fields, `CustomDirections(directions, n, bounded, normalized)`, since that's anyways a rather "advanced" use case.

mforets:
Thanks for the review.

PR #2094, opened 2020-04-01
WIP #2093 - Missing function in AbstractHyperrectangle notes

### LINKED ISSUE #2093: Missing function in AbstractHyperrectangle notes
In https://github.com/JuliaReach/LazySets.jl/blob/894f77bd2309a80e02d5be50acfb6dbc7c617947/src/Interfaces/AbstractHyperrectangle.jl#L21 the function `center` is missing; i think that one is also required.

mforets:
Closes #2093.

### DIFF
--- a/src/Interfaces/AbstractHyperrectangle.jl
+++ b/src/Interfaces/AbstractHyperrectangle.jl
@@ -19,14 +19,16 @@ Abstract type for hyperrectangular sets.
 See [`Hyperrectangle`](@ref) for a standard implementation of this interface.
 
 Hyperrectangular sets are particular cases of zonotopic sets, hence this interface
-is a subtype of the `AbstractZonotope` interface.
+is a subtype of the `AbstractZonotope` interface. As such, any concrete implementation
+of `AbstractHyperrectangle` should implement the methods required for that interface,
+see the notes in [`AbstractZonotope`] for that interface.
 
-Let `X` be a set such that `typeof(X)` is a concrete subtype of `AbstractHyperrectangle`
-and let `n` denote the dimension of `X`. Every concrete implementation of the
-`AbstractHyperrectangle` interface must define the following functions:
+In addition, every concrete implementation of the `AbstractHyperrectangle` interface
+must also define the functions described below. Let `X` be a set such that
+`typeof(X)` is a concrete subtype of `AbstractHyperrectangle` and let `n` denote
+the dimension of `X`.
 
-- `center(X)`                   -- return the center of `X`
-- `radius_hyperrectangle(X, i)` -- return the `n`-dimensional radius of `X`
+- `radius_hyperrectangle(X)` -- return the `n`-dimensional radius of `X`
 - `radius_hyperrectangle(X, i)` -- return the `i`-th component of `X`'s radius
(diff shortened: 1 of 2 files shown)


src/Interfaces/AbstractHyperrectangle.jl
```diff
@@ -18,13 +18,18 @@ Abstract type for hyperrectangular sets.
...
 See [`Hyperrectangle`](@ref) for a standard implementation of this interface.
 
-Every concrete `AbstractHyperrectangle` must define the following functions:
-- `radius_hyperrectangle(::AbstractHyperrectangle{N})` -- return the
-    hyperrectangle's radius, which is a full-dimensional vector
-- `radius_hyperrectangle(::AbstractHyperrectangle{N}, i::Int)` -- return the
-    hyperrectangle's radius in the `i`-th dimension
-- `isflat(::AbstractHyperrectangle{N})` -- determine whether the
-    hyperrectangle's radius is zero in some dimension
+Hyperrectangular sets are paricular cases of zonotopic sets, hence this interface
```
mforets, line 21:
```suggestion
Hyperrectangular sets are particular cases of zonotopic sets, hence this interface
```

schillic:
`center` is inherited from the parent interfaces (`AbstractZonotope`, which again interits it from `AbstractCentrallySymmetricPolytope`). We never mentioned these again recursively (for instance it is also not listed in `AbstractZonotope`).

mforets:
Those statements are correct, but it doesn't mean we can improve the docs. I see the purpose of this docs differently than you, it seems. If user X wants to write a new subtype of `AbstractHyperrectangle`, needs to implement `center`, so if this information is here, that is useful. We can also comment (i'll add it) that there are some functions which are also required by the supertype, does that make sense to you?

mforets:
Actually, in this PR i added the sentence,

> Hyperrectangular sets are particular cases of zonotopic sets, hence this interface is a subtype of the `AbstractZonotope` interface.

schillic:
I am generally fine with adding the methods of the parent interfaces. I just wanted to point out why it was missing and that this PR now makes the documentation inconsistent. So if you want to add this information, it should be added everywhere.

mforets:
Actually, it is not the case that all methods of the parent interfaces should be implemented. See

```julia
Every concrete `AbstractZonotope` must define the following functions:
- `genmat(::AbstractZonotope{N})` -- return the generator matrix
- `generators(::AbstractZonotope{N})` -- return an iterator over the generators
Since the functions `genmat` and `generators` can be defined in terms of each
other, it is sufficient to only genuinely implement one of them and let the
implementation of the other function call the fallback implementation
`genmat_fallback` resp. `generators_fallback`.
```

and

```julia
Every concrete `AbstractCentrallySymmetricPolytope` must define the following
functions:
- from `AbstractCentrallySymmetric`:
  - `center(::AbstractCentrallySymmetricPolytope{N})` -- return the
     center point
- from `AbstractPolytope`:
  - `vertices_list(::AbstractCentrallySymmetricPolytope{N})`
     -- return a list of all vertices
```
for example, `vertices_list` is not needed for `AbstractHyperrectangle`.

Issue #1983, opened 2020-02-17
Generic vertices list for hyperrectangular sets fails with static vectors


mforets:
```julia
julia> using LazySets, StaticArrays

julia> A = Hyperrectangle(low=SVector(0.0, 0.0), high=SVector(1.0, 1.0))
Hyperrectangle{Float64,SArray{Tuple{2},Float64,1,2},SArray{Tuple{2},Float64,1,2}}([0.5, 0.5], [0.5, 0.5])

julia> vertices_list(A)
setindex!(::SArray{Tuple{2},Float64,1,2}, value, ::Int) is not defined.

Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] setindex!(::SArray{Tuple{2},Float64,1,2}, ::Float64, ::Int64) at /home/mforets/.julia/packages/StaticArrays/1g9bq/src/indexing.jl:3
 [3] vertices_list(::Hyperrectangle{Float64,SArray{Tuple{2},Float64,1,2},SArray{Tuple{2},Float64,1,2}}) at /home/mforets/.julia/dev/LazySets/src/Interfaces/AbstractHyperrectangle.jl:197
 [4] top-level scope at In[51]:1
```

schillic:
The problem is that we copy the center and modify the copy. In addition we should convert the copy to a `Vector`.

https://github.com/JuliaReach/LazySets.jl/blob/c61beeacc7ad7eae8634a03f8e2fb7aef65bb8c9/src/Interfaces/AbstractHyperrectangle.jl#L182-L201

mforets:
> In addition we should convert the copy to a Vector.

I didn't check it yet, but that sounds like it would return `Vectors`? The `vertices_list` shouldn't change the vector representation.

PR #2095, opened 2020-04-02
Add explicit conversion in generic vertices list of hyperrectangular set 

mforets:
Related to https://github.com/JuliaReach/LazySets.jl/issues/1983. It doesn't close it, because the type of the vertices can (and should) be preserved. I'll address that on a separate PR.

### DIFF
--- a/src/Interfaces/AbstractHyperrectangle.jl
+++ b/src/Interfaces/AbstractHyperrectangle.jl
@@ -189,7 +189,7 @@ function vertices_list(H::AbstractHyperrectangle{N}) where {N<:Real}
     trivector = Vector{Int8}(undef, n)
     m = 1
     c = center(H)
-    v = copy(c)
+    v = convert(Vector{N}, copy(c))
     @inbounds for i in 1:n
         ri = radius_hyperrectangle(H, i)
         if iszero(ri)


src/Interfaces/AbstractHyperrectangle.jl
```diff
@@ -189,7 +189,7 @@ function vertices_list(H::AbstractHyperrectangle{N}) where {N<:Real}
     trivector = Vector{Int8}(undef, n)
     m = 1
     c = center(H)
-    v = copy(c)
+    v = convert(Vector{N}, copy(c))
```
schillic, line 192:
```suggestion
    VN = typeof(radius_hyperrectangle(H, 1))
    v = convert(VN, copy(c))
```

As an example why this is better, we may change the code below to return a sparse vector in the future.
https://github.com/JuliaReach/LazySets.jl/blob/894f77bd2309a80e02d5be50acfb6dbc7c617947/src/Interfaces/AbstractSingleton.jl#L67-L69

mforets: Hmm, I disagree. This is a generic fallback and `AbstractHyperrectangle{N}` does not have the radius/center as type parameters that we can readily use. Moreover the algorithm for sparse would be another one, see lines 192-200 and 209-221, we should iterate and fill the vectors differently, etc.

mforets: in other words, what i meant to say is that the implementation for sparse vectors requires other more fundamental changes in this function.

PR #2099, opened 2020-04-06
#2070 - Scaling from the right

### LINKED ISSUE #2070: Scaling from the right
```julia
julia> 1.0 * rand(Hyperrectangle)

LinearMap{Float64,Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}},Float64,SparseArrays.SparseMatrixCSC{Float64,Int64}}(
  [1, 1]  =  1.0
  [2, 2]  =  1.0, Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([-0.4431025443793669, 1.9758438962291769], [0.9698823931323284, 0.25346839195921167]))
 * 1.0
1

julia> rand(Hyperrectangle) * 1.0
MethodError: no method matching *(::Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}, ::Float64)
Closest candidates are:
  *(::Any, ::Any, !Matched::Any, !Matched::Any...) at operators.jl:529
  *(!Matched::Bool, ::T) where T<:AbstractFloat at bool.jl:110
  *(!Matched::Float64, ::Float64) at float.jl:405
  ...

Stacktrace:
 [1] top-level scope at In[57]:1
```

mforets:
Closes #2070.

I left the type annotation `map::Real` as in the rest of the file, although i don't know why we don't allow any scalar-like object ( = `Number`) in the `LinearMap` type. .. (it is probably for historical reasons :) )

### DIFF
--- a/src/LazyOperations/LinearMap.jl
+++ b/src/LazyOperations/LinearMap.jl
@@ -131,6 +131,11 @@ function *(map::Union{AbstractMatrix, UniformScaling, AbstractVector, Real}, X::
     return LinearMap(map, X)
 end
 
+# scaling from the right
+function *(X::LazySet, map::Real)
+    return LinearMap(map, X)
+end
+
 # convenience constructor from a vector
 function LinearMap(v::AbstractVector, X::LazySet)
     n = dim(X)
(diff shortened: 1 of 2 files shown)


mforets:
:) i don't see anything fundamental in constraining `<:Real` in the linear map definition (instead of `<:Number`) in consequence, in the methods defined by that type. the reason that this is using `<:Number` is purely historical.

PR #2100, opened 2020-04-06
#1578 - Center along a given dimension of AbstractHyperrectangle

### LINKED ISSUE #1578: Center along a given dimension of AbstractHyperrectangle
Similar to `radius_hyperrectangle(H::AbstractHyperrectangle, i::Int)` we could add `center(H::AbstractHyperrectangle, i::Int)` which for the usual cases is equivalent to `center(H)[i]` since the center of the hyperrectangle is actually stored, but there are other cases where the center has to be computed. The `center(H, i)` function is useful for early termination in some checks (see eg. membership test, https://github.com/JuliaReach/LazySets.jl/pull/1577).

SebastianGuadalupe:
Closes #1578.

### DIFF
--- a/src/Interfaces/AbstractHyperrectangle.jl
+++ b/src/Interfaces/AbstractHyperrectangle.jl
@@ -617,3 +617,21 @@ function volume(H::AbstractHyperrectangle{N}) where {N<:Real}
     vol = mapreduce(x -> 2x, *, radius_hyperrectangle(H))
     return vol
 end
+
+"""
+    center(H::AbstractHyperrectangle{N}, i::Int) where {N<:Real}
+
+Return the center along a given dimension of a hyperrectangular set.
+
+### Input
+
+- `H` -- hyperrectangular set
+- `i` -- dimension of interest
+
+### Output
+
+The center along a given dimension of the hyperrectangular set.
+"""
+function center(H::AbstractHyperrectangle{N}, i::Int) where {N<:Real}
+    return H.center[i]
+end
--- a/src/Interfaces/AbstractHyperrectangle.jl
+++ b/src/Interfaces/AbstractHyperrectangle.jl
@@ -632,6 +632,6 @@ Return the center along a given dimension of a hyperrectangular set.
 
 The center along a given dimension of the hyperrectangular set.
 """
-function center(H::AbstractHyperrectangle{N}, i::Int) where {N<:Real}
+@inline function center(H::AbstractHyperrectangle{N}, i::Int) where {N<:Real}
     return center(H)[i]
 end
--- a/src/Interfaces/AbstractHyperrectangle.jl
+++ b/src/Interfaces/AbstractHyperrectangle.jl
@@ -633,5 +633,5 @@ Return the center along a given dimension of a hyperrectangular set.
 The center along a given dimension of the hyperrectangular set.
 """
 function center(H::AbstractHyperrectangle{N}, i::Int) where {N<:Real}
-    return H.center[i]
+    return center(H)[i]
 end
--- a/src/Interfaces/AbstractHyperrectangle.jl
+++ b/src/Interfaces/AbstractHyperrectangle.jl
@@ -617,21 +617,3 @@ function volume(H::AbstractHyperrectangle{N}) where {N<:Real}
     vol = mapreduce(x -> 2x, *, radius_hyperrectangle(H))
     return vol
 end
-
-"""
-    center(H::AbstractHyperrectangle{N}, i::Int) where {N<:Real}
-
-Return the center along a given dimension of a hyperrectangular set.
-
-### Input
-
-- `H` -- hyperrectangular set
-- `i` -- dimension of interest
-
-### Output
-
-The center along a given dimension of the hyperrectangular set.
-"""
-@inline function center(H::AbstractHyperrectangle{N}, i::Int) where {N<:Real}
-    return center(H)[i]
-end
--- a/src/LazyOperations/SymmetricIntervalHull.jl
+++ b/src/LazyOperations/SymmetricIntervalHull.jl
@@ -147,8 +147,9 @@ Return the center of a symmetric interval hull of a convex set.
 
 The origin.
 """
-function center(sih::SymmetricIntervalHull{N}) where {N<:Real}
-    return zeros(N, dim(sih))
+@inline function center(sih::SymmetricIntervalHull{N}, i::Int) where {N<:Real}
+    @boundscheck _check_bounds(sih, i)
+    return zero(N)
 end
 
 
--- a/src/LazyOperations/SymmetricIntervalHull.jl
+++ b/src/LazyOperations/SymmetricIntervalHull.jl
@@ -126,7 +126,8 @@ Return the center along a given dimension of a symmetric interval hull of a conv
 
 The center along a given dimension of the symmetric interval hull of a convex set.
 """
-function center(sih::SymmetricIntervalHull{N}, i::Int) where {N<:Real}
+@inline function center(sih::SymmetricIntervalHull{N}, i::Int) where {N<:Real}
+    @boundscheck _check_bounds(sih, i)
     return zero(N)
 end
 
@@ -147,9 +148,8 @@ Return the center of a symmetric interval hull of a convex set.
 
 The origin.
 """
-@inline function center(sih::SymmetricIntervalHull{N}, i::Int) where {N<:Real}
-    @boundscheck _check_bounds(sih, i)
-    return zero(N)
+function center(sih::SymmetricIntervalHull{N}) where {N<:Real}
+    return zeros(N, dim(sih))
 end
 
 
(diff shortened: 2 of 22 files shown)


src/Interfaces/AbstractHyperrectangle.jl
```diff
@@ -617,3 +617,21 @@ function volume(H::AbstractHyperrectangle{N}) where {N<:Real}
...
+
+- `H` -- hyperrectangular set
+- `i` -- dimension of interest
+
+### Output
+
+The center along a given dimension of the hyperrectangular set.
+"""
+function center(H::AbstractHyperrectangle{N}, i::Int) where {N<:Real}
+    return H.center[i]
```
mforets, line 636: This doesn't work, because one can't guarantee that *any* implementation of an `AbstractHyperrectangle` contains a field named `center`. For example, think about an `Interval`, which does not have a `center` field, although one can construct its center by calling the `center` function.

mforets: Since `center` is a required field for any implementation of the `AbstractHyperrectangle` interface, you can fix this by returning `center(H)[i]`.

src/LazyOperations/SymmetricIntervalHull.jl
```diff
@@ -112,6 +112,24 @@ function radius_hyperrectangle(sih::SymmetricIntervalHull{N}) where {N<:Real}
...
+
+- `sih` -- symmetric interval hull of a convex set
+- `i` -- dimension of interest
+
+### Output
+
+The center along a given dimension of the symmetric interval hull of a convex set.
+"""
+function center(sih::SymmetricIntervalHull{N}, i::Int) where {N<:Real}
+    return zero(N)
```
mforets, line 130: This function, and also the one below for `ZeroSet`, would be returning `zero(N)` even if the index `i` is out of bounds ie. if it's greater than the dimension of the set. Though a user error, it doesn't harm to actually check that the index is correct and return and error otherwise. It can be fixed with a negligible tax in runtime using `@boundscheck` macro, see https://docs.julialang.org/en/v1/devdocs/boundscheck/
As explained in that docs, `@inbounds` will normally elide this check one layer further, eg. if `center(H, i)` is used in a loop that was annotated with `@inbounds`.

Then the implementation would be (note the `@inline` macro, i would suggest that we use it for the new function `center(X, i)` in other places of this PR as well):

```julia
@inline function _check_bounds(X, i)
    1 <= i <= dim(X) || throw(ArgumentError("there is no index at coordinate $i, since the set is of dimension $(dim(X))"))
end

@inline function center(sih::SymmetricIntervalHull{N}, i::Int) where {N<:Real}
    @boundscheck _check_bounds(S, i)
    return zero(N)
end
```

mforets: Since this new `_check_bounds` function is shared, we can put it in `Utils/helper_functions.jl`.

SebastianGuadalupe: should I inline every `center(X, i)` function?

mforets, line 130: we should also check bounds in this function

src/Sets/Interval.jl
```diff
@@ -586,3 +586,21 @@ The interval obtained by applying the numerical scale to the given interval.
...
+
+- `x` -- interval
+- `i` -- dimension of interest
+
+### Output
+
+The center along a given dimension of the interval.
+"""
+@inline function center(x::Interval{N}, i::Int) where {N<:Real}
+    return center(x)[i]
```
mforets, line 605: This unnecessarily creates a temporary array, i would suggest to implement it like so

```julia
@inline function center(x::Interval{N}, i::Int) where {N<:Real}
     @boundscheck i == 1
     return IntervalArithmetic.mid(x.dat)
end
```

src/Sets/Ball2.jl
```diff
@@ -279,6 +279,28 @@ function sample(B::Ball2{N, VN}, nsamples::Int=1;
 end
 
 
+# --- AbstractHyperrectangle interface functions ---
```
schillic, line 282:
```suggestion
# --- AbstractCentrallySymmetric interface functions ---
```

src/Sets/Ballp.jl
```diff
@@ -99,6 +99,28 @@ function center(B::Ballp{N}) where {N<:AbstractFloat}
 end
 
 
+# --- AbstractHyperrectangle interface functions ---
```
schillic, line 102:
```suggestion
# --- AbstractCentrallySymmetric interface functions ---
```

src/Sets/BallInf.jl
```diff
@@ -337,3 +337,21 @@ function volume(B::BallInf{N}) where {N<:Real}
...
+- `B` -- ball in the infinity norm
+- `i` -- dimension of interest
+
+### Output
+
+The center along a given dimension of the ball in the infinity norm.
+"""
+@inline function center(B::BallInf{N}, i::Int) where {N<:Real}
+    return B.center[i]
+end
```
schillic, line 357: This method is not needed as it is exactly the same as the `AbstractHyperrectangle` implementation (same for `Hyperrectangle`).

src/Sets/ZeroSet.jl
```diff
@@ -221,3 +221,26 @@ function translate(Z::ZeroSet{N}, v::AbstractVector{N}) where {N<:Real}
                                 "set by a $(length(v))-dimensional vector"
     return Singleton(v)
 end
+
+
+# --- AbstractHyperrectangle interface functions ---
```
schillic, line 226:
```suggestion
# --- AbstractCentrallySymmetric interface functions ---
```

src/Sets/BallInf.jl
mforets: Note that the lowered code is not *exactly* the same, because of the function call. But in practice it's fine to use the the interface functions in these cases where `center` is a field (same for `Singleton` types).

```julia
julia> H = rand(Hyperrectangle, dim=100);
julia> center_v1(H, i) = H.center[i]
julia> center_v2(H, i) = center(H)[i]

julia> @btime center_v1($H, 100)
  1.304 ns (0 allocations: 0 bytes)
-0.19825521890893702

@btime center_v2($H, 100)
  1.302 ns (0 allocations: 0 bytes)
-0.19825521890893702

julia> @code_lowered center_v1(H, 10)
CodeInfo(
1 ─ %1 = Base.getproperty(H, :center)
│   %2 = Base.getindex(%1, i)
└──      return %2
)

@code_lowered center_v2(H, 10)
CodeInfo(
1 ─ %1 = Main.center(H)
│   %2 = Base.getindex(%1, i)
└──      return %2
)
```

mforets:
a reminder that we also wanted to add this method in the `AbstractCentrallySymmetric` interface https://github.com/JuliaReach/LazySets.jl/issues/1578#issuecomment-609902336

src/Sets/Hyperrectangle.jl
```diff
@@ -167,6 +167,24 @@ function center(H::Hyperrectangle{N}) where {N<:Real}
...
+- `i` -- dimension of interest
+
+### Output
+
+The center along a given dimension of the hyperrectangle.
+"""
+@inline function center(H::Hyperrectangle{N}, i::Int) where {N<:Real}
+    return H.center[i]
+end
+
```
schillic, line 187:
```suggestion
```

src/Interfaces/AbstractHyperrectangle.jl
```diff
@@ -617,3 +617,21 @@ function volume(H::AbstractHyperrectangle{N}) where {N<:Real}
...
+- `H` -- hyperrectangular set
+- `i` -- dimension of interest
+
+### Output
+
+The center along a given dimension of the hyperrectangular set.
+"""
+@inline function center(H::AbstractHyperrectangle{N}, i::Int) where {N<:Real}
+    return center(H)[i]
+end
```
schillic, line 637: That sould be moved to `AbstractCentrallySymmetric`.

test/unit_SymmetricIntervalHull.jl
```diff
@@ -42,4 +42,7 @@ for N in [Float64, Rational{Int}, Float32]
     h = SymmetricIntervalHull(Singleton(N[1, 2]))
     @test high(h) == N[1, 2]
     @test low(h) == N[-1, -2]
+
+    #center
+    @test center(h) == N[0, 0]
```
schillic, line 47:
```suggestion
    # center
    @test center(h) == N[0, 0]
    @test center(h, 1) == zero(N)
```

test/unit_Interval.jl
```diff
@@ -20,6 +20,8 @@ for N in Ns
 
     @test dim(x) == 1
     @test center(x) == N[0.5]
+    @test center(x, 1) == N[0.5]
```
mforets, line 23:
```suggestion
    @test center(x, 1) == N(0.5)
```

PR #2104, opened 2020-04-08
Test for HPolyhedron's an_element method

schillic:
`an_element` was not tested.
The test fails for an `HPolytope` because the polyhedron is unbounded in the direction `[1, 0]` that we use in the implementation.
Also note that `an_element` returns `[Inf, Inf]`.
```julia
julia> N = Float64; P = HPolyhedron([HalfSpace(N[3//50, -4//10], N(1)),
                                     HalfSpace(N[-1//50, 1//10], N(-1))]);

julia> an_element(P)
2-element Array{Float64,1}:
 Inf
 Inf
```

### DIFF
--- a/test/unit_Polyhedron.jl
+++ b/test/unit_Polyhedron.jl
@@ -130,6 +130,11 @@ for N in [Float64]
     d = N[0, -1]
     @test σ(d, p) == N[0, 0]
 
+    # an_element
+    P = HPolyhedron([HalfSpace(N[3//50, -4//10], N(1)),
+                     HalfSpace(N[-1//50, 1//10], N(-1))])
+    @test an_element(P) ∈ P
+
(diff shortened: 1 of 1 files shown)


mforets:
here an element returns `[Inf, Inf]` as you pointed out -- ideally the element returned should be in R :) in this case one gets a solution solving the linear system,

```julia
A, b = tosimplehrep(P)
A \ b

2-element Array{Float64,1}:
 149.99999999999991
  19.999999999999986
```
(which is in the set) and it also works for rationals, giving
```
2-element Array{Rational{Int64},1}:
 150//1
  20//1
```

schillic:
Does this always work? Then we should change the implementation. I actually thought that we solve a feasibility LP already but we do not.

PR #2107, opened 2020-04-10
#1989 - Some linear map corner cases

### LINKED ISSUE #1989: Rectangular concrete linear map of Hyperplane/Line
`linear_map(M, P)` for `P` being a `Hyperplane` or a `Line` always creates a set of the same type. But this is not correct if `M` is not square (i.e., has different output dimension).
```
julia> M = [1.0 1.0]
1×2 Array{Float64,2}:
 1.0  1.0

julia> linear_map(M, rand(Line))  # tries to create a Line but then fails
ERROR: AssertionError: lines must be two-dimensional

julia> H = rand(Hyperplane)
Hyperplane{Float64}([-0.1514314278974606, -0.3351825271100233], 1.2796222453508572)

julia> linear_map(M, H)  # should be the one-dimensional universe
Hyperplane{Float64}([-0.24330697750374183], 1.2796222453508572)
```

mforets:
Closes #1989.

### DIFF
--- a/src/Sets/HalfSpace.jl
+++ b/src/Sets/HalfSpace.jl
@@ -487,7 +487,7 @@ function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::HalfSpace{N},
     if length(constraints) == 1
         return first(constraints)
     elseif isempty(constraints)
-        return Universe(size(M, 1))
+        return Universe{N}(size(M, 1))
     else
         return HPolyhedron(constraints)
     end
--- a/src/Sets/HalfSpace.jl
+++ b/src/Sets/HalfSpace.jl
@@ -486,6 +486,8 @@ function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::HalfSpace{N},
     constraints = _linear_map_hrep(M, P, algo)
     if length(constraints) == 1
         return first(constraints)
+    elseif isempty(constraints)
+        return Universe(size(M, 1))
     else
         return HPolyhedron(constraints)
     end
--- a/src/Sets/Hyperplane.jl
+++ b/src/Sets/Hyperplane.jl
@@ -432,9 +432,15 @@ end
 function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::Hyperplane{N},
                                  algo::AbstractLinearMapAlgorithm) where {N<:Real}
     constraints = _linear_map_hrep(M, P, algo)
-    @assert length(constraints) == 2
-    c = first(constraints)
-    return Hyperplane(c.a, c.b)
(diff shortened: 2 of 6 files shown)


src/Sets/Hyperplane.jl
```diff
@@ -432,9 +432,15 @@ end
...
     constraints = _linear_map_hrep(M, P, algo)
-    @assert length(constraints) == 2
-    c = first(constraints)
-    return Hyperplane(c.a, c.b)
+    if length(constraints) == 2
+        # assuming these constaraints define a hyperplane
+        c = first(constraints)
+        return Hyperplane(c.a, c.b)
+    elseif isempty(constraints)
+        return Universe(size(M, 1))
```
mforets, line 440:
```suggestion
        return Universe{N}(size(M, 1))
```

src/Sets/HalfSpace.jl
```diff
@@ -486,6 +486,8 @@ function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::HalfSpace{N},
     constraints = _linear_map_hrep(M, P, algo)
     if length(constraints) == 1
         return first(constraints)
+    elseif isempty(constraints)
+        return Universe(size(M, 1))
```
mforets, line 490:
```suggestion
        return Universe{N}(size(M, 1))
```

src/Sets/Line.jl
```diff
@@ -322,10 +322,16 @@ end
...
-     @assert length(constraints) == 2
-     c = first(constraints)
-     return Line(c.a, c.b)
+    constraints = _linear_map_hrep(M, P, algo)
+    if length(constraints) == 2
+        # assuming these constraints define a line
+        c = first(constraints)
+        return Line(c.a, c.b)
+    elseif isempty(constraints)
+        return Universe(size(M, 1))
```
mforets, line 331:
```suggestion
        return Universe{N}(size(M, 1))
```

src/Sets/Hyperplane.jl
```diff
@@ -432,9 +432,15 @@ end
 function _linear_map_hrep_helper(M::AbstractMatrix{N}, P::Hyperplane{N},
                                  algo::AbstractLinearMapAlgorithm) where {N<:Real}
     constraints = _linear_map_hrep(M, P, algo)
-    @assert length(constraints) == 2
-    c = first(constraints)
-    return Hyperplane(c.a, c.b)
+    if length(constraints) == 2
+        # assuming these constaraints define a hyperplane
```
schillic, line 436:
```suggestion
        # assuming these constraints define a hyperplane
```

mforets:
thanks for the review!

PR #2110, opened 2020-04-10
#1136 - Boundedness check for polyhedra using Stiemke's Theorem of alternatives

### LINKED ISSUE #1136: Boundedness check for polyhedra using Stiemke's Theorem of alternatives
See Mangasarian, Nonlinear Programming, p. 32.

Tags: unbounded polyhedron

SebastianGuadalupe:
Closes #1136.


src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -879,3 +879,42 @@ function an_element(P::AbstractPolyhedron{N}) where {N<:Real}
...
+function isbounded(P::AbstractPolyhedron{N}; solver=LazySets.default_lp_solver(N)) where {N<:Real}
+    A, b = tosimplehrep(P)
+    m, n = size(A)
+
+    if !isempty(nullspace(A))
+        return false
+    end
+
+    At = copy(transpose(A))
+    c = fill(1.0, m)
```
mforets, line 917: we can use `one(N)`, same on line 918  -- at least that's what i would expect to see, although we know that not all solvers support different `N`

mforets:
the new function looks good to me on the math side :tada: 

can you show some benchmarks for this function wrt the existing one?

we should think if we want to make this function the default for a general lazyset or not (https://github.com/JuliaReach/LazySets.jl/blob/master/src/Interfaces/LazySet.jl#L268)

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -879,3 +879,42 @@ function an_element(P::AbstractPolyhedron{N}) where {N<:Real}
...
+
+### Output
+
+`true` iff the polyhedron is bounded
+
+### Algorithm
+
+The algorithm is based on Stiemke's theorem of alternatives:
+
+Proposition 1. If ``ker(A)≠\{0\} \\implies P`` is unbounded.
```
mforets, line 902:
```suggestion
Proposition 1. If ``ker(A)≠\\{0\\} \\implies P`` is unbounded.
```

schillic:
> we should think if we want to make this function the default for a general lazyset or not

Does it apply to general sets or only to polyhedra?

mforets:
polyhedal sets; you need the simple `Ax <= b` representation. OTOH the current fallback applies to any set because it just requires support function evaluation. i would suggest that we keep things like that, perhaps adding the new algorithm as an option to `isbounded(::LazySet)`, the precondition of course being that `constraints_list` is applicable.

docs/src/lib/interfaces.md
```diff
@@ -177,6 +177,7 @@ constrained_dimensions(::AbstractPolyhedron)
 linear_map(::AbstractMatrix{N}, ::AbstractPolyhedron{N}) where {N<:Real}
 chebyshev_center(::AbstractPolyhedron{N}) where {N<:AbstractFloat}
 an_element(::AbstractPolyhedron{N}) where {N<:Real}
+isbounded(P::AbstractPolyhedron{N}) where {N}
```
mforets, line 180:
```suggestion
isbounded(::AbstractPolyhedron{N}) where {N}
```
(it is ok, though it is not needed here)

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -879,3 +879,42 @@ function an_element(P::AbstractPolyhedron{N}) where {N<:Real}
...
+function isbounded(P::AbstractPolyhedron{N}; solver=LazySets.default_lp_solver(N)) where {N<:Real}
+    A, b = tosimplehrep(P)
+    m, n = size(A)
+
+    if !isempty(nullspace(A))
+        return false
+    end
+
+    At = copy(transpose(A))
+    c = fill(1.0, m)
```
mforets, line 917:
```suggestion
    c = fill(one(N), m)
```

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -879,3 +879,42 @@ function an_element(P::AbstractPolyhedron{N}) where {N<:Real}
...
+    A, b = tosimplehrep(P)
+    m, n = size(A)
+
+    if !isempty(nullspace(A))
+        return false
+    end
+
+    At = copy(transpose(A))
+    c = fill(one(N), m)
+    lp = linprog(c, At, '=', zeros(n), 1.0, Inf, solver)
```
mforets, line 918:
```suggestion
    lp = linprog(c, At, '=', zeros(n), one(N), Inf, solver)
```

mforets:
we have to fix the dispatch too, because `isbounded(::HPolyhedron)` actually calls `isbounded_unit_dimension` and it shouldn't

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -879,3 +879,42 @@ function an_element(P::AbstractPolyhedron{N}) where {N<:Real}
...
+- `backend` -- (optional, default: `default_lp_solver(N)`) the backend used
+               to solve the linear program
+
+### Output
+
+`true` iff the polyhedron is bounded
+
+### Algorithm
+
+The algorithm is based on Stiemke's theorem of alternatives.
```
schillic, line 900: Can you give a reference?

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -879,3 +879,42 @@ function an_element(P::AbstractPolyhedron{N}) where {N<:Real}
...
+
```
schillic, line 901:
```suggestion

Let the polyhedron ``P`` be given in constraint form ``Ax ≤ b``.

```

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -879,3 +879,42 @@ function an_element(P::AbstractPolyhedron{N}) where {N<:Real}
...
+"""
+function isbounded(P::AbstractPolyhedron{N}; solver=LazySets.default_lp_solver(N)) where {N<:Real}
+    A, b = tosimplehrep(P)
+    m, n = size(A)
+
+    if !isempty(nullspace(A))
+        return false
+    end
+
+    At = copy(transpose(A))
```
schillic, line 916: Why the `copy`?

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -879,3 +879,42 @@ function an_element(P::AbstractPolyhedron{N}) where {N<:Real}
...
+
+`true` iff the polyhedron is bounded
+
+### Algorithm
+
+The algorithm is based on Stiemke's theorem of alternatives.
+
+Proposition 1. If ``ker(A)≠\\{0\\}``, then ``P`` is unbounded.
+
+Proposition 2. Assume that ``ker(A)={0}`` and ``P`` is non-empty.
```
schillic, line 904: In the code you do not check for emptiness. Does the code also work if `P` is empty? Maybe add such a test.

src/Interfaces/AbstractPolyhedron_functions.jl
mforets: we used it because `transpoe` is lazy

src/Interfaces/AbstractPolyhedron_functions.jl
mforets: the algorithm can be seen as a corollary on the theorem of alternatives (see [Farkas lemma](https://en.wikipedia.org/wiki/Farkas%27_lemma), though i don't have a specific reference for it. about the particular fm of the theorem of alternatives, there is 

Mangasarian, Olvi L. Nonlinear programming. Society for Industrial and Applied Mathematics, 1994.

src/Interfaces/AbstractPolyhedron_functions.jl
schillic: But isn't it lazy for a reason? I would expect that this is counterproductive. But okay...

src/Interfaces/LazySet.jl
```diff
@@ -300,6 +309,53 @@ function isbounded_unit_dimensions(S::LazySet{N}) where {N<:Real}
...
+
+### Output
+
+`true` iff the polyhedron is bounded
+
+### Algorithm
+
+We first check if the polyhedron has more than `max(dim(P), 1)` constraints,
+which is a necessary condition for boundedness.
+
```
schillic, line 331:
```suggestion
```

src/Interfaces/AbstractPolyhedron_functions.jl
mforets: those are most often used for matrix factorizations. i have checked in this case and there is no difference.

src/Interfaces/AbstractPolyhedron_functions.jl
mforets: the code doesn't check if the input is an empty set. i checked and the answer is the same as for `isbounded_unit_dimensions` (= false).

mforets: i will add this assumption to the docs nevertheless

schillic: I hope you mean `true`!
```julia
julia> isbounded_unit_dimensions(HPolyhedron([HalfSpace([1.], 0.), HalfSpace([-1.], 1.)]))
true
```

mforets: yes. though for some cases it gives an error and Stiemke gives "false"

```julia
p = HPolyhedron([HalfSpace(x <= 1, var), HalfSpace(x >= 2, var)])

__isbounded_unit_dimensions(p)
the support vector is undefined because the polyhedron is empty

__isbounded_stiemke(p)
false
```
it is tempting to make the emptiness check in the function, but the cost is relatively high,

```julia
p = convert(HPolyhedron, rand(Hyperrectangle, dim=10));
@btime isempty($p)
@btime __isbounded_stiemke($p)

  22.653 μs (79 allocations: 7.45 KiB)
  30.743 μs (100 allocations: 20.56 KiB)
true
```

mforets: we can also let the check be optional, defaulting to `true`. what are your thoughts?

schillic: Yes, getting a wrong answer by default doesn't sound good.

mforets: i had changed the docstring to state that the condition has to be checked by the user.

test/unit_Polyhedron.jl
```diff
@@ -42,6 +42,8 @@ for N in [Float64, Rational{Int}, Float32]
 
     # boundedness
     @test !isbounded(p_univ)
+    @test isbounded(p)
+    @test !isbounded(HPolyhedron([HalfSpace(N[1.0, 0.0], N(1))]))
```
schillic, line 48:
```suggestion
    @test !isbounded(HPolyhedron([HalfSpace(N[1, 0], N(1))]))
```

src/Interfaces/LazySet.jl
```diff
@@ -304,6 +313,64 @@ function isbounded_unit_dimensions(S::LazySet{N}) where {N<:Real}
...
+    A, b = tosimplehrep(P)
+    m, n = size(A)
+
+    if !isempty(nullspace(A))
+        return false
+    end
+
+    At = copy(transpose(A))
+    c = fill(one(N), m)
```
schillic, line 369:
```suggestion
    c = ones(N, m)
```

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -903,6 +903,35 @@ function an_element(P::AbstractPolyhedron{N}) where {N<:Real}
...
+
+### Algorithm
+
+We first check if the polyhedron has more than `max(dim(P), 1)` constraints,
+which is a necessary condition for boundedness.
+
+If so, we check boundedness via [`_isbounded_stiemke`](@ref).
+"""
+function isbounded(P::AbstractPolyhedron{N}; solver=LazySets.default_lp_solver(N)) where {N<:Real}
+    if length(P.constraints) <= max(dim(P), 1)
```
schillic, line 929: Not every `AbstractPolyhedron` has a field `constraints`.
But computing the constraints twice sounds expensive in general. I suggest to only apply this heuristics to `HPoly`s. Action points:
* Add another dispatch method `isbounded(P::HPoly, ...)`.
* Update the docstring.
* Add a test for non-`HPoly` polyhedron.

test/unit_Polyhedron.jl
```diff
@@ -42,7 +44,17 @@ for N in [Float64, Rational{Int}, Float32]
 
     # boundedness
     @test !isbounded(p_univ)
-
+    @test isbounded(p)
+    @test !isbounded(HPolyhedron([HalfSpace(N[1, 0], N(1))]))
+    
+    @test _isbounded_stiemke(p_univ)
+    @test _isbounded_stiemke(p)
+    @test !_isbounded_stiemke(HPolyhedron([HalfSpace(N[1.0, 0.0], N(1))]))
```
mforets, line 52:
```suggestion
    @test !_isbounded_stiemke(HPolyhedron([HalfSpace(N[1, 0], N(1))]))
```

test/unit_Polyhedron.jl
```diff
@@ -42,7 +44,17 @@ for N in [Float64, Rational{Int}, Float32]
...
+    @test isbounded(p)
+    @test !isbounded(HPolyhedron([HalfSpace(N[1, 0], N(1))]))
+    
+    @test _isbounded_stiemke(p_univ)
+    @test _isbounded_stiemke(p)
+    @test !_isbounded_stiemke(HPolyhedron([HalfSpace(N[1, 0], N(1))]))
+
+    @test _isbounded_unit_dimensions(p_univ)
+    @test _isbounded_unit_dimensions(p)
+    @test !_isbounded_unit_dimensions(HPolyhedron([HalfSpace(N[1.0, 0.0], N(1))]))
```
mforets, line 56:
```suggestion
    @test !_isbounded_unit_dimensions(HPolyhedron([HalfSpace(N[1, 0], N(1))]))
```

src/Interfaces/AbstractPolyhedron_functions.jl
mforets: ok. i'll change it to use to use `constraints_list(P)` here and pass an HPolyhedron to the new algorithm Stiemke.

schillic: > Add a test for non-`HPoly` polyhedron.

This might actually be impossible because all other types probably already define `isbounded`.

PR #2112, opened 2020-04-11
Update .travis.yml

mforets:
trying this as there was a message "Unknown key codecov: true" in another PR, https://travis-ci.org/github/JuliaReach/LazySets.jl/jobs/673596018/config#rccb_LazySets.jl:.travis.yml@9bbea1af87d6c033.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 ## Documentation: http://docs.travis-ci.com/user/languages/julia/
-codecov: true
 
 language: julia
 


schillic:
There was no update at https://codecov.io/gh/JuliaReach/LazySets.jl/branches so I am not sure we should do this.

schillic:
The logs are back. There was no code execution for coverage. But this is not surprising because [the script](https://github.com/JuliaReach/LazySets.jl/blob/928505dcf14810150de61d38032ba3447f38d155/.travis.yml) does not contain any command for that. Do you want to add a command or should we close?

PR #2123, opened 2020-04-14
Simpler Intersection for Lines

### LINKED ISSUE #697: Improve intersection of lines
In line-line intersection we compute `a \ b` and try...catch for an exception (zero division). We should better check in advance if the lines are parallel. See http://geomalgorithms.com/a05-_intersect-1.html#Line-Intersections

birm:
fixes #697

Really, the change is just checking if the lines intersect first. I'm not sure if that's what was desired, but I'm willing to amend as desired.

In case you care about my tests/benchmarks:
```julia
julia> using LazySets

julia> function simplerIntersect(L1::Line{N},L2::Line{N}) where {N<:Real}
         # are the lines parallel?
         if ((L1.a[1]==0 && L2.a[1] == 0) || (L1.a[2]==0 && L2.a[2] == 0) || (L1.a[1]*L2.a[2]) == (L2.a[1]*L1.a[2]) )
             # are they the same line?
             if (L1.b == L2.b)
                 return L1
             else
                 # lines are parallel but not identical
                 return EmptySet{N}(dim(L1))
             end
         else
             b = [L1.b, L2.b]
             a = [transpose(L1.a); transpose(L2.a)]
             return Singleton(a \ b)
         end
       end
simplerIntersect (generic function with 1 method)

julia> a = Line([1,1], 1)
Line{Int64,Array{Int64,1}}([1, 1], 1)

julia> b = Line([-1,1], 1)
Line{Int64,Array{Int64,1}}([-1, 1], 1)

julia> # obvious cross
       @assert intersection(a,b) == simplerIntersect(a,b)

julia> # parallel, same
       c = Line([1,1],[2,2])
Line{Float64,Array{Float64,1}}([-1.0, 1.0], 0.0)

julia> d = Line([2,2],[3,3])
Line{Float64,Array{Float64,1}}([-1.0, 1.0], 0.0)

julia> @assert intersection(d,c) == simplerIntersect(d,c)

julia> # parallel, not same
       e = Line([1,0],[2,1])
Line{Float64,Array{Float64,1}}([-1.0, 1.0], -1.0)

julia> @assert intersection(d,e) == simplerIntersect(d,e)

julia> # vert
       f = Line([1,0],0)
Line{Int64,Array{Int64,1}}([1, 0], 0)

julia> g = Line([2,0],0)
Line{Int64,Array{Int64,1}}([2, 0], 0)

julia> @assert intersection(f,g) == simplerIntersect(f,g)

julia> # horiz
       f = Line([0,1],0)
Line{Int64,Array{Int64,1}}([0, 1], 0)

julia> g = Line([0,2],0)
Line{Int64,Array{Int64,1}}([0, 2], 0)

julia> @assert intersection(f,g) == simplerIntersect(f,g)

julia> # both
       f = Line([1,0],0)
Line{Int64,Array{Int64,1}}([1, 0], 0)

julia> g = Line([0,2],0)
Line{Int64,Array{Int64,1}}([0, 2], 0)

julia> @assert intersection(f,g) == simplerIntersect(f,g)

julia> # randoms
       for i in 1:1000000
         r1 = rand(Line)
         r2 = rand(Line)
         @assert intersection(r1,r2) == simplerIntersect(r1,r2)
       end

julia> using BenchmarkTools

julia> @benchmark simplerIntersect(rand(Line),rand(Line))
BenchmarkTools.Trial: 
  memory estimate:  880 bytes
  allocs estimate:  14
  --------------
  minimum time:     743.451 ns (0.00% GC)
  median time:      788.736 ns (0.00% GC)
  mean time:        897.050 ns (8.98% GC)
  maximum time:     56.384 μs (98.25% GC)
  --------------
  samples:          10000
  evals/sample:     144

julia> @benchmark intersection(rand(Line),rand(Line))
BenchmarkTools.Trial: 
  memory estimate:  880 bytes
  allocs estimate:  14
  --------------
  minimum time:     765.157 ns (0.00% GC)
  median time:      822.795 ns (0.00% GC)
  mean time:        938.838 ns (9.20% GC)
  maximum time:     38.390 μs (96.95% GC)
  --------------
  samples:          10000
  evals/sample:     127
```

### DIFF
--- a/src/ConcreteOperations/intersection.jl
+++ b/src/ConcreteOperations/intersection.jl
@@ -70,8 +70,9 @@ Line{Float64,Array{Float64,1}}([1.0, 1.0], 1.0)
 """
 function intersection(L1::Line{N}, L2::Line{N}
                      ) where {N<:Real}
+    det = L1.a[1]*L2.a[2] - L1.a[2]*L2.a[1]
     # are the lines parallel?
-    if isapprox(L1.a[1]*L2.a[2], L2.a[1]*L1.a[2])
+    if isapproxzero(det)
         # are they the same line?
         if isapprox(L1.b, L2.b)
             return L1
@@ -80,9 +81,9 @@ function intersection(L1::Line{N}, L2::Line{N}
             return EmptySet{N}(dim(L1))
         end
     else
-        b = [L1.b, L2.b]
-        a = [transpose(L1.a); transpose(L2.a)]
-        return Singleton(a \ b)
+        x = (L1.b*L2.a[2] - L1.a[2]*L2.b)/det
+        y = (L1.a[1]*L2.b - L1.b*L2.a[1])/det
+        return Singleton([x, y])
     end
 end
 
--- a/src/ConcreteOperations/intersection.jl
+++ b/src/ConcreteOperations/intersection.jl
@@ -56,6 +56,12 @@ If the lines are identical, the result is the first line.
 If the lines are parallel and not identical, the result is the empty set.
 Otherwise the result is the only intersection point.
 
+### Algorithm
+
+We first check whether the lines are parallel.
+If not, we use [Cramer's rule](https://en.wikipedia.org/wiki/Cramer%27s_rule)
+to compute the intersection point.
+
 ### Examples
 
 The line ``y = -x + 1`` intersected with the line ``y = x``:
--- a/src/ConcreteOperations/intersection.jl
+++ b/src/ConcreteOperations/intersection.jl
@@ -71,9 +71,9 @@ Line{Float64,Array{Float64,1}}([1.0, 1.0], 1.0)
 function intersection(L1::Line{N}, L2::Line{N}
                      ) where {N<:Real}
     # are the lines parallel?
-    if (L1.a[1]*L2.a[2]) == (L2.a[1]*L1.a[2])
+    if isapprox(L1.a[1]*L2.a[2], L2.a[1]*L1.a[2])
         # are they the same line?
-        if (L1.b == L2.b)
+        if isapprox(L1.b, L2.b)
             return L1
         else
             # lines are parallel but not identical
--- a/src/ConcreteOperations/intersection.jl
+++ b/src/ConcreteOperations/intersection.jl
@@ -70,22 +70,19 @@ Line{Float64,Array{Float64,1}}([1.0, 1.0], 1.0)
 """
 function intersection(L1::Line{N}, L2::Line{N}
                      ) where {N<:Real}
-    b = [L1.b, L2.b]
-    a = [transpose(L1.a); transpose(L2.a)]
-    try
-        # results in LAPACKException or SingularException if parallel
-        return Singleton(a \ b)
-    catch e
-        @assert e isa LAPACKException || e isa SingularException "unexpected " *
-            "$(typeof(e)) from LAPACK occurred while intersecting lines:\n$e"
-        # lines are parallel
-        if an_element(L1) ∈ L2
-            # lines are identical
+    # are the lines parallel?
+    if (L1.a[1]*L2.a[2]) == (L2.a[1]*L1.a[2])
+        # are they the same line?
+        if (L1.b == L2.b)
             return L1
         else
             # lines are parallel but not identical
             return EmptySet{N}(dim(L1))
         end
+    else
+        b = [L1.b, L2.b]
+        a = [transpose(L1.a); transpose(L2.a)]
+        return Singleton(a \ b)
     end
 end
 


mforets:
Hi @birm, thanks for your first contribution in this library! I haven't yet looked into details, but i think that your PR is in the right direction.

Here are three comments, to get started:

- Equality checks are discouraged for possibly floating-point values (important: this is a general "fact" in scientific computing, in general). So instead of a check like `foo == bar`, it is better to check if the numbers on each side are *approximately equal*. Now in LazySets, we have a global tolerance so here it would be best to do `isapprox(foo, bar)`. Please check the documentaion of [_isapprox](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Utils/comparisons.jl#L201) and ask if you have questions.

- When you use BenchmarkTools, use interpoaltion like so

```julia
L1 = rand(...); L2 = rand(...)
@benchmark intersection($L1, $L2)
```
See the README of BeenchmarkTools for details.

- Is the new function type-stable? We can use the macro `@inferred` from the `Test` module in Julia Base.

birm:
Re approx, done with 0d3a694

Re Profiling, here's the result of what you've requested:
```julia
julia> using BenchmarkTools

julia> L1 = rand(Line)
Line{Float64,Array{Float64,1}}([-1.0544065169456562, -0.32176871852782946], -1.7118686137056296)

julia> L2 = rand(Line)
Line{Float64,Array{Float64,1}}([-0.9328386923710373, 1.266576182049872], 0.5450159708710848)

julia> @benchmark simplerIntersect($L1, $L2)
BenchmarkTools.Trial: 
  memory estimate:  624 bytes
  allocs estimate:  10
  --------------
  minimum time:     540.232 ns (0.00% GC)
  median time:      585.408 ns (0.00% GC)
  mean time:        649.514 ns (7.44% GC)
  maximum time:     23.370 μs (96.92% GC)
  --------------
  samples:          10000
  evals/sample:     190

julia> @benchmark intersection($L1, $L2)
BenchmarkTools.Trial: 
  memory estimate:  624 bytes
  allocs estimate:  10
  --------------
  minimum time:     557.866 ns (0.00% GC)
  median time:      604.481 ns (0.00% GC)
  mean time:        674.646 ns (7.48% GC)
  maximum time:     22.883 μs (96.55% GC)
  --------------
  samples:          10000
  evals/sample:     186
```

Re type stability, I've gotten this message. I don't think I've changed return types though.
```
@inferred simplerIntersect(r1,r2)
ERROR: return type Singleton{Float64,Array{Float64,1}} does not match inferred return type Union{EmptySet{Float64}, Line{Float64,Array{Float64,1}}, Singleton{Float64,Array{Float64,1}}}
```

mforets:
> Re type stability, I've gotten this message. I don't think I've changed return types though.

Ok, this is trickier and not the purpose of this issue. As it turns out, the function is not type stable because (by design) we choose to return either an empty set, or an interval, or a line, depending on the values. For maximum performance in Julia functions should be stable. A workaround would be to refactor the intersection check into smaller pieces; preallocate the singleton, and so on, but this would take us too far from the purpose of this PR, so I'd rather say to forget about this.

---

Regarding the computation of the intersection in lines 83-85, i did some tests and saw that it is more efficient to directly hard-code the solution of the system using [Cramer's rule](https://en.wikipedia.org/wiki/Cramer%27s_rule). Could you try this?

It should be something like the following, where the `b1`, `b2` and the `aij` should be appropriately defined by the inputs:

```julia
# ....
# once you know that Δ is not zero
x = (b1*a22 - a12*b2)/Δ
y = (a11*b2 - b1*a21)/Δ
return Singleton([x, y])
```

mforets:
I forgot to mention that I defined `Δ` as being `Δ = a11 * a22 - a12 * a21`, and the check for the lines being parallel should now be `isapproxzero(Δ)`.

birm:
Yes, that is faster. I was earlier trying to minimize changes, but I've changed it to that.
I can use `Δ` instead of `del` if you prefer, but I find `del` cleaner

```julia
julia> L1 = rand(Line)
Line{Float64,Array{Float64,1}}([-2.1003553978239853, 0.8522753059803063], -0.6447691460541012)

julia> L2 = rand(Line)
Line{Float64,Array{Float64,1}}([0.7600442745779465, 1.039140438096269], 0.7135721135995625)

julia> @benchmark simplerIntersect($L1, $L2)
BenchmarkTools.Trial: 
  memory estimate:  624 bytes
  allocs estimate:  10
  --------------
  minimum time:     536.905 ns (0.00% GC)
  median time:      579.098 ns (0.00% GC)
  mean time:        659.783 ns (8.22% GC)
  maximum time:     43.153 μs (98.42% GC)
  --------------
  samples:          10000
  evals/sample:     189

julia> @benchmark simplerIntersect2($L1, $L2)
BenchmarkTools.Trial: 
  memory estimate:  128 bytes
  allocs estimate:  3
  --------------
  minimum time:     66.898 ns (0.00% GC)
  median time:      77.909 ns (0.00% GC)
  mean time:        89.619 ns (8.37% GC)
  maximum time:     5.538 μs (98.33% GC)
  --------------
  samples:          10000
  evals/sample:     979

```

mforets:
> Yes, that is faster

Looks great! There is a considerable speedup.

> I can use Δ instead of del if you prefer, but I find del cleaner

I'm fine with using `Δ` or `det` for determinant, why would you choose `del` (mind the ell), this is the shortcut of another word or is a typo?

---

PR looks good to me, let's wait for CI to finish (can take a while :) ).

birm:
I was using del as short for delta; changed to det, since you're right, that's much clearer.

mforets:
I've approved this PR, and it's now awaiting for other potential reviewers. Thanks again for your contribution!

mforets:
Thanks again for your contribution @birm ! It doesn't seem like you submitted the PR through issue hunt (https://issuehunt.io/r/JuliaReach/LazySets.jl/issues/697), though this issue had a bounty.

Issue #1895, opened 2020-01-05
Extend LazySet to non-convex sets


schillic:
We plan to extend the `LazySet` interface to non-convex sets. Since this is a major breaking change, we will then release version 2 of this library.

For a smooth transition, we accept that code that used to receive a `LazySet` input before may continue to work for non-convex sets but the output is undefined (i.e., there may be an error, or there will be no error but the result may be inconsistent, or it may just work). We will gradually clarify these places, but the transition may happen before all places have been updated. Below we list the progress.

### not checked

* `ConcreteOperations/`
* `Interfaces/`

### checked

* `Approximations/`
* `Arrays/`
* `Assertions/`
* `Initialization/`
* `LazyOperations/` (#1837)
* `Plotting/`
* `Parallel/`
* `Sets/`
* `Utils/`
* single files (`convert.jl`, `init.jl`, `LazySets.jl`)

### to be discussed

* [x] `Plotting/mesh.jl`:
  We `@assert` that `constraints_list` is applicable. I think this method requires that the input is convex. So everything should be fine.
* [x] `Approximations/overapproximate.jl`:
  I think `overapproximate(cap::Intersection{N, <:LazySet, <:AbstractPolyhedron{N}}` should still work.

### to be changed

* [x] iterative refinement (#1855)
* [x] inclusion between line segment and lazy set (#1856)
* [x] docstrings of lazy operations (#1837)
* [ ] replace "convex set" by "set" in explanations of input parameters taking a `LazySet`
* [x] `convert.jl`: `convert(::Type{Interval}, S::LazySet{N})` only works for convex sets
* [x] `hausdorff_distance`
* [x] `underapproximate`
* [x] `overapproximate(::LazySet, Type{<:Interval})` returns `convert(Interval, S)`, which only works for convex sets (~~#1976~~)

schillic:
> should there be also `NonConvexSet <: LazySet` ?

No, `ConvexSet` is already only a one-sided approximation ("definitely convex"). You cannot guarantee that a generally non-convex set is not accidentally convex. We may actually not have a use case for `ConvexSet` in the end (once we have replaced them all), but that is too far in the future.

schillic:
Closing as superseded by #3047.

PR #2132, opened 2020-04-20
#1357 - WIP Add star set representation

### LINKED ISSUE #1357: Add star set representation

mforets:
Closes #1357.

---

## References

- Parsimonious, Simulation based Verification of Linear Systems [pdf](https://www.cs.unc.edu/~psd/files/research/2016/psd-mv-cav-16.pdf) << initial paper describing the set representation
- Simulation-Equivalent Reachability of Large Linear Systems with Inputs [pdf](https://www.cs.unc.edu/~psd/files/research/2017/sb-psd-cav-2017.pdf) << stars are generalized such that the number of vertices can be greater than the dimension. the representation is applied for (simulation based) reachability on linear systems

Applications in neural verification:

- Star-Based Reachability Analysis of Deep Neural Networks  [pdf](http://www.verivital.com/research/tran2019fm.pdf)
- Star-Based Reachability Analysis of Deep Neural Networks (extended version, contains proofs)
- Constrained Polynomial Zonotopes [pdf](https://arxiv.org/pdf/2005.08849.pdf) << in relation to other set representations


mforets:
cc'ing @tomerarnon  to whom this set rep may be interesting.

i wrote this branch long ago and it would need to be  rebased and since this setrep is non-convex this related to https://github.com/JuliaReach/LazySets.jl/issues/1895

comments about concrete functionality to add are welcome, that we can add in follow up PRs.

tomerarnon:
Yes, this is certainly interesting. I can't comment yet as to concrete functionality since I am not yet familiar enough. @changliuliu might already have thoughts on this though.

changliuliu:
Wow, nice to know you are going to add star set representation! It can be used in both exact reachability and approximated reachability analysis as a new data structure. 

There is a paper on how to use star set for reachability analysis: [Star-Based Reachability Analysis of Deep Neural Networks](http://www.verivital.com/research/tran2019fm.pdf). The solver is called NNV.

For NeuralVerification.jl, the star set representation can be used to upgrade exactReach and Ai2. The affine transformation of a star set would be more efficient than a polytope, hence can boost efficiency in exactReach. The star set is less constraining than zonotope (which is required to be symmetric), hence can lead to smaller over approximation error.

castrong:
Hi @mforets I was wondering if there is still a plan to incorporate a star set representation into LazySets? I am working on implementing the [NNV solver](http://www.verivital.com/research/tran2019fm.pdf) for [NeuralVerification.jl](https://github.com/sisl/NeuralVerification.jl) which relies heavily on star sets.

mforets:
Hi @castrong, we are very interested in adding such set representation to the library. But no one is currently working on it so this PR has stalled. Do you have an understanding of star sets? What additional methods do you think we should be adding to this PR? Linear maps, minkowski sum, ... ? I'm also fine merging the PR as is and add features incrementally.

Apart from these considerations, do you think we should have two structs, `Star` and `GeneralizedStar`, or only one should do?

castrong:
For NNV the functions we are likely to use are:

(1) affine maps of a star set (results in another star set)
(2) intersect a star set with a halfspace (results in another star set)
(3) ReLU applied to a single dimension of the star set (results in one or two star sets)
(4) ReLU applied to every dimension of the star set
(5) Subset - check whether a star set is a subset of another star set or polytope

I believe these should not be too complex - several are detailed on page 4 of the [Star-Based Reachability paper](http://www.verivital.com/research/tran2019fm.pdf), I'd be happy to discuss each function in detail! 

I actually don't have too much experience outside of NNV with star sets, maybe @changliuliu or @tomerarnon would have insight into what other functionality would be most useful in other applications? 

I'm a bit new to the internals of LazySets, what would be the advantage of abstracting a star set into a GeneralizedStar and a Star? I only know of one representation for star sets in case it is to accommodate different representations. Would it make any sense for star set to be an AbstractPolytope since any polytope can be represented by a star set and vice versa? Although there can be many star set representations for the same polytope.

schillic:
> what would be the advantage of abstracting a star set into a GeneralizedStar and a Star? I only know of one representation for star sets in case it is to accommodate different representations.

The advantage of having many set representations is that you can write more efficient algorithms for subclasses. The disadvantage is that you can lose track of what algorithms are available :smiley:
It is probably useful to add representations in steps and on demand. @mforets already removed `GeneralizedStar` from this PR.

> Would it make any sense for star set to be an AbstractPolytope since any polytope can be represented by a star set and vice versa?

Yes, as long as we can effectively compute the vertices (possibly by computing the constraints).

mforets:
Does someone have the extended version of the article with proofs? I get a 404 error with Ref [22] http://www.taylortjohnson.com/research/tran2019fm_extended.pdf

To transform a star to a polyhedron i think one can take `c = 0` and let `v1 = e1, ..., vm = em` the canonical vectors of R^n, with the predicate being the polyhedron itself. Conversely, the stars are non-convex in general so asking for a conversion from a star to a polyhedron is not possible. Hence i doubt about setting `Star <: AbstractPolytope`, given that `AbstractPolytope <: AbstractPolyhedron`.

About the plan, i think in this PR we can just define the type and move on with implementations of (1), (2) and (5) in separate PRs. About (3) and (4), I'm not sure yet if they should be added in this library or in NV.jl directly,: from a design perspective, LazySets.jl shouldn't reason about networks and layers, that is application-specific. That said, there is a lazy ReLU type as a mathematical transformation on sets, so it seems reasonable to add some functionality if we have such types.

mforets:
> Would it make any sense for star set to be an AbstractPolytope since any polytope can be represented by a star set and vice versa?

Does the converse hold? Maybe I'm missing something. (an `AbstractPolytope` in LazySets is a bounded convex polyhedron)

schillic:
> Does the converse hold?

Figure 1 [here](https://arxiv.org/pdf/2005.08849) indicates that this is not the case.

mforets:
> I'm a bit new to the internals of LazySets, what would be the advantage of abstracting a star set into a GeneralizedStar and a Star?

In the work tran2019fm, they restrict the predicate to be linear constraints. In the implementation side, it makes sense to have one set type, call it eg. `Star`, such that the predicates are indeed a polyhedron. This allows to use the type information on the constraints defining the predicate and results in better code. On the other hand, i was wondering if there are other works where the predicate is not a polyhedron but something more general. In that case we shall define a `GeneralizedStar` , in which the type of the predicate is not limited to a polyhedron but can be anything, we just add a type parameter without constraining it.

(The main advantage of defining a `AbstractStar` layer is that this allows for generic programming and general fallbacks: you can write generic functions that work by calling getter functions of any star implementation (say, `center(S)`, `predicate(S)`, ..). This avoids code duplication, plus you get "for free" lots of methods for any future concrete implementation of `AbstractStar` that implements the interface.)

schillic:
I suggest to use `PolyhedralStar` or similar for your definition of `Star` because at least in my reference above a star is not restricted to a polyhedron.

mforets:
An observation: if we take the predicate polyhedron `P` to be the unit ball in the infinity norm, then we get a zonotope.

mforets:
> Does someone have the extended version of the article with proofs?

I got it (thanks @schillic). It contains an appendix with the proofs.

castrong:
Ah interesting observation @mforets could the Zonotope object then be a subclass of a `PolyhedralStar`? For implementing NNV PolyhedralStars are the only variety of Star which would be needed, I hadn't considered non Polyhedral ones so a general star makes a lot of sense in that case thank you for the explanation.

src/Sets/Stars.jl
```diff
@@ -0,0 +1,118 @@
...
+function affine_map(M::AbstractMatrix, X::AbstractStarSet, v::AbstractVector)
+    # .......
+end
+
+function intersection(X::AbstractStarSet, H::HalfSpace)
+    # .......
+end
+
+# symmetric method
+intersection(H::HalfSpacev, X::AbstractStarSet) = intersection(X, H)
```
castrong, line 28: Typo (extra v on the end of HalfSpace)?

src/Interfaces/AbstractStar.jl
```diff
@@ -0,0 +1,9 @@
+"""
+    AbstractStar{N} <: LazySet{N}
+
+Abstract supertype for all star set types.
+"""
+abstract type AbstractStar{N} <: LazySet{N} end
```
mforets, line 6: About the type hierarchy, `AbstractPolyhedron` could inherit from `AbstractStar`.

src/Sets/Star.jl
```diff
@@ -0,0 +1,142 @@
...
+The predicate function is implemented as a conjunction of linear constraints.
+This definition can be found in reference [1].
+
+### Examples
+
+This example is drawn from Example 1 in [1]. Consider the two-dimensional plane
+``\\mathbb{R}^2``. Let
+
+```jldoctest star_constructor
+julia> V = [[1.0, 0.0], [0.0, 1.0]]
```
mforets, line 28: Here is another example, taken from Fig. 1 in "Star-based reachability analysis of DNNs"

```julia
c = zeros(2)
V = [1 0; 0 1.]
P = BallInf(zeros(2), 1.0)
S = Star(c, V, P)
```

Same example but instantiating with a polygon in constraint representation. Please note that `convert(HPolytope, BallInf(...))` will produce a polytope whose normal vectors are `SingleEntryVector` (SEV, ie. an efficient representation for vectors proportional to canonical basis vectors of R^n); the problem is that the new constraint obtained by intersection is not in general a SEV, hence we can't push onto the same array. OTOH `overapproximate(..., HPolygon)` creates constraints with `Vector{..}` normal vectors.

c = zeros(2)
V = [1 0; 0 1.]
P = overapproximate(BallInf(zeros(2), 1.0), HPolygon)
S = Star(c, V, P)
H = rand(HalfSpace)

intersection!(S, H) # in-place : add one more constraint to X

intersection(S, H)  # return a new star
```

File: src/Sets/Star.jl [hunk 4]
```diff
@@ -0,0 +1,142 @@
...
+end
+
+# getter functions
+center(X::Star) = X.c
+basis(X::Star) = X.V
+predicate(X::Star) = X.P
+
+dim(X::Star) = length(X.c)
+
+function linear_map!(M::AbstractMatrix, X::Star, c′, V′)
```
mforets, line 63: For linear map and affine map, I added alternative in-place implementations which do not allocate memory.

src/Sets/Star.jl
```diff
@@ -0,0 +1,142 @@
...
+
+    c = center(X)
+    V = basis(X)
+    P = predicate(X)
+    c′ = M * c + v
+    V′ = M * V
+    return Star(c′, V′, P)
+end
+
+function addconstraint!(X::Star, H::HalfSpace)
```
mforets, line 112: I think that some functionality similar to polyedral types is useful, such as those used to handle constraints. It is used in the implementation of the concrete intersection below.

src/Sets/Star.jl
```diff
@@ -0,0 +1,142 @@
...
+julia>
+```
+
+### References
+
+- [1] Duggirala, Parasara Sridhar, and Mahesh Viswanathan.
+      *Parsimonious, simulation based verification of linear systems.*
+      International Conference on Computer Aided Verification. Springer, Cham, 2016.
+"""
+struct Star{N, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}, PT<:AbstractPolyhedron{N}} <: AbstractStar{N}
```
mforets, line 49: In the initial implementation we stored the basis as a vector of vectors. In the new one I propose to store the basis as a matrix, because the affine map and the intersection operation work with the full matrix, and partly because storing as a matrix we can make use of Level 2 BLAS operations. (The basis vectors of zonotopes are also stored as the columns of a matrix).

src/Sets/Star.jl
```diff
@@ -0,0 +1,142 @@
...
+    c::VN # center
```
mforets, line 50: About the type parameter I chose to use `PT <: AbstractPolyhedron`, I think this is better than restricting to an `HPolyhedron`. We could even let any `LazySet`, but i'm not sure if this is useful.

mforets:
@castrong your comments if you have any on the methods I added yesterday are most welcome.

---

About:

> (5) Subset - check whether a star set is a subset of another star set or polytope

If we can compute the support function of a star set, then the inclusion on a polytope can be obtained with the general algorithm
[_issubset_constraints_list](https://github.com/JuliaReach/LazySets.jl/blob/master/src/ConcreteOperations/issubset.jl#L295). But better to keep this PR short so I've opened https://github.com/JuliaReach/LazySets.jl/issues/2431. For general polyhedral predicate `P` the support function of the star set is obtained solving a linear program. We can also specialize for `BallInf` predicates, etc.

src/Sets/Star.jl
```diff
@@ -0,0 +1,142 @@
...
+    c = center(X)
+    V = basis(X)
+    c′ = M * c
+    V′ = M * V
+    P = predicate(X)
+    return Star(c′, V′, P)
+end
+
+# compute the affine map M * X + c′ overriding c′
+function affine_map!(M::AbstractMatrix{N}, X::Star, c′, V′) where {N}
```
castrong, line 88: Could affine_map! and affine_map use linear_map! and linear_map or would that always require allocating a star twice (and so be less efficient)?

castrong:
@mforets I took a detailed read through and checked the implementations against the paper and my notes and they looked good, it's very clean! I'll pull this branch try some things out with it.

castrong:
I also just realized an overapproximation from a star set to a Hyperrectangle is needed for NNV. The bounds in each dimension are used to tell whether passing the star set through a ReLU function will have an effect on that dimension. For an n-dimensional star set, you can solve 2*n LPs in order to find the tight Hyperrectangle over approximation.

src/Sets/Star.jl
```diff
@@ -0,0 +1,142 @@
...
+    @assert dim(X) == size(M, 2) == length(c′) "an affine map of size $(size(M)) and " *
```
castrong, line 89: Why should dim(x) and size(M, 2) have to match the length of c'? For instance if we have a 1-d star, with affine map from 1-d-->2d with M = [1; 0] and c = [1; 1] we will have dim(x) = 1, size(M, 2) = 1, and length(c') = 2.

castrong: I think length(c') should match size(M, 1) instead

src/Sets/Star.jl
```diff
@@ -0,0 +1,142 @@
...
+    c = center(X)
+    V = basis(X)
+    mul!(c′, M, c, one(N), one(N))
+    mul!(V′, M, V)
+    P′ = predicate(X)
+    return Star(c′, V′, P′)
+end
+
+function affine_map(M::AbstractMatrix, X::Star, v::AbstractVector)
+    @assert dim(X) == size(M, 2) == length(v) "an affine map of size $(size(M)) and " *
```
castrong, line 101: same comment as line 89, I think that size(M, 1) == length(v) should be checked instead of size(M, 2)

mforets:
@castrong  thanks for your comments and your code contributions. I look forward to rounding up the PR this week.

castrong:
@mforets I now realize the implementations of overapprox and subset I sent along will probably be done using the σ and ρ functions, would sending along implementations of those functions instead be helpful?

mforets:
the function `temp_overapprox(set::Star; optimizer=GLPK.Optimizer)` shall be `overapproxiamte(S, Hyperrectangle)`, while the signature of `issubset(star::Star, polytope::AbstractPolytope; optimizer=GLPK.Optimizer)` looks good to me. we should also implement `ρ` for star sets. if you write the function for `ρ` (and `σ` but this one is optional) it would be helpful yes, i will review it and incorporate it into this PR.

castrong:
Here is an implementation of σ and ρ. Here, σ will just give back a single point - I wasn't sure how to modify it to find the appropriate set of points if multiple have the same ρ value. 

```
function ρ(ℓ, star::Star; optimizer=GLPK.Optimizer)
    model = Model(optimizer)
    @variable(model, x[1:dim(star.P)])
    add_set_constraint!(model, star.P, x)
    @objective(model, Max, ℓ' * (star.V * x + star.c))
    optimize!(model)
    return objective_value(model)
end

function σ(ℓ, star::Star; optimizer=GLPK.Optimizer)
    model = Model(optimizer)
    @variable(model, x[1:dim(star.P)])
    add_set_constraint!(model, star.P, x)
    @objective(model, Max, ℓ' * (star.V * x + star.c))
    optimize!(model)
    return star.V * value(x) + star.c
end
```

castrong:
Hi @mforets is there anything else I can do to help with this PR / any other bits of code or implementation that could be useful for me to write?

castrong:
Hi @mforets I'm just checking back in on this PR to see if there's anything I can help with to get it all wrapped up :).

mforets:
@castrong I'll readdress this PR real soon (in part due summer break in december- january this took more than usua, sorry).

mforets:
> summer break in december- january

(yes, it's summer i'm in south america :sun_with_face: :uruguay: :smile: )

castrong:
Okay no worries, sounds great and I hope the summer break is going well! 🌞🌞🌞

mforets:
- about the of the support function ([this comment](https://github.com/JuliaReach/LazySets.jl/pull/2132#issuecomment-741343285))  we can delay the computation of the support function over the predicate -- see [this PR](https://github.com/JuliaReach/LazySets.jl/pull/2557). it has the advantage that we an efficient implementation of rho over the basis when it is available, and fallback to the LP in the general case

- about sigma: "I wasn't sure how to modify it to find the appropriate set of points if multiple have the same ρ value" << LazySets only returns 1 value, we haven't implemented (nor needed in practice) to get all the support vectors for a given direction

mforets:
update :new: after some discussion we saw that the generalized star with a polyhedral predicate is best defined as an alias for the AffineMap (with a constraint on the type of the matrix)... which already exists! :)   this way we reuse all existing methods. there is a new constructor `Star`, getter functions and `LazySets.STAR` for dispatch and to implement better methods than the fallback (several ideas in this thread so i won't close it yet).  See https://github.com/JuliaReach/LazySets.jl/pull/2578.

mforets:
Fixed in https://github.com/JuliaReach/LazySets.jl/pull/2665.

Issue #2125, opened 2020-04-14
Improve intersection for line segments


mforets:
Currently, `intersection` uses the default for polytopes. A new algorithm that dispatches on a pair of line segments should be implemented.

May be useful:

- http://geomalgorithms.com/a05-_intersect-1.html#Line-Intersections

mforets:
See also this comment https://github.com/JuliaReach/LazySets.jl/issues/262#issuecomment-369352449

IssueHuntBot:
[@mforets](https://issuehunt.io/u/mforets) has funded $10.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/105701832/issues/2125) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/105701832/issues/2125).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

IssueHuntBot:
[@mforets](https://issuehunt.io/u/mforets) has rewarded $9.00 to [@birm](https://issuehunt.io/u/birm). [See it on IssueHunt](https://issuehunt.io/repos/105701832/issues/2125)

- :moneybag: Total deposit: $10.00
- :tada: Repository reward(0%): $0.00
- :wrench: Service fee(10%): $1.00

PR #2137, opened 2020-04-22
intersection between line and line segment

### LINKED ISSUE #262: Intersection of lines and line segments
See also the book [Graphics Gems](
http://inis.jinr.ru/sl/vol1/CMC/Graphics_Gems_1,ed_A.Glassner.pdf) (eg. Goldman's chapter about Intersection of two lines in three-space, p. 304).

birm:
Fixes #262 

Draft due to an error I don't understand when testing:

```julia
function i1(a::LineSegment, b::Line)
  # cast a as line
  ap = Line(a.p, a.q)
  # find intersection between a' and b
  m = intersection(ap, b)
  if m == b
    # if this equals b, then all of the segment is the intersection
    return a
  elseif typeof(m) <: Singleton && m.element ∈ a
    # if the intersection between lines is in the segment
    return m
  else
    # no intersection
    return EmptySet(max(dim(a), dim(b)))
  end
end

# handle reverse case
function i1(a::Line, b::LineSegment)
  return i1(b,a)
end

## tests
s1 = LineSegment([-5,-5], [5,5])
t = Line(s1.p, s1.q)
# same as s1
l1 = Line([0,0], [5,5])
# intersects out of s1
l2 = Line([10,10], [6,0])
# parallel to s1
l3 = Line([0,1], [5,6])
# intersects within
l4 = Line([5,-5], [-5,5])


i1(s1, l1)
i1(s1, l2)
i1(s1, l3)
i1(s1, l4)
```
throws 

```julia
julia> i1(s1, l4)
ERROR: MethodError: no method matching iterate(::LineSegment{Int64,Array{Int64,1}})
Closest candidates are:
  iterate(::Core.SimpleVector) at essentials.jl:603
  iterate(::Core.SimpleVector, ::Any) at essentials.jl:603
  iterate(::ExponentialBackOff) at error.jl:253
  ...
Stacktrace:
 [1] in(::Array{Float64,1}, ::LineSegment{Int64,Array{Int64,1}}) at ./operators.jl:1041
 [2] i1(::LineSegment{Int64,Array{Int64,1}}, ::Line{Float64,Array{Float64,1}}) at ./REPL[159]:9
 [3] top-level scope at REPL[170]:1
```
In the second and fourth test case. I'm probably missing something silly.

### DIFF
--- a/src/ConcreteOperations/intersection.jl
+++ b/src/ConcreteOperations/intersection.jl
@@ -90,37 +90,41 @@ function intersection(L1::Line{N}, L2::Line{N}
 end
 
 """
-    intersection(X::LineSegment{N}, hp::Line{N}
-                ) where {N<:Real}
+    intersection(a::LineSegment{N}, b::Line{N}) where {N<:Real}
+
 Compute the intersection of a line and a line segment.
+
 ### Input
+
 - `a`  -- LineSegment
 - `b` -- Line
+
 ### Output
+
 If the sets do not intersect, the result is the empty set.
 Otherwise the result is the singleton or line segment that describes the intersection.
 """
 
 function intersection(a::LineSegment{N}, b::Line{N}) where {N<:Real}
-  # cast a as line
-  ap = Line(a.p, a.q)
-  # find intersection between a' and b
-  m = intersection(ap, b)
-  if m == b
-    # if this equals b, then all of the segment is the intersection
-    return a
-  elseif typeof(m) <: Singleton && m.element ∈ a
-    # if the intersection between lines is in the segment
-    return m
-  else
-    # no intersection
-    return EmptySet(max(dim(a), dim(b)))
-  end
+    # cast a as line
+    ap = Line(a.p, a.q)
+    # find intersection between a' and b
+    m = intersection(ap, b)
+    if m == ap
+        # if this equals a, then all of the segment is the intersection
+        return a
+    elseif typeof(m) <: Singleton && m.element ∈ a
+        # if the intersection between lines is in the segment
+        return m
+    else
+        # no intersection
+        return EmptySet{N}(2)
+    end
 end
 
 # symmetric method
-function intersection(a::Line, b::LineSegment)
-  return intersection(b,a)
+function intersection(a::Line{N}, b::LineSegment{N}) where {N<:Real}
+    return intersection(b,a)
 end
 
 """
--- a/src/ConcreteOperations/intersection.jl
+++ b/src/ConcreteOperations/intersection.jl
@@ -90,6 +90,40 @@ function intersection(L1::Line{N}, L2::Line{N}
 end
 
 """
+    intersection(X::LineSegment{N}, hp::Line{N}
+                ) where {N<:Real}
+Compute the intersection of a line and a line segment.
+### Input
+- `a`  -- LineSegment
+- `b` -- Line
+### Output
+If the sets do not intersect, the result is the empty set.
+Otherwise the result is the singleton or line segment that describes the intersection.
+"""
+
+function intersection(a::LineSegment{N}, b::Line{N}) where {N<:Real}
+  # cast a as line
+  ap = Line(a.p, a.q)
+  # find intersection between a' and b
+  m = intersection(ap, b)
+  if m == b
+    # if this equals b, then all of the segment is the intersection
+    return a
+  elseif typeof(m) <: Singleton && m.element ∈ a
+    # if the intersection between lines is in the segment
+    return m
+  else
+    # no intersection
+    return EmptySet(max(dim(a), dim(b)))
+  end
+end
+
+# symmetric method
+function intersection(a::Line, b::LineSegment)
+  return intersection(b,a)
+end
+
+"""
     intersection(H1::AbstractHyperrectangle{N},
                  H2::AbstractHyperrectangle{N}
                 ) where {N<:Real}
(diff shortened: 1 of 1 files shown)


src/ConcreteOperations/intersection.jl
```diff
@@ -89,6 +89,40 @@ function intersection(L1::Line{N}, L2::Line{N}
...
+"""
+
+function intersection(a::LineSegment{N}, b::Line{N}) where {N<:Real}
+  # cast a as line
+  ap = Line(a.p, a.q)
+  # find intersection between a' and b
+  m = intersection(ap, b)
+  if m == b
+    # if this equals b, then all of the segment is the intersection
+    return a
```
schillic, line 111: I would rather check for `m == ap` (or swap the arguments to `intersection`) because the implementation returns the first argument if they are equal (you could hence even check for `===`) and I am not sure what happens if they are only approximately equal.

src/ConcreteOperations/intersection.jl
```diff
@@ -89,6 +89,40 @@ function intersection(L1::Line{N}, L2::Line{N}
...
+"""
+    intersection(X::LineSegment{N}, hp::Line{N}
+                ) where {N<:Real}
+Compute the intersection of a line and a line segment.
+### Input
+- `a`  -- LineSegment
+- `b` -- Line
+### Output
+If the sets do not intersect, the result is the empty set.
+Otherwise the result is the singleton or line segment that describes the intersection.
```
schillic, line 101: Something happened to the empty lines (I have this problem when I copy code from Github).
The top line should also be a copy of the function signature (probably a copy-paste error here).

src/ConcreteOperations/intersection.jl
```diff
@@ -89,6 +89,40 @@ function intersection(L1::Line{N}, L2::Line{N}
...
+"""
+
+function intersection(a::LineSegment{N}, b::Line{N}) where {N<:Real}
+  # cast a as line
+  ap = Line(a.p, a.q)
```
schillic, line 106: In this library, `Line` is represented by `ax = b` where `a` is a vector and `b` is a scalar, while `LineSegment` is represented by the two end points (vectors). Hence this constructor does not work. I could not find code to overapproximate a `LineSegment` by a `Line`, so that would need to be added first (can be done in the same PR).

Ideally you would not have to create the `Line` object but would call a helper function that accepts `a` and `b` directly. But this is a micro optimization.

src/ConcreteOperations/intersection.jl
```diff
@@ -89,6 +89,40 @@ function intersection(L1::Line{N}, L2::Line{N}
...
+  m = intersection(ap, b)
+  if m == b
+    # if this equals b, then all of the segment is the intersection
+    return a
+  elseif typeof(m) <: Singleton && m.element ∈ a
+    # if the intersection between lines is in the segment
+    return m
+  else
+    # no intersection
+    return EmptySet(max(dim(a), dim(b)))
```
schillic, line 117: `Line` and `LineSegment` are defined to be two-dimensional in this library, so you can do `EmptySet{N}(2)`.

src/ConcreteOperations/intersection.jl
birm: Maybe I'm misunderstanding, but there is a line constructor which takes two points through which it passes -- https://github.com/JuliaReach/LazySets.jl/blob/cd6856388521869b89e2700393502cf326527eb3/src/Sets/Line.jl#L50

schillic: Oh, you are right :+1: I completely forgot about that, sorry!

schillic:
> Draft due to an error I don't understand when testing:

The problem is here: `in(::Array{Float64,1}, ::LineSegment{Int64,Array{Int64,1}})`: your `LineSegment` has `Int` entries. You need to make them `Float64`.

src/ConcreteOperations/intersection.jl
```diff
@@ -89,6 +89,44 @@ function intersection(L1::Line{N}, L2::Line{N}
...
+
+function intersection(a::LineSegment{N}, b::Line{N}) where {N<:Real}
+    # cast a as line
+    ap = Line(a.p, a.q)
+    # find intersection between a' and b
+    m = intersection(ap, b)
+    if m == ap
+        # if this equals a, then all of the segment is the intersection
+        return a
+    elseif typeof(m) <: Singleton && m.element ∈ a
```
schillic, line 116:
```suggestion
    elseif m isa Singleton && m.element ∈ a
```
I think this is equivalent but more readable(?)

birm:
I can add line segment with line segment in this or another PR if desired.

mforets:
Thanks for your contribution @birm ! I would prefer to add new functionality in a separate PR please. @schillic beat me on reviewing this one, but looks good for me as well, will merge asap. I see that you left the "approximately parallel" case to be handled by the intersection of lines, which is ok.

With respect to the intersection between line segments, that is a particularly more challenging algorithm though (to handle all the special cases properly, i.e. with a given tolerance). The corresponding issue is [this one](https://github.com/JuliaReach/LazySets.jl/issues/2125).

A bunch of excellent references for this problem are:

- https://www.geeksforgeeks.org/check-if-two-given-line-segments-intersect/
- O'Rourke's book [Computational Geometry in C](http://crtl-i.com/PDF/comp_c.pdf); note that the C implementation is [available on github](https://github.com/w8r/orourke-compc/blob/master/segseg/segseg.c).

This contribution would be truly amazing, which would naturally continue implementing a polygon intersection algorithm ("clipping") for a `VPolygon`.

mforets:
Thanks a lot!

Issue #2139, opened 2020-04-24
Do not depend on Plots.jl for testing


giordano:
`Plots.jl` is a huge dependency.  As far as I can see, in [`test/unit_plot.jl`](https://github.com/JuliaReach/LazySets.jl/blob/c12e313e445f0e350250761c409f387f9c75c1e0/test/unit_plot.jl) you are not even testing anything.  I recommend you to look at what [I do](https://github.com/JuliaPhysics/Measurements.jl/blob/2ef7fadf7c2644408cff555781b9c10adfca26f6/test/plots.jl) in `Measurements.jl` to test plotting functionalities simply using `RecipesBase.jl`, so that you can remove `Plots.jl` and actually test something meaningful.

mforets:
Thanks for the tip @giordano, we'll consider it. In this library a full run of CI usually takes like ~40minutes on Travis :tired_face: So if this can reduce such runtimes while having some tests for the plotting stuff it would already be a win.

schillic:
We also need `Plots` for the documentation. I am not sure that this would resolve #2098 (I just commented there).

giordano:
> In this library a full run of CI usually takes like ~40minutes on Travis tired_face So if this can reduce such runtimes while having some tests for the plotting stuff it would already be a win.

Well, not sure if it'll reduce sensibly from 40 minutes, but in a smaller project I was able to shave off 3 minutes per job by moving from Plots to RecipesBase

> We also need Plots for the documentation.

That's one job out of the 5 total, no?

giordano:
Also, consider caching artifacts in CI: https://discourse.julialang.org/t/recommendation-cache-julia-artifacts-in-ci-services/35484.  That would help in jobs for Julia v1.3+

mforets:
thanks @giordano for all the tips :100:

PR #2140, opened 2020-04-24
Bump RecipesBase and Julia versions

schillic:
This resolves #2098.

CC @giordano

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -39,7 +39,7 @@ RecipesBase = "0.6, 0.7, 0.8, 1.0"
 Reexport = "0.2"
 Requires = "0.5, 1.0"
 TaylorModels = "0.3"
-julia = "1.0, 1.1, 1.2, 1.3, 1.4"
+julia = "1.0, 1.1, 1.2, 1.3"
 
 [extras]
 CDDLib = "3391f64e-dcde-5f30-b752-e11513730f60"
--- a/Project.toml
+++ b/Project.toml
@@ -35,11 +35,11 @@ MiniQhull = "0.1"
 Optim = "0.15, 0.16, 0.17, 0.18, 0.19, 0.20"
 Plots = "0.25, 0.26, 0.27, 0.28, 0.29"
 Polyhedra = "0.5"
-RecipesBase = "0.6, 0.7, 0.8"
+RecipesBase = "0.6, 0.7, 0.8, 1.0"
 Reexport = "0.2"
 Requires = "0.5, 1.0"
 TaylorModels = "0.3"
-julia = "1.0, 1.1, 1.2, 1.3"
+julia = "1.0, 1.1, 1.2, 1.3, 1.4"
 
 [extras]
 CDDLib = "3391f64e-dcde-5f30-b752-e11513730f60"
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ os:
 
 julia:
   - 1.0
-  - 1.2
   - 1.3
+  - 1.4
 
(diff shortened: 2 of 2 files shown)


Project.toml
```diff
@@ -35,11 +35,11 @@ MiniQhull = "0.1"
 Optim = "0.15, 0.16, 0.17, 0.18, 0.19, 0.20"
 Plots = "0.25, 0.26, 0.27, 0.28, 0.29"
 Polyhedra = "0.5"
-RecipesBase = "0.6, 0.7, 0.8"
+RecipesBase = "0.6, 0.7, 0.8, 1.0"
 Reexport = "0.2"
 Requires = "0.5, 1.0"
 TaylorModels = "0.3"
-julia = "1.0, 1.1, 1.2, 1.3"
+julia = "1.0, 1.1, 1.2, 1.3, 1.4"
```
giordano, line 42: Note that this is equivalent to
```suggestion
julia = "1.0"
```

schillic: As soon as v1.5 arrives, it is not equivalent anymore. This is a philosophical question (do you want to support non-existing Julia versions at the risk of not working then or do you want to play safe at the cost of having to make a new release for every Julia version).

giordano:
f you want to slightly reduce CI waiting time in pull requests, consider adding this to Travis:
```yaml
branches:
  only:
  - master
  - gh-pages # For building documentation
  - /^testing-.*$/ # testing branches
  - /^v[0-9]+\.[0-9]+\.[0-9]+$/ # version tags
```
This will run CI only on `master`, `gh-pages`, tags and branches whose name starts with `testing-`.  In this way pull requests opened from branches in this repository won't spawn 4 extra (mostly useless) CI jobs.

As I mentioned in #2139, consider also caching artifacts to further reduce CI time: https://discourse.julialang.org/t/recommendation-cache-julia-artifacts-in-ci-services/35484

giordano:
Caching artifacts will drastically reduce the probability of getting errors like
```
curl: (22) The requested URL returned error: 429 Too Many Requests
ERROR: Unable to automatically install 'MbedTLS' from '/home/travis/.julia/packages/MbedTLS_jll/hP8le/Artifacts.toml'
```
like in https://travis-ci.org/github/JuliaReach/LazySets.jl/jobs/679085636

schillic:
> If you want to slightly reduce CI waiting time in pull requests, consider adding this to Travis:

Thanks, this is helpful :+1:
I opened #2141 for that.

Project.toml
giordano: > As soon as v1.5 arrives, it is not equivalent anymore. 

Well, I can install LazySets v1.34.0 in Julia 1.4.1

> do you want to support non-existing Julia versions at the risk of not working then

Minor releases are non-breaking releases by semver definition, so there is (nearly) no risk of breaking.

> do you want to play safe at the cost of having to make a new release for every Julia version

As I said above, I can install LazySets v1.34.0 despite the fact [`Project.toml`](https://github.com/JuliaReach/LazySets.jl/blob/2c637151709be5968415449769da3d42bb86be25/Project.toml#L42) doesn't mention Julia 1.4

schillic: You are right. These version specifiers are so confusing because they differ for version 0.x and y.x where y != 0 (see [here](https://julialang.github.io/Pkg.jl/v1/compatibility/#Caret-specifiers-1)). So we would need to use a `~` to disallow newer minor versions.
I will keep things as they are for now because this was just a meta comment but we will consider it. Thanks for the pointer!

PR #2146, opened 2020-05-01
Generalize 1D zonotope constraints_list

schillic:
This fixes a regression uncovered in https://github.com/JuliaReach/MathematicalPredicates.jl/pull/19.

### DIFF
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -498,12 +498,17 @@ function constraints_list(Z::AbstractZonotope{N}; check_full_rank::Bool=true
     # special handling of 1D case
     if n == 1
         if p > 1
-            error("1D-zonotope constraints currently only support a single " *
-                  "generator")
+            Z_1g = remove_zero_generators(Z)
+            if ngens(Z_1g) > 1
+                error("1D-zonotope constraints currently only support a " *
+                      "single generator")
+            end
+        else
+            Z_1g = Z
         end
 
-        c = center(Z)[1]
-        g = G[:, 1][1]
+        c = center(Z_1g)[1]
+        g = genmat(Z_1g)[:, 1][1]
         constraints = [LinearConstraint([N(1)], c + g),
                        LinearConstraint([N(-1)], g - c)]
         return constraints
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -508,7 +508,7 @@ function constraints_list(Z::AbstractZonotope{N}; check_full_rank::Bool=true
         end
 
         c = center(Z_1g)[1]
-        g = genmat(Z_1g)[:, 1][1]
+        g = genmat(Z_1g)[1, 1]
         constraints = [LinearConstraint([N(1)], c + g),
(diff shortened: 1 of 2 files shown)


src/Interfaces/AbstractZonotope.jl
```diff
@@ -498,12 +498,17 @@ function constraints_list(Z::AbstractZonotope{N}; check_full_rank::Bool=true
...
+                      "single generator")
+            end
+        else
+            Z_1g = Z
         end
 
-        c = center(Z)[1]
-        g = G[:, 1][1]
+        c = center(Z_1g)[1]
+        g = genmat(Z_1g)[:, 1][1]
```
mforets, line 511:
```suggestion
        g = genmat(Z_1g)[1, 1]
```
(doesn't create the temporary array)

```
julia> m = hcat(1.0)
1×1 Array{Float64,2}:
 1.0

julia> @allocated m[:, 1][1]
7469351

julia> @allocated m[1, 1]
1664
```

mforets:
For the `n=1` case, i would try keeping the maximum of the first row of the generators matrix. This covers the case with `p > 1` and should be cheaper than calling `remove_zero_generators` (and we can remove the error too).

schillic:
> For the n=1 case, i would try keeping the maximum in absolute value of the first row of the generators matrix.

I think it should be the sum of the absolute values.

mforets:
Indeed :) it's `sum(abs.(x))`.

src/Interfaces/AbstractZonotope.jl
```diff
@@ -497,13 +497,8 @@ function constraints_list(Z::AbstractZonotope{N}; check_full_rank::Bool=true
...
     # special handling of 1D case
     if n == 1
-        if p > 1
-            error("1D-zonotope constraints currently only support a single " *
-                  "generator")
-        end
-
-        c = center(Z)[1]
-        g = G[:, 1][1]
+        c = center(Z, 1)
```
mforets, line 500: Zonotopes don't have this `center(Z, 1)` function (just tried it in `master`).

src/Interfaces/AbstractZonotope.jl
```diff
@@ -497,13 +497,8 @@ function constraints_list(Z::AbstractZonotope{N}; check_full_rank::Bool=true
...
+        g = sum(abs.(G[1, :]))
```
mforets, line 501:
```suggestion
        g = sum(abs.(view(G, 1, :)))
```

src/Interfaces/AbstractZonotope.jl
schillic: Oh, thanks :sweat:

test/unit_Zonotope.jl
```diff
@@ -189,6 +189,10 @@ for N in [Float64, Rational{Int}, Float32]
     B = BallInf(zeros(N, 3), N(1))  # equivalent to Z
     constraints = constraints_list(Z)
     @test constraints isa Vector{<:HalfSpace{N}} && length(constraints) == 6
+
+    # 1D projection removes zero generators automatically
```
schillic, line 193:
```suggestion
    # 1D projection works correctly even with zero generators
```

mforets:
this is good to go from my side

PR #2159, opened 2020-05-30
Inherit all functions for AbstractCentrallySymmetricPolytope

schillic:
Background: `center(rand(Ball2), 1)` works but `center(rand(BallInf), 1)` did not work - scandalous!


schillic:
The error seems unrelated (a recent `Polyhedra` update). I will try to fix that elsewhere first.

PR #2160, opened 2020-05-30
Distance between two boxes

schillic:
I need this for some other code base but I prototyped it here so I thought I should add it here as well. I was not sure about the function name `distance`. [It seems to be a standard distance notion and not have a specific name.](https://en.wikipedia.org/wiki/Distance#Distances_between_sets_and_between_a_point_and_a_set) I am open to suggestions.

The algorithm consists of two steps: compute a shortest vector and then take the norm of that vector. Is there interest in the first part as a separate function?

Requires #2159

### DIFF
--- a/src/Approximations/distance.jl
+++ b/src/Approximations/distance.jl
@@ -4,7 +4,11 @@ export distance
     distance(H1::AbstractHyperrectangle{N}, H2::AbstractHyperrectangle{N};
              p::Real=2) where {N<:Real}
 
-Compute the standard distance between two hyperrectangular sets.
+Compute the standard distance between two hyperrectangular sets, defined as
+
+```math
+    \\inf_{x \\in H_1, y \\in H_2} \\{ d(x, y) \\}.
+```
 
 ### Input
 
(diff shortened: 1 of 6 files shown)


src/Approximations/distance.jl
```diff
@@ -0,0 +1,42 @@
...
+        # find relative position in dimension i
+        # (if c1 == c2, the results are equivalent independent of the branch)
+        if center(H1, i) >= center(H2, i)
+            lhs = low(H1, i)
+            rhs = high(H2, i)
+        else
+            lhs = low(H2, i)
+            rhs = high(H1, i)
+        end
+        if lhs <= rhs
```
mforets, line 35: here it's probably preferable to use a robust check. otherwise, you can have `lhs` which is bigger than `rhs` by a tiny amount (say, smaller than the `ztol`) but still not return `zero(N)`. i would find that incompatible with the rest of the library.

mforets:
About the name, his notion is not a distance in the usual sense as the wikipedia article shows (eg. it doesn't satisfy the triangle inequality). I'm fine to use `distance` here so long as we link to Hausdorff distance function..

> The algorithm consists of two steps: compute a shortest vector and then take the norm of that vector. Is there interest in the first part as a separate function?

as you prefer. we can refactor if that function is useful. actually we could have `distance` function which dispatches on a vector and a set, which corresponds to this notion iiuc.

PR #2163, opened 2020-05-31
#2152 - Vertices list of a UnionSet and ConvexHull

### LINKED ISSUE #2152: Vertices list of a UnionSetArray
See [this comment](https://github.com/JuliaReach/ReachabilityAnalysis.jl/pull/169#discussion_r424468043).

schillic:
Closes #2152.


schillic:
> the linked issue (JuliaReach/ReachabilityAnalysis.jl#169 (comment)) was about obtaining the list of vertices of a ConvexHullArray.

Duh, the issue #2152 is called "Vertices list of a UnionSetArray" -.-

I was also concerned about the convex hull but I just copied it from your code. What about removing the arguments for the convex hull altogether? IMO `convex_hull` should be called explicitly by the user or else a `ConvexHullArray` should be created in the first place.

mforets:
ok. i think it's also fine to keep the option.

Issue #2105, opened 2020-04-09
Use tolerance in hyperrectangles intersection


mforets:
https://github.com/JuliaReach/LazySets.jl/blob/db28fc77f8e8eb4b634409c22c61eb7318a659be/src/ConcreteOperations/intersection.jl#L133

schillic:
[A similar comment](https://github.com/JuliaReach/LazySets.jl/pull/2164#discussion_r432971560) here:

https://github.com/JuliaReach/LazySets.jl/blob/db28fc77f8e8eb4b634409c22c61eb7318a659be/src/ConcreteOperations/intersection.jl#L133

PR #2164, opened 2020-05-31
#2133 - Use center(P, i) method

### LINKED ISSUE #2133: Use center(P, i) method
This is a follow up of https://github.com/JuliaReach/LazySets.jl/issues/1578.

Ideas where this method would be useful:

- https://github.com/JuliaReach/LazySets.jl/issues/1578#issuecomment-519036574
- https://github.com/JuliaReach/LazySets.jl/issues/1578#issue-477838116

schillic:
Closes #2133.

### DIFF
--- a/src/ConcreteOperations/intersection.jl
+++ b/src/ConcreteOperations/intersection.jl
@@ -209,24 +209,16 @@ function intersection(H1::AbstractHyperrectangle{N},
                       H2::AbstractHyperrectangle{N}
                      ) where {N<:Real}
     n = dim(H1)
-    c1 = center(H1)
-    c2 = center(H2)
-    r1 = radius_hyperrectangle(H1)
-    r2 = radius_hyperrectangle(H2)
-    high = Vector{N}(undef, n)
-    low = Vector{N}(undef, n)
+    v_high = Vector{N}(undef, n)
+    v_low = Vector{N}(undef, n)
     for i in 1:n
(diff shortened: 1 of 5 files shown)


src/ConcreteOperations/intersection.jl
```diff
@@ -209,24 +209,16 @@ function intersection(H1::AbstractHyperrectangle{N},
...
-        high1 = c1[i] + r1[i]
-        low1 = c1[i] - r1[i]
-        high2 = c2[i] + r2[i]
-        low2 = c2[i] - r2[i]
-        high[i] = min(high1, high2)
-        low[i] = max(low1, low2)
-        if high[i] < low[i]
+        v_high[i] = min(high(H1, i), high(H2, i))
+        v_low[i] = max(low(H1, i), low(H2, i))
+        if v_high[i] < v_low[i]
```
mforets, line 217:
```suggestion
        if _leq(v_high[i], v_low[i])
```

schillic: I am not sure this is correct (note the strict inequality). I will move this to #2105.

mforets: yep, this case should be considered carefully. if `v_low` is approx equal (but higher) than `v_high` then the Hyperrectangle will refuse to build it. OTOH, in such cases we may want the difference to be interpreted as zero.

PR #2175, opened 2020-06-08
Fix assertion in linear_map_inverse

schillic:
This fixes a silly mistake from #2171.

### DIFF
--- a/src/Interfaces/AbstractPolyhedron_functions.jl
+++ b/src/Interfaces/AbstractPolyhedron_functions.jl
@@ -690,7 +690,7 @@ end
 
 function linear_map_inverse(Minv::AbstractMatrix{N},
                             P::AbstractPolyhedron{N}) where {N}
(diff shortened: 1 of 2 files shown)


src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -690,7 +690,7 @@ end
 
 function linear_map_inverse(Minv::AbstractMatrix{N},
                             P::AbstractPolyhedron{N}) where {N}
-    @assert size(Minv, 2) == dim(P) "a linear map of size $(size(Minv)) " *
+    @assert size(Minv, 1) == dim(P) "a linear map of size $(size(Minv)) " *
```
mforets, line 693: It's the same.. :)  If `Minv` is invertible => `m = n`.

mforets: It would be more sensible to check that `Minv` is square though, since `_linar_map_inverse_hrep` doesn't make that check i think.

PR #2177, opened 2020-06-11
Approximate rectification with zonotopes

### LINKED ISSUE #2155: Implement rectification operation using zonotopes
cf. Fig. 7 in AI2 paper

SebastianGuadalupe:
Closes #2155


src/Interfaces/AbstractZonotope.jl
```diff
@@ -525,3 +525,52 @@ function constraints_list(Z::AbstractZonotope{N}; check_full_rank::Bool=true
...
+
+This function implements [Theroem 3.1, 1].
+
+[1] *Singh, G., Gehr, T., Mirman, M., Püschel, M., & Vechev, M. (2018). Fast
+and effective robustness certification. In Advances in Neural Information
+Processing Systems (pp. 10802-10813).*
+"""
+function rectify!(Z::AbstractZonotope{N}) where {N}
+    c = Z.center
+    G = Z.generators
```
schillic, line 552: An `AbstractZonotope` generally does not have fields `center` and `generators`. You have to restrict yourself to `Zonotope` or convert to that type first.

src/Interfaces/AbstractZonotope.jl
```diff
@@ -525,3 +525,52 @@ function constraints_list(Z::AbstractZonotope{N}; check_full_rank::Bool=true
     end
     return constraints
 end
+
+"""
+    rectify(Z::AbstractZonotope{N}) where {N}
```
schillic, line 530: I would add the documentation to `rectify` (and not to `rectify!`).

src/Interfaces/AbstractZonotope.jl
```diff
@@ -525,3 +525,52 @@ function constraints_list(Z::AbstractZonotope{N}; check_full_rank::Bool=true
...
+
+- `Z` -- zonotope
+
+### Output
+
+The zonotope obtained by rectifying `Z`.
+
+### Algorithm
+
+This function implements [Theroem 3.1, 1].
```
schillic, line 544:
```suggestion
This function implements [Theorem 3.1, 1].
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -525,3 +525,52 @@ function constraints_list(Z::AbstractZonotope{N}; check_full_rank::Bool=true
...
+        else
+            λ = ux / (ux - lx)
+            μ = - λ * lx / 2
+            c[i] = c[i] * λ + μ
+            G[i, :] = G[i, :] .* λ
+            Gnew[i, i] = μ
+        end
+    end
+    Z = Zonotope(c, hcat(G, Gnew))
+    return remove_zero_generators(Z)
```
schillic, line 573: We had some discussion about this in the past and I think we concluded that we do not remove zero generators (a user can do that manually).

In the other case, it would be more efficient to create the `Zonotope` only once.
```suggestion
    Z = Zonotope(c, remove_zero_columns(hcat(G, Gnew)))
    return Z
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -525,3 +525,52 @@ function constraints_list(Z::AbstractZonotope{N}; check_full_rank::Bool=true
...
+
+### Algorithm
+
+This function implements [Theroem 3.1, 1].
+
+[1] *Singh, G., Gehr, T., Mirman, M., Püschel, M., & Vechev, M. (2018). Fast
+and effective robustness certification. In Advances in Neural Information
+Processing Systems (pp. 10802-10813).*
+"""
+function rectify!(Z::AbstractZonotope{N}) where {N}
```
schillic, line 550: This function may be misleading because it modifies the zonotope but it still returns a new one. I would have expected that no new zonotope is allocated.

src/Interfaces/AbstractZonotope.jl
```diff
@@ -525,3 +525,52 @@ function constraints_list(Z::AbstractZonotope{N}; check_full_rank::Bool=true
...
+            c[i] = c[i] * λ + μ
+            G[i, :] = G[i, :] .* λ
+            Gnew[i, i] = μ
+        end
+    end
+    Z = Zonotope(c, hcat(G, Gnew))
+    return remove_zero_generators(Z)
+end
+
+rectify(Z::AbstractZonotope) = rectify!(copy(Z))
```
schillic, line 576: Better move these methods to the `Approximations` module because they are not exact.

mforets: About the place of this function, in fact i suggest to place this in `Zonotope.jl`. An inconvenience that i find by dispatching on top level set interfaces like that is that there are little assumptions you can make on the type, and you end up hard-coding `Vector{N}` and so on which is in many cases something one *doesn't* wan't to do, because that may change the type of the input zonotope. 

For example, `split` or `reduce_order` zonotope functions are *much* faster if you use small statically sized arrays, and we have versions that specialize on those ([here](https://juliaarrays.github.io/StaticArrays.jl/stable/pages/api/#Indexing-1) and [here](https://github.com/JuliaReach/ReachabilityAnalysis.jl/blob/master/src/Flowpipes/setops.jl#L604) . Taking the vec rep and the mat rep from the dispatch on the abstract type is possible but cumbersome and anit-julian.

In sum, a possible pattern is to mimic the one we have in place the concrete 
linear map function `linear_map` as shown below.

```julia
# in file AbstractZonotope.jl
#  fallback
rectify(Z::AbstractZonotope) = _rectify(Zonotope(center(Z), genmat(Z))

# in file Zonotope.jl, use getter function
function _rectify(Z::Zonotope)
    # .... new method
end

# another _rectify for another zonotope type
function _rectify(Z::Zonotope)
    # .... new method
end
```

src/Interfaces/AbstractZonotope.jl
mforets: I agree with the proposed change. In fact i would only check  `Gnew`, if this is possible

```julia
hcat(G, remove_zero_columns(Gnew))
```

mforets: In fact it would be a better algorithm to just store the indices `i` where  `Gnew` is different from zero.

src/Interfaces/AbstractZonotope.jl
mforets: good point.  we should either modify and return `Z` or change the signature to `rectify!(Zout, Z)`. i prefer the former if possible

src/Approximations/overapproximate.jl
```diff
@@ -1635,3 +1635,66 @@ function overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
     end
     return Zonotope(c, G)
 end
+
+"""
+    overapproximate(r::Rectification{N, <:Zonotope{N}}, ::Type{<:AbstractZonotope}) where {N}
```
mforets, line 1640: remember to add this new function to the docs

src/Approximations/overapproximate.jl
```diff
@@ -1635,3 +1635,66 @@ function overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
...
+
```
mforets, line 1641: also, the docstring doesn't match the function's signature because of the last  change

src/Approximations/overapproximate.jl
```diff
@@ -1635,3 +1635,66 @@ function overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
...
+Overapproximation of the rectification of a zonotopic set.
+
+### Input
+
+- `r` -- Lazy Rectification of a zonotope
```
mforets, line 1646:
```suggestion
- `r` -- lazy rectification of a zonotopic set
```

src/Approximations/overapproximate.jl
```diff
@@ -1635,3 +1635,66 @@ function overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
...
+            end
+            push!(row_idx, i)
+            push!(μ_idx, μ)
+        end
+    end
+
+    q = length(row_idx)
+    Gnew = zeros(N, n, q)
+    j = 1
+    for i in row_idx
```
mforets, line 1694:
```suggestion
    @inbounds for i in row_idx
```

src/Approximations/overapproximate.jl
```diff
@@ -1635,3 +1635,66 @@ function overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
...
+Overapproximation of the rectification of a zonotopic set.
+
+### Input
+
+- `r` -- Lazy Rectification of a zonotope
+- `Zonotope` -- type for dispatch
+
+### Output
+
+The zonotope overapproximation of the set obtained by rectifying `Z`.
```
mforets, line 1651:
```suggestion
A zonotope overapproximation of the set obtained by rectifying `Z`.
```
because such overapproximation is not unique :)

src/Approximations/overapproximate.jl
SebastianGuadalupe: is it needed for a multiple dispatch?

mforets: what i meant is to "register" this function in the appropriate place inside `docs/...`

mforets: since we build docs with an option like `force=true` this implies that if there is function which has a docstring and it *has not* been added to the docs then they will fail.

schillic: In this case we only have the function `overapproximate` in the docs (and not each individual method), so it will work.

src/Approximations/overapproximate.jl
```diff
@@ -1635,3 +1635,66 @@ function overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
...
+    G = genmat(Z)
+    n, m = size(G)
+    H = overapproximate(Z, Hyperrectangle)
+    row_idx = Vector{Int}()
+    μ_idx = Vector{N}()
+
+    @inbounds for i in 1:n
+        lx, ux = H.center[i] - H.radius[i], H.center[i] + H.radius[i]
+        if !_leq(lx, zero(N))
+            nothing
```
schillic, line 1673: `G[i, :]` stays undefined here.

src/Approximations/overapproximate.jl
```diff
@@ -1635,3 +1635,66 @@ function overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
...
+    Z = copy(r.X)
+    c = center(Z)
+    G = genmat(Z)
+    n, m = size(G)
+    H = overapproximate(Z, Hyperrectangle)
+    row_idx = Vector{Int}()
+    μ_idx = Vector{N}()
+
+    @inbounds for i in 1:n
+        lx, ux = H.center[i] - H.radius[i], H.center[i] + H.radius[i]
```
schillic, line 1671:
```suggestion
        lx, ux = low(H, i), high(H, i)
```

test/unit_overapproximate.jl
```diff
@@ -120,6 +120,14 @@ for N in [Float64, Rational{Int}, Float32]
...
     am = AffineMap(A, X, b)
     @test overapproximate(am, Hyperrectangle) == Hyperrectangle(N[3, 5, 7], N[2, 3, 4])
+
+    # rectification of a zonotope
+    Z = Zonotope(N[0, 0], N[1 1 1; 1 1 0])
+    @test overapproximate(Rectification(Z), Zonotope) == Zonotope(N[3/4, 1/2], N[1/2 1/2 1/2 3/4 0; 1/2 1/2 0 0 1/2])
+    Z = Zonotope(N[-2, -2], N[1 1; 1 0])
+    @test overapproximate(Rectification(Z), Zonotope) == convert(Zonotope, Singleton(N[0, 0]))
+    Z = Zonotope(N[-2, 2], N[1 1; 1 0])
+    @test overapproximate(Rectification(Z), Zonotope) == remove_zero_generators(Zonotope([0, 2.], [0. 0; 1 0]))
```
schillic, line 130:
```suggestion
    @test overapproximate(Rectification(Z), Zonotope) == Zonotope(N[0, 2], hcat(N[0, 1]))
```

src/Approximations/overapproximate.jl
SebastianGuadalupe: What do you mean? G is defined in line 1664

SebastianGuadalupe:
I added again the `remove_zero_columns` function to the output because the algorithm is very likely to generate zero generators in the lines 1674-1678.

src/Approximations/overapproximate.jl
schillic: Oh, I see.

src/Approximations/overapproximate.jl
```diff
@@ -1635,3 +1635,66 @@ function overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
...
+        if !_leq(lx, zero(N))
+            nothing
+        elseif _leq(ux, zero(N)) || isapproxzero(lx)
```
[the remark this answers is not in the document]
schillic:
```julia
julia> Z = Zonotope([1.0, 0], hcat([1.0, 0]))
Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.0, 0.0], [1.0; 0.0])

julia> W = overapproximate(Rectification(Z), Zonotope)
Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([0.0, 0.0], Array{Float64}(undef,2,0))
```
`Z` is the Cartesian product of the interval [0, 2] and the singleton [0] (a flat hyperrectangle) that is completely contained in the positive quadrant. Hence `W` should be equivalent to `Z`. With this bugfix:
```julia
julia> W = overapproximate(Rectification(Z), Zonotope)
Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.0, 0.0], [1.0; 0.0])
```

schillic:
Apparently there is a breaking test in Julia v1.0. But that part should not be affected by this PR. Maybe there was some package update.

```julia
LazySets.Polygon: Test Failed at /home/travis/build/JuliaReach/LazySets.jl/test/unit_Polygon.jl:449
  Expression: n == length(p2.constraints) == length(po2.constraints) == length(po2.constraints)
   Evaluated: 5 == 4 == 4 == 4
Stacktrace:
 [1] top-level scope at /home/travis/build/JuliaReach/LazySets.jl/test/unit_Polygon.jl:449
```

I am fine with merging here and investigating the breaking test separately.

src/Approximations/overapproximate.jl
```diff
@@ -1635,3 +1635,71 @@ function overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
...
+### Algorithm
+
+This function implements [Theorem 3.1, 1].
+
+[1] *Singh, G., Gehr, T., Mirman, M., Püschel, M., & Vechev, M. (2018). Fast
+and effective robustness certification. In Advances in Neural Information
+Processing Systems (pp. 10802-10813).*
+"""
+function overapproximate(r::Rectification{N, <:AbstractZonotope{N}}, ::Type{<:Zonotope}) where {N}
+    Z = copy(r.X)
```
mforets, line 1662: @SebastianGuadalupe can you rebase and use the getter function for `r.X`?

PR #2179, opened 2020-06-12
#2178 - Add getter function for lazy rectification

### LINKED ISSUE #2178: Getter function for lazy rectification
Proposed name: `set`, which returns the `X` field.

SebastianGuadalupe:
Closes #2178.

### DIFF
--- a/src/LazyOperations/Rectification.jl
+++ b/src/LazyOperations/Rectification.jl
@@ -116,7 +116,7 @@ isconvextype(::Type{<:Rectification}) = false
 """
     set(r::Rectification)
 
-Return the set of a rectification.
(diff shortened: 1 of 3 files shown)


src/LazyOperations/Rectification.jl
```diff
@@ -113,6 +113,23 @@ end
 isoperationtype(::Type{<:Rectification}) = true
 isconvextype(::Type{<:Rectification}) = false
 
+"""
+    set(r::Rectification)
+
+Return the set of a rectification.
```
schillic, line 119: I find "the set" ambiguous. Maybe the *original* set?

Issue #2187, opened 2020-06-21
Error in concrete intersection of HPolygon


mforets:
```julia
julia> h = Hyperrectangle([0.9806984373998077, 0.8502036865893472], [0.002214407802010432, 0.0007748971207262034])
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([0.9806984373998077, 0.8502036865893472], [0.002214407802010432, 0.0007748971207262034])

julia> B = Ball2([1.0, 1.0], 0.15)
Ball2{Float64,Array{Float64,1}}([1.0, 1.0], 0.15)

julia> p = overapproximate(B, 1e-4);

julia> isempty(LazySets.intersection(HPolygon(constraints_list(h)), p))
true

julia> isempty(intersection(h, p))
false

julia> LazySets.isdisjoint(h, p)
false

julia> any(v -> v ∈ p, vertices_list(h))
true
```

mforets:
Related to this function

```julia
A = rand(Zonotope)
X = rand(Zonotope)
intersection(overapproximate(X + A, 1e-3), overapproximate(X + B, 1e-3))

AssertionError: the constraints must have the same normal direction

Stacktrace:
 [1] macro expansion at /home/mforets/.julia/dev/LazySets/src/Assertions/Assertions.jl:23 [inlined]
 [2] is_tighter_same_dir_2D(::HalfSpace{Float64,Array{Float64,1}}, ::HalfSpace{Float64,Array{Float64,1}}; strict::Bool) at /home/mforets/.julia/dev/LazySets/src/Sets/HalfSpace.jl:460
 [3] is_tighter_same_dir_2D at /home/mforets/.julia/dev/LazySets/src/Sets/HalfSpace.jl:459 [inlined]
 [4] intersection(::HPolygon{Float64,Array{Float64,1}}, ::HPolygon{Float64,Array{Float64,1}}, ::Bool) at /home/mforets/.julia/dev/LazySets/src/ConcreteOperations/intersection.jl:515
 [5] intersection(::HPolygon{Float64,Array{Float64,1}}, ::HPolygon{Float64,Array{Float64,1}}) at /home/mforets/.julia/dev/LazySets/src/ConcreteOperations/intersection.jl:469
 [6] top-level scope at In[134]:3

```

PR #2191, opened 2020-06-24
Add floating-point tolerance to is_tighter_same_dir_2D

schillic:
The function `is_tighter_same_dir_2D` is called from here:

https://github.com/JuliaReach/LazySets.jl/blob/d16f98bb4072debccd12b24b7dec784396f80364/src/Interfaces/AbstractHPolygon.jl#L309-L313

Recently (#2172) that code was made more robust wrt. floating-point precision. But the function `is_tighter_same_dir_2D` was forgotten.


mforets:
tihs could be related to https://github.com/JuliaReach/LazySets.jl/issues/2187 ?

schillic:
> tihs could be related to #2187 ?

I still get the same behavior, sorry :disappointed:

Issue #718, opened 2018-10-06
Convert expressions to sets


mforets:
In SX.jl we have 

```julia
function convert(::Type{LazySets.HalfSpace{N}}, expr::Expr; vars::Vector{SymEngine.Basic}=Basic[]) where {N}
```

(long term) Proposal: pull out these methods to a proper module in LazySets. This functions depend on SymEngine for the symbolic parts.

mforets:
To write down some thoughts on this idea:

- Add a `@var` macro to declare variables (this can be found elsewhere). 
- Add a generic `@set` macro to transform an expression into a set type given some heuristics,

    - interval if it is 1d
    - line
    - hpolygon
    - halfspace
    - hpolytope
    - hpolyhedron
 
- The ambien dimension is given by the number of declare variables. Add functions `clear_vars()`, `get_vars()`, `set_vars(...)`.

Example using multiple variables:

```julia
@var x[1:10] # 10-dimensional ambient space
@set x[1]  + x[3] <= 10 # creates a half-space with normal vector [1, 0, 1, 0, .. , 0]
```

- More examples:

```julia
@var x y
@set(x <= y, y == 0) # probably easier to parse, but i like the next one; creates a hpolyhedron
@set x <= y && y == 0 # not sure about the y == 0 or just let y = 0
@set A*x <= b # creates a hpolytope or hpolyhedron depending on its boundedness
@set Ax <= b # fancy, see MathematicalSystems.add_asterisk
```

- not sure if we want to use `y == 0` or just use `y = 0` as in math

mforets:
cc: @ueliwechsler

ueliwechsler:
> @set x <= y && y == 0

Parsing-wise this should be fine if you use `==`. With  `=` it would be a more annoying to parse.

I might have some time to play around with the this in march if you are not already on it. But I am not sure if I understand what this macro is supposed to be at the end.
Is it a convenience constructor for `HPoyltopes` and similar sets or is it more used for sequential creation of polytopic sets? Because for the first in the form `@set A*x <= b` , there is no need for variables.
What would be an MPV for this macro?

mforets:
[...]

PR #2193, opened 2020-06-24
Add conversions from MT to Hyperplane and HalfSpace

mforets:
This PR is one step towards converting expressions to sets (https://github.com/JuliaReach/LazySets.jl/issues/718). Here there's no macro, but two new constructors for `HalfSpace` and `Hyperplane` that can handle symbolic expressions. The interesting part is that one can do, for instance,

```julia
julia> @variables x[1:10]
(Operation[x₁, x₂, x₃, x₄, x₅, x₆, x₇, x₈, x₉, x₁₀],)

julia> HalfSpace(2x[1] + 5x[4] ≥ -10., x)
HalfSpace{Float64,Array{Float64,1}}([-2.0, 0.0, 0.0, -5.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0], 10.0)
```
to define a halfspace. Note that `<=` and `>=` work as well as `≥`, and they are normalized to the internal representation of `HalfSpace`. The tests show more examples.

If no second argument ("ambient variables") is given, the expression is parsed according to the value returned by `get_variables`. One should be careful because in some cases it "reverts the order" (presumably, there is no order :) ). For example,

```julia
julia> vars  = @variables x y

julia> expr = 2x ≤ 5y - 1
(2x < 5y - 1) | (2x == 5y - 1)

julia> get_variables(expr)
2-element Array{Operation,1}:
 x
 y

julia> expr = 2x ≥ 5y - 1
(5y - 1 < 2x) | (5y - 1 == 2x)

julia> get_variables(expr)
2-element Array{Operation,1}:
 y
 x
```

---

Special thanks to @dpsanders who suggested this route and helped with MT :100:


mforets:
The tests pass in v1.4 but fail in julia v1.0. I don't think we can do much about that because it is a compat issue with the new optional dependency added in this PR..

```julia
$ julia --check-bounds=yes --color=yes -e "if VERSION < v\"0.7.0-DEV.5183\"; Pkg.test(\"${JL_PKG}\", coverage=true); else using Pkg; Pkg.test(coverage=true); end"
   Testing LazySets
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package SymbolicUtils [d1185830]:
 SymbolicUtils [d1185830] log:
 ├─possible versions are: [0.1.0-0.1.2, 0.2.0-0.2.1, 0.3.0-0.3.4, 0.4.0-0.4.2] or uninstalled
 ├─restricted by compatibility requirements with SpecialFunctions [276daf66] to versions: uninstalled
 │ └─SpecialFunctions [276daf66] log:
 │   ├─possible versions are: [0.7.0-0.7.2, 0.8.0, 0.9.0, 0.10.0-0.10.3] or uninstalled
 │   └─restricted to versions 0.8.0 by an explicit requirement, leaving only versions 0.8.0
 └─restricted by compatibility requirements with ModelingToolkit [961ee093] to versions: 0.3.3-0.3.4 — no versions left
   └─ModelingToolkit [961ee093] log:
     ├─possible versions are: [0.0.1-0.0.2, 0.1.0, 0.2.0, 0.4.0, 0.5.0, 0.6.0-0.6.1, 0.6.3-0.6.4, 0.7.0-0.7.2, 0.8.0, 0.9.0-0.9.1, 0.10.0, 1.0.0-1.0.3, 1.1.0-1.1.3, 1.2.0-1.2.10, 1.3.0-1.3.3, 1.4.0-1.4.3, 2.0.0, 3.0.0-3.0.2, 3.1.0-3.1.1, 3.2.0, 3.3.0, 3.4.0, 3.5.0, 3.6.0-3.6.4, 3.7.0-3.7.1, 3.8.0, 3.9.0, 3.10.0-3.10.2] or uninstalled
     └─restricted to versions 3.10.0-3 by an explicit requirement, leaving only versions 3.10.0-3.10.2
```

mforets:
Until we solve this version issue in LazySets.jl, this functionality is now available in RA (https://github.com/JuliaReach/ReachabilityAnalysis.jl/pull/248).

Project.toml
```diff
@@ -32,6 +32,7 @@ JuMP = "0.21"
 Makie = "0.9"
 MathProgBase = "0.7"
 MiniQhull = "0.1"
+ModelingToolkit = "3.10"
```
mforets, line 35:
```suggestion
ModelingToolkit = "0.4, 3.10"
```

Project.toml
```diff
@@ -32,6 +32,7 @@ JuMP = "0.21"
 Makie = "0.9"
 MathProgBase = "0.7"
 MiniQhull = "0.1"
+ModelingToolkit = "0.4, 3.10"
```
schillic, line 35:
```suggestion
ModelingToolkit = "1.1.3, 3.10"
```
Not sure if this helps but this is the last version for Julia v1.0.

src/Sets/HalfSpace.jl
```diff
@@ -519,3 +519,119 @@ end
...
+        a, b = expr.args[2].args
+        sexpr = simplify(a - b)
+
+    elseif (expr.op == |) && (expr.args[1].op == >)
+        a, b = expr.args[1].args
+        sexpr = simplify(b - a)
+
+    elseif (expr.op == |) && (expr.args[2].op == >)
+        a, b = expr.args[2].args
+        sexpr = simplify(b - a)
```
schillic, line 621: I do not understand these cases.

src/Sets/Hyperplane.jl
```diff
@@ -478,3 +478,79 @@ function translate(hp::Hyperplane{N}, v::AbstractVector{N}; share::Bool=false
...
+Return the hyperplane given by a symbolic expression.
+
+### Input
+
+- `expr` -- symbolic expression that describes a hyperplane
+- `vars` -- (optional, default: `get_variables(expr)`), if an array of variables is given,
+            use those as the ambient variables in the set with respect to which derivations
+            take place; otherwise, use only the variables which appear in the given
+            expression (but be careful because the order may change; see `Notes` below fo details)
+- `N`    -- (optional, default: `Float64`) the numeric type of the returned half-space
```
schillic, line 500:
```suggestion
- `N`    -- (optional, default: `Float64`) the numeric type of the returned hyperplane
```

src/Sets/Hyperplane.jl
```diff
@@ -478,3 +478,79 @@ function translate(hp::Hyperplane{N}, v::AbstractVector{N}; share::Bool=false
...
+
+Return the hyperplane given by a symbolic expression.
+
+### Input
+
+- `expr` -- symbolic expression that describes a hyperplane
+- `vars` -- (optional, default: `get_variables(expr)`), if an array of variables is given,
+            use those as the ambient variables in the set with respect to which derivations
+            take place; otherwise, use only the variables which appear in the given
+            expression (but be careful because the order may change; see `Notes` below fo details)
```
schillic, line 499: There is no `Notes` section and also no example.

src/Sets/Hyperplane.jl
```diff
@@ -478,3 +478,79 @@ function translate(hp::Hyperplane{N}, v::AbstractVector{N}; share::Bool=false
...
+
+julia> Hyperplane(x[1] == x[2], x)
+Hyperplane{Float64,Array{Float64,1}}([1.0, -1.0, 0.0, 0.0], -0.0)
+```
+
+### Algorithm
+
+It is assumed that the expression is of the form
+`EXPR0: α*x1 + ⋯ + α*xn + γ CMP β*x1 + ⋯ + β*xn + δ`,
+where `CMP` is `==`.
```
schillic, line 531:
```suggestion
`EXPR0: α*x1 + ⋯ + α*xn + γ == β*x1 + ⋯ + β*xn + δ`.
```

src/Sets/Hyperplane.jl
```diff
@@ -478,3 +478,79 @@ function translate(hp::Hyperplane{N}, v::AbstractVector{N}; share::Bool=false
...
+    # simplify to the form a*x + β == 0
+    a, b = expr.args
+    sexpr = simplify(a - b)
+
+    # compute the linear coefficients by taking first order derivatives
+    coeffs = [N(α.value) for α in gradient(sexpr, collect(vars))]
+
+    # get the constant term by expression substitution
+    dvars = Dict(to_symbolic(vi) => zero(N) for vi in vars)
+    β = -N(ModelingToolkit.SymbolicUtils.substitute(to_symbolic(sexpr), dvars, fold=true))
```
schillic, line 551: Maybe share this code with `HalfSpace` for easier refactorings.

mforets: yes indeed. i can do that in a follow up PR. we can also check that the expression correctly defines a half-space (see [this issue](https://github.com/SciML/ModelingToolkit.jl/issues/470)). 

otoh the implementation i using some internal functions (`ModelingToolkit.SymbolicUtils`) that were needed at the time of writing this PR, however, they have bee fixed in a more recent ModelingToolkit release.

src/Sets/HalfSpace.jl
mforets: this is because i saw that sometimes the inequality is stored as the first argument but for some other cases it is stored as the second argument. i can go through examples in our next LazySets devs call!

src/Sets/HalfSpace.jl
```diff
@@ -519,3 +519,119 @@ end
...
+
+Return the half-space given by a symbolic expression.
+
+### Input
+
+- `expr` -- symbolic expression that describes a half-space
+- `vars` -- (optional, default: `get_variables(expr)`), if an array of variables is given,
+            use those as the ambient variables in the set with respect to which derivations
+            take place; otherwise, use only the variables which appear in the given
+            expression (but be careful because the order may change; see `Notes` below fo details)
```
mforets, line 541:
```suggestion
            expression (but be careful because the order may change; see the examples below for details)
```

src/Sets/Hyperplane.jl
```diff
@@ -478,3 +478,78 @@ function translate(hp::Hyperplane{N}, v::AbstractVector{N}; share::Bool=false
...
+
+Return the hyperplane given by a symbolic expression.
+
+### Input
+
+- `expr` -- symbolic expression that describes a hyperplane
+- `vars` -- (optional, default: `get_variables(expr)`), if an array of variables is given,
+            use those as the ambient variables in the set with respect to which derivations
+            take place; otherwise, use only the variables which appear in the given
+            expression (but be careful because the order may change; see `Notes` below fo details)
```
mforets, line 499:
```suggestion
            expression (but be careful because the order may change; in doubt, pass `vars` explicitly)
```

PR #2194, opened 2020-06-24
Always use constraint representation in linear_map_cartesian_product

schillic:
This allows to compute the linear map with a singular matrix of an unbounded polyhedral Cartesian product (and should often be faster in the bounded case).


schillic:
> the implementation doesn't seem to use anything particular abut cartesian poducts. so we could later generalize this idea to other lazysets. for example, we can't compute the concrete linear map of a minkowski sum, although their list of constraints is available

True, and I did not write this method (just modified it) :smile:
There is #996 that sounds like what you suggest.

PR #2197, opened 2020-06-25
#2089 - Use tolerance in membership test of Polyhedron

### LINKED ISSUE #2089: Use tolerance in membership test of Polyhedron
https://github.com/JuliaReach/LazySets.jl/blob/894f77bd2309a80e02d5be50acfb6dbc7c617947/src/Interfaces/AbstractPolyhedron_functions.jl#L50

SebastianGuadalupe:
Closes #2089.

### DIFF
--- a/test/unit_Polyhedron.jl
+++ b/test/unit_Polyhedron.jl
@@ -51,6 +51,7 @@ for N in [Float64, Rational{Int}, Float32]
 
     # membership
     @test N[5 / 4, 7 / 4] ∈ p && N[4, 1] ∉ p
+    @test [Inf, Inf] ∉ p
 
     # constrained dimensions
     @test constrained_dimensions(p) == [1, 2]
@@ -133,7 +134,7 @@ for N in [Float64]
     # an_element
     P = HPolyhedron([HalfSpace(N[3//50, -4//10], N(1)),
                      HalfSpace(N[-1//50, 1//10], N(-1))])
-    @test an_element(P) ∈ P
+    @test_broken an_element(P) ∈ P
 
     # boundedness
     @test isbounded(p)
--- a/test/unit_Polyhedron.jl
+++ b/test/unit_Polyhedron.jl
@@ -51,7 +51,6 @@ for N in [Float64, Rational{Int}, Float32]
 
     # membership
     @test N[5 / 4, 7 / 4] ∈ p && N[4, 1] ∉ p
-    @test [Inf, Inf] ∉ p
 
     # constrained dimensions
     @test constrained_dimensions(p) == [1, 2]
@@ -130,6 +129,9 @@ for N in [Float64]
     @test σ(d, p) == N[-1, 1]
     d = N[0, -1]
     @test σ(d, p) == N[0, 0]
+
+    # membership
+    @test [Inf, Inf] ∉ p
 
     # an_element
     P = HPolyhedron([HalfSpace(N[3//50, -4//10], N(1)),
--- a/test/unit_Polyhedron.jl
+++ b/test/unit_Polyhedron.jl
@@ -136,7 +136,7 @@ for N in [Float64]
     # an_element
     P = HPolyhedron([HalfSpace(N[3//50, -4//10], N(1)),
                      HalfSpace(N[-1//50, 1//10], N(-1))])
-    @test_broken an_element(P) ∈ P
+    @test_broken an_element(P) ∈ P # see LazySets.jl/pull/2197
 
     # boundedness
     @test isbounded(p)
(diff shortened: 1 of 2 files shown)


mforets:
```julia
LazySets.Polyhedron: Test Failed at /home/travis/build/JuliaReach/LazySets.jl/test/unit_Polyhedron.jl:136
  Expression: an_element(P) ∈ P
   Evaluated: [Inf, Inf] ∈ HPolyhedron{Float64,Array{Float64,1}}(HalfSpace{Float64,Array{Float64,1}}[HalfSpace{Float64,Array{Float64,1}}([0.06, -0.4], 1.0), HalfSpace{Float64,Array{Float64,1}}([-0.02, 0.1], -1.0)])
Stacktrace:
 [1] top-level scope at /home/travis/build/JuliaReach/LazySets.jl/test/unit_Polyhedron.jl:136
 [2] include(::String) at ./client.jl:439
 [3] macro expansion at /home/travis/build/JuliaReach/LazySets.jl/test/runtests.jl:85 [inlined]
 [4] macro expansion at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Test/src/Test.jl:1113 [inlined]
 [5] macro expansion at /home/travis/build/JuliaReach/LazySets.jl/test/runtests.jl:85 [inlined]
 [6] macro expansion at ./util.jl:175 [inlined]
 [7] top-level scope at /home/travis/build/JuliaReach/LazySets.jl/test/runtests.jl:85
```

schillic:
This is the failing test:
```julia
julia> P = HPolyhedron([HalfSpace(N[3//50, -4//10], N(1)),
                        HalfSpace(N[-1//50, 1//10], N(-1))]);

julia> an_element(P)
2-element Array{Float64,1}:
 Inf
 Inf

julia> an_element(P) ∈ P
false
```

schillic:
Checking for NaN imitates the previous behavior. I am not convinced that this behavior was good but we can address this later.

SebastianGuadalupe:
> Checking for NaN imitates the previous behavior. I am not convinced that this behavior was good but we can address this later.

Are there other cases where the dot product gives you `NaN` other than when the point checked is `Inf`?

schillic:
The problem is that in the example below I would argue that the point should not be contained in `P1`.

```julia
julia> P1 = HPolyhedron([HalfSpace([1.0, 0.0], 0.0)]);

julia> P2 = HPolyhedron([HalfSpace([-1.0, 0.0], 0.0)]);

julia> [Inf, Inf] ∈ P1
true

julia> [Inf, Inf] ∈ P2
true
```

Note that in 1D the behavior is as I would expect it:

```julia
julia> P1 = HPolyhedron([HalfSpace([1.0], 0.0)]);

julia> P2 = HPolyhedron([HalfSpace([-1.0], 0.0)]);

julia> [Inf] ∈ P1
false

julia> [Inf] ∈ P2
true
```

But as I said this is the behavior in `master`, so we can merge this.

mforets:
Can we mark the test as broken?  This can be done with `@test_broken`.

The transformation rule/convention states that

`dot(c.a, x) > c.b <=> !_leq(dp, c.b)`.

schillic:
We can just use a vector with finite entries.

mforets:
> The problem is that in the example below I would argue that the point should not be contained in P1.

yes, i agree. this is in fact an edge case because

```julia
P1 = HPolyhedron([HalfSpace([1.0, 0.0], 0.0)]) # x <= 0
dot(P1.constraints[1].a, [Inf, Inf])
NaN

[Inf, Inf] ∈ P1
true
```

however note that

```julia
P1 = HPolyhedron([HalfSpace([1.0, 1e-20], 0.0)]) # x <= 0
dot(P1.constraints[1].a, [Inf, Inf])
Inf

[Inf, Inf] ∈ P1
false
```

mforets:
in sum i think that making such runtime checks to handle special cases like 0 * Inf is not a good idea (at least, for the "default" behavior of this function).

so i propose to:

- rervert commit "check for NaN"
- keep the test with infinities but mark it as currently consistently broken, `@test_broken !([Inf, Inf] ∈ P1)`
- add a new test that passes, with finite entries 
- add a new test that passes, with infinite entries but non-zero normal vector

mforets:
@schillic  does that list make sense to you?

schillic:
I would be fine with items 1 and 3 only but as you prefer.

schillic:
Still missing:
> * add a new test that passes, with finite entries

SebastianGuadalupe:
There is already a test with finite entries:

mforets:
ok, thanks for checking. then this lgtm after tests pass

test/unit_Polyhedron.jl
```diff
@@ -130,10 +130,13 @@ for N in [Float64]
...
     @test σ(d, p) == N[0, 0]
 
+    # membership
+    @test [Inf, Inf] ∉ p
+
     # an_element
     P = HPolyhedron([HalfSpace(N[3//50, -4//10], N(1)),
                      HalfSpace(N[-1//50, 1//10], N(-1))])
-    @test an_element(P) ∈ P
+    @test_broken an_element(P) ∈ P
```
mforets, line 139:
```suggestion
    @test_broken an_element(P) ∈ P # see LazySets.jl/pull/2197
```

mforets:
tests passed. thanks @SebastianGuadalupe !

PR #2199, opened 2020-06-25
Apply tricks for 2d zonotope vertex enumeration

yupbank:
when generators are all in first/third quadrant

### DIFF
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -407,6 +407,11 @@ factors are ``ξ_i ∈ \\{-1, 1\\}``.
 There are at most ``2^p`` distinct vertices. Use the flag `apply_convex_hull` to
 control whether a convex hull algorithm is applied to the vertices computed by
 this method; otherwise, redundant vertices may be present.
+
+We use a trick to speed up enumerating vertices of 2-dimensional zonotopic
+sets with all generators in the first quadrant or third quadrant (same sign).
+Namely, sort the generators in angle and add them clockwise in increasing
+order and anticlockwise in decreasing order.
 """
 function vertices_list(Z::AbstractZonotope{N};
                        apply_convex_hull::Bool=true) where {N<:Real}
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -425,7 +425,7 @@ function vertices_list(Z::AbstractZonotope{N};
         if sum(abs, G) == abs(sum(G))
             sorted_G = sortslices(G, dims=2, by=x->atan(x[2], x[1]))
             index = ones(N, p, 2*p)
-            for i in 1:p
+            @inbounds for i in 1:p
                 index[i, i+1:i+p-1] *= -1
             end
             index[:, 1] *= -1
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -422,11 +422,15 @@ function vertices_list(Z::AbstractZonotope{N};
         return [c]
     end
     if n == 2
-      if all(G .> 0) | all(G .< 0)
-           return hcat(-sum(G, dims=2),
-                        2*cumsum(sortslices(G,dims=2,by=x->x[2]/x[1]), dims=2)[:, 1:size(G,2)-1] .- sum(G, dims=2),
-                        2*cumsum(sortslices(G,dims=2,by=x->x[1]/x[2]), dims=2) .- sum(G, dims=2)
-           ) .+ c
+      if same_sign(G)
+        sorted_G = sortslices(G, dims=2,by=x->atan(x[2], x[1]))
+        index = ones(p, 2*p)
+        for i in 1:p
+          index[i, i+1:i+p-1] *= -1
+        end
+        index[:, 1] *= -1
+        V = sorted_G * index .+ c
+        return [V[:, i] for i in 1:2*p]
       end
     end
 
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -432,7 +432,7 @@ function vertices_list(Z::AbstractZonotope{N};
             @inbounds for i in 1:p
                 index[i, i+1:i+p-1] .= -one(N)
             end
-            index[:, 1] = -one(N)
+            index[:, 1] .= -one(N)
             V = sorted_G * index .+ c
             return [V[:, i] for i in 1:2*p]
         end
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -418,10 +418,10 @@ function vertices_list(Z::AbstractZonotope{N};
     end
     if n == 2
       if all(G .> 0) | all(G .< 0)
-	    return hcat(zeros(N, n, 1),
-                        cumsum(sortslices(G,dims=2,by=x->x[2]/x[1]), dims=2)[:, 1:size(G,2)-1],
-                        cumsum(sortslices(G,dims=2,by=x->x[1]/x[2]), dims=2)
-	    )
+           return hcat(-sum(G, dims=2),
+                        2*cumsum(sortslices(G,dims=2,by=x->x[2]/x[1]), dims=2)[:, 1:size(G,2)-1] .- sum(G, dims=2),
+                        2*cumsum(sortslices(G,dims=2,by=x->x[1]/x[2]), dims=2) .- sum(G, dims=2)
+           ) .+ c
       end
     end
 
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -422,16 +422,16 @@ function vertices_list(Z::AbstractZonotope{N};
         return [c]
     end
     if n == 2 && p > 2 && apply_convex_hull
-      if sum(abs, G) == abs(sum(G))
-        sorted_G = sortslices(G, dims=2,by=x->atan(x[2], x[1]))
-        index = ones(p, 2*p)
-        for i in 1:p
-          index[i, i+1:i+p-1] *= -1
+        if sum(abs, G) == abs(sum(G))
+            sorted_G = sortslices(G, dims=2,by=x->atan(x[2], x[1]))
+            index = ones(p, 2*p)
+            for i in 1:p
+                index[i, i+1:i+p-1] *= -1
+            end
+            index[:, 1] *= -1
+            V = sorted_G * index .+ c
+            return [V[:, i] for i in 1:2*p]
         end
-        index[:, 1] *= -1
-        V = sorted_G * index .+ c
-        return [V[:, i] for i in 1:2*p]
-      end
     end
 
     vlist = Vector{Vector{N}}()
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -423,7 +423,7 @@ function vertices_list(Z::AbstractZonotope{N};
     end
     if n == 2 && p > 2 && apply_convex_hull
         if sum(abs, G) == abs(sum(G))
-            sorted_G = sortslices(G, dims=2,by=x->atan(x[2], x[1]))
+            sorted_G = sortslices(G, dims=2, by=x->atan(x[2], x[1]))
             index = ones(p, 2*p)
             for i in 1:p
                 index[i, i+1:i+p-1] *= -1
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -421,7 +421,7 @@ function vertices_list(Z::AbstractZonotope{N};
     if p == 0
         return [c]
     end
-    if n == 2
+    if n == 2 & p > 2
       if same_sign(G)
         sorted_G = sortslices(G, dims=2,by=x->atan(x[2], x[1]))
         index = ones(p, 2*p)
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -412,9 +412,17 @@ function vertices_list(Z::AbstractZonotope{N};
                        apply_convex_hull::Bool=true) where {N<:Real}
     c = center(Z)
     G = remove_zero_columns(genmat(Z))
-    p = size(G, 2)
+    n, p = size(G)
     if p == 0
         return [c]
+    end
+    if n == 2
+      if all(G .> 0) | all(G .< 0)
+	    return hcat(zeros(N, n, 1),
+                        cumsum(sortslices(G,dims=2,by=x->x[2]/x[1]), dims=2)[:, 1:size(G,2)-1],
+                        cumsum(sortslices(G,dims=2,by=x->x[1]/x[2]), dims=2)
+	    )
+      end
     end
 
     vlist = Vector{Vector{N}}()
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -426,7 +426,7 @@ function vertices_list(Z::AbstractZonotope{N};
             sorted_G = sortslices(G, dims=2, by=x->atan(x[2], x[1]))
             index = ones(N, p, 2*p)
             @inbounds for i in 1:p
-                index[i, i+1:i+p-1] *= -1
+                index[i, i+1:i+p-1] .= -one(N)
             end
             index[:, 1] *= -1
             V = sorted_G * index .+ c
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -421,7 +421,7 @@ function vertices_list(Z::AbstractZonotope{N};
     if p == 0
         return [c]
     end
-    if n == 2 & p > 2
+    if n == 2 & p > 2 & apply_convex_hull
       if same_sign(G)
         sorted_G = sortslices(G, dims=2,by=x->atan(x[2], x[1]))
         index = ones(p, 2*p)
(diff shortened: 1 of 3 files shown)


mforets:
Thanks for your contribution! 

Here are some comments to get started:

- As per disussion on gitter, seems that we can generally check  `all(G .> 0) | all(G .< 0)` very efficiently by fusing both checks in one loop over the elements of `G`.

- Could you add a small description of the algorithm? This can be added in the `### Notes` section of the docstring of this  function.

- Why is `zeros(N, n, 1)` always added? For instance, `Z = Zonotope([1.0, 1.0], 0.1 * Matrix(1.0I, 2, 2))` satisfies the assumption on the generators's sign but the origin is not a vertex.

- The call `cumsum(sortslices(G,dims=2,by=x->x[2]/x[1]), dims=2)[:, 1:size(G,2)-1]` allocates a new vector.

mforets:
Thanks for the update. I think now i understand how the algorithm works. Also thanks for the link https://math.stackexchange.com/questions/3356418/how-can-i-generate-boundary-in-2d-space-from-sum-of-finite-vectors/3356460#3356460

---


 About the fourth point in my last comment, i think there are some performance improvements that we can consider. First, i think that we can sort the generators only once. There are also some temporaries that can be cached such as the sum over all columns. We can consider using the internal `right_turn` for the comparisons which i presume may be faster than comparing by the quotients. I'll extend these comments when i have some free time.

For reference below i post the results so far that shows the new method in this branch has much better scalability. Nice!
It does not perform so well on small order zonotopes though. But my guess is that addressing some of the suggestions in the previous paragraph will improve this. In the comparison i also tried using static arrays. There are no perf difference (the algorithms are not specialized for these..).

```julia
function same_sign(G)
    s = sign(first(G))
    for j in size(G, 2)
        for i in size(G, 1)
            s != sign(G[i, j]) && return false
        end
    end
    return true
end

# new function
function _vertices_list_1d_v0(Z)
    G = Z.generators
    c = Z.center
    @assert same_sign(G)

    aux = sum(G, dims=2)
    A = -aux
    B = 2*cumsum(sortslices(G,dims=2,by=x->x[2]/x[1]), dims=2)[:, 1:size(G,2)-1] .- aux
    C = 2*cumsum(sortslices(G,dims=2,by=x->x[1]/x[2]), dims=2) .- aux
    V = hcat(A, B, C) .+ c
    [V[:, i] for i in 1:size(V, 2)]
end
```

Results:

```julia
for p in [2, 4, 6, 8, 10]
    println("p = $p")
    Z = Zonotope(rand(2), rand(2, p))
    Zs = Zonotope(SVector{2}(Z.center), SMatrix{2,p}(Z.generators))
    
    println(" - old - ")
    @btime vertices_list($Z)
    @btime vertices_list($Zs)

    println(" - new - ")
    @btime _vertices_list_1d_v0($Z)
    @btime _vertices_list_1d_v0($Zs)
    println(" ")
end

p = 2
 - old - 
  2.474 μs (45 allocations: 2.91 KiB)
  2.421 μs (53 allocations: 2.84 KiB)
 - new - 
  8.274 μs (68 allocations: 4.05 KiB)
  11.780 μs (70 allocations: 4.00 KiB)
 
p = 4
 - old - 
  10.142 μs (162 allocations: 12.66 KiB)
  9.418 μs (194 allocations: 12.66 KiB)
 - new - 
  8.572 μs (72 allocations: 5.11 KiB)
  12.296 μs (74 allocations: 5.53 KiB)
 
p = 6
 - old - 
  40.495 μs (597 allocations: 55.31 KiB)
  39.151 μs (725 allocations: 56.31 KiB)
 - new - 
  8.890 μs (76 allocations: 6.20 KiB)
  12.815 μs (78 allocations: 7.28 KiB)
 
p = 8
 - old - 
  169.002 μs (2329 allocations: 249.17 KiB)
  160.239 μs (2841 allocations: 257.17 KiB)
 - new - 
  9.192 μs (80 allocations: 7.23 KiB)
  13.395 μs (84 allocations: 9.66 KiB)
 
p = 10
 - old - 
  729.468 μs (9243 allocations: 1.09 MiB)
  692.103 μs (11291 allocations: 1.14 MiB)
 - new - 
  9.590 μs (84 allocations: 8.33 KiB)
  13.908 μs (88 allocations: 12.06 KiB)
```

yupbank:
I made some changes to the `sorting` twice part.

Now we have 
```julia
function same_sign(G)
    s = sign(first(G))
    for j in size(G, 2)
        for i in size(G, 1)
            s != sign(G[i, j]) && return false
        end
    end
    return true
end

# new function
function _vertices_list_1d_v0(Z)
    G = Z.generators
    c = Z.center
    @assert same_sign(G)

    aux = sum(G, dims=2)
    A = -aux
    B = 2*cumsum(sortslices(G,dims=2,by=x->x[2]/x[1]), dims=2)[:, 1:size(G,2)-1] .- aux
    C = 2*cumsum(sortslices(G,dims=2,by=x->x[1]/x[2]), dims=2) .- aux
    V = hcat(A, B, C) .+ c
    [V[:, i] for i in 1:size(V, 2)]
end

# new function
function _vertices_list_1d_v2(Z)
    G = Z.generators
    c = Z.center
    @assert same_sign(G)
    p = size(G, 2)

    sorted_G = sortslices(G, dims=2,by=x->atan(x[2], x[1]))
    index = ones(p, 2*p)
    for i in 1:p
        index[i, i+1:i+p-1] *= -1
    end
    index[:, 1] *= -1
    V = sorted_G * index .+ c
    return [V[:, i] for i in 1:2*p]
end
```

And the performance part
``` Julia 
for p in [2, 3, 4, 5, 6, 8, 10]
    println("p = $p")
    Z = Zonotope(rand(2), rand(2, p))
    Zs = Zonotope(SVector{2}(Z.center), SMatrix{2,p}(Z.generators))
    
    println(" - old - ")
    @btime vertices_list($Z)
    @btime vertices_list($Zs)

    println(" - new - ")
    @btime _vertices_list_1d_v0($Z)
    @btime _vertices_list_1d_v0($Zs)
    println(" ")
    
    println(" - v2 - ")
    @btime _vertices_list_1d_v2($Z)
    @btime _vertices_list_1d_v2($Zs)
    println(" ")
end
```

gives 

```
p = 2
 - old - 
  3.164 μs (45 allocations: 2.83 KiB)
  4.012 μs (53 allocations: 2.83 KiB)
 - new - 
  13.267 μs (93 allocations: 4.73 KiB)
  17.200 μs (77 allocations: 4.05 KiB)
 
 - v2 - 
  5.604 μs (42 allocations: 2.69 KiB)
  4.704 μs (38 allocations: 2.73 KiB)
 
p = 3
 - old - 
  6.374 μs (89 allocations: 6.14 KiB)
  7.758 μs (105 allocations: 6.14 KiB)
 - new - 
  13.745 μs (97 allocations: 5.36 KiB)
  18.372 μs (79 allocations: 4.83 KiB)
 
 - v2 - 
  6.120 μs (47 allocations: 3.38 KiB)
  5.096 μs (42 allocations: 3.50 KiB)
 
p = 4
 - old - 
  11.797 μs (164 allocations: 12.47 KiB)
  14.911 μs (196 allocations: 12.72 KiB)
 - new - 
  14.401 μs (101 allocations: 5.86 KiB)
  18.115 μs (81 allocations: 5.64 KiB)
 
 - v2 - 
  6.719 μs (52 allocations: 4.13 KiB)
  5.679 μs (46 allocations: 4.38 KiB)
 
p = 5
 - old - 
  24.082 μs (310 allocations: 25.84 KiB)
  30.022 μs (374 allocations: 26.34 KiB)
 - new - 
  14.792 μs (105 allocations: 6.48 KiB)
  18.684 μs (83 allocations: 6.55 KiB)
 
 - v2 - 
  6.816 μs (57 allocations: 4.92 KiB)
  5.949 μs (50 allocations: 5.31 KiB)
 
p = 6
 - old - 
  47.217 μs (599 allocations: 54.36 KiB)
  60.108 μs (727 allocations: 56.36 KiB)
 - new - 
  14.088 μs (109 allocations: 6.98 KiB)
  17.940 μs (87 allocations: 7.52 KiB)
 
 - v2 - 
  7.519 μs (62 allocations: 5.83 KiB)
  6.510 μs (54 allocations: 6.39 KiB)
 
p = 8
 - old - 
  193.140 μs (2331 allocations: 245.20 KiB)
  246.819 μs (2843 allocations: 257.20 KiB)
 - new - 
  14.363 μs (117 allocations: 8.11 KiB)
  19.336 μs (91 allocations: 9.67 KiB)
 
 - v2 - 
  8.488 μs (72 allocations: 7.81 KiB)
  7.570 μs (62 allocations: 8.86 KiB)
 
p = 10
 - old - 
  858.423 μs (9245 allocations: 1.08 MiB)
  1.058 ms (11293 allocations: 1.14 MiB)
 - new - 
  14.871 μs (125 allocations: 9.23 KiB)
  19.255 μs (95 allocations: 12.08 KiB)
 
 - v2 - 
  10.250 μs (82 allocations: 10.06 KiB)
  9.230 μs (70 allocations: 11.72 KiB)
```

mforets:
nice :+1: this is in the right direction. 

did you try if it makes a difference to sort using the right hand side rule? i think it would be: `sorted_G = sortslices(G, dims=2,lt=right_turn)`.

mforets:
The new version using a matrix-matrix product after sorting, to avoid sorting twice, is neat :100:   I thought about it and don't have further suggestions for improvement :smile: 

---

From my side this PR looks good. Below are some further considerations which pertain mostly to the "library code" aspects of LazySets.jl, and as such i don't mind if we address this on a follow up PR (but always opening a new issue(s) so we don't forget).

- Outsource the code to check the positivity condition to an independent function, eg. `same_sign`. Add an optional kwarg that defaults to true for this check. Also consider the alternative version given below.

- Outsource to its own function the code to compute the vertices used in this PR, eg `_vertices_list_2D_positive`. This is will be an internal (non-exported) function. Users may want to directly call this function if they know what they are doing ;)

- Faster implementation for checking that all elements in `G` have the same sign (ignoring "0"):

```
function same_sign_v2(G)
    # find sign of first non-zero element
    idx = findfirst(!iszero, G)
    if isnothing(idx)
        return true
    end
    ns = -sign(G[idx])

    # check if all other elements have the same sign
    @inbounds for j in 1:size(G, 2)
        for i in 1:size(G, 1)
            sign(G[i, j]) == ns && return false
        end
    end
    return true
end
```

yupbank:
had a better same sign implementation.
```
function same_sign_v3(x)
    abs(sum(x)) == sum(abs, x)
end
```
tested against `same_sign_v2`

```
using StaticArrays
function same_sign_v2(G)
    # find sign of first non-zero element
    idx = findfirst(!iszero, G)
    if isnothing(idx)
        return true
    end
    ns = -sign(G[idx])

    # check if all other elements have the same sign
    @inbounds for j in 1:size(G, 2)
        for i in 1:size(G, 1)
            sign(G[i, j]) == ns && return false
        end
    end
    return true
end

function same_sign_v3(x)
    abs(sum(x)) == sum(abs, x)
end

for p in [2, 3, 4, 5, 6, 8, 10]
  Z = rand(p, p*10)
  Zs=  SMatrix{p, p*10}(Z)
  println(" - v2 - ")
  @btime same_sign_v2($Z)
  @btime same_sign_v2($Zs)

    println(" - v3 - ")
    @btime same_sign_v3($Z)
    @btime same_sign_v3($Zs)
    println(" ")
end
```
with performance result

```
- v2 - 
  46.120 ns (0 allocations: 0 bytes)
  34.497 ns (0 allocations: 0 bytes)
 - v3 - 
  21.952 ns (0 allocations: 0 bytes)
  0.035 ns (0 allocations: 0 bytes)
 
 - v2 - 
  93.802 ns (0 allocations: 0 bytes)
  75.476 ns (0 allocations: 0 bytes)
 - v3 - 
  34.239 ns (0 allocations: 0 bytes)
  0.035 ns (0 allocations: 0 bytes)
 
 - v2 - 
  149.769 ns (0 allocations: 0 bytes)
  130.070 ns (0 allocations: 0 bytes)
 - v3 - 
  46.752 ns (0 allocations: 0 bytes)
  0.035 ns (0 allocations: 0 bytes)
 
 - v2 - 
  221.069 ns (0 allocations: 0 bytes)
  192.487 ns (0 allocations: 0 bytes)
 - v3 - 
  54.182 ns (0 allocations: 0 bytes)
  0.035 ns (0 allocations: 0 bytes)
 
 - v2 - 
  307.256 ns (0 allocations: 0 bytes)
  273.436 ns (0 allocations: 0 bytes)
 - v3 - 
  64.323 ns (0 allocations: 0 bytes)
  0.035 ns (0 allocations: 0 bytes)
 
 - v2 - 
  523.293 ns (0 allocations: 0 bytes)
  478.846 ns (0 allocations: 0 bytes)
 - v3 - 
  116.965 ns (0 allocations: 0 bytes)
  0.035 ns (0 allocations: 0 bytes)
 
 - v2 - 
  800.096 ns (0 allocations: 0 bytes)
  761.536 ns (0 allocations: 0 bytes)
 - v3 - 
  147.019 ns (0 allocations: 0 bytes)
  0.035 ns (0 allocations: 0 bytes)
 ```

File: src/Interfaces/AbstractZonotope.jl [hunk 1]
```diff
@@ -407,15 +407,32 @@ factors are ``ξ_i ∈ \\{-1, 1\\}``.
...
+      if sum(abs, G) == abs(sum(G))
+        sorted_G = sortslices(G, dims=2,by=x->atan(x[2], x[1]))
+        index = ones(p, 2*p)
+        for i in 1:p
+          index[i, i+1:i+p-1] *= -1
+        end
+        index[:, 1] *= -1
+        V = sorted_G * index .+ c
+        return [V[:, i] for i in 1:2*p]
+      end
```
schillic, line 434: Please use indentation with four spaces.

src/Interfaces/AbstractZonotope.jl
```diff
@@ -407,15 +407,32 @@ factors are ``ξ_i ∈ \\{-1, 1\\}``.
...
     c = center(Z)
     G = remove_zero_columns(genmat(Z))
-    p = size(G, 2)
+    n, p = size(G)
     if p == 0
         return [c]
     end
+    if n == 2 && p > 2 && apply_convex_hull
+        if sum(abs, G) == abs(sum(G))
+            sorted_G = sortslices(G, dims=2,by=x->atan(x[2], x[1]))
```
schillic, line 426:
```suggestion
            sorted_G = sortslices(G, dims=2, by=x->atan(x[2], x[1]))
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -407,15 +407,32 @@ factors are ``ξ_i ∈ \\{-1, 1\\}``.
...
+            index = ones(p, 2*p)
```
schillic, line 427:
```suggestion
            index = ones(N, p, 2*p)
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -407,15 +407,32 @@ factors are ``ξ_i ∈ \\{-1, 1\\}``.
...
+            for i in 1:p
```
schillic, line 428:
```suggestion
            @inbounds for i in 1:p
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -407,15 +407,32 @@ factors are ``ξ_i ∈ \\{-1, 1\\}``.
...
+                index[i, i+1:i+p-1] *= -1
```
schillic, line 429:
```suggestion
                index[i, i+1:i+p-1] .= -one(N)
```
This is cheaper than reading the 1 and multiplying with -1 (you know the result already).
Benchmark with `p = 20`:
```julia
julia> @time for i in 1:p
       index[i, i+1:i+p-1] .= -1.0
       end;
  0.000017 seconds (81 allocations: 3.469 KiB)

julia> @time for i in 1:p
       index[i, i+1:i+p-1] *= -1
       end;
  0.000028 seconds (81 allocations: 10.656 KiB)
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -407,15 +407,32 @@ factors are ``ξ_i ∈ \\{-1, 1\\}``.
...
+            end
+            index[:, 1] *= -1
```
schillic, line 431:
```suggestion
            index[:, 1] = -one(N)
```

mforets:
> I do not fully understand the algorithm. Could you add a short comment how index works?

My 2cents: there is a nice plot on the stackexchange link [here](https://github.com/JuliaReach/LazySets.jl/pull/2199#issuecomment-650161509) that illustrates how it works. The algorithm uses the fact that you can walk through the vertices of the zonotope by taking a path over the generators, as in the first implementation (with the `cumsum`), in clock-wise fashion, that's why they are first sorted, then accumulated. The implementation with the index matrix is a clever trick to use matrix products for the accumulation instead of calling `cumsum` twice.

schillic:
> there is a nice plot on the stackexchange link here that illustrates how it works. The algorithm uses the fact that you can walk through the vertices of the zonotope by taking a path over the generators, as in the first implementation (with the cumsum), in clock-wise fashion, that's why they are first sorted, then accumulated. The implementation with the index matrix is a clever trick to use matrix products for the accumulation instead of calling cumsum twice.

Okay, but it would be good to have this information in the code documentation for later.

src/Interfaces/AbstractZonotope.jl
```diff
@@ -407,15 +407,36 @@ factors are ``ξ_i ∈ \\{-1, 1\\}``.
...
         return [c]
     end
+    if n == 2 && p > 2 && apply_convex_hull
+        if sum(abs, G) == abs(sum(G))
+            sorted_G = sortslices(G, dims=2, by=x->atan(x[2], x[1]))
+            index = ones(N, p, 2*p)
+            @inbounds for i in 1:p
+                index[i, i+1:i+p-1] .= -one(N)
+            end
+            index[:, 1] = -one(N)
```
schillic, line 435: Sorry, my bad...
```suggestion
            index[:, 1] .= -one(N)
```

mforets:
test pass.. merging!

PR #2203, opened 2020-06-28
#2202 - Extend isdisjoint for Julia v1.5 

### LINKED ISSUE #2202: Extend isdisjoint for julia v1.5
```julia
[mforets@localhost ~]$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.0-beta1.0 (2020-05-28)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using LazySets

julia> isdisjoint
WARNING: both LazySets and Base export "isdisjoint"; uses of it in module Main must be qualified
ERROR: UndefVarError: isdisjoint not defined

```

mforets:
Closes #2202.

### DIFF
--- a/src/ConcreteOperations/isdisjoint.jl
+++ b/src/ConcreteOperations/isdisjoint.jl
@@ -49,13 +49,24 @@ function is_intersection_empty(X::LazySet{N},
 end
 
 # --- alias ---
+@static if VERSION <= v"1.5"
+    """
+        isdisjoint(X, Y)
 
-"""
-    isdisjoint(X, Y)
+    An alternative name for `is_intersection_empty(X, Y)`.
+    """
+    const isdisjoint = is_intersection_empty
 
-An alternative name for `is_intersection_empty(X, Y)`.
-"""
-const isdisjoint = is_intersection_empty
+else
+
+    """
+        isdisjoint(X, Y)
+
+    An alternative name for `is_intersection_empty(X, Y)`.
+    """
+    Base.isdisjoint(X::LazySet, Y::LazySet) = is_intersection_empty(X, Y)
+    Base.isdisjoint(X::LazySet, Y::LazySet, witness::Bool) = is_intersection_empty(X, Y, witness)
+end
 
 # --- AbstractHyperrectangle ---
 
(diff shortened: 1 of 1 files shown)


mforets:
this idea is OK but i think in that case the code should extend `isdisjoint`. in other words,

```julia
const is_intersection_empty = isdisjoint
```
and redefine the LazySets functions as new methods for `isdisjoint`.

src/ConcreteOperations/isdisjoint.jl
```diff
@@ -49,13 +49,24 @@ function is_intersection_empty(X::LazySet{N},
...
-const isdisjoint = is_intersection_empty
+else
+
+    """
+        isdisjoint(X, Y)
+
+    An alternative name for `is_intersection_empty(X, Y)`.
+    """
+    Base.isdisjoint(X::LazySet, Y::LazySet) = is_intersection_empty(X, Y)
+    Base.isdisjoint(X::LazySet, Y::LazySet, witness::Bool) = is_intersection_empty(X, Y, witness)
```
mforets, line 68:
```suggestion
    Base.isdisjoint(X::LazySet, Y::LazySet, args...; kwargs...) = is_intersection_empty(X, Y, args...; kwargs...)
```

mforets:
> In addition, this is not equivalent. We have methods with more arguments.

true; i've fixed this in the last commit.

> Note that you need to check for VERSION < v"1.5" instead.

i did `<= v"1.4"` on purpose, note that on Version 1.5.0-beta1.0 (2020-05-28), i get

```
julia> VERSION < v"1.5"
true
```

mforets:
for this to work on v1.5 we still need to add some methods, eg. those which dispatch on `Complement`, which is atm not a lazy set.

mforets:
>  I was wondering if VERSION >= v"1.5" works but unfortunately I do not have the v1.5 binary and the downloads are all down right now

for the comparison you dont need to download it, as `VERSION` is just a global const.

mforets:
i've added the missing methods in the last commit. the rest of the docs errors are seemingly unrelated and we can address them in a follow up PR.

PR #2205, opened 2020-07-03
Faster Zonotope inclusion in Hyperrectangle

### LINKED ISSUE #1555: Zonotope inclusion in hyperrectangle
See Lemma 3.1 [here](https://arxiv.org/pdf/1901.01006).
It remains to be seen if this method is more efficient than the current support-function based implementation. Conceptually they do the same, but the support-function implementation involves an additional matrix-vector multiplication.

SebastianGuadalupe:
Closes #1555.

### DIFF
--- a/src/ConcreteOperations/issubset.jl
+++ b/src/ConcreteOperations/issubset.jl
@@ -1059,16 +1059,13 @@ Algorithm based on Lemma 3.1 of [1]
 function ⊆(Z::AbstractZonotope{N}, H::AbstractHyperrectangle{N}) where {N<:Real}
     c = center(Z)
     G = genmat(Z)
-    n = dim(Z)
-    Gn = copy(c)
-    Gp = copy(c)
-    for i=1:n
-        aux = sum(norm.(G[i, :]))
-        Gp[i] = Gp[i] + aux
-        Gn[i] = Gn[i] - aux
-    end
+    n, m = size(G)
     for i = 1:n
-        if !_leq(Gp[i], high(H, i)) || !_geq(Gn[i], low(H, i))
+        aux = zero(N)
+        @inbounds for j in 1:m
+            aux += abs(G[i, j])
+        end
+        if !_leq(c[i] + aux, high(H, i)) || !_geq(c[i] - aux, low(H, i))
             return false
         end
     end
--- a/src/ConcreteOperations/issubset.jl
+++ b/src/ConcreteOperations/issubset.jl
@@ -1060,10 +1060,9 @@ function ⊆(Z::AbstractZonotope{N}, H::AbstractHyperrectangle{N}) where {N}
     c = center(Z)
     g = genmat(Z)
     n = dim(Z)
-    m = ngens(Z)
     Gn = copy(c)
     Gp = copy(c)
-    for i=1:m
+    for i=1:n
        Gp[i] = Gp[i] + sum(norm.(g[i, :]))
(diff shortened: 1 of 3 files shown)


src/ConcreteOperations/issubset.jl
```diff
@@ -1027,3 +1027,48 @@ function ⊆(X::CartesianProductArray{N},
...
+### Algorithm
+
+Algorithm based on Lemma 3.1 of [1]
+
+[1] Mitchell, I. M., Budzis, J., & Bolyachevets, A. (2019, April). Invariant,
+ viability and discriminating kernel under-approximation via zonotope scaling.
+ In Proceedings of the 22nd ACM International Conference on Hybrid Systems:
+ Computation and Control (pp. 268-269).
+"""
+function ⊆(Z::AbstractZonotope{N}, H::AbstractHyperrectangle{N}) where {N}
```
schillic, line 1059:
```suggestion
function ⊆(Z::AbstractZonotope{N}, H::AbstractHyperrectangle{N}) where {N<:Real}
```
This fixes the ambiguity error.

src/ConcreteOperations/issubset.jl
```diff
@@ -1027,3 +1027,48 @@ function ⊆(X::CartesianProductArray{N},
...
+    c = center(Z)
+    g = genmat(Z)
```
schillic, line 1061: Can you rename `g` to `G` to be consistent with the library?

src/ConcreteOperations/issubset.jl
```diff
@@ -1027,3 +1027,48 @@ function ⊆(X::CartesianProductArray{N},
...
+    n = dim(Z)
+    Gn = copy(c)
+    Gp = copy(c)
+    for i=1:n
+       Gp[i] = Gp[i] + sum(norm.(g[i, :]))
+       Gn[i] = Gn[i] - sum(norm.(g[i, :]))
+    end
```
schillic, line 1068: I think `Gp = c + ∑_i g_i` and `Gn = c - ∑_i g_i`. It is more efficient to first compute the sum only once and then add/subtract if to/from `c`.
(Also: use four spaces for indentation.)

src/ConcreteOperations/issubset.jl
```diff
@@ -1027,3 +1027,48 @@ function ⊆(X::CartesianProductArray{N},
...
+    result = true
+    for i=1:n
+       result = result && _leq(Gp[i], high(H)[i]) && _geq(Gn[i], low(H)[i])
+    end
+    return result
```
schillic, line 1073: This loop can be left as soon as `result` becomes `false`. In fact you do not need this variable. Also, you can use `low(H, i)` and `high(H, i)` for efficiency.
```suggestion
    for i = 1:n
        if !_leq(Gp[i], high(H, i)) || !_geq(Gn[i], low(H, i))
            return false
        end
    end
    return true
```

src/ConcreteOperations/issubset.jl
schillic: What I meant here was to have only one edit in the loop and do the add/subtract in the end.

schillic: On second thought, what you did now could actually be more performant because it does not allocate another vector :+1:

mforets:
Merging.. I changed the title to "Faster ..." because the method existed; this is a (very neat) optimization.

PR #2214, opened 2020-07-08
#2213 - Port zonotope splitting implementation from RA

### LINKED ISSUE #2213: Port zonotope splitting implementation from RA
See https://github.com/JuliaReach/ReachabilityAnalysis.jl/blob/master/src/Flowpipes/setops.jl#L470

- Efficiency (small gain), but allows to use SArrays
- Recursive method which allow to split given a set of generators, see `_split(Z::Zonotope, gens::AbstractVector, n::AbstractVector)`

SebastianGuadalupe:
Closes #2213.


src/Sets/Zonotope.jl
```diff
@@ -371,19 +372,129 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+- `gens` -- vector of indices of the generators to be splitted
+- `n` -- vector of integers describing the number of partitions in the
+         corresponding generator
+
+### Output
+
+The zonotopes obtained by splitting `Z` into `2^{n_i}` zonotopes for each
+generator `i` such that their union is `Z` and their intersection is
+possibly non-empty.
+"""
```
mforets, line 483: An example or two here would help. Could you add some?

src/Sets/Zonotope.jl
```diff
@@ -7,6 +7,7 @@ export Zonotope,
        remove_zero_generators
 
 using LazySets.Arrays: _vector_type, _matrix_type
+using StaticArrays: SMatrix, SVector, MMatrix, MVector
```
schillic, line 10: You need to work with `Requires` for optional dependencies. In a nutshell, create a file `init_StaticArrays` [here](https://github.com/JuliaReach/LazySets.jl/tree/master/src/Initialization) and wrap your code in a macro + helper function. An example for `init_Expokit.jl` is [here](https://github.com/JuliaReach/LazySets.jl/blob/a581de8c90d8e3c6d9689a7d724f9c77d7347225/src/LazyOperations/ExponentialMap.jl#L13-L18). It is a bit lengthy to describe this process. We can discuss further in Gitter if you need help.

src/Sets/Zonotope.jl
```diff
@@ -405,6 +404,11 @@ function split!(Z₁::Zonotope, Z₂::Zonotope, Z::Zonotope, j::Int)
     return Z₁, Z₂
 end
 
+function load_static_arrays()
+return quote
+
+using StaticArrays: SMatrix, SVector, MMatrix, MVector
```
schillic, line 410:
```suggestion
using .StaticArrays: SMatrix, SVector, MMatrix, MVector
```
I think you need to do this.

docs/src/lib/sets/Zonotope.md
```diff
@@ -15,6 +15,8 @@ ngens(::Zonotope)
 togrep(::Zonotope)
 reduce_order(::Zonotope, ::Union{Integer, Rational})
 split(::Zonotope, ::Int)
+split(Z::Zonotope, gens::AbstractVector, n::AbstractVector)
+split(Z::Zonotope{N, SVector{n, N}, <:SMatrix{n, p, N}}, j::Int) where {N, n, p}
```
mforets, line 19: there is a docs build error, https://travis-ci.org/github/JuliaReach/LazySets.jl/jobs/706883297#L557

doing `using StaticArrays` in the docs setup may help. otherwise, just remove line 19..

schillic: That would be here I guess:
https://github.com/JuliaReach/LazySets.jl/blob/a581de8c90d8e3c6d9689a7d724f9c77d7347225/docs/make.jl#L3

schillic: Hm, I just tried that and it still gives the same error.

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,179 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+    G₁ = MMatrix{n, p}(G)
+    G₂ = MMatrix{n, p}(G)
+
+    @inbounds for i in 1:n
+        α = G[i, j] / 2
+        c₁[i] = c[i] - α
+        c₂[i] = c[i] + α
+        G₁[i, j] = α
+        G₂[i, j] = α
+    end
```
schillic, line 462: This part could be shared.

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,179 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+"""
+    split(Z::Zonotope, gens::AbstractVector{Int}, n::AbstractVector{Int})
+
+Return a vector zonotopes obtained form splitting in the given generators the
+given zonotope.
+
+### Input
+
+- `Z` -- zonotope
+- `gens` -- vector of indices of the generators to be splitted
```
schillic, line 479:
```suggestion
- `gens` -- vector of indices of the generators to be split
```

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,179 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+julia> Z = Zonotope([1.0, 0.0], [0.1 0.0; 0.0 0.1])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.0, 0.0], [0.1 0.0; 0.0 0.1])
+
+julia> split(Z, [1], [1])
+2-element Array{Zonotope{Float64,Array{Float64,1},Array{Float64,2}},1}:
+ Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([0.95, 0.0], [0.05 0.0; 0.0 0.1])
+ Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.05, 0.0], [0.05 0.0; 0.0 0.1])
+```
+Here, the first vector in the arguments corresponds to the zonotope's
+generator to be splitted, and the second vector corresponds to the exponent of
```
schillic, line 503:
```suggestion
generator to be split, and the second vector corresponds to the exponent of
```

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,179 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+`2^n` times that the zonotope will be splitted along the corresponding generator.
```
schillic, line 504:
```suggestion
`2^n` times that the zonotope will be split along the corresponding generator.
```

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,179 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.025, -0.075], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.025, -0.025], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.025, 0.025], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.025, 0.075], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.075, -0.075], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.075, -0.025], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.075, 0.025], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.075, 0.075], [0.025 0.0; 0.0 0.025])
+```
+Here the zonotope is splitted along both of its generators 4 times.
```
schillic, line 531:
```suggestion
 the zonotope is split along both of its generators four times.
```

docs/src/lib/sets/Zonotope.md
schillic: > otherwise, just remove line 19

I agree, remove that line and the documentation for that method because it is actually identical to the old one I think.

mforets: yes, indeed

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,179 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
-    G₁ = copy(G)
-    G₁[:, j] = Gj_half
-    G₂ = copy(G₁)
 
+    c₁ = similar(c)
+    G₁ = similar(G)
     Z₁ = Zonotope(c₁, G₁)
+
+    c₂ = similar(c)
+    G₂ = similar(G)
```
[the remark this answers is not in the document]
mforets: i think that was to share code (with / without sarrays). also, the `split!` method is interesting on its own, there are cases when it's useful to pass preallocated arrays in advance.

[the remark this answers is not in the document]
schillic: But at the moment I think there is no shared code.

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,179 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+
+```math
+Z^* = ⟨c, (g^{(1,…,j-1)}, g^{(j+1,…, p)})⟩.
+```
+
+[1] *Althoff, M., Stursberg, O., & Buss, M. (2008). Reachability analysis of
+nonlinear systems with uncertain parameters using conservative linearization.
+In Proc. of the 47th IEEE Conference on Decision and Control.*
+"""
+function split(Z::Zonotope{N, SV, SM}, j::Int) where {N, n, p, SV<:SVector{n, N}, SM<:SMatrix{n, p, N}}
```
schillic, line 446: I tried this out but for me it does not choose this method.
```julia
julia> Z
Zonotope{Float64,SArray{Tuple{2},Float64,1,2},SArray{Tuple{2,2},Float64,2,4}}([0.7504943082957674, 0.36528999613841906], [-0.5419307531431913 1.1459414507001595; -0.08228095036507727 -0.4595899030827082])

julia> @which split(Z, 1)
split(Z::Zonotope, j::Int64) in LazySets at .julia/dev/LazySets/src/Sets/Zonotope.jl:374
```
Are you sure about the signature? In my example, I think it should have the signatures `SV<:SVector{Tuple{n}, N, 1, n}` and `SM<:SMatrix{Tuple{n, p}, N, n, p}` (not 100% sure this is valid syntax).

mforets: StaticArrays has some convenience type aliases, eg. 

```julia
julia> MVector{2, Float64}
MArray{Tuple{2},Float64,1,2}

julia> MMatrix{2, 3, Float64}
MArray{Tuple{2,3},Float64,2,L} where L
```

src/init.jl
```diff
@@ -6,6 +6,7 @@ function __init__()
     @require MiniQhull = "978d7f02-9e05-4691-894f-ae31a51d76ca" include("Initialization/init_MiniQhull.jl")
     @require Optim = "429524aa-4258-5aef-a3af-852621145aeb" include("Initialization/init_Optim.jl")
     @require Polyhedra = "67491407-f73d-577b-9b50-8179a7c68029" include("Initialization/init_Polyhedra.jl")
+    @require StaticArrays = "67491407-f73d-577b-9b50-8179a7c68029" include("Initialization/init_StaticArrays.jl")
```
mforets, line 9: this uuid is not correct. i can fix it together with other commit that simplifies the optional code.

src/Sets/Zonotope.jl
mforets: > I tried this out but for me it does not choose this method.

i think this was because the uuid was wrong under `__init__`

mforets:
A comparison between static vs non-static implementation:

```julia
julia> using Revise, LazySets, StaticArrays, BenchmarkTools
[ Info: Precompiling LazySets [b4f0291d-fe17-52bc-9479-3d1a343d9043]

julia> Z = Zonotope(@SVector(rand(2)), @SMatrix(rand(2, 2)))
Zonotope{Float64,SArray{Tuple{2},Float64,1,2},SArray{Tuple{2,2},Float64,2,4}}([0.6088108765998801, 0.5135386852090658], [0.666497673074505
6 0.9542751369371558; 0.5158563660350037 0.5883345234682646])

julia> split(Z, 2)
(Zonotope{Float64,SArray{Tuple{2},Float64,1,2},SArray{Tuple{2,2},Float64,2,4}}([0.13167330813130218, 0.21937142347493344], [0.666497673074
5056 0.4771375684685779; 0.5158563660350037 0.2941672617341323]), Zonotope{Float64,SArray{Tuple{2},Float64,1,2},SArray{Tuple{2,2},Float64,
2,4}}([1.085948445068458, 0.8077059469431981], [0.6664976730745056 0.4771375684685779; 0.5158563660350037 0.2941672617341323]))

julia> @btime split($Z, 2);
  35.733 ns (4 allocations: 160 bytes)

julia> Z2 = Zonotope(Vector(Z.center), Matrix(Z.generators));

julia> split(Z2, 2)
(Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([0.13167330813130218, 0.21937142347493344], [0.6664976730745056 0.4771375684685779; 0
.5158563660350037 0.2941672617341323]), Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.085948445068458, 0.8077059469431981], [0.66
64976730745056 0.4771375684685779; 0.5158563660350037 0.2941672617341323]))

julia> @btime split($Z2, 2);
  120.027 ns (4 allocations: 416 bytes)
```

runtime in master using regular arrays (errors with sarrays)

```julia
julia> @btime split($Z2, 2);
  196.267 ns (9 allocations: 704 bytes)
```

src/Sets/Zonotope.jl
mforets: ill mark this as resolved, please reopen if it's not the case.

docs/src/lib/sets/Zonotope.md
mforets: i've removed the docs for sarrays since its a dup of the existing one.

src/Sets/Zonotope.jl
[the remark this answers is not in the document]
mforets: with the new commit there's shared code. some comments

- `similar` is used because that produces an MArray (mutable statically sized array) from an SArray
- `copyto!` works regardless of the array type
- in the end we need to convert to SArray again so that is the particular method (this is optional, but conventional)

docs/src/lib/sets/Zonotope.md
mforets, line 19:
```suggestion
```

src/Sets/Zonotope.jl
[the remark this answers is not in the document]
schillic: I still do not see the purpose of doing `similar` followed by `copyto!`. Why not do `copy` immediately? I do not see a use case for passing dummy zonotopes instead of creating the zonotopes inside the method.

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,128 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+    c₂, G₂ = Z₂.center, Z₂.generators
+    copyto!(G₁, G)
+    copyto!(G₂, G)
+
+    @inbounds for i in 1:n
+        α = G[i, j] / 2
+        c₁[i] = c[i] - α
+        c₂[i] = c[i] + α
+        G₁[i, j] = α
+        G₂[i, j] = α
```
schillic, line 402: You can remove this line and after the loop do `copyto!(G₂, G₁)`.

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,128 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.0, 0.0], [0.1 0.0; 0.0 0.1])
+
+julia> split(Z, [1], [1])
+2-element Array{Zonotope{Float64,Array{Float64,1},Array{Float64,2}},1}:
+ Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([0.95, 0.0], [0.05 0.0; 0.0 0.1])
+ Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.05, 0.0], [0.05 0.0; 0.0 0.1])
+```
+Here, the first vector in the arguments corresponds to the zonotope's
+generator to be split, and the second vector corresponds to the exponent of
+`2^n` times that the zonotope will be split along the corresponding generator.
```
schillic, line 453:
```suggestion
`2^n` parts that the zonotope will be split into along the corresponding generator.
```

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,128 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.025, -0.075], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.025, -0.025], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.025, 0.025], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.025, 0.075], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.075, -0.075], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.075, -0.025], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.075, 0.025], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.075, 0.075], [0.025 0.0; 0.0 0.025])
+```
+Here the zonotope is split along both of its generators four times.
```
schillic, line 480:
```suggestion
Here the zonotope is split along both of its generators, each time into four parts.
```

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,128 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+function _split_ret(Z₁::Zonotope{N, SV, SM}, Z₂::Zonotope{N, SV, SM}) where {N, n, p, SV<:MVector{n, N}, SM<:MMatrix{n, p, N}}
+    Z₁ = Zonotope(SVector{n}(Z₁.center), SMatrix{n, p}(Z₁.generators))
+    Z₂ = Zonotope(SVector{n}(Z₂.center), SMatrix{n, p}(Z₂.generators))
     return Z₁, Z₂
 end
+
+end end  # quote / load_static_arrays
+
+"""
+    split(Z::Zonotope, gens::AbstractVector{Int}, n::AbstractVector{Int})
```
schillic, line 421: Can we use a different name for `n`, which is usually a scalar for the dimension? Maybe `splits`?

src/Sets/Zonotope.jl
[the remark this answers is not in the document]
mforets: The idea is that the in-place version `split!` receives preallocated arrays (zonotope center / generators matrices) and overrides whatever is such arrays with new values. 

> Why not do copy immediately? Does that not work for MArray?

Copying an SArray creates an SArray, but we want to modify values so we need an intermediate MArray to work with. See also [here](https://juliaarrays.github.io/StaticArrays.jl/latest/pages/api/#Mutable-arrays:-MVector,-MMatrix-and-MArray-1).

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,128 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.025, 0.025], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.025, 0.075], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.075, -0.075], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.075, -0.025], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.075, 0.025], [0.025 0.0; 0.0 0.025])
+Zonotope{Float64,Array{Float64,1},Array{Float64,2}}([1.075, 0.075], [0.025 0.0; 0.0 0.025])
+```
+Here the zonotope is split along both of its generators four times.
+"""
+function split(Z::Zonotope, gens::AbstractVector, n::AbstractVector)
```
schillic, line 482: I think the code is general for `AbstractZonotope`.

src/Sets/Zonotope.jl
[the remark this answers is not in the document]
mforets: About output preallocation, using the pattern `split` / `split!` is comparable to https://docs.julialang.org/en/v1/manual/performance-tips/index.html#Pre-allocating-outputs-1.

test/unit_Zonotope.jl
```diff
@@ -160,6 +160,10 @@ for N in [Float64, Rational{Int}, Float32]
     Z = Zonotope(N[0, 0], N[1 1; -1 1])
     Z1, Z2 = split(Z, 1) # in this case the splitting is exact
     @test Z1 ⊆ Z && Z2 ⊆ Z
+    Z1, Z2, Z3, Z4 = split(Z, [1, 2], [1, 1])
+    @test Z1 ⊆ Z && Z2 ⊆ Z && Z3 ⊆ Z && Z4 ⊆ Z
+    Z = Zonotope(SVector{2}(N[0, 0]), SMatrix{2, 2}(N[1 1; -1 1]))
+    @test Z1 ⊆ Z && Z2 ⊆ Z
```
schillic, line 166: I guess you wanted to split `Z` as well?

src/Sets/Zonotope.jl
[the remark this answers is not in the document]
schillic: Ah, I see, makes sense now.

src/Sets/Zonotope.jl
mforets: yes indeed, that should be generalized, as well as the code for `split(Z, j)`.

mforets: i didn't pretend working further in this PR, but it's fine either way.

test/unit_Zonotope.jl
SebastianGuadalupe: Yes, I forgot!

src/Sets/Zonotope.jl
schillic: We can keep the other `split` specific but here it is a matter of changing the signature only.

src/Sets/Zonotope.jl
mforets: we could also use `partition`, `numsplits` , `nsplits`, .. this is a non-keyword arg though.

src/Sets/Zonotope.jl
mforets: is there a particular reason to keep the `split(Z, j)` specific for Zonotopes? `split(Z::Zonotope, gens::AbstractVector, n::AbstractVector)` is a generalization.

schillic: You sounded like you want to merge this one soon.

mforets: the code has got in much better shape after each iteration, it's not bad to do it once again :D

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,128 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+    @assert length(gens) == length(n) "the number of generators doesn't match the" *
+    " number of indicated partitions ($(length(gens)) and $(length(n)))"
+    @assert length(gens) <= ngens(Z) "the number of generators to split is greater" *
+    " than the number of generators of the zonotope (($(length(gens)) and $(ngens(Z)))"
+    Zs = [Z]
+    for i = 1:length(gens)
+        for j = 1:n[i]
+            km = length(Zs)
+            for k = 1:km
+                append!(Zs, split(Zs[k], gens[i]))
```
schillic, line 492:
```suggestion
    for (i, g) in enumerate(gens)
        for j = 1:n[i]
            km = length(Zs)
            for k = 1:km
                append!(Zs, split(Zs[k], g))
```

src/Sets/Zonotope.jl
```diff
@@ -371,19 +371,128 @@ nonlinear systems with uncertain parameters using conservative linearization.
...
+            end
+            deleteat!(Zs, 1:km)
```
schillic, line 494: I guess it is faster to collect the new zonotopes in a fresh array and replace that array in the end.

SebastianGuadalupe: I opened an issue to do it in the future: https://github.com/JuliaReach/LazySets.jl/issues/2224

mforets:
Thanks @SebastianGuadalupe to help porting this code from RA. Also, we can now begin adding optional parts for use with StaticArrays. Should we merge this? @schillic

schillic:
Before merging, can you add a test for the two split methods for one of the other zonotopic types (e.g., `BallInf`), please?

There is an issue with mixed array types. I guess that `_split_ret` for the two other combinations (where only one component is an `MArray`) would be needed. You can either fix it or open an issue.
We should also open an issue that in the conversion to zonotopes from static arrays we should use a static generator matrix.
```julia
julia> B = BallInf(SVector{2}(1., 1.), 1.)
BallInf{Float64,SArray{Tuple{2},Float64,1,2}}([1.0, 1.0], 1.0)

julia> convert(Zonotope, B)
Zonotope{Float64,SArray{Tuple{2},Float64,1,2},Array{Float64,2}}([1.0, 1.0], [1.0 0.0; 0.0 1.0])
# note that only the center is a static vector; the generators are a normal matrix

julia> split(B, [1], [1])
ERROR: MethodError: Cannot `convert` an object of type 
  Zonotope{Float64{},MArray{Tuple{2},Float64,1,2},Array{Float64{},2}} to an object of type 
  Zonotope{Float64{},SArray{Tuple{2},Float64,1,2},Array{Float64{},2}}
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:171
Stacktrace:
 [1] setindex!(::Array{Zonotope{Float64,SArray{Tuple{2},Float64,1,2},Array{Float64,2}},1}, ::Zonotope{Float64,MArray{Tuple{2},Float64,1,2},Array{Float64,2}}, ::Int64) at ./array.jl:826
 [2] _append! at ./array.jl:968 [inlined]
 [3] append! at ./array.jl:960 [inlined]
 [4] _split(::Zonotope{Float64,SArray{Tuple{2},Float64,1,2},Array{Float64,2}}, ::Array{Int64,1}, ::Array{Int64,1}) at .julia/dev/LazySets/src/Sets/Zonotope.jl:402
 [5] split(::BallInf{Float64,SArray{Tuple{2},Float64,1,2}}, ::Array{Int64,1}, ::Array{Int64,1}) at .julia/dev/LazySets/src/Interfaces/AbstractZonotope.jl:652
```

PR #2215, opened 2020-07-08
#941 - Add convex_hull of an array of polytopes

### LINKED ISSUE #941: convex_hull of polytopes

SebastianGuadalupe:
Closes #941.

### DIFF
--- a/docs/src/lib/binary_functions.md
+++ b/docs/src/lib/binary_functions.md
@@ -53,7 +53,7 @@ convex_hull(::HPoly{N}, ::HPoly{N}) where {N<:Real}
 convex_hull(::VPolytope{N}, ::VPolytope{N}) where {N<:Real}
 convex_hull(::VPolygon{N}, ::VPolygon{N}) where {N<:Real}
 convex_hull(::Vector{VN}) where {N<:Real, VN<:AbstractVector{N}}
-convex_hull(::UnionSetArray{N, PT}) where {N, PT<:AbstractPolytope{N}}
+convex_hull(::UnionSetArray{N, PT}; kwargs...) where {N, PT<:AbstractPolytope{N}}
 monotone_chain!
 ```
 
--- a/docs/src/lib/binary_functions.md
+++ b/docs/src/lib/binary_functions.md
@@ -53,7 +53,7 @@ convex_hull(::HPoly{N}, ::HPoly{N}) where {N<:Real}
 convex_hull(::VPolytope{N}, ::VPolytope{N}) where {N<:Real}
 convex_hull(::VPolygon{N}, ::VPolygon{N}) where {N<:Real}
 convex_hull(::Vector{VN}) where {N<:Real, VN<:AbstractVector{N}}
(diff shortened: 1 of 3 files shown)


src/ConcreteOperations/convex_hull.jl
```diff
@@ -476,3 +476,21 @@ function monotone_chain!(points::Vector{VN}; sort::Bool=true
...
+Compute the convex hull of an array of polytopes
+
+### Input
+
+- `U` -- UnionSetArray of polytopes
+
+### Output
+
+List of vectors containing the vertices of the convex hull of an array of
+the given polytopes.
```
schillic, line 491:
```suggestion
    convex_hull(U::UnionSetArray{N, <:AbstractPolytope{N}}) where {N}
    
Compute the convex hull of a union of a finite number of polytopes.

### Input

- `U` -- UnionSetArray of polytopes

### Output

A list of the vertices of the convex hull.
```

mforets:
docs are green now... merging!

docs/src/lib/binary_functions.md
```diff
@@ -53,6 +53,7 @@ convex_hull(::HPoly{N}, ::HPoly{N}) where {N<:Real}
 convex_hull(::VPolytope{N}, ::VPolytope{N}) where {N<:Real}
 convex_hull(::VPolygon{N}, ::VPolygon{N}) where {N<:Real}
 convex_hull(::Vector{VN}) where {N<:Real, VN<:AbstractVector{N}}
+convex_hull(U::UnionSetArray{N, PT}) where {N, PT<:AbstractPolytope{N}}
```
mforets, line 56:
```suggestion
convex_hull(::UnionSetArray{N, PT}) where {N, PT<:AbstractPolytope{N}}
```

mforets: @SebastianGuadalupe  note that the name of the variables are not needed here.

mforets:
> I am not sure if removing the kwargs was good, but fine smiley

i agree; the kwargs can be used to pass a backend for  taking the convex hull, or additional options.

mforets:
> i agree; the kwargs can be used to pass a backend for taking the convex hull, or additional options.

@SebastianGuadalupe  could you please try adding those again?

PR #2220, opened 2020-07-15
Improved order reduction methods

mforets:
Similarly to https://github.com/JuliaReach/LazySets.jl/pull/2214, this PR ports some functionality from RA about zonotope order reduction methods. These methods are more efficient than in LazySets#master, and there are implementations for static arrays, that we should properly wrap into optional code as in #2214.


src/Sets/Zonotope.jl
```diff
@@ -387,3 +364,143 @@ function split(Z::Zonotope, j::Int)
...
+
+    # r is bigger than the order of Z => don't reduce
+    (r * n >= p) && return Z
+
+    # this algorithm sort generators by decreasing 2-norm
+    indices = Vector{Int}(undef, p)
+    _weighted_gens!(indices, G, _COMB03)
+
+    # the first m generators have greatest weight
+    m = floor(Int, n * (r - 1))
```
mforets, line 470: in fact if `r = 1` => `m = 0` and the generators dont need to be sorted

src/Sets/Zonotope.jl
```diff
@@ -329,30 +329,149 @@ function scale!(α::Real, Z::Zonotope)
...
+
+function load_reduce_order_static()
+return quote
+
+# implementation for static arrays
+function _interval_hull(G::SMatrix{n, p, N, L}, indices) where {n, p, N, L}
+    Lred = zeros(MMatrix{n, n, N})
+    @inbounds for i in 1:n
+        for j in indices
+            Lred[i, i] += abs(G[i, j])
```
schillic, line 460:
```suggestion
    Lred = zeros(MMatrix{n, n, N})
    @inbounds for i in 1:n
        for j in indices
            Lred[i, i] = abs(G[i, j])
```
This is more efficient.

src/Sets/Zonotope.jl
```diff
@@ -329,30 +329,149 @@ function scale!(α::Real, Z::Zonotope)
...
+end
+
+# compute interval hull of the generators of G (= columns) corresponding to `indices`
+function _interval_hull(G::AbstractMatrix{N}, indices) where {N}
+    n, p = size(G)
+    Lred = zeros(N, n, n)
+    @inbounds for i in 1:n
+        for j in indices
+            Lred[i, i] += abs(G[i, j])
+        end
```
schillic, line 440:
```suggestion
    Lred = zeros(N, n, n)
    @inbounds for i in 1:n
        for j in indices
            Lred[i, i] = abs(G[i, j])
        end
```
This is more efficient.

src/Sets/Zonotope.jl
```diff
@@ -329,30 +329,149 @@ function scale!(α::Real, Z::Zonotope)
...
+# ==================================
+
+# These methods split the given zonotope Z into two zonotopes, K and L, where
+# K contains the most "representative" generators and L contains the generators
+# that are reduced, Lred using a box overapproximation
+struct GIR05 <: AbstractReductionMethod end
+struct COMB03 <: AbstractReductionMethod end
+
+const _GIR05 = GIR05()
+const _COMB03 = COMB03()
```
schillic, line 343: We do not need these. You can just pass the algorithm that you already have in the `_reduce_order` methods.

src/Sets/Zonotope.jl
```diff
@@ -329,30 +329,149 @@ function scale!(α::Real, Z::Zonotope)
...
+    m = floor(Int, n * (r - 1))
+
+    # compute interval hull of L
+    Lred = _interval_hull(G, view(indices, (m+1):p))
+
+    isone(r) && return Zonotope(c, Lred)
+
+    Gred = _hcat_KLred(G, view(indices, 1:m), Lred)
+    return Zonotope(c, Gred)
+end
```
schillic, line 397: The two methods `_reduce_order` are identical if you pass the algorithm to `_weighted_gens!`. So they should be merged.

src/Sets/Zonotope.jl
```diff
@@ -329,30 +329,149 @@ function scale!(α::Real, Z::Zonotope)
...
+# that are reduced, Lred using a box overapproximation
+struct GIR05 <: AbstractReductionMethod end
+struct COMB03 <: AbstractReductionMethod end
+
+const _GIR05 = GIR05()
+const _COMB03 = COMB03()
+
+# Implements zonotope order reduction method from [COMB03]
+# We follow the notation from [YS18]
+function _reduce_order(Z::Zonotope, r::Number, ::COMB03)
```
schillic, line 347: Should we make the algorithm the first argument?

src/Sets/Zonotope.jl
schillic: This was incorrect, sorry.

src/Sets/Zonotope.jl
mforets: I suggest to rename `_reduce_order` to `reduce_order` since the underscore is not needed; the method is kept as the third argument.

src/Sets/Zonotope.jl
mforets: I double checked and this is indeed correct, giving a factor x5 .. x10 speedup for such case.

src/Sets/Zonotope.jl
```diff
@@ -329,30 +329,149 @@ function scale!(α::Real, Z::Zonotope)
...
+
+# given an n x p matrix G and a vector of m integer indices with m <= p,
+# concatenate the columns of G given by `indices` with the matrix Lred
+function _hcat_KLred(G::AbstractMatrix, indices, Lred::AbstractMatrix)
+    K = view(G, :, indices)
+    return hcat(K, Lred)
+end
+
+function load_reduce_order_static()
+return quote
```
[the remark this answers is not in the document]
mforets: Added a test

src/Sets/Zonotope.jl
schillic: This is a rather rare case because you would probably use `box_approximation` instead, but okay :+1:

schillic:
`COMB03` and `GIR05` need to be added to the docs. There is also a failing test.

schillic (approved):

src/Sets/Zonotope.jl
mforets: That's true, though it's convenient to rely on `reduce_order` being fast in other place (such as [here](https://github.com/JuliaReach/ReachabilityAnalysis.jl/blob/3c3d4a62b514d24773d3dee53b798e0200b4e1bf/src/Algorithms/GLGM06/post.jl#L39)). 

BTW after merging this PR I'll remove a bunch of repeated code in RA and tag releases on both repos.

Issue #175, opened 2018-01-23
Clarify user interface for applying lazy and concrete operations


schillic:
In #148 we added functions like `minkowski_sum`.
(The following is copied from the discussion started in #174.)

Why did we not add a `MinkowskiSum` constructor for `Zonotope`s?
Answer: Because we wanted to reserve the `MinkowskiSum` constructor for the lazy operation.

I think that currently the design is inconsistent. For some lazy operations we actually do a concrete operation (like `LinearMap(lm::LinearMap)`, `MinkowskiSum(E::EmptySet, S::LazySet)`) and for some concrete operations and only some types we have another function (like `minkowski_sum`). Why should the user know what to use? (in particular in an application where you do not know the set type in advance!)

I was thinking if we should have this:
```julia
minkowski_sum(S1::LazySet, S2::LazySet)
    return MinkowskiSum(S1, S2)
end
minkowski_sum(Z1::Zonotope, Z2::Zonotope)
    # what we did here
end
minkowski_sum(E::EmptySet, S::LazySet)
    return S
end
minkowski_sum(T1::Type1, T2::Type2) # other type combinations where concrete implemenations are simple
    ...
end
```
Then we could remove the specific `MinkowskiSum` constructors like `MinkowskiSum(E::EmptySet, S::LazySet)`. The user would use this new function instead and "under the hood" we decide what happens (lazy or concrete).

But now I realized that this is a design decision and I do not have an opinion here. Previously you had to call the constructor `MinkowskiSum` and it either created a lazy `MinkowskiSum` object or, in special cases, skipped it (`EmptySet`). Now we have the mixture (sometimes you have to call the constructor and sometimes it is concrete; sometimes you have to call this function to get the concrete implementation).

The question is if we want to give the user the freedom to choose between lazy and concrete set operations. I see three options for who decides between lazy and concrete operations:
1. *We* decide: We only provide one interface (namely the `MinkowskiSum` constructor). In that case we should remove the function `minkowski_sum` wrote and add a respective constructor.
2. *The user* decides: The constructor `MinkowskiSum` is always lazy; the function `minkowski_sum` is always concrete if possible (and lazy otherwise).
3. *The user* **can** decide but *we* provide also a default choice.

We currently prefer the most general answer (3.), but it requires some discussion. Example implementation:
The constructor `MinkowskiSum` is always lazy; the function `minkowski_sum` gets a new argument from the domain {"lazy", "concrete", "default"} with default value "default".
[...]

PR #2228, opened 2020-07-17
#2226 - Add concretize function

### LINKED ISSUE #2226: Add concretize function
Add a function `concretize` that, given a set, constructs a concrete set by applying the concrete counterpart (e.g., `minkowski_sum` for a `MinkowskiSum` object) and (recursively!) calling `concretize` on the operands.

_Originally posted by @schillic in https://github.com/JuliaReach/LazySets.jl/issues/175#issuecomment-534467842_

This is a first step for #175.

schillic:
Closes #2226.


mforets:
`concretize` can be though as an expression transformation from CamelCase to underscore (the module [StringCases.jl](https://github.com/djsegal/StringCases.jl) contains such transformations). so i would suggest to consider using a generated function (or a macro, `@concretize`) that implements such rule once (ie. dispatch on `LazySet`)

schillic:
> StringCases.jl

I doubt that this works because of the type parameters and/or different arguments.

> `@concretize`

I think this is orthogonal. The functionality here works for *existing* variables while the macro would be invoked *at construction time*. Currently I need the functionality given here. Maybe it can be reused for the macro but the macro is definitely more work.

schillic:
Thanks for the review. I will add your comments to #175.

PR #2230, opened 2020-07-17
Update .travis.yml

mforets:


.travis.yml
```diff
@@ -27,13 +27,24 @@ branches:
   - /^testing-.*$/  # testing branches
   - /^v[0-9]+\.[0-9]+\.[0-9]+$/  # version tags
 
+matrix:
+  include:
+    - julia: 1.0
+      env: TEST_SUITE='using Pkg; Pkg.build("LazySets"); Pkg.test("LazySets"; coverage=true); Pkg.add(PackageSpec(name="ModelingToolkit", version="3.10.2"))'; DOCS=false
```
schillic, line 33:
```suggestion
      env: TEST_SUITE='using Pkg; Pkg.build("LazySets"); Pkg.test("LazySets"; coverage=true)'; DOCS=false
```

schillic:
`ModelingToolkit` has a new release and now it seems to work again (saw it in a different PR) :roll_eyes:
We can still merge here for stability reasons but maybe we should use the new version already?

mforets:
> We can still merge here for stability reasons but maybe we should use the new version already?

i can update it later (with other changes, probably)

schillic:
I think #2234 is the better solution.

PR #2234, opened 2020-07-18
Update Project.toml

mforets:


Project.toml
```diff
@@ -32,7 +32,7 @@ JuMP = "0.21"
 Makie = "0.9"
 MathProgBase = "0.7"
 MiniQhull = "0.1"
-ModelingToolkit = "1.1.3, 3.10"
+ModelingToolkit = "1.1.3, =3.10.2"
```
schillic, line 35:
```suggestion
ModelingToolkit = "1.1.3, ~3.10"
```
This would allow newer patch versions.

PR #2237, opened 2020-07-18
#2117 - Add type parameters for polynomial zonotope

### LINKED ISSUE #2117: Add type parameters for polynomial zonotope

schillic:
Closes #2117.

### DIFF
--- a/src/Sets/PolynomialZonotope.jl
+++ b/src/Sets/PolynomialZonotope.jl
@@ -6,10 +6,7 @@ export PolynomialZonotope, dim, σ,
        minkowski_sum
 
 """
-    PolynomialZonotope{N, VTC<:AbstractVector{N},
-                       VMTE<:AbstractVector{<:AbstractMatrix{N}},
-                       VMTF<:AbstractVector{<:AbstractMatrix{N}},
-                       MTG<:AbstractMatrix{N}}
+    PolynomialZonotope{N, VT, VMT, MT}
 
 Type that represents a polynomial zonotope.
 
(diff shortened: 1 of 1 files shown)


src/Sets/PolynomialZonotope.jl
```diff
@@ -6,7 +6,10 @@ export PolynomialZonotope, dim, σ,
        minkowski_sum
 
 """
-    PolynomialZonotope{N}
+    PolynomialZonotope{N, VTC<:AbstractVector{N},
```
mforets, line 9: my preferred choice would be:

```
struct PolynomialZonotope{N, VT<:AbstractVector{N}, MT<:AbstractMatrix{N}}
... # use VT and MT
end
```
because you can reason about polynomial zonotopes with regular / sparse / static arrays , = dispatch on `VT` and `MT`. somtimes it is convenient to use array views and it's pertinent to generalize the abstract vector container as well.

schillic: > somtimes it is convenient to use array views and it's pertinent to generalize the abstract vector container as well.

I also started with removing the type restrictions completely but the problem is that we somehow need to bind `N` and for that we need a container type.

schillic: If a struct can have unused type parameters (not tested), we could determine the `N` parameter in the constructor via `typeof(center[1])`. Should I try?

schillic: I did that now. What do you think?

PR #2238, opened 2020-07-18
#700 - Documentation for issubset

### LINKED ISSUE #700: Use issubset for function names
Function signatures in `/src/issubset.jl` are written using the `⊆` operator. But if you ask for the help in the REPL for `?issubset` it won't show those of lazysets. 

We should rather use `issubset` for the signatures, which will make both `?issubset` and `?⊆` be complete.

schillic:
Closes #700.

### DIFF
--- a/docs/src/lib/binary_functions.md
+++ b/docs/src/lib/binary_functions.md
@@ -104,6 +104,7 @@ pontryagin_difference
 ## Subset check
 
 ```@docs
+issubset
 ⊆(::LazySet{N}, ::LazySet{N}, ::Bool=false) where {N<:Real}
 ⊆(::LazySet{N}, ::AbstractHyperrectangle{N}, ::Bool=false) where {N<:Real}
 ⊆(::AbstractPolytope{N}, ::LazySet{N}, ::Bool=false) where {N<:Real}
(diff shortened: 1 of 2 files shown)


mforets:
- document `issubste` generically (like we do for `+`)
- keep `⊆` in the code

schillic:
done

docs/src/lib/binary_functions.md
```diff
@@ -101,6 +101,7 @@ pontryagin_difference
 ## Subset check
 
 ```@docs
+issubset
```
mforets, line 104: i guess `issubset(::LazySet, ...)` so it doesn't print the methods from Base (`[1] issubset(a::BitSet, b::BitSet) in Base at bitset.jl:419`, ...)

schillic: Did you try it? I didn't but I think `Documenter` only shows local module code because of this line:
https://github.com/JuliaReach/LazySets.jl/blob/8a7212b45032806529773007e46a91d4c1217710/docs/make.jl#L9

schillic: I tried it and it works in Julia versions before v1.5. For v1.5 it does not work at all (does not find the associated function) but I could not find a quick solution, so I suggest that we fix that later in #2310.

mforets (approved):

mforets:
@schillic  should we merge this?

PR #2241, opened 2020-07-19
#1227 - Efficient subset check between CartesianProducts

### LINKED ISSUE #1227: Efficient subset check between CartesianProducts
This is the particular binary case of https://github.com/JuliaReach/LazySets.jl/pull/1225.

schillic:
Closes #1227.

Additionally this PR adds a subset check between `CartesianProductArray`s of different block structure by using the fallback implementation.

~Tests were already present.~

```julia
# two non-subset blocks

julia> @time X ⊆ Y
  0.000024 seconds (111 allocations: 7.453 KiB)
false

julia> @time ⊆(X, Y, true)
  0.000032 seconds (151 allocations: 9.891 KiB)
(false,   [1]  =  -0.989371
  [2]  =  0.890032
  [3]  =  1.15984
  [4]  =  2.03698)

### this branch

julia> @time X ⊆ Y
  0.000003 seconds
false

julia> @time ⊆(X, Y, true)
  0.000005 seconds (4 allocations: 272 bytes)
(false, [-0.9893707748634932, 0.6090025573953656, 0.6279955152649557, 0.38677706906864234])


# first block is a subset

julia> @time X ⊆ Y
  0.000025 seconds (129 allocations: 8.672 KiB)
false

julia> @time ⊆(X, Y, true)
  0.000030 seconds (163 allocations: 10.703 KiB)
(false,   [1]  =  -0.427312
  [2]  =  0.890032
  [3]  =  1.15984
  [4]  =  2.03698)

### this branch

julia> @time X ⊆ Y
  0.000003 seconds
false

julia> @time ⊆(X, Y, true)
    0.000005 seconds (6 allocations: 384 bytes)
(false, [-0.23845042779673375, 0.6386840809336349, 1.159841586170895, 0.6386840809336349])


# both blocks are subsets

julia> @time X ⊆ Y
  0.000025 seconds (147 allocations: 9.891 KiB)
true

julia> @time ⊆(X, Y, true)
  0.000026 seconds (149 allocations: 10.000 KiB)
(true, Float64[])

### this branch

julia> @time X ⊆ Y
  0.000003 seconds
true

julia> @time ⊆(X, Y, true)
  0.000004 seconds (6 allocations: 336 bytes)
(true, Float64[])
```

### DIFF
--- a/docs/src/lib/binary_functions.md
+++ b/docs/src/lib/binary_functions.md
@@ -122,6 +122,7 @@ pontryagin_difference
 ⊆(::LazySet{N}, ::Universe{N}, ::Bool=false) where {N<:Real}
 ⊆(::Universe{N}, ::LazySet{N}, ::Bool=false) where {N<:Real}
 ⊆(::LazySet{N}, ::Complement{N}, ::Bool=false) where {N<:Real}
+⊆(::CartesianProduct{N}, ::CartesianProduct{N}, ::Bool=false) where {N<:Real}
 ⊆(::CartesianProductArray{N}, ::CartesianProductArray{N}, ::Bool=false) where {N<:Real}
 ```
 
--- a/src/ConcreteOperations/issubset.jl
+++ b/src/ConcreteOperations/issubset.jl
@@ -985,7 +985,9 @@ witness.
 
 This algorithm requires that the two Cartesian products share the same block
 structure.
-Depending on the value of `check_block_equality`, we check this property.
+If `check_block_equality` is activated, we check this property and, if it does
+not hold, we use a fallback implementation based on conversion to constraint
+representation (assuming that the sets are polyhedral).
 
 ### Algorithm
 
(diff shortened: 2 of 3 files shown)


mforets:
there is a problem to extend `⊆(::CartesianProduct{N}, ::CartesianProduct{N}, ::Bool=false) ..` without dispatch on the algorithm, in the sense that the new method doesn't apply for cases which have different block structure, but which are polytopic, so the fallback implementation (`_issubset_constraints_list`) applies.

one approach would be to handle the case where the blocks dims don't match instead of throwing an error. otoh, this pr shows that we don't test `x ⊆ y` if `x = rand(Interval) × rand(BallInf)`  and `y = rand(BallInf) × rand(Interval)`.

schillic:
True, but the same holds for the `CPA` method. I copied that behavior as suggested in the issue. I can fix them here.

mforets:
sorry, what i meant in my comment is that if we merge this PR there is a regression as some subset checks no longer work as the one i gave. 

> True, but the same holds for the CPA method.

ok, i see. then we should fix that too :)

schillic:
> ok, i see. then we should fix that too :)

done :rescue_worker_helmet:

mforets:
cool! 

> return invoke(⊆, Tuple{LazySet{N}, LazySet{N}, Bool}, X, Y, witness)

i think that using internal functions is better than invoke, but we can fix that later.


> julia> @time ⊆(X, Y, true)
    0.000005 seconds (6 allocations: 384 bytes)

when measuring small times (like hundreds of ms or smaller), `@btime` from BenchmarkTools is better because the differences shown can be just variation between runs.

schillic:
> when measuring small times (like hundreds of ms or smaller), @btime from BenchmarkTools is better because the differences shown can be just variation between runs.

Yeah, I mainly looked at the memory usage here.

PR #2252, opened 2020-07-20
#1769 - Faster box approximation of VPolytope

### LINKED ISSUE #1769: Faster box approximation of VPolytope
The box approximation from a vertex representation can be computed more efficiently than evaluating the support function in all `2n` directions on all `m` vertices. The idea is to create the box incrementally: Start with the first vertex and then extend the bounds for each new vertex.
```julia
low = copy(vertices[1])
high = copy(vertices[1])
for v in vertices[2:m]
    for i in 1:n
        if v[i] < high[i]
            high[i] = v[i]
        elseif v[i] > low[i]
            low[i] = v[i]
        end
    end
end
return Hyperrectangle(low=low, high=high)
```
The crucial part is the `elseif`: If a vertex extends the upper bound in dimension `i`, it cannot extend the lower bound in the same dimension.

In fact, this idea generalizes to parallelotopes with fixed directions.

schillic:
Closes #1769.

In 2D this implementation is less efficient for a moderate number of vertices due to the additional conversion from `low`/`high` in the `Hyperrectangle` constructor. I guess in higher dimensions it will be more useful. Should I remove the dispatch for `VPolygon` or add an `invoke` of the old method based on the number of vertices?

```julia
julia> P1 = VPolytope([[0.0, 0.0], [2.0, 1.0], [1.0, 2.0]]);
julia> P2 = VPolytope([rand(2) for _ in 1:100]);
julia> P3 = VPolytope([rand(2) for _ in 1:1000]);

# this branch

julia> @time box_approximation(P1)
[0.0, 0.0][2.0, 2.0]
  0.000008 seconds (5 allocations: 416 bytes)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([1.0, 1.0], [1.0, 1.0])
julia> @time box_approximation(P2)
[0.001830275380493962, 0.005165832293896333][0.9791389830947983, 0.9914106144943142]
    0.000009 seconds (5 allocations: 416 bytes)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([0.49048462923764613, 0.4982882233941053], [0.48865435385715217, 0.49312239110020895])
julia> @time box_approximation(P3)
  0.000024 seconds (5 allocations: 416 bytes)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([0.4991620587806421, 0.4983966049474178], [0.49909969477169713, 0.4978833437247012])

# master

julia> @time box_approximation(P1)
  0.000004 seconds (3 allocations: 224 bytes)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([1.0, 1.0], [1.0, 1.0])
julia> @time box_approximation(P2)
  0.000008 seconds (3 allocations: 224 bytes)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([0.49048462923764613, 0.4982882233941053], [0.48865435385715217, 0.49312239110020895])
julia> @time box_approximation(P3)
  0.000049 seconds (3 allocations: 224 bytes)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([0.4991620587806421, 0.4983966049474178], [0.49909969477169713, 0.4978833437247012])
```

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -1513,6 +1513,24 @@ function overapproximate(am::AbstractAffineMap{N, <:AbstractHyperrectangle{N}},
     return Hyperrectangle(center_MXv, radius_MX)
 end
 
+function overapproximate(P::Union{VPolytope, VPolygon},
+                         ::Type{<:Hyperrectangle}) where {N<:Real}
+    n = dim(P)
+    vlist = vertices_list(P)
+    low = copy(vlist[1])
+    high = copy(vlist[1])
+    @inbounds for v in @view vlist[2:length(vlist)]
+        for i in 1:n
+            if v[i] > high[i]
+                high[i] = v[i]
+            elseif v[i] < low[i]
+                low[i] = v[i]
+            end
+        end
+    end
+    return Hyperrectangle(low=low, high=high)
+end
+
 """
     overapproximate(X::LazySet{N}, ::Type{<:Zonotope},
                     dir::AbstractDirections{N};
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -1514,21 +1514,30 @@ function overapproximate(am::AbstractAffineMap{N, <:AbstractHyperrectangle{N}},
 end
 
 function overapproximate(P::Union{VPolytope, VPolygon},
-                         ::Type{<:Hyperrectangle}) where {N<:Real}
+                         ::Type{<:Hyperrectangle})
     n = dim(P)
     vlist = vertices_list(P)
-    low = copy(vlist[1])
-    high = copy(vlist[1])
-    @inbounds for v in @view vlist[2:length(vlist)]
-        for i in 1:n
-            if v[i] > high[i]
-                high[i] = v[i]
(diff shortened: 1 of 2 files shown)


mforets:
i had a look, see [this notebook](https://github.com/mforets/escritoire/blob/master/2020/Week31/VPolytope_BoxOA.ipynb). in low dims (usually <= 10) static arrays are much faster that regular arrays so it'd be nice that our code supports them. it is not the case with the new method but see the alternative implementation `ba2` in that notebook. it is also faster for dim 2 and different number of vertices. for dimension 10, `ba1` is faster, though i didn't make exhaustive checks!

mforets:
> In 2D this implementation is less efficient for a moderate number of vertices

here `ba1` is faster than the version in master, while it allocates a bit more. see cells [15] and [9] in the notebook.

schillic:
I added the ideas from your implementation with minor modifications.

Side note: I always get one more allocation than in your notebook (but even when I use your exact version, so this is not a problem of my modifications). Maybe you want to double-check...

mforets (approved):
lgtm :tada:

PR #2266, opened 2020-07-23
Remove redundant option from condition

schillic:


mforets:
there was a reason for this.. @yupbank  do you remember?

schillic:
Indeed, there was [this commit](https://github.com/JuliaReach/LazySets.jl/pull/2199/commits/00d8e6ed6055fd3a7dd4e838739a6b5c6e3a2371). But it is not clear to me why the condition is needed.

yupbank:
the expected behavior of `apply_convex_hull = false` is to have all the 2^n points, this is just to ensure the same behavior. 

no sure who would be interested in those though.

schillic:
> the expected behavior (based on existing implementation) of apply_convex_hull = false is to have all the 2^n points

Okay, my interpretation was that the purpose of the option is to be able to skip the costly convex hull, and not that you want to have redundant vertices if it is deactivated. @mforets what do you think?

schillic:
This was already addressed in the refactoring in #2303.

PR #2268, opened 2020-07-24
Iteration protocol for OctDirections using a regular vector

mforets:


mforets:
for context, `BoxDirections` allows to specify the vector type (sparse, one-hot vector, regular), but for octdirections we only had sparse. this is useful for template hull intersections.

src/Approximations/template_directions.jl
```diff
@@ -350,11 +350,38 @@ function Base.iterate(od::OctDirections{N, SparseVector{N, Int}}, state::Tuple)
...
+
 function Base.iterate(od::OctDirections{N, SparseVector{N, Int}}, state::Int) where {N}
     # continue with box directions
     return iterate(BoxDirections{N, SparseVector{N, Int}}(od.n), state)
 end
 
+# ----------------------------------
+# implementation with regular arrays
+# ----------------------------------
+function Base.iterate(od::OctDirections{N, Vector{N}}) where {N}
```
schillic, line 365: I guess these methods can also serve as fallback implementation for any vector type (by removing `Vector{N}`).

mforets: one case that i would be interested is immutable vector types such as static arrays, but they would need a new implementation.

Issue #2245, opened 2020-07-19
Constrained dimensions of a line


mforets:
first implementation:

```julia
"""
    constrained_dimensions(L::Line)

Return the indices in which a line is constrained.

### Input

- `L` -- line

### Output

A vector of ascending indices `i` such that the line is constrained in dimension
`i`.

### Algorithm

For each coordinate ``i``, vectors of the form ``x_i = p_i + λ n_i`` are constrained
(i.e. they belong to the line and are bounded) if and only if ``n_i`` is zero.
Hence, this function returns all indices of the normal vector ``n`` for which
the ``i``-th coordinate is nonzero.

### Examples

The line ``y = 5`` in two dimensions can be written as ``p = [0, 5]`` and
``n = [1, 0]``. This line constrains dimension ``2``.

```jldoctest
julia> constrained_dimensions(Line([0, 5.], [1, 0.]))
1-element Array{Int64,1}:
 2
```
"""
function constrained_dimensions(L::Line)
    return findall(isapproxzero, L.n)
end
```

**Comment by schillic** [comment]
I think (where `d` is the dimension):
* If the only nonzero entry is in dim `i`, `constrained_dimensions` should return `[1, ..., i-1, i+1, ..., d]`.
* If there is more than one nonzero entry in `L.n`, `constrained_dimensions` should return `1:d`.

**Comment by mforets** [comment]
it all depends on the exact meaning of what this function is supposed to do; from the docstring, "Return the indices in which a line is constrained." and "constrained" as opposed to universal, the function does what it's supposed to return. the fact that the line can't in general be seen as a cartesian product is an additional constraint.

**Comment by schillic** [comment]
> Furthermore we should check consistency with the fallback implementation.

I just checked and this is not the case.

```julia
julia> L = Line([-1, 2, 3.], [3, 0, -1.]);

julia> constrained_dimensions(L)
3-element Array{Int64,1}:
 1
 2
 3

julia> constrained_dimensions2(L)  # implementation given in the OP
1-element Array{Int64,1}:
 2
```

It might very well be that we have to change the documentation and/or the name of the function to clarify the meaning.

PR #2271, opened 2020-07-25
#2245 - Add constrained dimensions of a line

mforets:
Closes https://github.com/JuliaReach/LazySets.jl/issues/2245.

---

Adding #2245 as i had done it already. Here the interpretation of a "constrained dimension" is that the set is bounded along the i-the dimension, otherwise it is "unconstrained". In fact, as discussed in #2245, for half-spaces the interpretation is the same. 

On the other hand, one can't blindly apply the `constrained_dimensions` to the BFPSV19 algorithm implemented in Reachability.jl (not yet available in RA) because it requires equality with the projection of the set along the constrained dimensions, times the cartesian prod with the universal set in the unconstrained dimensions.. using a line, one only gets an overapproximation.

### DIFF
--- a/docs/src/lib/sets/Line.md
+++ b/docs/src/lib/sets/Line.md
@@ -24,6 +24,7 @@ normalize!(::Line, ::Real=2.0)
 distance(::AbstractVector, ::Line, ::Real=2.0)
 distance(::AbstractSingleton, ::Line, ::Real=2.0)
 linear_map(::AbstractMatrix, ::Line)
+constrained_dimensions(::Line)
 ```
 Inherited from [`LazySet`](@ref):
 * [`norm`](@ref norm(::LazySet, ::Real))
--- a/src/Sets/Line.jl
+++ b/src/Sets/Line.jl
@@ -495,3 +495,39 @@ function _linear_map(M::AbstractMatrix, L::Line)
     Md = M * L.d
     return Line(Mp, Md)
 end
+
+"""
+    constrained_dimensions(L::Line)
+
+Return the indices in which a line is constrained.
+
+### Input
+
+- `L` -- line
+
+### Output
+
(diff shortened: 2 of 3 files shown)


schillic:
> In fact, as discussed in #2245, for half-spaces the interpretation is the same.

```julia
julia> L = Line(ones(3), [1., 0., 1.]);

julia> constrained_dimensions(L)  # implementation in master
3-element Array{Int64,1}:
 1
 2
 3

julia> function constrained_dimensions2(L::Line)  # implementation proposed here
           return findall(LazySets.isapproxzero, L.d)
       end

julia> constrained_dimensions2(L)
1-element Array{Int64,1}:
 2
```
The definition in `master` returns those dimensions `i` such that at least one constraint is nonzero in dimension `i`. 
```julia
julia> constraints_list(L)
4-element Array{HalfSpace{Float64,Array{Float64,1}},1}:
 HalfSpace{Float64,Array{Float64,1}}([0.0, 1.0, 0.0], 1.0)
 HalfSpace{Float64,Array{Float64,1}}([-0.0, -1.0, -0.0], -1.0)
 HalfSpace{Float64,Array{Float64,1}}([-0.7071067811865475, 0.0, 0.7071067811865475], 0.0)
 HalfSpace{Float64,Array{Float64,1}}([0.7071067811865475, -0.0, -0.7071067811865475], -0.0)
```

mforets:
Ideas for names:

-> `nonuniversal_dimensions`: those dimensions i such that at least one constraint is nonzero in dimension i.

-> `bounded_dimensions`: the set is bounded along the i-th dimension (meaning that both rho(ei, X) and rho(-ei, X) are bounded).

Examples:

- `x >= 0` in 2D the non-universal dimensions are [1]  and the bounded dimensions are [].

mforets:
Closing as stale.

PR #2273, opened 2020-07-25
#2063 - Add concrete in-place linear map for zonotope

### LINKED ISSUE #2063: Inplace linear map for zonotopic sets
An implementation is given here https://github.com/mforets/ReachabilityAnalysis.jl/blob/master/src/Flowpipes/setops.jl#L11

mforets:
Closes #2063.

### DIFF
--- a/src/Sets/Zonotope.jl
+++ b/src/Sets/Zonotope.jl
@@ -406,3 +406,24 @@ function _split(Z::Zonotope, gens::AbstractVector, n::AbstractVector)
     end
     return Zs
 end
+
+"""
+    linear_map!(Zout::Zonotope, M::AbstractMatrix, Z::Zonotope)
+
+Compute the concrete linear map of a zonotope storing the result in `Zout`.
+
+### Input
+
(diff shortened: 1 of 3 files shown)


src/Sets/Zonotope.jl
```diff
@@ -406,3 +406,24 @@ function _split(Z::Zonotope, gens::AbstractVector, n::AbstractVector)
...
+
+- `Zout` -- zonotope (output)
+- `M`    -- matrix
+- `Z`    -- zonotope
+
+### Output
+
+The zonotope `Zout`, which is modified in-place.
+"""
+function linear_map!(Zout::Zonotope, M::AbstractMatrix, Z::Zonotope)
```
schillic, line 425: Does this also work if `Zout === Z`?

mforets: no; in `mul!(Y, A, B)`, `Y` must not be aliased with either `A` or `B` (see `?mul!`)

schillic: Makes sense.

PR #2284, opened 2020-07-27
#2283 - Add concrete projection of a VPolytope

### LINKED ISSUE #2283: Concrete projection of a VPolytope
... should return a VPolytope

```julia
X = rand(VPolytope, dim=3)
project(X, [2])

AssertionError: package 'Polyhedra' not loaded (it is required for executing `default_polyhedra_backend`)

Stacktrace:
 [1] macro expansion at /home/mforets/.julia/dev/LazySets/src/Assertions/Assertions.jl:23 [inlined]
 [2] require(::Symbol; fun_name::String, explanation::String) at /home/mforets/.julia/dev/LazySets/src/init.jl:37
 [3] default_polyhedra_backend(::VPolytope{Float64,Array{Float64,1}}, ::Type{T} where T) at /home/mforets/.julia/dev/LazySets/src/Interfaces/AbstractPolyhedron_functions.jl:24
 [4] tohrep(::VPolytope{Float64,Array{Float64,1}}) at /home/mforets/.julia/dev/LazySets/src/Sets/VPolytope.jl:518
 [5] constraints_list(::VPolytope{Float64,Array{Float64,1}}) at /home/mforets/.julia/dev/LazySets/src/Sets/VPolytope.jl:378
 [6] constrained_dimensions(::VPolytope{Float64,Array{Float64,1}}) at /home/mforets/.julia/dev/LazySets/src/Interfaces/AbstractPolyhedron_functions.jl:120
 [7] project(::VPolytope{Float64,Array{Float64,1}}, ::Array{Int64,1}) at /home/mforets/.julia/dev/LazySets/src/Approximations/decompositions.jl:638
 [8] top-level scope at In[22]:2
```

SebastianGuadalupe:
Closes #2283.

### DIFF
--- a/src/Approximations/decompositions.jl
+++ b/src/Approximations/decompositions.jl
@@ -665,5 +665,5 @@ function project(V::Union{<:VPolygon{N}, <:VPolytope{N}},
     n = dim(V)
     M = projection_matrix(block, n, N)
     πvertices = broadcast(v -> M * v, vertices_list(V))
-    return basetype(V)(πvertices)
+    return VPolytope(πvertices)
 end
--- a/src/Approximations/decompositions.jl
+++ b/src/Approximations/decompositions.jl
@@ -659,3 +659,11 @@ function project(H::AbstractHyperrectangle, block::AbstractVector{Int})
     πr = radius_hyperrectangle(H)[block]
     return Hyperrectangle(πc, πr, check_bounds=false)
 end
+
+function project(V::Union{<:VPolygon{N}, <:VPolytope{N}},
+                 block::AbstractVector{Int}) where {N}
+    n = dim(V)
+    M = projection_matrix(block, n, N)
+    πvertices = broadcast(v -> M * v, vertices_list(V))
+    return basetype(V)(πvertices)
+end
--- a/test/unit_Polygon.jl
+++ b/test/unit_Polygon.jl
@@ -391,6 +391,10 @@ for N in [Float64, Float32, Rational{Int}]
     m = N[4 0; 6 2; 4 4]'
     P = VPolygon(m)
     @test is_cyclic_permutation(vertices_list(P), [N[4, 0], N[6, 2], N[4, 4]])
+
+    # test concrete projection
+    V = VPolygon([N[0, 1], N[1, 0], N[-1, 0]])
+    @test project(V, [1]) == VPolygon([N[-1], N[1]])
 end
 
 function same_constraints(v::Vector{<:LinearConstraint{N}})::Bool where N<:Real
--- a/test/unit_Polygon.jl
+++ b/test/unit_Polygon.jl
@@ -395,6 +395,8 @@ for N in [Float64, Float32, Rational{Int}]
     # test concrete projection
     V = VPolygon([N[0, 1], N[1, 0], N[-1, 0]])
     @test project(V, [1]) == VPolygon([N[-1], N[1]])
+    V = VPolygon([N[1, 0], N[1, 1]])
+    @test project(V, [1]) == VPolygon([N[1], N[1]])
 end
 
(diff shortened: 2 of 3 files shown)


test/unit_Polygon.jl
```diff
@@ -391,6 +391,10 @@ for N in [Float64, Float32, Rational{Int}]
     m = N[4 0; 6 2; 4 4]'
     P = VPolygon(m)
     @test is_cyclic_permutation(vertices_list(P), [N[4, 0], N[6, 2], N[4, 4]])
+
+    # test concrete projection
+    V = VPolygon([N[0, 1], N[1, 0], N[-1, 0]])
+    @test project(V, [1]) == VPolygon([N[-1], N[1]])
```
schillic, line 397: Can you add a test for `V = VPolygon([N[1, 0], N[1, 1]]); project(V, [1])`? This shows that in general you can get duplicates. I wonder if we should add an option to apply a convex hull.

mforets: > I wonder if we should add an option to apply a convex hull.

note that the VPolygon constructor has such option but the VPolytope does not (such option would require loading Polyhedra).

src/Approximations/decompositions.jl
```diff
@@ -659,3 +659,11 @@ function project(H::AbstractHyperrectangle, block::AbstractVector{Int})
     πr = radius_hyperrectangle(H)[block]
     return Hyperrectangle(πc, πr, check_bounds=false)
 end
+
+function project(V::Union{<:VPolygon{N}, <:VPolytope{N}},
+                 block::AbstractVector{Int}) where {N}
+    n = dim(V)
+    M = projection_matrix(block, n, N)
+    πvertices = broadcast(v -> M * v, vertices_list(V))
+    return basetype(V)(πvertices)
```
schillic, line 668: Projecting a `VPolygon` would probably give you an `Interval`. You need to check the dimension.

schillic: I suggest to exclude the `VPolygon` case here. The implementation can be made more efficiently if you know that you go to 1D. In particular, you do not need to create the list of vertices and then run a convex hull. Instead you can keep track of upper and lower bound and in the end return an `Interval`.

mforets: >  I am surprised that this does not crash.

VPolygon constructor doesn't check dimension (unless you pass a vertices matrix). I think it's a good idea to return an interval if the output dimension of `M` is `1`. we could also return a VPolytope.

schillic: For now we can return a `VPolytope` and open an issue.

mforets: > I suggest to exclude the VPolygon case here.

yes, though the implementation in this pr is better than the fallback implementation in master i guess? (which converts to constraints list). that can be handled as a follow up issue?

schillic: The current implementation is wrong. If you change it to `VPolytope` that is fine.

test/unit_Polygon.jl
```diff
@@ -391,6 +391,12 @@ for N in [Float64, Float32, Rational{Int}]
...
+    V = VPolygon([N[1, 0], N[1, 1]])
+    @test project(V, [1]) == VPolygon([N[1], N[1]])
```
schillic, line 399:
```suggestion
    V = VPolygon([N[0, 1], N[1, 0], N[-1, 0]])
    @test project(V, [1]) == VPolytope([N[-1], N[1]])
    V = VPolygon([N[1, 0], N[1, 1]])
    @test project(V, [1]) == VPolytope([N[1], N[1]])
```
I guess...

mforets:
the error is unrelated (building MiniQhull)

PR #2286, opened 2020-07-28
#241 - Macro for commutative binary functions

### LINKED ISSUE #241: Add macro to create symmetric functions
```
@symmetric function myfunction(arg1::T1, arg2::T2)
    (code)
end
```
should produce
```
function myfunction(arg1::T1, arg2::T2)
    (code)
end
function myfunction(arg2::T2, arg1::T1)
    (code)
end
```
Careful: There are method ambiguities for `myfunction(::T1, ::T1)` and `myfunction(::T2, ::T2)`.

mforets:
Closes #241.


src/ConcreteOperations/intersection.jl
```diff
@@ -19,17 +17,10 @@ Return the intersection of a singleton with another set.
 If the sets intersect, the result is `S`.
 Otherwise, the result is the empty set.
 """
-function intersection(S::AbstractSingleton{N},
-                      X::LazySet{N}
-                     ) where {N<:Real}
-    return element(S) ∈ X ? S : EmptySet{N}(dim(S))
-end
+intersection(S::AbstractSingleton{N}, X::LazySet{N}) where {N<:Real}
```
mforets, line 20: this can be probably further improved, such that Julia can actually document one of the methods (say the original method).

schillic: Documentation seems to work in newer Julia versions, right?

mforets:
> for the a single intersection function and let Julia find the relevant methods automatically.

i didn't understand, the different `intersection` methods often have different docstrings.

mforets:
> The documentation does not seem to work in v1.0. 

:/ it's a pity

schillic:
> > for the a single intersection function and let Julia find the relevant methods automatically.
> 
> i didn't understand, the different `intersection` methods often have different docstrings.

I meant to replace the whole list [here](https://github.com/JuliaReach/LazySets.jl/blob/master/docs/src/lib/binary_functions.md#intersection-of-two-sets) by the single line
```
intersection
```

mforets:
i kind of like to have this list, because it's a nice way to see all the different specializations in one paragaph.

src/ConcreteOperations/intersection.jl
schillic, line 20:
```suggestion
function intersection(S::AbstractSingleton{N}, X::LazySet{N}) where {N<:Real} end
```

PR #2287, opened 2020-07-28
update vpolytope projection

mforets:
follow up of https://github.com/JuliaReach/LazySets.jl/pull/2284, adds some dimensional checks and (runtime) dispatch. the drawback of this approach is the type-instability, but it has all other benefits such as specialized methods on the returned type.

### DIFF
--- a/src/Approximations/decompositions.jl
+++ b/src/Approximations/decompositions.jl
@@ -665,5 +665,17 @@ function project(V::Union{<:VPolygon{N}, <:VPolytope{N}},
     n = dim(V)
     M = projection_matrix(block, n, N)
     πvertices = broadcast(v -> M * v, vertices_list(V))
-    return VPolytope(πvertices)
+
+    m = size(M, 1)
+    if m == 1
+        # convex_hull in 1d returns the minimum and maximum points, in that order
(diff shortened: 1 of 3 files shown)


src/Approximations/decompositions.jl
```diff
@@ -665,5 +665,17 @@ function project(V::Union{<:VPolygon{N}, <:VPolytope{N}},
...
+
+    m = size(M, 1)
+    if m == 1
+        # convex_hull in 1d returns the minimum and maximum points, in that order
+        aux = convex_hull(πvertices)
+        a = first(aux[1])
+        b = length(aux) == 1 ? a : first(aux[2])
+        return Interval(a, b)
+    elseif m == 2
+        return VPolygon(convex_hull(πvertices))
```
schillic, line 677:
```suggestion
        return VPolygon(πvertices; apply_convex_hull=true)
```
The constructor already applies a convex hull. I think it is clearer to specify the option explicitly, though.

mforets:
> the drawback of this approach is the type-instability, but it has all other benefits such as specialized methods on the returned type.

some middle ground is to use type-stable (non-exported) functions for more advanced use cases.

Issue #2209, opened 2020-07-05
Faster vertex enumeration of 2d zonotopes


mforets:
Idea: generalize https://github.com/JuliaReach/LazySets.jl/pull/2199#issuecomment-650632317.

mforets:
cc: @yupbank

yupbank:
can you assign this to me? I can't assign :(

mforets:
related : https://github.com/JuliaReach/LazySets.jl/issues/1404

PR #2288, opened 2020-07-28
General 2d Zonotope  vertex enumeration with linear complexity

yupbank:
The algorithm can be summarized in the following steps:
  - Partition the generators into four quadrants using this angle function $f: R^2 \to [0, 360)$. 
  - Enumerate within each quadrant using the previously implemented method, with a polygon representation.
  - Reduce over the polygons using `minkowski_sum` 

addressing  https://github.com/JuliaReach/LazySets.jl/issues/2209


src/Interfaces/AbstractZonotope.jl
```diff
@@ -419,6 +419,18 @@ https://math.stackexchange.com/q/3356460
 To avoid cumulative sum from both directions separately, we build a 2d index matrix
 to sum generators for both directions in one matrix-vector product.
 """
+
+@inline function _angles(point::AbstractVector{N})  where{N<:Real}
+    (atan(point[2], point[1]) /π * 180 + 360)%360
+end
+
+function _enum(G::AbstractMatrix{N}, sorted::Bool=true) where{N<:Real}
```
schillic, line 427: Can you use a more specific function name, please? :)
It helps to understand the functionality a bit and to avoid accidentally calling it in other contexts.
We often just use the caller name plus some addition, like `_vertices_list_zonotope_enum`, but other names are also fine.

src/Interfaces/AbstractZonotope.jl
```diff
@@ -419,6 +419,18 @@ https://math.stackexchange.com/q/3356460
 To avoid cumulative sum from both directions separately, we build a 2d index matrix
 to sum generators for both directions in one matrix-vector product.
 """
+
+@inline function _angles(point::AbstractVector{N})  where{N<:Real}
```
schillic, line 423: I think you need to keep this docstring attached to the `vertices_list` method.

Also, is the algorithm still the same or would you update the docstring?

schillic:
I do not really understand this algorithm, but ignoring the center and only adding it in the end (which you do), you could exploit that the zonotope is centrally symmetric and only enumerate the vertices in two quadrants (and then obtain the symmetric vertices in the other two quadrants by just negating them), right? (Maybe you already do that :sweat_smile:)

src/Interfaces/AbstractZonotope.jl
mforets: see https://github.com/JuliaReach/LazySets.jl/pull/2303, which prepares the stage for the new 2d method here https://github.com/JuliaReach/LazySets.jl/pull/2303/files#diff-42ebbc16f90bae7a08e04f097c7aeb9eR506

src/Sets/Zonotope.jl
```diff
@@ -406,3 +406,36 @@ function _split(Z::Zonotope, gens::AbstractVector, n::AbstractVector)
...
+
+@inline function _angles(point::AbstractVector{N})  where{N<:Real}
+    (atan(point[2], point[1]) /π * 180 + 360)%360
+end
+
+function _single_quadrant_vertices_enum(G::AbstractMatrix{N}, sorted::Bool=true) where{N<:Real}
+    if !sorted
+        G = sortslices(G, dims=2, by=_angles)
+     end
+    return VPolygon(2*cumsum(hcat(G, -G), dims=2) .- sum(G, dims=2))
```
yupbank, line 418: this one line is used to replace https://github.com/JuliaReach/LazySets.jl/blob/master/src/Interfaces/AbstractZonotope.jl#L433-L438 
from my testing, when the number of generators on the scale of 100, this implementation would be much faster. as the size of index in previous implementation growths in the scale of `p^2`

yupbank:
```julia
2*cumsum(hcat(G, -G), dims=2) .- sum(G, dims=2)
```
is the same as 
```julia
           p = size(G, 2)
            index = ones(p, 2*p)
            @inbounds for i in 1:p
                index[i, i+1:i+p-1] .= -1
            end
            index[:, 1] .= -1
            G * index
```

src/Sets/Zonotope.jl
```diff
@@ -533,3 +508,35 @@ function _vertices_list_iterative(c::AbstractVector{N}, G::AbstractMatrix{N}; ap
...
+end
+
+function _vertices_list_2D(c::AbstractVector{N}, G::AbstractMatrix{N}, apply_convex_hull::Bool) where {N<:Real}
```
schillic, line 523:
```suggestion
function _vertices_list_2D(c::AbstractVector{N}, G::AbstractMatrix{N}; apply_convex_hull::Bool) where {N<:Real}
```
This fixes the build error.

src/Sets/Zonotope.jl
```diff
@@ -554,29 +554,36 @@ end
...
+    return VPolygon(2*cumsum(hcat(G, -G), dims=2) .- sum(G, dims=2))
+end
 
-    sorted_G = sortslices(G, dims=2, by=x->atan(x[2], x[1]))
-    index = ones(N, p, 2*p)
-    @inbounds for i in 1:p
-        index[i, i+1:i+p-1] .= -one(N)
+function _vertices_list_2D(c::AbstractVector{N}, G::AbstractMatrix{N}; apply_convex_hull::Bool) where {N}
+    if apply_convex_hull
+        return _vertices_list_iterative(c, G, apply_convex_hull)
```
mforets, line 570:
```suggestion
        return _vertices_list_iterative(c, G, apply_convex_hull=apply_convex_hull)
```
see https://travis-ci.org/github/JuliaReach/LazySets.jl/jobs/730769538#L678

src/Sets/Zonotope.jl
```diff
@@ -554,29 +554,36 @@ end
...
-function _vertices_list_2D(c::AbstractVector{N}, G::AbstractMatrix{N}; apply_convex_hull::Bool) where {N}
-    if same_sign(G)
-        return _vertices_list_2D_positive(c, G)
-    else
-        # FIXME generalized 2D vertices list function is not implemented yet
-        # See LazySets#2209
-        return _vertices_list_iterative(c, G, apply_convex_hull=apply_convex_hull)
-    end
+@inline function _angles(point::AbstractVector{N})  where{N}
+    (atan(point[2], point[1]) / π * 180 + 360) % 360
```
mforets, line 558:
```suggestion
    (atand(point[2], point[1]) + 360) % 360
```

src/Sets/Zonotope.jl
```diff
@@ -554,29 +554,36 @@ end
...
-    end
+@inline function _angles(point::AbstractVector{N})  where{N}
+    (atand(point[2], point[1]) + 360) % 360
 end
 
-function _vertices_list_2D_positive(c::AbstractVector{N}, G::AbstractMatrix{N}) where {N}
-    n, p = size(G)
-
-    # TODO special case p = 1 or p = 2 ?
+function _single_quadrant_vertices_enum(G::AbstractMatrix{N}, sorted::Bool=true) where{N}
```
schillic, line 561: I would set the default value of `sorted` to `false` and pass `true` explicitly at the (currently only) call site below.

mforets (approved):

PR #2306, opened 2020-08-02
Clean up underapproximate

mforets:


src/Approximations/underapproximate.jl
```diff
@@ -14,16 +14,27 @@ Compute the underapproximation of a convex set by sampling support vectors.
 
 The `VPolytope` obtained by taking the convex hull of the support vectors of `X`
 along the directions determined by `dirs`.
+
+### Notes
+
+Since the support vectors are not always unique, this algorithm may return
+a strict underapproximation even if the set can be exactly approximated using
+the given template. For example,
```
mforets, line 22: adding an example here may be relevant

src/Approximations/underapproximate.jl
```diff
@@ -14,16 +14,27 @@ Compute the underapproximation of a convex set by sampling support vectors.
...
 """
 function underapproximate(X::LazySet{N}, dirs::AbstractDirections;
-                          apply_convex_hull::Bool=false) where {N<:Real}
-    @assert LazySets.dim(X) == LazySets.dim(dirs)
-    vinner = Vector{Vector{N}}(undef, length(dirs)) # TODO could be generalized to eltype(dirs)
+                          apply_convex_hull::Bool=false) where {N}
+    @assert dim(X) == dim(dirs) "the dimension of the set, $(dim(X)), doesn't match " *
+                                "the dimension of the template directions, $(dim(dirs))"
+    vinner = Vector{Vector{N}}(undef, length(dirs))
```
mforets, line 28: the problem of using `eltype(dirs)` is that such vector type is not always preserved. for instance if it's a SEV, the support vectors can be full. there are of course cases where keeping the vector type is relevant.

schillic (approved):
I suggest that we merge this.

mforets:
ok let's merge when CI finishes

in the last commit i added an in-place version, with the purpose of generalizing the output array type

Issue #2265, opened 2020-07-23
Use iterator methods `constraints` and `vertices`


schillic:
After merging #2263 and to make use of  #2264 we should internally refactor `constraints_list` to `constraints` and `vertices_list` to `vertices` whenever we just need an iterator.

PR #2313, opened 2020-08-05
#2312 - Fix robust membership test for HalfSpace

### LINKED ISSUE #2312: Inconsistent membership check
```julia
julia> using LazySets

julia> o = [0.07768723948819561, -0.5762273280928935, 0.28897399484750297, 1.9299362784322858];

julia> H = HalfSpace([-0.09291863543681655, -0.2176689899601838, -0.07453829739226348, 0.048948632014371496], 0.1911363393469332);

julia> P = HPolyhedron([H]);

julia> o ∈ H
false

julia> o ∈ P
true
```

reported https://github.com/sisl/NeuralVerification.jl/issues/144

mforets:
Closes #2312.

### DIFF
--- a/test/unit_HalfSpace.jl
+++ b/test/unit_HalfSpace.jl
@@ -133,7 +133,7 @@ for N in [Float64, Float32]
     @test normalize(hs1, N(Inf)) == HalfSpace(N[1//2, 1], N(3//2))
 end
 
-for N in Float64
+for N in [Float64]
     # test robustness of membership function (see LazySets#2312)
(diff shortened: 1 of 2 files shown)


schillic:
Side note: Once we have #2265, we can remove this method altogether and use the fallback implementation, which should be equivalent.

test/unit_HalfSpace.jl
```diff
@@ -133,6 +133,13 @@ for N in [Float64, Float32]
     @test normalize(hs1, N(Inf)) == HalfSpace(N[1//2, 1], N(3//2))
 end
 
+for N in Float64
```
mforets, line 136:
```suggestion
for N in [Float64]
```

PR #2315, opened 2020-08-05
Add HPolytope with MT + let macro work on tuples

mforets:
The PR contains two updates:

- add macro for `HPolytope`

- in master one has to extract the tuple argument

```julia
vars, = @variables x[1:2]
HalfSpace(x[1] <= 1, x)
```
with this branch we can do

```julia
vars = @variables x[1:2]
HalfSpace(x[1] <= 1, x)
```

### DIFF
--- a/src/Initialization/init_ModelingToolkit.jl
+++ b/src/Initialization/init_ModelingToolkit.jl
@@ -4,8 +4,38 @@ eval(quote
                             simplify,
                             Operation
 
+   """
+       _vec(vars::NTuple{L, Union{<:Operation, <:Vector{Operation}}}) where {L}
+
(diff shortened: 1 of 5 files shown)


src/Initialization/init_ModelingToolkit.jl
```diff
@@ -4,8 +4,38 @@ eval(quote
...
+
+   A vector of `Operation` obtained by concatenating each tuple component.
+
+   ## Examples
+
+   ```julia
+   julia> vars = @variables x[1:2] y
+   (Operation[x₁, x₂], y)
+
+   julia> _vec(vars)
```
schillic, line 26: I guess you need this...
```suggestion
   julia> LazySets._vec(vars)
```

PR #2317, opened 2020-08-07
Improve zonotope - hyperplane disjointness check

mforets:
the implementation was sub-optimal and it was not robust. for benchmarks see https://github.com/mforets/escritoire/blob/master/2020/Week32/ZG_Disjointness.ipynb . done jointly with @SebastianGuadalupe

### DIFF
--- a/src/ConcreteOperations/isdisjoint.jl
+++ b/src/ConcreteOperations/isdisjoint.jl
@@ -463,10 +463,10 @@ function _is_intersection_empty(Z::Zonotope{N}, H::Hyperplane{N}, ::Val{false}) 
     n, p = size(G)
     abs_sum = zero(N)
     if p > 0
-        @inbounds for i in 1:n
+        @inbounds for j in 1:p
             aux = zero(N)
-            @simd for j in 1:p
-                aux += H.a[j] * G[j, i]
+            @simd for i in 1:n
+                aux += H.a[i] * G[i, j]
             end
             abs_sum += abs(aux)
         end
--- a/src/ConcreteOperations/isdisjoint.jl
+++ b/src/ConcreteOperations/isdisjoint.jl
@@ -418,8 +418,7 @@ end
 
 
 """
-    is_intersection_empty(Z::Zonotope{N}, H::Hyperplane{N}, witness::Bool=false
-                         ) where {N<:Real}
+    is_intersection_empty(Z::Zonotope{N}, H::Hyperplane{N}, witness::Bool=false) where {N<:Real}
 
 Check whether a zonotope and a hyperplane do not intersect, and otherwise
 optionally compute a witness.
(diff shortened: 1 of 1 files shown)


src/ConcreteOperations/isdisjoint.jl
```diff
@@ -446,33 +445,37 @@ center, and ``g_i`` are the zonotope's generators.
...
+
+    n, p = size(G)
+    abs_sum = zero(N)
+    if p > 0
+        @inbounds for i in 1:n
+            aux = zero(N)
+            @simd for j in 1:p
+                aux += H.a[i] * G[j, i]
+            end
+            abs_sum += abs(aux)
```
schillic, line 471: You can pull out the multiplication with `H.a[i]` which is a constant here.
And I think it should be `G[i, j]`, right?
```suggestion
            aux = zero(N)
            @simd for j in 1:p
                aux += G[i, j]
            end
            abs_sum += abs(H.a[i] * aux)
```

src/ConcreteOperations/isdisjoint.jl
```diff
@@ -446,33 +445,37 @@ center, and ``g_i`` are the zonotope's generators.
...
+
+function _is_intersection_empty(Z::Zonotope{N}, H::Hyperplane{N}, ::Val{false}) where {N}
+    c, G = Z.center, Z.generators
+    v = H.b - dot(H.a, c)
+
+    n, p = size(G)
+    abs_sum = zero(N)
+    if p > 0
+        @inbounds for i in 1:n
+            aux = zero(N)
```
schillic, line 467: Another idea might be to first check whether `H.a[i]` is not zero. For sparse vectors this might actually pay off. But I am less sure about this suggestion...
```suggestion
            if isapproxzero(H.a[i])
                continue
            end
            aux = zero(N)
```

src/ConcreteOperations/isdisjoint.jl
mforets: good catch, that should have been `H.a[j] * G[j, i]`. the multiplication is from the left because we want to sum `|<a, g>|` for each generator `g`.

src/ConcreteOperations/isdisjoint.jl
```diff
@@ -446,33 +445,37 @@ center, and ``g_i`` are the zonotope's generators.
...
+            @simd for j in 1:p
+                aux += H.a[i] * G[j, i]
```
mforets, line 469:
```suggestion
                aux += H.a[j] * G[j, i]
```

src/ConcreteOperations/isdisjoint.jl
schillic: I think this is wrong. `j` runs through the number of columns but you use it for the rows, and vice versa for `i`.

mforets: i've changed the variable's names to match what we usually use.

src/ConcreteOperations/isdisjoint.jl
mforets: true that if `H.a` is sparse in the inner loop we can iterate over `H.a.nzind`.

mforets: i'm fine using keeping the current implementation. it can be specialized further on that use case by dispatching on the vector type of the hyperplane.

mforets: SEV can also be improved, etc

PR #2328, opened 2020-08-11
#2327 - Hyperplane-Hyperplane intersection emptiness

### LINKED ISSUE #2327: Hyperplane-Hyperplane intersection emptiness
```julia
julia> isdisjoint(rand(Hyperplane), rand(Hyperplane))
ERROR: the support vector for the hyperplane with normal direction [0.85425817454693, -0.8751753682278215]
is not defined along a direction [-0.5262298487743711, 1.0888813935440376]
```
This should be a simple system of linear equations.

schillic:
Closes #2327.

### DIFF
--- a/src/ConcreteOperations/isdisjoint.jl
+++ b/src/ConcreteOperations/isdisjoint.jl
@@ -744,12 +744,31 @@ function is_intersection_empty(hp::Union{Hyperplane{N}, Line2D{N}},
     return is_intersection_empty_helper_hyperplane(hp, X, witness)
 end
 
-# disambiguation
 function is_intersection_empty(hp1::Union{Hyperplane{N}, Line2D{N}},
                                hp2::Union{Hyperplane{N}, Line2D{N}},
                                witness::Bool=false
                               ) where {N<:Real}
-    return is_intersection_empty_helper_hyperplane(hp1, hp2, witness)
+    if isequivalent(hp1, hp2)
+        res = false
+        if witness
+            w = an_element(hp1)
+        end
+    else
+        cap = intersection(hp1, hp2)
+        res = cap isa EmptySet
+        if !res && witness
+            w = an_element(cap)
+        end
+    end
+    if witness
+        if res
+            return (true, N[])
+        else
+            return (false, w)
+        end
+    else
+        return res
+    end
 end
 
 # disambiguation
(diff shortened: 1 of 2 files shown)


src/ConcreteOperations/isdisjoint.jl
```diff
@@ -744,12 +744,31 @@ function is_intersection_empty(hp::Union{Hyperplane{N}, Line2D{N}},
...
 function is_intersection_empty(hp1::Union{Hyperplane{N}, Line2D{N}},
                                hp2::Union{Hyperplane{N}, Line2D{N}},
                                witness::Bool=false
                               ) where {N<:Real}
-    return is_intersection_empty_helper_hyperplane(hp1, hp2, witness)
+    if isequivalent(hp1, hp2)
+        res = false
+        if witness
+            w = an_element(hp1)
+        end
```
schillic, line 755: This case is needed because `intersection(hp1, hp2)` would return an `HPolyhedron` and then `an_element` would return `[Inf, Inf]`.

src/ConcreteOperations/isdisjoint.jl
```diff
@@ -744,12 +744,31 @@ function is_intersection_empty(hp::Union{Hyperplane{N}, Line2D{N}},
...
+    else
+        cap = intersection(hp1, hp2)
+        res = cap isa EmptySet
```
schillic, line 758: I first tried solving a linear system, but that throws an exception if there is no intersection.

This implementation is not really efficient, but at least it does not crash as in `master`...

mforets: to check disjointness of Line2D there is [this approach](https://github.com/JuliaReach/LazySets.jl/blob/master/src/ConcreteOperations/intersection.jl#L86)

mforets: for the n-dimensional hyperplane case, there is a full length explanation [here](https://math.stackexchange.com/questions/2360603/how-to-find-the-intersection-of-two-hyperplanes-in-n-dimensions)

schillic: > to check disjointness of Line2D there is this approach

Yep, that will be called by `intersection` :sweat_smile:

Issue #2264, opened 2020-07-23
Specialized iterator methods `constraints` and `vertices`


schillic:
After merging #2263 we can add efficient methods.

Trivial cases:
* [ ] iterator with no element:
  * [ ] `constraints` for `Universe`
  * [ ] `vertices` for `EmptySet`
* [ ] iterator over a single element: `vertices` for `AbstractSingleton`
* [ ] iterator over a list:
  * [ ] `constraints` for `AbstractHPolygon`/`HPoly`
  * [ ] `vertices` for `VPolygon`/`VPolytope`

Non-exhaustive list of interesting cases:
* [ ] `AbstractHyperrectangle` (@schillic has a draft implementation)

PR #2331, opened 2020-08-12
#2264 - vertices iterator for AbstractHyperrectangle

schillic:
See #2264.

I was surprised that the iterator is generally slower. It is faster if it is not exhausted, though. I checked everything and I would conclude that the additional allocations are due to the way iteration is implemented: creation of a new `Tuple` in each iteration.

Defining the generator implementation:

```julia
function vertices2(H::AbstractHyperrectangle)
    n = dim(H)
    c = center(H)
    r = radius_hyperrectangle(H)
    p = Iterators.product(fill([1, -1], n)...)
    return Base.Generator(x -> c .+ r .* x, p)
end
```

```julia
julia> H = rand(Hyperrectangle, dim=10);

julia> function f(H)
           for v in LazySets.vertices2(H)
           end
       end;

julia> function g(H)
           for v in vertices(H)
           end
       end;

julia> function h(H)
           for v in vertices_list(H)
           end
       end;

julia> function f2(H, k)
           for (i, v) in enumerate(LazySets.vertices2(H))
               if i == k
                   return
               end
           end
       end;

julia> function g2(H, k)
           for (i, v) in enumerate(vertices(H))
               if i == k
                   return
               end
           end
       end;

julia> function h2(H, k)
           for (i, v) in enumerate(vertices_list(H))
               if i == k
                   return
               end
           end
       end;

# generate all vertices
julia> @btime f($H)
  400.304 μs (6151 allocations: 816.67 KiB)

julia> @btime g($H)
  62.453 μs (2052 allocations: 192.31 KiB)

julia> @btime h($H)
  50.713 μs (1027 allocations: 168.38 KiB)

# generate only k vertices
julia> @btime f2($H, 2)
  1.055 μs (15 allocations: 1.78 KiB)

julia> @btime g2($H, 2)
  249.454 ns (9 allocations: 752 bytes)

julia> @btime h2($H, 2)  # h/h2 are equivalent independent of k
  50.362 μs (1027 allocations: 168.38 KiB)

julia> @btime g2($H, 3)
  323.396 ns (12 allocations: 992 bytes)

julia> @btime f2($H, 4)
  1.448 μs (23 allocations: 2.88 KiB)

julia> @btime g2($H, 4)
  395.905 ns (15 allocations: 1.20 KiB)
```
The last two runs show that Incrementing `k` leads to three new allocations of size 240 bytes. One of them is of course the additional vertex that is generated. The two other allocations are probably the `Tuple`s; at least I do not see where else the code would allocate anything depending on `k`.

### DIFF
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -364,6 +364,7 @@ radius(::AbstractHyperrectangle, ::Real=Inf)
 σ(::AbstractVector, ::AbstractHyperrectangle)
 ρ(::AbstractVector, ::AbstractHyperrectangle)
 ∈(::AbstractVector, ::AbstractHyperrectangle)
+vertices(::AbstractHyperrectangle)
 vertices_list(::AbstractHyperrectangle)
 constraints_list(::AbstractHyperrectangle{N}) where {N}
 high(::AbstractHyperrectangle)
--- a/src/Interfaces/AbstractHyperrectangle.jl
+++ b/src/Interfaces/AbstractHyperrectangle.jl
@@ -146,6 +146,128 @@ end
 
 
 """
+    vertices(H::AbstractHyperrectangle)
+
+Construct an iterator over the vertices of a hyperrectangular set.
+
+### Input
+
+- `H` -- hyperrectangular set
+
+### Output
+
+An iterator over the vertices.
+
+### Algorithm
+
+We implement two different iterators, depending on whether the set is flat in
+some dimension.
+"""
+function vertices(H::AbstractHyperrectangle)
+    n = dim(H)
+    start = -1
+    @inbounds for i in 1:n
+        if iszero(radius_hyperrectangle(H, i))
+            start = i  # found flat dimension
+            break
+        end
+    end
+    if start == -1
+        # no flat dimensions -> use more efficient implementation
+        return HyperrectangleVertexIteratorNonflat(H)
+    end
+    flats = falses(n)
+    flats[start] = true
+    n_flats = 1
+    # find remaining flat dimensions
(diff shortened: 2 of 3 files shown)


mforets:
> I was surprised that the iterator is generally slower

have you considered using a generator? in Julia v1.6 one can directly map iterators, such that the vertices iterartor hyperrectangles is simply:

```julia
julia> p = Iterators.product(fill([1, -1], 2)...)
Base.Iterators.ProductIterator{Tuple{Vector{Int64},Vector{Int64}}}(([1, -1], [1, -1]))

julia> Iterators.map(x -> rand(2) .* x .+ rand(2), p)
Base.Generator{Base.Iterators.ProductIterator{Tuple{Vector{Int64},Vector{Int64}}},var"#27#28"}(var"#27#28"(), Base.Iterators.ProductIterator{Tuple{Vector{Int64},Vector{Int64}}}(([1, -1], [1, -1])))

julia> first(ans)
2-element Vector{Float64}:
 1.0472048628849284
 1.9230998571087827
```
the idea is to use the former method, see [here](https://github.com/JuliaReach/LazySets.jl/blob/v1.10.0/src/AbstractHyperrectangle.jl#L55). 

we could otherwise try with `Base.Generator` (or IterTools.jl) as suggested in [this question](https://discourse.julialang.org/t/why-no-base-iterators-map/14280)

mforets:
whatever method we choose we should be sure that the vector is not changed, eg if `H` has sarray center and radius, the vertices should also be sarrays.

schillic:
I played around with your suggestion of a generator. I updated the results in the OP (`f` is the new implementation). The generator implementation is much worse. Maybe I misunderstood something there.

PR #2333, opened 2020-08-13
Fix problems with Cartesian Product Array algorithm for overapproximating with Zonotope

SebastianGuadalupe:


mforets:
looks good but i would suggest that we fix the bug in 1 separate PR that we'll quickly merge. 

about using a lazy projection have in mind that this is sometimes useful (eg if the set is hyperrectangular, or in general if the concrete projection is easy to compute). so i have my doubts about the proposed change. maybe we can simply add a kwarg so that the user can decide the approach.

src/Approximations/overapproximate.jl
```diff
@@ -1722,7 +1722,6 @@ continuous dynamics (Doctoral dissertation).
 """
 function _overapproximate_zonotope_cpa(X::LazySet, dir::Type{<:AbstractDirections})
     n = dim(X)
-
     # overapproximate 2D blocks
     if n > 1
         nblocks = Int(floor(n/2))
```
mforets, line 1727: shouldnt we remove nblocks and use n below?

mforets:
we can merge this and address https://github.com/JuliaReach/LazySets.jl/pull/2333#discussion_r470191901 in a follow up PR

src/Approximations/overapproximate.jl
```diff
@@ -1722,11 +1722,9 @@ continuous dynamics (Doctoral dissertation).
 """
 function _overapproximate_zonotope_cpa(X::LazySet, dir::Type{<:AbstractDirections})
     n = dim(X)
-
     # overapproximate 2D blocks
     if n > 1
-        nblocks = Int(floor(n/2))
-        πX_2D = [project(X, [i, i+1]) for i in 1:2:nblocks]
+        πX_2D = [project(X, [i, i+1]) for i in 1:2:n]
```
schillic, line 1727: Will this work for odd `n`? Is there such a test? If not, please add one.

SebastianGuadalupe: Yes, that is addressed here: https://github.com/JuliaReach/LazySets.jl/blob/1bee735828f19d0c167975f7fd9a2758a5811dfe/src/Approximations/overapproximate.jl#L1734

schillic: But that comes only later.

```julia
julia> collect([i, i+1] for i in 1:2:9)
5-element Array{Array{Int64,1},1}:
 [1, 2]
 [3, 4]
 [5, 6]
 [7, 8]
 [9, 10]
```
If you do `project(X, [9, 10])` for a 9-dimensional set, I think this should crash.

SebastianGuadalupe: You are right, and there are tests with odd dimensionality, it errored  :)

schillic: After your commit it works for odd `n`, but does it for even `n`?

SebastianGuadalupe: Yes! it should:
```julia
julia> collect([i, i+1] for i in 1:2:10)
5-element Array{Array{Int64,1},1}:
 [1, 2]
 [3, 4]
 [5, 6]
 [7, 8]
 [9, 10]
```

schillic: But you are only going up to `n-1`, so for `n = 10` you are now doing `1:2:(10-1)`.

SebastianGuadalupe: Yes, I edited the comment (sorry) and it is the same result

schillic: Oh indeed!

Issue #2082, opened 2020-03-24
Type instability in hyperrectangle split


mforets:
```julia
julia> using LazySets

julia> H = rand(Hyperrectangle);

julia> split(H, [2, 2])
4-element Array{Hyperrectangle{Float64,VNC,VNR} where VNR<:AbstractArray{Float64,1} where VNC<:AbstractArray{Float64,1},1}:
 Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([-0.5857521505943779, -0.250035294092884], [0.7155884238793997, 0.2603711149488523])
 Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([0.8454246971644215, -0.250035294092884], [0.7155884238793997, 0.2603711149488523]) 
 Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([-0.5857521505943779, 0.2707069358048206], [0.7155884238793997, 0.2603711149488523])
 Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([0.8454246971644215, 0.2707069358048206], [0.7155884238793997, 0.2603711149488523]) 
```

```julia
julia> @code_warntype split(H, [2, 2])
Variables
  #self#::Core.Compiler.Const(split, false)
  H::Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}
  num_blocks::Array{Int64,1}
  radius::Array{Float64,1}
  total_number::Int64
  lo::Array{Float64,1}
  hi::Array{Float64,1}
  centers::Array{StepRangeLen{Float64,R,S} where S where R,1}
  @_9::Union{Nothing, Tuple{Tuple{Int64,Int64},Tuple{Int64,Int64}}}
  result::Array{Hyperrectangle{Float64,VNC,VNR} where VNR<:AbstractArray{Float64,1} where VNC<:AbstractArray{Float64,1},1}
  @_11::Union{Nothing, Tuple{Tuple{Int64,Any},Tuple{Int64,Any}}}
  i@_12::Int64
  m::Int64
  @_14::Int64
  center::Any
  @_16::Int64
  i@_17::Int64

Body::Array{Hyperrectangle{Float64,VNC,VNR} where VNR<:AbstractArray{Float64,1} where VNC<:AbstractArray{Float64,1},1}
1 ──        Core.NewvarNode(:(radius))
│           Core.NewvarNode(:(total_number))
│           Core.NewvarNode(:(lo))
[...]

PR #2335, opened 2020-08-14
Better diameter and split methods for Interval

schillic:
I found it annoying that `split(rand(Interval), 2)` does not work and `split(rand(Interval), [2])` returns a list of `Hyperrectangle`s.

The first commit brings a faster implementation of `diameter` that I first thought I would use but then I decided to just use `IntervalArithmetic.mince(·)`.

x-ref: #2082

### DIFF
--- a/src/Sets/Interval.jl
+++ b/src/Sets/Interval.jl
@@ -656,7 +656,7 @@ A real number representing the diameter.
 In one dimension all norms are the same.
 """
 function diameter(x::Interval, p::Real=Inf)
-    return norm(max(x) - min(x), p)
+    return norm(IntervalArithmetic.diam(x.dat), p)
 end
 
 """
--- a/src/Sets/Interval.jl
+++ b/src/Sets/Interval.jl
@@ -683,7 +683,7 @@ function split(x::Interval, k::Int)
     return [Interval(x2) for x2 in mince(x.dat, k)]
 end
 
-if VERSION < v"1.1"
+@static if VERSION < v"1.1"
     # IntervalArithmetic.mince() is not available -> define it here
     function mince(x::IntervalArithmetic.Interval, n)
         width = (x.hi - x.lo) / n
--- a/src/Sets/Interval.jl
+++ b/src/Sets/Interval.jl
@@ -655,3 +655,31 @@ In one dimension all norms are equivalent.
 function diameter(x::Interval, p::Real=Inf)
     return norm(max(x) - min(x), p)
 end
+
+"""
+    split(x::Interval, num_blocks)
+
+Partition an interval into uniform sub-intervals.
+
+### Input
+
+- `x`          -- interval
+- `num_blocks` -- number of blocks, possibly wrapped in a vector
+
+### Output
+
+A list of `Interval`s.
+"""
+function split(x::Interval, num_blocks::AbstractVector{Int})
+    @assert length(num_blocks) == 1 "an interval can only be split along one " *
+        "dimension"
+    return split(x, num_blocks[1])
+end
+
+function split(x::Interval, num_blocks::Int)
+    @assert num_blocks > 0 "can only split into a positive number of intervals"
+    if num_blocks == 1
+        return [x]
+    end
+    return [Interval(x2) for x2 in IntervalArithmetic.mince(x.dat, num_blocks)]
+end
--- a/src/Sets/Interval.jl
+++ b/src/Sets/Interval.jl
@@ -656,7 +656,7 @@ A real number representing the diameter.
 In one dimension all norms are the same.
 """
 function diameter(x::Interval, p::Real=Inf)
-    return norm(IntervalArithmetic.diam(x.dat), p)
+    return IntervalArithmetic.diam(x.dat)
 end
 
 """
--- a/src/Sets/Interval.jl
+++ b/src/Sets/Interval.jl
@@ -640,7 +640,7 @@ end
 """
     diameter(x::Interval, [p]::Real=Inf)
 
-Compute the diameter of an interval.
+Compute the diameter of an interval, defined as ``\\Vert b - a\\Vert`` in the ``p`-norm, where ``a`` (resp. ``b``) are the minimum (resp. maximum) of the given interval.
 
 ### Input
 
--- a/src/Sets/Interval.jl
+++ b/src/Sets/Interval.jl
@@ -1,5 +1,8 @@
 import IntervalArithmetic
 using IntervalArithmetic: AbstractInterval
+if VERSION >= v"1.1"
+    using IntervalArithmetic: mince
+end
 import Base: +, -, *, ∈, ⊆, rand, min, max
 
 export Interval,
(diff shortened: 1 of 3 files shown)


mforets:
> I found it annoying that split(rand(Interval), 2) does not work and split(rand(Interval), [2]) returns a list of Hyperrectangles.

the "problem" is that we have one `split` for hyperrectangular sets and anothr `split` for zonotopic sets.

src/Sets/Interval.jl
```diff
@@ -633,3 +633,53 @@ function rectify(x::Interval{N}) where {N<:Real}
...
+
+"""
+    split(x::Interval, num_blocks)
+
+Partition an interval into uniform sub-intervals.
+
+### Input
+
+- `x`          -- interval
+- `num_blocks` -- number of blocks, possibly wrapped in a vector
```
mforets, line 667: perhaps say that here "block" means "sub-interval". not sure that "block" is standard though, i woud just use `k` ..

schillic: I copied the name `num_blocks` from the `AbstractHyperrectangle` method. But I agree that in this case we should just call it `k` because it is not a keyword argument.

schillic:
> > I found it annoying that split(rand(Interval), 2) does not work and split(rand(Interval), [2]) returns a list of Hyperrectangles.
> 
> the "problem" is that we have one `split` for hyperrectangular sets and anothr `split` for zonotopic sets.

You probably refer to

```julia
julia> split(rand(Interval), 2)
ERROR: AssertionError: cannot split a zonotope with 1 generators along index 2
```

I agree that this is also confusing but not what I meant. I meant that it is required that you pass a vector of length 1 for the method for hyperrectangular sets. But yes, this change actually also resolves this possible confusion with the zonotope method by hiding it, so win-win :smile:

schillic:
The build error is in v1.0 where `IntervalArithmetic.mince(·)` is not available (it requires v1.1). I guess I will just copy over [the current implementation from `IntervalArithmetic`](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/a48c26a837fb7e8e00138cef75064061551ae59d/src/intervals/arithmetic.jl#L578-L581) for older Julia versions.

schillic:
> The build error is in v1.0 where `IntervalArithmetic.mince(·)` is not available (it requires v1.1). I guess I will just copy over [the current implementation from `IntervalArithmetic`](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/a48c26a837fb7e8e00138cef75064061551ae59d/src/intervals/arithmetic.jl#L578-L581) for older Julia versions.

Unfortunately this implementation uses a function that was not available in older versions, so I ended up manually hacking something :sweat_smile:

src/Sets/Interval.jl
```diff
@@ -1,5 +1,8 @@
 import IntervalArithmetic
 using IntervalArithmetic: AbstractInterval
+if VERSION >= v"1.1"
```
mforets, line 3:
```suggestion
@static if VERSION >= v"1.1"
```

src/Sets/Interval.jl
```diff
@@ -633,3 +636,65 @@ function rectify(x::Interval{N}) where {N<:Real}
         return Interval(zero(N), max(x.dat.hi, zero(N)))
     end
 end
+
+"""
+    diameter(x::Interval, [p]::Real=Inf)
+
+Compute the diameter of an interval.
```
mforets, line 643:
```suggestion
Compute the diameter of an interval, defined as ``\\Vert b - a\\Vert`` in the ``p`-norm, where ``a`` (resp. ``b``) are the minimum (resp. maximum) of the given interval.
```

src/Sets/Interval.jl
```diff
@@ -633,3 +636,65 @@ function rectify(x::Interval{N}) where {N<:Real}
...
+    @assert length(k) == 1 "an interval can only be split along one dimension"
+    return split(x, k[1])
+end
+
+function split(x::Interval, k::Int)
+    @assert k > 0 "can only split into a positive number of intervals"
+    return [Interval(x2) for x2 in mince(x.dat, k)]
+end
+
+if VERSION < v"1.1"
```
mforets, line 686:
```suggestion
@static if VERSION < v"1.1"
```

src/Sets/Interval.jl
```diff
@@ -633,3 +636,65 @@ function rectify(x::Interval{N}) where {N<:Real}
...
+### Output
+
+A real number representing the diameter.
+
+### Notes
+
+In one dimension all norms are equivalent.
+"""
+function diameter(x::Interval, p::Real=Inf)
+    return norm(max(x) - min(x), p)
```
mforets, line 659: i've noted that the correct (wrt the floating-point standard) implementation used directed rounding, see here https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/a48c26a837fb7e8e00138cef75064061551ae59d/src/intervals/arithmetic.jl#L494

so i'm wondering if we could add two functions, one `diameter(::Interval)` which just calls IntervalArithmeticl.diam, and another generalized function with two arguments .

mforets: like

```julia
diameter(x::Interval) = IntervalArithmetic.diam(x)
diameter(x::Interval, p::Real) = norm(max(x) - min(x), p)
```

schillic: Hm, in that case I would suggest to always use `norm(IntervalArithmetic.diam(x), p)`, unless the interval is empty (in which case I would rather expect the result `0`).

(By the way, I know that the `norm` is redundant for numbers but it costs the same, so I kept it. I thought that maybe there are "invalid" `p`s that you can pass and then it might be preferable to get an error instead of silently ignoring them.)

src/Sets/Interval.jl
```diff
@@ -633,3 +636,65 @@ function rectify(x::Interval{N}) where {N<:Real}
...
+- `x` -- interval
+- `p` -- (optional, default: `Inf`) norm
+
+### Output
+
+A real number representing the diameter.
+
+### Notes
+
+In one dimension all norms are equivalent.
```
mforets, line 656:
```suggestion
In one dimension all norms are the same.
```
( technically all norms are equivalent in R^n )

src/Sets/Interval.jl
mforets: > Hm, in that case I would suggest to always use norm(IntervalArithmetic.diam(x), p), 

ok, makes sense :+1: 

> unless the interval is empty (in which case I would rather expect the result 0).

well since we dont check that intervals are well formed i wouldn't make additional checks  about emptiness in this function...? just solve the construction issue .

```julia
julia> Interval(2, 1)
Interval{Float64,IntervalArithmetic.Interval{Float64}}([2, 1])

julia> diam(Interval(2, 1).dat)
-1.0
```

mforets: but the call the `norm` is needless; we can check that `p` actually satisfies `1 <= p <= Inf` that's all

schillic: Oh, I was referring to [this line](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/a48c26a837fb7e8e00138cef75064061551ae59d/src/intervals/arithmetic.jl#L492) and thought that it would be triggered because I thought that `isempty` would be meant for flat intervals. But it is not :+1:

PR #2336, opened 2020-08-17
Fix area for non-polygons

### LINKED ISSUE #2305: Area gives wrong result for non-polygons
`area(::LazySet)` should convert to `VPolygon` to ensure that the vertices are sorted.

```julia
julia>  area(BallInf(zeros(2), 1.0))

0.0

julia> volume(BallInf(zeros(2), 1.0))
4.0
```

mforets:
Closes #2305.

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -153,7 +153,7 @@ are guaranteed to be sorted in counter-clockwise fashion.
 """
 function convert(::Type{VPolygon}, X::LazySet)
     @assert dim(X) == 2 "set must be two-dimensional for conversion"
-    return VPolygon(vertices_list(concretize(X)))
+    return VPolygon(vertices_list(X))
 end
 
 """
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -133,7 +133,7 @@ end
 """
     convert(::Type{VPolygon}, X::LazySet)
 
-Generic conversion to polygon in vertex representation
+Generic conversion to polygon in vertex representation.
 
 ### Input
 
@@ -147,7 +147,7 @@ The 2D set represented as a polygon.
 ### Algorithm
 
 First we concretize `X` obtaining a (possibly) new representation `Xc`; then we
-wrap the result into a polygon in vertex representation by compuing the list of
+wrap the result into a polygon in vertex representation by computing the list of
 vertices of `Xc`. The vertices of the resulting polygon in vertex representation
 are guaranteed to be sorted in counter-clockwise fashion.
 """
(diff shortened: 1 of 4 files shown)


src/convert.jl
```diff
@@ -130,6 +130,32 @@ function convert(::Type{VPolygon}, P::AbstractPolytope)
...
+### Algorithm
+
+First we concretize `X` obtaining a (possibly) new representation `Xc`; then we
+wrap the result into a polygon in vertex representation by computing the list of
+vertices of `Xc`. The vertices of the resulting polygon in vertex representation
+are guaranteed to be sorted in counter-clockwise fashion.
+"""
+function convert(::Type{VPolygon}, X::LazySet)
+    @assert dim(X) == 2 "set must be two-dimensional for conversion"
+    return VPolygon(vertices_list(concretize(X)))
```
mforets, line 156: @schillic what do you think about using `concretize` here?

mforets: i would prefer to not use it, but note that, for example,

```julia
julia> vertices_list(rand(2, 10) * rand(Zonotope, dim=10))
ERROR: AssertionError: package 'Polyhedra' not loaded (it is required for executing `remove_redundant_vertices`)
Stacktrace:
 [1] macro expansion at /home/mforets/.julia/dev/LazySets/src/Assertions/Assertions.jl:23 [inlined]
....

julia> vertices_list(concretize(rand(2, 10) * rand(Zonotope, dim=10)))
36-element Array{Array{Float64,1},1}:
 [-31.25354193606883, -22.148084939907815]
 [-30.664022201688585, -23.27061463165326]
...
```

schillic: Agreed, I would delegate this decision to `vertices_list` instead. We could let `vertices_list` use `concretize`, possibly only if `Polyhedra` is not loaded.

mforets, line 156:
```suggestion
    return VPolygon(vertices_list(X))
```

src/convert.jl
```diff
@@ -130,6 +130,32 @@ function convert(::Type{VPolygon}, P::AbstractPolytope)
...
+### Output
+
+The 2D set represented as a polygon.
+
+### Algorithm
+
+First we concretize `X` obtaining a (possibly) new representation `Xc`; then we
+wrap the result into a polygon in vertex representation by computing the list of
+vertices of `Xc`. The vertices of the resulting polygon in vertex representation
+are guaranteed to be sorted in counter-clockwise fashion.
```
schillic, line 152: The docstring is outdated.
```suggestion
We compute the list of vertices of `X` and wrap the result in a polygon in
vertex representation, which guarantees that the vertices are sorted in
counter-clockwise fashion.
```

PR #2344, opened 2020-08-24
Fix membership test for line segments

mforets:
Closes https://github.com/JuliaReach/LazySets.jl/issues/1485.

The test was inspired from https://discourse.julialang.org/t/less-than-approx-equals/45436. Note that without the change in this PR, the membership test fails irrespective of the LazySets tolerances.


mforets:
The PR also closes https://github.com/JuliaReach/LazySets.jl/issues/1485 which asks for a test that changes a tolerance.

schillic (approved):

Issue #698, opened 2018-10-03
Concrete intersection between VPolygons


mforets:
https://www.swtestacademy.com/intersection-convex-polygons-algorithm/

schillic:
Currently we fall back to `Polyhedra`. Should we instead convert to H-rep if `Polyhedra` is not loaded?

mforets:
intersection of convex polygons was implemented in https://github.com/JuliaReach/LazySets.jl/pull/2345 so i think we should close this. i kept the reference in the wiki : https://github.com/JuliaReach/LazySets.jl/wiki/References#polygon-intersection

PR #2345, opened 2020-08-24
Add Sutherland–Hodgman VPolygon intersection algorithm

mforets:
This is a standard algorithm, see https://en.wikipedia.org/wiki/Sutherland%E2%80%93Hodgman_algorithm. The code follows that of http://www.rosettacode.org/wiki/Sutherland-Hodgman_polygon_clipping#Julia.

Related to https://github.com/JuliaReach/LazySets.jl/issues/698.

### DIFF
--- a/src/ConcreteOperations/intersection.jl
+++ b/src/ConcreteOperations/intersection.jl
@@ -661,7 +661,7 @@ end
     intersection(P1::VPolytope{N},
                  P2::VPolytope{N};
                  [backend]=default_polyhedra_backend(P1, N),
-                 [prunefunc]=removevredundancy!) where {N<:Real}
+                 [prunefunc]=removevredundancy!) where {N}
 
 Compute the intersection of two polytopes in vertex representation.
 
@@ -678,10 +678,35 @@ Compute the intersection of two polytopes in vertex representation.
 
 A `VPolytope`.
 """
-function intersection(P1::VPolytope{N},
-                      P2::VPolytope{N};
-                      backend=default_polyhedra_backend(P1, N),
-                      prunefunc=removevredundancy!) where {N<:Real}
+function intersection(P1::Union{VPolygon{N}, VPolytope{N}},
+                      P2::Union{VPolygon{N}, VPolytope{N}};
+                      backend=nothing,
+                      prunefunc=nothing) where {N}
+    n = dim(P1)
+    @assert n == dim(P2) "expected polytopes with the equal dimensions but they " *
+                         "are $(dim(P1)) and $(dim(P2)) respectively"
+
+    # fast path for one and two-dimensional sets
+    if n == 1
+        Q1 = overapproximate(P1, Interval)
+        Q2 = overapproximate(P2, Interval)
+        Pint = intersection(Q1, Q2)
+        return convert(VPolytope, Pint)
+    elseif n == 2
+        Q1 = convert(VPolygon, P1)
+        Q2 = convert(VPolygon, P2)
+        Pint = intersection(Q1, Q2)
+        return convert(VPolytope, Pint)
+    end
+
+    if isnothing(backend)
+        backend = default_polyhedra_backend(P1, N)
+    end
+    if isnothing(prunefunc)
+        prunefunc = removevredundancy!
+    end
+
+    # general case: convert to half-space representation
     Q1 = polyhedron(P1; backend=backend)
     Q2 = polyhedron(P2; backend=backend)
     Pint = Polyhedra.intersect(Q1, Q2)
--- a/src/ConcreteOperations/intersection.jl
+++ b/src/ConcreteOperations/intersection.jl
@@ -695,8 +695,8 @@ Compute the intersection of two polygons in vertex representation.
 
 ### Input
 
-- `P1`        -- polytope in vertex representation
-- `P2`        -- polytope in vertex representation
+- `P1` -- polygon in vertex representation
+- `P2` -- polygon in vertex representation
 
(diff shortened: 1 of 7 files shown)


src/ConcreteOperations/intersection.jl
```diff
@@ -676,21 +676,44 @@ Compute the intersection of two polytopes in vertex representation.
 
 ### Output
 
-A `VPolygon` if both arguments are `VPolygon`s, and a `VPolytope` otherwise.
+A `VPolytope`.
 """
-function intersection(P1::Union{VPolytope{N}, VPolygon{N}},
-                      P2::Union{VPolytope{N}, VPolygon{N}};
+function intersection(P1::VPolytope{N},
+                      P2::VPolytope{N};
```
schillic, line 682: Was it necessary to downgrade this method? Because I think we now lost the intersection for mixed types. I would expect that Julia would use your other method automatically because it is more specific.

src/ConcreteOperations/intersection.jl
```diff
@@ -676,21 +676,44 @@ Compute the intersection of two polytopes in vertex representation.
...
                       backend=default_polyhedra_backend(P1, N),
                       prunefunc=removevredundancy!) where {N<:Real}
-    Q1 = polyhedron(convert(VPolytope, P1); backend=backend)
-    Q2 = polyhedron(convert(VPolytope, P2); backend=backend)
+    Q1 = polyhedron(P1; backend=backend)
+    Q2 = polyhedron(P2; backend=backend)
     Pint = Polyhedra.intersect(Q1, Q2)
-    prunefunc(Pint)
```
schillic, line 688: Now the `prunefunc` parameter is not used anymore. Why?

src/ConcreteOperations/intersection.jl
```diff
@@ -676,21 +676,44 @@ Compute the intersection of two polytopes in vertex representation.
...
+
+"""
+    intersection(P1::VPolygon, P2::VPolygon)
+
+Compute the intersection of two polygons in vertex representation.
+
+### Input
+
+- `P1`        -- polytope in vertex representation
+- `P2`        -- polytope in vertex representation
```
schillic, line 699:
```suggestion
- `P1` -- polygon in vertex representation
- `P2` -- polygon in vertex representation
```

src/ConcreteOperations/intersection.jl
mforets: i see, it wasn't intentional.

src/ConcreteOperations/intersection.jl
mforets: i think that for mixed types they should use `_intersection_vrep`, i can try changing that if u agree. 

how about VPolytope's in 2D? method `_intersection_vrep` is much faster (from 1-2us to 1-2ms) so i guess we it's also OK to use the method for polygons in that case? and return a VPolytope.

schillic: > i can try changing that if u agree

:+1:

> how about VPolytope's in 2D?

as you prefer

mforets:
if the intersection is empty the new function should rather return `EmptySet` -- for more advanced users, type stability is ensured using `_intersection_vrep` directly.

src/ConcreteOperations/intersection.jl
```diff
@@ -710,11 +713,16 @@ clipping algorithm](https://en.wikipedia.org/wiki/Sutherland%E2%80%93Hodgman_alg
...
+function intersection(P1::VPolygon{N}, P2::VPolygon{N};
+                      apply_convex_hull::Bool=true) where {N}
     v1 = vertices_list(P1)
     v2 = vertices_list(P2)
     v12 = _intersection_vrep(v1, v2)
-    return VPolygon(v12)
+    if isempty(v12)
+        return EmptySet{N}(2)
+    else
+        return VPolygon(v12, apply_convex_hull=apply_convex_hull)
```
schillic, line 724: IMO `VPolygon`s should always be created using a convex hull unless the algorithm guarantees that result anyway.

mforets: yes , i agree. i left the option as `true` by default.

src/ConcreteOperations/intersection.jl
```diff
@@ -678,10 +678,35 @@ Compute the intersection of two polytopes in vertex representation.
...
-function intersection(P1::VPolytope{N},
-                      P2::VPolytope{N};
-                      backend=default_polyhedra_backend(P1, N),
-                      prunefunc=removevredundancy!) where {N<:Real}
+function intersection(P1::Union{VPolygon{N}, VPolytope{N}},
+                      P2::Union{VPolygon{N}, VPolytope{N}};
+                      backend=nothing,
+                      prunefunc=nothing) where {N}
+    n = dim(P1)
+    @assert n == dim(P2) "expected polytopes with the equal dimensions but they " *
```
schillic, line 686:
```suggestion
    @assert n == dim(P2) "expected polytopes with equal dimensions but they " *
```

schillic (approved):
I still think the option `apply_convex_hull` is misleading because you may want to set it to `false` to save time, but that will violate our type assumption of `VPolygon`.

mforets:
> I still think the option apply_convex_hull is misleading because you may want to set it to false to save time, but that will violate our type assumption of VPolygon.

yes, but if an application just requires the vertices of the intersection (without any particular sorting assumption) it's reasonable to pass `apply_convex_hull=false`, instead of digging & using internal methods to get the vertices directly.

PR #2276, opened 2020-07-25
#2219 Remove N <: Real for zonotopic sets

mforets:
see #2219

PR #2357, opened 2020-09-11
#2356 - InverseLinearMap

### LINKED ISSUE #2356: Inverse linear map set operation

dfcaporale:
Closes #2356


src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+
+The support function in the given direction.
+If the direction has norm zero, the result depends on the wrapped set.
+
+### Notes
+
+If ``L = M^{-1}⋅S``, where ``M`` is a matrix and ``S`` is a convex set, it follows
+that ``ρ(d, L) = ρ((M^T)^{-1} d, S)`` for any direction ``d``.
+"""
+function ρ(d::AbstractVector, lm::InverseLinearMap)
```
mforets, line 189: `lm` -> `ilm`

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+- `ilm` -- inverse linear map
+
+### Output
+
+The support vector in the given direction.
+If the direction has norm zero, the result depends on the wrapped set.
+
+### Notes
+
+If ``L = M^{-1}⋅S``, where ``M`` is a matrix and ``S`` is a convex set, since
```
mforets, line 159: can you use ``L = M^{-1}⋅X``? so the notation in the docstring and the code are the same.

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+(M^T)^{-1}=(M^{-1})^T, it follows that ``σ(d, L) = M^{-1}⋅σ((M^T)^{-1} d, S)``
+for any direction ``d``.
+"""
+function σ(d::AbstractArray{N,1}, ilm::InverseLinearMap{N,S,NM,MAT}) where {N<:Real, S<:LazySet{N}, NM, MAT}
+    return ilm.M \ σ(transpose(ilm.M) \ d, ilm.X) #TODO: Check this formula
```
mforets, line 164:
```suggestion
    return ilm.M \ σ(transpose(ilm.M) \ d, ilm.X)
```
the formula lgtm

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+The support vector in the given direction.
+If the direction has norm zero, the result depends on the wrapped set.
+
+### Notes
+
+If ``L = M^{-1}⋅S``, where ``M`` is a matrix and ``S`` is a convex set, since
+(M^T)^{-1}=(M^{-1})^T, it follows that ``σ(d, L) = M^{-1}⋅σ((M^T)^{-1} d, S)``
+for any direction ``d``.
+"""
+function σ(d::AbstractArray{N,1}, ilm::InverseLinearMap{N,S,NM,MAT}) where {N<:Real, S<:LazySet{N}, NM, MAT}
```
mforets, line 163:
```suggestion
function σ(d::AbstractVector, ilm::InverseLinearMap{N,S,NM,MAT}) where {N<:Real, S<:LazySet{N}, NM, MAT}
```
(it is the same)

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+### Output
+
+The ambient dimension of the inverse linear map.
+"""
+function dim(ilm::InverseLinearMap)
+    return size(ilm.M, 1)
+end
+
+"""
+    σ(d::AbstractVector{N}, ilm::InverseLinearMap{N}) where {N<:Real}
```
mforets, line 143:
```suggestion
    σ(d::AbstractVector{N}, ilm::InverseLinearMap{N}) where {N}
```
it is not necessary to annotate `N<:Real`. 

we can also remove `N` altogether, here or in anther PR.

please note that the comment about removing `N<:Real` applies to all other occurrences in this file.

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+If ``L = M^{-1}⋅S``, where ``M`` is a matrix and ``S`` is a convex set, since
+(M^T)^{-1}=(M^{-1})^T, it follows that ``σ(d, L) = M^{-1}⋅σ((M^T)^{-1} d, S)``
+for any direction ``d``.
+"""
+function σ(d::AbstractArray{N,1}, ilm::InverseLinearMap{N,S,NM,MAT}) where {N<:Real, S<:LazySet{N}, NM, MAT}
+    return ilm.M \ σ(transpose(ilm.M) \ d, ilm.X) #TODO: Check this formula
+end
+
+"""
+    ρ(d::AbstractVector{N}, ilm::InverseLinearMap{N}; kwargs...) where {N<:Real}
```
mforets, line 168:
```suggestion
    ρ(d::AbstractVector, ilm::InverseLinearMap)
```

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+
+Return the support function of the inverse linear map.
```
mforets, line 170:
```suggestion
Return the support function of the inverse linear map of a set.
```

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+
+### Input
+
+- `d`      -- direction
+- `ilm`    -- inverse linear map
+- `kwargs` -- additional arguments that are passed to the support function
```
mforets, line 176: i dont see kwargs passed in the body of the function

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+"""
+function ρ(d::AbstractVector, lm::InverseLinearMap)
+    return ρ(transpose(lm.M) \ d, lm.X)
+end
+
+
+"""
+    ∈(x::AbstractVector{N}, ilm::InverseLinearMap{N}) where {N<:Real}
+
+Check whether a given point is contained in an inverse linear map of a convex set.
```
mforets, line 197:
```suggestion
Check whether a given point is contained in the inverse linear map of a convex set.
```
(it is unique)

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+
+### Input
+
+- `x`   -- point/vector
+- `ilm` -- inverse linear map of a convex set
```
mforets, line 202:
```suggestion
- `ilm` -- inverse linear map of a set
```
i dont think that the set has to be convex

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+```jldoctest
+julia> ilm = LinearMap([0.5 0.0; 0.0 -0.5], BallInf([0., 0.], 1.));
+
+julia> [1.0, 1.0] ∈ ilm
+false
+julia> [0.1, 0.1] ∈ ilm
+true
+```
+"""
+function ∈(x::AbstractVector, lm::InverseLinearMap)
```
mforets, line 224:
```suggestion
function ∈(x::AbstractVector, ilm::InverseLinearMap)
```
and also in the body of the function

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+    return lm.M * x ∈ lm.X
+end
+
+"""
+    an_element(ilm::InverseLinearMap{N})::Vector{N} where {N<:Real}
```
mforets, line 229:
```suggestion
    an_element(ilm::InverseLinearMap)
```

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+
+An element in the inverse linear map.
+It relies on the `an_element` function of the wrapped set.
+"""
+function an_element(lm::InverseLinearMap)
+    return lm.M \ an_element(lm.X)
+end
+
+"""
+    vertices_list(ilm::InverseLinearMap{N}; prune::Bool=true)::Vector{Vector{N}} where {N<:Real}
```
mforets, line 247:
```suggestion
    vertices_list(ilm::InverseLinearMap{N}; prune::Bool=true)  where {N}
```

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+
+Return the list of vertices of a (polyhedral) inverse linear map.
+
+### Input
+
+- `ilm`   -- inverse linear map
+- `prune` -- (optional, default: `true`) if true removes redundant vertices
```
mforets, line 254:
```suggestion
- `prune` -- (optional, default: `true`) if `true` removes redundant vertices
```

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+    # for a zero map, the result is just the list containing the origin
+    if iszero(ilm.M)
+        return [zeros(N, dim(ilm))]
+    end
+
+    # collect low-dimensional vertices lists
+    vlist_X = vertices_list(ilm.X)
+
+    # create resulting vertices list
+    vlist = Vector{Vector{N}}()
```
mforets, line 275: `Vector{Vector{N}}()` works but is a bad idea for library code because the vertices of `X` don't need to be stored as regular arrays . better if we use the type of the vertices of `X` i guess that's `eltype(vlist_X)`.

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+A list of vertices.
+
+### Algorithm
+
+We assume that the underlying set `X` is polyhedral.
+Then the result is just the inverse linear map applied to the vertices of `X`.
+"""
+function vertices_list(ilm::InverseLinearMap{N}; prune::Bool=true) where {N}
+    # for a zero map, the result is just the list containing the origin
+    if iszero(ilm.M)
```
mforets, line 267: i think that we can remove this branch, because it is a rare case (zero matrix) and the simplfication is covered by the convex hull below anyways.

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+    sizehint!(vlist, length(vlist_X))
+    for v in vlist_X
+        push!(vlist, ilm.M \ v)
+    end
+
+    return prune ? convex_hull(vlist) : vlist
+end
+
+"""
+    constraints_list(ilm::InverseLinearMap{N}) where {N<:Real}
```
mforets, line 285:
```suggestion
    constraints_list(ilm::InverseLinearMap)
```
please update the function's signature too

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+
+We fall back to a concrete set representation and apply `linear_map_inverse`.
+"""
+function constraints_list(ilm::InverseLinearMap{N}) where {N}
+    return constraints_list(linear_map_inverse(ilm.M, ilm.X))
+end
+
+
+"""
+    linear_map(M::AbstractMatrix{N}, ilm::InverseLinearMap{N}) where {N<:Real}
```
mforets, line 312:
```suggestion
    linear_map(M::AbstractMatrix, ilm::InverseLinearMap)
```
please also update the function's signature

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+
+- `M`   -- matrix
+- `ilm` -- inverse linear map
+
+### Output
+
+The polytope representing the linear map of the lazy inverse linear map of a set.
+"""
+function linear_map(M::AbstractMatrix{N}, ilm::InverseLinearMap{N}) where {N<:Real}
+    #TODO: Check if linear_map needs matrix of Int64
```
mforets, line 326:
```suggestion
```
i think that this computation shall work independently of the numeric type.

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+    return linear_map(M * inv(ilm.M), ilm.X)
```
mforets, line 327: shall we add a note in the docstring that this function is inefficient? as it requires computing the concrete inverse, which is what the new type is supposed to avoid.

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+end
+
+function concretize(ilm::InverseLinearMap)
+    #TODO: Check if linear_map needs matrix of Int64
+    return linear_map(inv(ilm.M), concretize(ilm.X))
```
mforets, line 332: same comments as in the concrete linear map apply here :)

src/LazyOperations/InverseLinearMap.jl
dfcaporale: If I change the signature from
```
function σ(d::AbstractVector{N}, ilm::InverseLinearMap{N,S,NM,MAT}) where {N<:Real, S<:LazySet{N}, NM, MAT}
```
to
```
function σ(d::AbstractVector{N}, ilm::InverseLinearMap{N,S,NM,MAT}) where {N, S<:LazySet{N}, NM, MAT}
```
I got an error of ambiguity with another methods:

```
MethodError: σ(::Array{Float64,1}, ::InverseLinearMap{Float64,BallInf{Float64,Array{Float64,1}},Float64,Array{Float64,2}}) is ambiguous. Candidates:
  σ(d::AbstractArray{N,1}, am::AbstractAffineMap{N,S} where S<:LazySet{N}) where N<:Real in LazySets at C:\Users\danie\.julia\dev\LazySets\src\Interfaces\AbstractAffineMap.jl:71
  σ(d::AbstractArray{N,1}, ilm::InverseLinearMap{N,S,NM,MAT}) where {N, S<:LazySet{N}, NM, MAT} in LazySets at C:\Users\danie\.julia\dev\LazySets\src\LazyOperations\InverseLinearMap.jl:163
Possible fix, define
  σ(::AbstractArray{N,1}, ::InverseLinearMap{N,S,NM,MAT}) where {N<:Real, S<:LazySet{N}, NM, MAT}

Stacktrace:
 [1] top-level scope at show.jl:641
 [2] include_string(::Function, ::Module, ::String, ::String) at .\loading.jl:1091
```

Suggestion?

src/LazyOperations/InverseLinearMap.jl
dfcaporale: Removed

src/LazyOperations/InverseLinearMap.jl
dfcaporale: Good point... I remove convex just in case... 👍

src/LazyOperations/InverseLinearMap.jl
dfcaporale: OK. I will test this

src/LazyOperations/InverseLinearMap.jl
dfcaporale: OK

src/LazyOperations/InverseLinearMap.jl
mforets: ok, i see. then i suggest to keep it like it was, and revisit after we do https://github.com/JuliaReach/LazySets.jl/pull/2276

src/LazyOperations/InverseLinearMap.jl
dfcaporale: You are right. The point is that the set and the map have to be of the same number type.

I mean, this works:
```
> M = [1 2 3; 2 3 1; 3 1 2];
> X = BallInf([1, 1, 1], 2);
> lm = linear_map(M, X)
```
but it doesn't:
```
> M = [1. 2 3; 2 3 1; 3 1 2]; # promoted to Float64
> X = BallInf([1, 1, 1], 2);
> lm = linear_map(M, X)
```
In the last case, type number in the set should me changed:
```
> M = [1. 2 3; 2 3 1; 3 1 2]; # promoted to Float64
> X = BallInf([1., 1, 1], 2.); # promoted to Float64
> linear_map(M, X)
```

mforets: yes, but we can address this on a follow up PR. in the meantime, i would just suggest that you remove `N<:Real` here.

src/LazyOperations/InverseLinearMap.jl
mforets: @dfcaporale  here and in other places, we usually mark conversations as resolved only *after* they have been fixed. otherwise, it is easy to forget what still needs attention :)

mforets: if you want to comment that you agree (but will do later), the thumbsup :+1:  should be fine!

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,333 @@
...
+
+The application of a `InverseLinearMap` to a `ZeroSet` or an `EmptySet` is
+simplified automatically.
+
+```jldoctest constructors
+julia> InverseLinearMap(A, ZeroSet{Int}(3))
+ZeroSet{Int64}(3)
+```
+"""
+struct InverseLinearMap{N<:Real, S<:LazySet{N},
```
mforets, line 59: can you try removing `N<:Real` here ? i don't see a compelling reason to do so since it is a new type.

src/LazyOperations/InverseLinearMap.jl
dfcaporale: ```sizehint!``` needs a collection as the first entry. Then ```vlist = eltype(vlist_X)``` doesn't work.
I fixed ```vlist``` definition to:
```vlist = Vector{eltype(vlist_X)}();```
Then, I think we are at the same point than before...

Related to this function: does it need ```Polyhedra```?

File: src/LazyOperations/InverseLinearMap.jl (hunk 22 above)
Comment (dfcaporale) [reply src/LazyOperations/InverseLinearMap.jl:59]: Yes. I'm fine with the change.

File: src/LazyOperations/InverseLinearMap.jl (hunk 15 above)
Comment (mforets) [reply src/LazyOperations/InverseLinearMap.jl:275]: > vlist = Vector{eltype(vlist_X)}();

that's precisely what i meant above by " use the type of the vertices of X": `vlist` is a vector of vectors. it is ok to use the outer `Vector`  for storage.

File: src/LazyOperations/InverseLinearMap.jl (hunk 2 above)
Comment (dfcaporale) [reply src/LazyOperations/InverseLinearMap.jl:159]: You are right. What happened is that I modified the file localy with your comments and didn't want to loose them. I'll be more careful in the following.

**Comment by mforets** [comment]
looks good so far. it remains to add the new functions to the documentation.

File: src/LazyOperations/InverseLinearMap.jl [hunk 23]
```diff
@@ -0,0 +1,331 @@
...
+```jldoctest constructors
+julia> A = [1 2 3; 2 3 1; 3 1 2];
+
+julia> X = BallInf([0, 0, 0], 1);
+
+julia> ilm = InverseLinearMap(A, X)
+InverseLinearMap{Int64,BallInf{Int64,Array{Int64,1}},Int64,Array{Int64,2}}([1 2 3; 2 3 1; 3 1 2], BallInf{Int64,Array{Int64,1}}([0, 0, 0], 1))
+```
+
+Applying a linear map to a `InverseLinearMap` object combines the two maps into
```
mforets, line 39:
```suggestion
Applying an inverse linear map to a `InverseLinearMap` object combines the two maps into
```

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,331 @@
...
+- `ilm` -- inverse linear map
+
+### Output
+
+The polytope representing the linear map of the lazy inverse linear map of a set.
+
+### Notes
+
+This function is inefficient in the sense that it requires computing the
+concrete inverse of M, which is what InverseLinearMap is supposed to avoid.
```
mforets, line 323:
```suggestion
concrete inverse of M, which is what `InverseLinearMap` is supposed to avoid.
```

src/LazyOperations/InverseLinearMap.jl
schillic: So does the set have to be convex? If so, this should be clarified in the struct docs.

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,325 @@
...
+- `ilm` -- inverse linear map
+
+### Output
+
+The support vector in the given direction.
+If the direction has norm zero, the result depends on the wrapped set.
+
+### Notes
+
+If ``L = M^{-1}⋅X``, where ``M`` is a matrix and ``X`` is a convex set, since
```
schillic, line 153: Does `X` need to be convex?

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,325 @@
...
+
+A list of vertices.
+
+### Algorithm
+
+We assume that the underlying set `X` is polyhedral.
+Then the result is just the inverse linear map applied to the vertices of `X`.
+"""
+function vertices_list(ilm::InverseLinearMap; prune::Bool=true)
+    # collect low-dimensional vertices lists
```
schillic, line 261:
```suggestion
    # collect vertices list of wrapped set
```

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,325 @@
...
+Return the linear map of a lazy inverse linear map.
+
+### Input
+
+- `M`   -- matrix
+- `ilm` -- inverse linear map
+
+### Output
+
+The polytope representing the linear map of the lazy inverse linear map of a set.
```
schillic, line 312:
```suggestion
The set representing the linear map of the lazy inverse linear map of a set.
```

schillic (commented):
I am not 100% sure about the scope of this type. Originally there was the idea of having non-invertible matrices and in some places this is still present. But I think the majority of places assumes invertibility. So I went for that interpretation.

mforets:
thanks for the review! 

> Originally there was the idea of having non-invertible matrices and in some places this is still present. 
> But I think the majority of places assumes invertibility. 

yes. i'm assuming invertibility. it's interesting for cases in which we are given A, but want to compute the support function of A^{-1}X only for a number of directions much smaller than the dimension of A.

mforets:
the typical example is in the Phi_1 and Phi_2 functions in conservative discretization (def [here](https://juliareach.github.io/ReachabilityAnalysis.jl/dev/lib/discretize/#Exponentiation))

src/LazyOperations/InverseLinearMap.jl
mforets: it doesn't . i'll change this

src/LazyOperations/InverseLinearMap.jl
```diff
@@ -0,0 +1,325 @@
...
+# EmptySet is absorbing for LinearMap
+function InverseLinearMap(M::AbstractMatrix, ∅::EmptySet)
+    return ∅
+end
+
+# --- AbstractAffineMap interface functions ---
+
+function matrix(ilm::InverseLinearMap)
+    return ilm.M
+end
```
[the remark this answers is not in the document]
mforets: makes sense :+1:

mforets:
i think that all review comments have been addressed

schillic (approved):
:100:

PR #2359, opened 2020-09-17
Specialized sf for zonotopes

mforets:
we shall refactor the `_abs_sum` function, which specializes for some types of input directions, then use it in the support function of zonotopic sets, which currently uses a slower fallback.


src/ConcreteOperations/isdisjoint.jl
```diff
@@ -479,11 +480,13 @@ end
     return abs_sum
 end
 
-@inline function _isdisjoint_zonotope_hyperplane(G::AbstractMatrix{N}, a::SparseVector{N}) where {N}
+# computes ‖a^T G‖₁ for a being a sprase vector
```
schillic, line 483:
```suggestion
# computes ‖a^T G‖₁ for `a` being a sparse vector
```

src/ConcreteOperations/isdisjoint.jl
```diff
@@ -479,11 +480,13 @@ end
...
+@inline function _abs_sum(a::SparseVector{N}, G::AbstractMatrix{N}) where {N}
```
schillic, line 484:
```suggestion
@inline function _abs_sum(a::AbstractSparseVector{N}, G::AbstractMatrix{N}) where {N}
```

src/ConcreteOperations/isdisjoint.jl
```diff
@@ -479,11 +480,13 @@ end
...
     return sum(abs, transpose(a) * G)
 end
 
-@inline function _isdisjoint_zonotope_hyperplane(G::AbstractMatrix{N}, a::SingleEntryVector{N}) where {N}
+# computes ‖a^T G‖₁ for a having only one nonzero element
```
schillic, line 488:
```suggestion
# computes ‖a^T G‖₁ for `a` having only one nonzero element
```

schillic (approved):

mforets:
This PR is ready for review. The key part is that now we refactor a computation used in a couple of places, namely ||d^T G||_1, therefore, we can use dispatch as will on the new internal function `_abs_sum(d, G)`. Regarding benchmarks, i've uploaded this notebook: https://nbviewer.jupyter.org/github/mforets/escritoire/blob/master/2020/Week44/Specialized_support_function_Zonotope.ipynb

Please note that the new fallback for `_abs_sum(a::AbstractVector{N}, G::AbstractMatrix{N})` has 0 allocations; it  implements the computation by "rolling our own loop". On the other hand, as it is shown in the last part of the notebook, using `sum(abs, transpose(d) * G)` is better if the dimensions are large (100 or more). If may want to make a branch, but i prefer to defer that to another, follow up issue. Also, instead of testing rho in isolation we can see how it pefrorms with other surrounding code, eg, in a reachability computation.

PR #2361, opened 2020-09-21
#2360 - Add new line constructor in 2d

### LINKED ISSUE #2360: Add constructor for Line and Line2D with vector and float
In the Line2D PR the constructor of Line and Line2D of the form `ax = b` was removed.  I think we should re-add this particular constructor, checking that the length of `a` is two.

Note this causes issues, MathematicalSystems.jl#master tests don't pass: https://travis-ci.org/github/JuliaReach/MathematicalSystems.jl

mforets:
Closes #2360.

### DIFF
--- a/src/Sets/Line.jl
+++ b/src/Sets/Line.jl
@@ -82,6 +82,49 @@ function Line(; from::AbstractVector, to::AbstractVector, normalize=true)
 end
 
 """
+    Line(a::AbstractVector{N}, b::N; normalize=true) where {N}
+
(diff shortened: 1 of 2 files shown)


schillic (approved):
Another way is to create a `Line2D` and then `convert` to a `Line`. However, that is currently missing:
```julia
julia> convert(Line, rand(Line2D))
ERROR: MethodError: Cannot `convert` an object of type 
  Line2D{Float64,Array{Float64,1}} to an object of type 
  Line
```
It would be great to have that work as well.

PR #2375, opened 2020-10-06
Special case for order 1 zonotopes

mforets:
This PR makes two changes:

1) speedup computation of vertices of a 2d zonotope of order 1, it goes from ~2us (master) to ~500ns (this branch). (*)

2) fix the `_vertices_list_iterative` such that the type of the vertices corresponds to the type of the center of the zonotope. before, it was hardcoding `Vector{N}`, which is a bad idea that prevents from obtaining an SArray output if the input is SArray.

(*) Using static arrays, runs in ~60ns.

cc @yupbank

### DIFF
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -438,7 +438,11 @@ function vertices_list(Z::AbstractZonotope{N};
         return vertices_list(convert(Interval, Z))
 
     elseif n == 2
-        return _vertices_list_2D(c, G, apply_convex_hull=apply_convex_hull)
+        if p == 1
+            return _vertices_list_iterative_ord1(c, G, apply_convex_hull=apply_convex_hull)
+        else
+            return _vertices_list_2D(c, G, apply_convex_hull=apply_convex_hull)
+        end
 
     else
         return _vertices_list_iterative(c, G, apply_convex_hull=apply_convex_hull)
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -426,7 +426,7 @@ this method; otherwise, redundant vertices may be present.
 function vertices_list(Z::AbstractZonotope{N};
                        apply_convex_hull::Bool=true) where {N<:Real}
     c = center(Z)
-    G = remove_zero_columns(genmat(Z))
+    G = genmat(Z)
     n, p = size(G)
 
     # empty generators => sole vertex is the center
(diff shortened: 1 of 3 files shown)


src/Interfaces/AbstractZonotope.jl
```diff
@@ -426,7 +426,7 @@ this method; otherwise, redundant vertices may be present.
 function vertices_list(Z::AbstractZonotope{N};
                        apply_convex_hull::Bool=true) where {N<:Real}
     c = center(Z)
-    G = remove_zero_columns(genmat(Z))
+    G = genmat(Z)
```
schillic, line 429: This changes the result for `apply_convex_hull=false` in 2D. What is the motivation for this change? I do not see why 2D should behave differently.

```julia
julia> Z1 = Zonotope([0.0, 0.0], zeros(2, 1));
julia> Z2 = Zonotope([0.0, 0.0], zeros(2, 2));
julia> Z3 = Zonotope([0.0, 0.0], zeros(2, 4));

# this branch
julia> vertices_list(Z1; apply_convex_hull=false)
 [0.0, 0.0]
 [0.0, 0.0]

julia> vertices_list(Z2; apply_convex_hull=false)
 [0.0, 0.0]
 [0.0, 0.0]
 [0.0, 0.0]
 [0.0, 0.0]

julia> vertices_list(Z3; apply_convex_hull=false)
8-element Array{Array{Float64,1},1}:
 [0.0, 0.0]
 [0.0, 0.0]
 [0.0, 0.0]
 [0.0, 0.0]
 [0.0, 0.0]
 [0.0, 0.0]
 [0.0, 0.0]
 [0.0, 0.0]

# master
julia> vertices_list(Z1; apply_convex_hull=false)
1-element Array{Array{Float64,1},1}:
 [0.0, 0.0]
julia> vertices_list(Z2; apply_convex_hull=false)
1-element Array{Array{Float64,1},1}:
 [0.0, 0.0]
julia> vertices_list(Z3; apply_convex_hull=false)
1-element Array{Array{Float64,1},1}:
 [0.0, 0.0]
```

mforets: the motivation is for performance reasons, because remove_zero_columns doesn't need to be applied in general; in 2d, the convex hull will reduce such vertices efficiently.

schillic (approved):
I did not look at the build error...

mforets:
the build error was a doctest that was failing, and i;ve sent the correction now (tests pass locally).  

in fact, the returned array is "better" in the sense that the points are sorted in CCW order, while in master it is not the case. (didn't actually check why, since i would have also expected that they are sorted).

PR #2377, opened 2020-10-06
Read polygons from file

mforets:
This is code we had from https://github.com/JuliaReach/ReachabilityBenchmarks/blob/master/src/plot_GEN.jl , but it is useful quite frequently, eg. to read files generated by other tools.

### DIFF
--- a/src/Utils/helper_functions.jl
+++ b/src/Utils/helper_functions.jl
@@ -468,7 +468,7 @@ A list of polygons in vertex representation.
 ### Notes
 
 The `x` and `y` coordinates of each vertex should be separated by an empty space,
-and polygons are separated by empty lines. For example:
+and polygons are separated by empty lines (even the last polygon). For example:
 
 ```julia
 1.01 1.01
--- a/src/Utils/helper_functions.jl
+++ b/src/Utils/helper_functions.jl
@@ -450,3 +450,68 @@ end
 function _isupwards(vec)
     return vec[2] > 0 || (vec[2] == 0 && vec[1] > 0)
 end
+
+
+"""
+    read_gen(filename)
+
+Read a sequence of polygons stored in vertex representation (gen format).
+
+### Input
+
+- `filename` -- path of the file containing the polygons
+
+### Output
+
+A list of polygons in vertex representation.
+
+### Notes
+
+The `x` and `y` coordinates of each vertex should be separated by an empty space,
+and polygons are separated by empty lines. For example:
+
+```julia
+1.01 1.01
+0.99 1.01
+0.99 0.99
+1.01 0.99
+
+0.908463 1.31047
+0.873089 1.31047
+0.873089 1.28452
+0.908463 1.28452
+
+
+```
+This is parsed as
+
+```julia
+2-element Array{VPolygon{Float64,Array{Float64,1}},1}:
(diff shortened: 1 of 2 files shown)


src/Utils/helper_functions.jl
```diff
@@ -450,3 +450,68 @@ end
...
+0.99 0.99
+1.01 0.99
+
+0.908463 1.31047
+0.873089 1.31047
+0.873089 1.28452
+0.908463 1.28452
+
+
+```
```
schillic, line 485: Can we remove one of the empty lines?

src/Utils/helper_functions.jl
```diff
@@ -450,3 +450,68 @@ end
...
+This is parsed as
+
+```julia
+2-element Array{VPolygon{Float64,Array{Float64,1}},1}:
+ VPolygon{Float64,Array{Float64,1}}([[1.01, 1.01], [0.99, 1.01], [0.99, 0.99], [1.01, 0.99]])
+ VPolygon{Float64,Array{Float64,1}}([[0.908463, 1.31047], [0.873089, 1.31047], [0.873089, 1.28452], [0.908463, 1.28452]])
+```
+
+The input file should end with at least one empty line before the end of files
+(this is to detect the last polygon).
```
schillic, line 495:
```suggestion
```

src/Utils/helper_functions.jl
```diff
@@ -450,3 +450,68 @@ end
...
+- `filename` -- path of the file containing the polygons
+
+### Output
+
+A list of polygons in vertex representation.
+
+### Notes
+
+The `x` and `y` coordinates of each vertex should be separated by an empty space,
+and polygons are separated by empty lines. For example:
```
schillic, line 471:
```suggestion
and polygons are separated by empty lines (even the last polygon). For example:
```

src/Utils/helper_functions.jl
```diff
@@ -450,3 +450,68 @@ end
...
+        for line in eachline(f)
+          if !isempty(line)
+              push!(Mi, map(x -> parse(Float64, x), split(line)))
+              new_polygon = true
+          elseif isempty(line) && new_polygon
+              push!(P, VPolygon(Mi))
+              Mi = Vector{Vector{Float64}}()
+              new_polygon = false
+          end
+        end
```
schillic, line 514: The indentation needs to be fixed here.

schillic (approved):
I do not think that this format is restricted to (2D) polygons. Should we generally return `VPolytope`s?

src/Utils/helper_functions.jl
mforets: hmm im 99% sure it needs both empty lines to work properly.

(i didn't want to spend a lot of time adapting this function, but feel free to improve it).

src/Utils/helper_functions.jl
mforets: thanks ,i'll fix it

mforets: thanks ,i'll fix it

mforets:
> I do not think that this format is restricted to (2D) polygons. Should we generally return VPolytopes?

good idea. i'll create a follow-up issue.

mforets:
(error is unrelated)

Issue #2248, opened 2020-07-20
Fix conversion of SA hyperrectangular sets to zonotopes


mforets:
see [this comment](https://github.com/JuliaReach/LazySets.jl/pull/2214#issuecomment-659858766)

---

There is an issue with mixed array types. I guess that `_split_ret` for the two other combinations (where only one component is an `MArray`) would be needed. You can either fix it or open an issue.
We should also open an issue that in the conversion to zonotopes from static arrays we should use a static generator matrix.
```julia
julia> B = BallInf(SVector{2}(1., 1.), 1.)
BallInf{Float64,SArray{Tuple{2},Float64,1,2}}([1.0, 1.0], 1.0)

julia> convert(Zonotope, B)
Zonotope{Float64,SArray{Tuple{2},Float64,1,2},Array{Float64,2}}([1.0, 1.0], [1.0 0.0; 0.0 1.0])
# note that only the center is a static vector; the generators are a normal matrix

julia> split(B, [1], [1])
ERROR: MethodError: Cannot `convert` an object of type 
  Zonotope{Float64{},MArray{Tuple{2},Float64,1,2},Array{Float64{},2}} to an object of type 
  Zonotope{Float64{},SArray{Tuple{2},Float64,1,2},Array{Float64{},2}}
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:171
Stacktrace:
 [1] setindex!(::Array{Zonotope{Float64,SArray{Tuple{2},Float64,1,2},Array{Float64,2}},1}, ::Zonotope{Float64,MArray{Tuple{2},Float64,1,2},Array{Float64,2}}, ::Int64) at ./array.jl:826
 [2] _append! at ./array.jl:968 [inlined]
 [3] append! at ./array.jl:960 [inlined]
 [4] _split(::Zonotope{Float64,SArray{Tuple{2},Float64,1,2},Array{Float64,2}}, ::Array{Int64,1}, ::Array{Int64,1}) at .julia/dev/LazySets/src/Sets/Zonotope.jl:402
 [5] split(::BallInf{Float64,SArray{Tuple{2},Float64,1,2}}, ::Array{Int64,1}, ::Array{Int64,1}) at .julia/dev/LazySets/src/Interfaces/AbstractZonotope.jl:652
```

PR #2382, opened 2020-10-12
Improve AbstractHyperrectangular -> Zonotope 2D convert methods for static arrays

mforets:
This PR improves the runtime performance of abstract hyperrectangle -> zonotope convert methods, in particular for static arrays in two dimensions. These changes are related to https://github.com/JuliaReach/LazySets.jl/issues/2248, though it doesn't yet fix the underlying issue there (which should be fixed in the fallback), i'll address them in a separate PR.

Reference notebook: https://nbviewer.jupyter.org/github/mforets/escritoire/blob/master/2020/Week42/LazySets%232248.ipynb

The performance improvement in 2D is ~2.5x with less allocs. But more importantly, if the input hyperrectangular set has static array vectors, now the output has that type too.

For applications that want the maximum performance (and that you don't have/care about hyperrectangles with flat sides), I added the unexported function `LazySets._convert_2D_static`, for which the conversion is immediate as seen in the tests.

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -280,8 +280,12 @@ Converts a hyperrectangular set to a zonotope.
 
 A zonotope.
 """
-function convert(::Type{Zonotope}, H::AbstractHyperrectangle{N}) where {N}
-    if isflat(H)
+function convert(ZT::Type{Zonotope}, H::AbstractHyperrectangle{N}) where {N}
+    if dim(H) == 2
+        return _convert_2D(ZT, H)
+    end
+
+    if prune && isflat(H)
         r = radius_hyperrectangle(H)
         n = length(r)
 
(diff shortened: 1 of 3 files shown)


src/convert.jl
```diff
@@ -280,8 +280,12 @@ Converts a hyperrectangular set to a zonotope.
...
 A zonotope.
 """
-function convert(::Type{Zonotope}, H::AbstractHyperrectangle{N}) where {N}
-    if isflat(H)
+function convert(ZT::Type{Zonotope}, H::AbstractHyperrectangle{N}) where {N}
+    if dim(H) == 2
+        return _convert_2D(ZT, H)
+    end
+
+    if prune && isflat(H)
```
mforets, line 288:
```suggestion
    if isflat(H)
```

mforets:
the appveyor build fail is unrelated ([stacktrace here](https://ci.appveyor.com/project/mforets/lazysets-jl/builds/35703256#L500))

schillic (approved):
A test with static arrays would be good.

mforets:
> A test with static arrays would be good.

:+1:  i'll add some.

PR #2400, opened 2020-10-29
#2399 - Review vertices list of lazy linear map

### LINKED ISSUE #2399: Review vertices list of lazy linear map
```julia
using LazySets, StaticArrays, BenchmarkTools

U = SA[1 2; 0 1] * Hyperrectangle(SA[0, 0.], SA[0.1, 0.1])

# master
@btime vertices_list($U)
  275.497 ns (9 allocations: 864 bytes)
 [0.30000000000000004, 0.1]
 [0.1, 0.1]
 [-0.30000000000000004, -0.1]
 [-0.1, -0.1]
```
This returns an array of regular arrays, but I expected an array of static arrays. The reason is that this operation is used in a chain of other operations (say, involving other static arrays), so converting to `Vector{Float64}` is bad for performance (from this point on we loose all the speedup and type-stability).

This is easy to fix, though: 

```julia
_apply(U) = broadcast(x -> U.M*x, vertices_list(U.X)

  80.704 ns (3 allocations: 384 bytes)
4-element Array{SArray{Tuple{2},Float64,1,2},1}:
 [0.30000000000000004, 0.1]
 [0.1, 0.1]
 [-0.1, -0.1]
 [-0.30000000000000004, -0.1]
```

mforets:
Closes #2399.

----

Reference notebook: [here](https://nbviewer.jupyter.org/github/mforets/escritoire/blob/master/2020/Week41/Area_Overlap.ipynb#New-method) (in the context of an application)

### DIFF
--- a/src/LazyOperations/LinearMap.jl
+++ b/src/LazyOperations/LinearMap.jl
@@ -347,13 +347,13 @@ function an_element(lm::LinearMap{N})::Vector{N} where {N<:Real}
 end
 
 """
-    vertices_list(lm::LinearMap{N}; prune::Bool=true)::Vector{Vector{N}} where {N<:Real}
+    vertices_list(lm::LinearMap; prune::Bool=true)
 
 Return the list of vertices of a (polyhedral) linear map.
 
 ### Input
 
-- `lm` -- linear map
+- `lm`    -- linear map
 - `prune` -- (optional, default: `true`) if true removes redundant vertices
 
 ### Output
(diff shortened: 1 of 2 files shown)


src/LazyOperations/LinearMap.jl
```diff
@@ -362,24 +362,15 @@ A list of vertices.
...
 """
-function vertices_list(lm::LinearMap{N}; prune::Bool=true)::Vector{Vector{N}} where {N<:Real}
-    # for a zero map, the result is just the list containing the origin
-    if iszero(lm.M)
-        return [zeros(N, dim(lm))]
-    end
-
-    # collect low-dimensional vertices lists
+function vertices_list(lm::LinearMap; prune::Bool=true)
+    # collect list of vertices of the wrapped set
```
mforets, line 369: i remove the branch that checks if the user wants to compute `M * X` for `M` the identically null matrix. this case is handled by the prune.

mforets: ... and also because it introduces a type instability whenever the type of the vertices is not the same as `Vector{N}`.

schillic (approved):

PR #2402, opened 2020-11-02
Adapt TagBot

schillic:


mforets (approved):

mforets:
shall we apply this to all other repos?

schillic:
I think so. It is not necessary but recommended.

PR #2418, opened 2020-11-16
#2417 - Specialize concrete minkowski sum for singletons

### LINKED ISSUE #2417: Concrete Minkowski sum of Singletons
```julia
julia> minkowski_sum(rand(Singleton), rand(Singleton))
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([1.5157916104619913, 0.9172086575800644], [0.0, 0.0])
```
This should return a `Singleton` instead.

mforets:
Closes #2417.

### DIFF
--- a/src/ConcreteOperations/minkowski_sum.jl
+++ b/src/ConcreteOperations/minkowski_sum.jl
@@ -244,6 +244,32 @@ function minkowski_sum(Z1::AbstractZonotope{N}, Z2::AbstractZonotope{N}) where {
 end
 
 """
+    minkowski_sum(X::AbstractSingleton, Y::AbstractSingleton)
+
+Concrete Minkowski sum of a pair of singletons.
+
+### Input
+
+- `X` -- singleton
+- `Y` -- singleton
+
+### Output
+
+A singleton
+
+### Algorithm
+
+The singleton obtained by summing the elements in `X` and `Y`.
+"""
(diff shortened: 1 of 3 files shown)


src/ConcreteOperations/minkowski_sum.jl
```diff
@@ -243,6 +243,32 @@ function minkowski_sum(Z1::AbstractZonotope{N}, Z2::AbstractZonotope{N}) where {
...
+
+The singleton obtained by summing the elements in `X` and `Y`.
+"""
+function minkowski_sum(X::AbstractSingleton, Y::AbstractSingleton)
+    return Singleton(element(X) + element(Y))
+end
+
+minkowski_sum(X::AbstractSingleton, Y::ZeroSet) = X
+minkowski_sum(X::ZeroSet, Y::AbstractSingleton) = Y
+minkowski_sum(X::ZeroSet, Y::ZeroSet) = X
```
schillic, line 270: I wonder if we should add dimension checks.

schillic (approved):

src/ConcreteOperations/minkowski_sum.jl
mforets: ok :+1:  as we usually we with `@assert` in this library

schillic (approved):
Ultimately this could actually be the way to implement operations for all set types: have a single `minkowski_sum` method with the error message and then do the real work in `_minkowski_sum` methods.

Issue #2219, opened 2020-07-14
Remove N<:Real annotation


mforets:
We should remove `N<:Real` in most (if not all places). This is an "[anti-pattern](https://white.ucc.asn.au/2020/04/19/Julia-Antipatterns.html)" that we carry from earlier versions of the codebase. Probably for LazySets v2.0 though :up: 

* [x] zonotopic sets (#2276)
* [x] remaining interfaces (#2420)
* [x] `Approximations` (#2422, #2425, #2428)
* [x] concrete sets (#2430, #2435, #2438)
* [x] lazy operations (#2421)
* [ ] concrete operations

PR #2421, opened 2020-11-19
#2219 -- Remove N <: Real for set operations

mforets:
See #2219.


src/LazyOperations/ConvexHull.jl
```diff
@@ -33,12 +33,12 @@ julia> typeof(c)
...
 ```
 """
-struct ConvexHull{N<:Real, S1<:LazySet{N}, S2<:LazySet{N}} <: LazySet{N}
+struct ConvexHull{N, S1<:LazySet{N}, S2<:LazySet{N}} <: LazySet{N}
     X::S1
     Y::S2
 
     # default constructor with dimension check
-    function ConvexHull(X::S1, Y::S2) where {N<:Real, S1<:LazySet{N},
+    function ConvexHull(X::S1, Y::S2) where {N, S1<:LazySet{N},
```
schillic, line 41: In Julia v1.0 this causes an error:
```julia
Unreachable reached at 0x7f8ac74e4628
signal (4): Illegal instruction
in expression starting at /home/travis/build/JuliaReach/LazySets.jl/test/unit_ConvexHull.jl:1
```
Note that the tests first run `Intersection` but there you kept the annotation.

The error would be fixed like this, and this might actually be better because it does not require the same types:
```julia
function ConvexHull(X::S1, Y::S2) where {S1<:LazySet, S2<:LazySet}
    N = promote_type(eltype(X), eltype(Y))
```

mforets: great that you found this! i made the change.

schillic: I guess it is required in all the binary lazy operations unfortunately.

src/LazyOperations/AffineMap.jl
```diff
@@ -104,7 +104,7 @@ function AffineMap(M::UniformScaling, X::LazySet, v::AbstractVector)
 end
 
 # convenience constructor from a scalar
-function AffineMap(α::N, X::LazySet, v::AbstractVector) where {N<:Real}
+function AffineMap(α::N, X::LazySet, v::AbstractVector) where {N}
```
schillic, line 107: I think this annotation should stay.

schillic (commented):
Looks good so far!

src/LazyOperations/AffineMap.jl
mforets, line 107:
```suggestion
function AffineMap(α::N, X::LazySet, v::AbstractVector) where {N<:Real}
```

src/LazyOperations/ConvexHull.jl
```diff
@@ -33,15 +33,14 @@ julia> typeof(c)
...
+struct ConvexHull{N, S1<:LazySet{N}, S2<:LazySet{N}} <: LazySet{N}
     X::S1
     Y::S2
 
     # default constructor with dimension check
-    function ConvexHull(X::S1, Y::S2) where {N<:Real, S1<:LazySet{N},
-                                             S2<:LazySet{N}}
-        @assert dim(X) == dim(Y) "sets in a convex hull must have the same " *
-            "dimension"
+       function ConvexHull(X::S1, Y::S2) where {S1<:LazySet, S2<:LazySet}
```
mforets, line 41:
```suggestion
       function ConvexHull(X::S1, Y::S2) where {N, S1<:LazySet{N}, S2<:LazySet{N}}
```

src/LazyOperations/ConvexHull.jl
mforets: ok, i see what happened.. removing the `N` altogether was not intentional here, i just wanted to remove `N<:Real` as far as this PR is concerned.

i agree that it would be nice to create lazy set operations with different numeric types. however, i'm not sure whether changing the inner constructor is a good idea, we may have to actually convert the data to the promoted type, and maybe such generalized outer constructor is better.

schillic: Here is another version that works:
```julia
function ConvexHull(X::LazySet{N}, Y::LazySet{N}) where {N}
    @assert dim(X) == dim(Y) "sets in a convex hull must have the same dimension"
    return new{N, typeof(X), typeof(Y)}(X, Y)
end
```
`typeof` can always be inferred and hence is the same wrt. run time.
```julia
julia> B = rand(BallInf);

julia> @btime ConvexHull($B, $B)
  19.871 ns (0 allocations: 0 bytes) # this branch
  19.698 ns (0 allocations: 0 bytes) # proposal
```

schillic: It worked, but needs to be changed for `CartesianProduct`, `Intersection`, and `MinkowskiSum` as well :sweat_smile:

src/LazyOperations/MinkowskiSum.jl
```diff
@@ -19,16 +19,14 @@ Type that represents the Minkowski sum of two convex sets.
...
     X::S1
     Y::S2
 
     # default constructor with dimension check
-    function MinkowskiSum(X::S1, Y::S2) where {N<:Real, S1<:LazySet{N},
-                                               S2<:LazySet{N}}
-        @assert dim(X) == dim(Y) "sets in a Minkowski sum must have the " *
-            "same dimension"
-        return new{N, S1, S2}(X, Y)
+    function MinkowskiSum(X::LazySet{N}, Y::Lazyset{N}) where {N}
```
schillic, line 27:
```suggestion
    function MinkowskiSum(X::LazySet{N}, Y::LazySet{N}) where {N}
```

src/LazyOperations/ConvexHull.jl
schillic: Sorry, forgot about the `UnionSet` :disappointed:

schillic (approved):
:100:

mforets:
image

PR #2427, opened 2020-11-24
Refer to ReachabilityAnalysis.jl

goretkin:
From the README at https://github.com/JuliaReach/Reachability.jl , new users should use ReachabilityAnalysis


mforets:
Oh, good catch @goretkin, thanks. The LazySets tutorials part need an thourough update.

schillic (approved):
Thanks!

PR #2438, opened 2020-12-10
#2219 - Remove 'N<:Real' for some set types

schillic:
See #2219.

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -1132,3 +1132,7 @@ function convert(::Type{Zonotope}, cpa::CartesianProductArray{N, AZ}) where
     G = reduce(blockdiag, sparse.(genmat.(arr)))
     return Zonotope(c, G)
 end
+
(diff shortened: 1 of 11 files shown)


src/convert.jl
```diff
@@ -1132,3 +1132,7 @@ function convert(::Type{Zonotope}, cpa::CartesianProductArray{N, AZ}) where
     G = reduce(blockdiag, sparse.(genmat.(arr)))
     return Zonotope(c, G)
 end
+
+convert(::Type{HPolytope}, P::HPolyhedron) = HPolytope(copy(constraints_list(P)))
```
mforets, line 1136: i'm surprised that these functions copy; i didn't expect it to be the case, see for example https://github.com/JuliaReach/LazySets.jl/blob/master/src/convert.jl#L177

mforets (approved):
Awesome!

src/convert.jl
schillic: Just to be clear, I didn't change things here, just moved the code.

PR #2439, opened 2020-12-10
Generalize _isapprox to arrays

schillic:


src/Utils/comparisons.jl
```diff
@@ -250,56 +250,37 @@ function _isapprox(x::N, y::M; kwargs...) where {N<:Real, M<:Real}
     return _isapprox(promote(x, y)...; kwargs...)
 end
 
-# generic "dense"
-function _isapprox(x::AbstractVector{N}, y::AbstractVector{N};
+# numeric arrays
+function _isapprox(A::AbstractArray{N}, B::AbstractArray{N};
```
mforets, line 254: does it work if the arrays don't support linear indexing? (https://docs.julialang.org/en/v1/base/arrays/#Base.IndexStyle)

mforets (approved):

src/Utils/comparisons.jl
schillic: I think so. The way I understand the documentation, Julia will define suitable conversion for arrays that only support Cartesian indexing.
> Julia's internal indexing machinery will automatically (and invisibly) recompute all indexing operations into the preferred style.

PR #2444, opened 2020-12-13
refactor abs sum

mforets:


src/ConcreteOperations/isdisjoint.jl
```diff
@@ -456,41 +456,7 @@ function _is_intersection_empty(Z::AbstractZonotope, H::Union{Hyperplane, Line2D
...
-    abs_sum = zero(N)
-    @inbounds for j in 1:p
-        abs_sum += abs(G[i, j])
-    end
-    abs_sum *= v
-    return abs_sum
-end
-
-function _is_intersection_empty(Z::AbstractZonotope, H::Union{Hyperplane, Line2D}, ::Val{true})
+function _is_intersection_empty(Z::Zonotope, H::Hyperplane, ::Val{true})
```
mforets, line 459:
```suggestion
function _is_intersection_empty(Z::AbstractZonotope, H::Union{Hyperplane, Line2D}, ::Val{true})
```

schillic (approved):

src/Arrays/matrix_vector_operations.jl
```diff
@@ -1,4 +1,39 @@
-export inner
+export inner,
+       _abs_sum
+
+# computes ‖a^T G‖₁
```
schillic, line 4: I am a bit confused but I think that this formula is not correct. It should be `∞` (see [here](https://en.wikipedia.org/wiki/Matrix_norm#Special_cases)). Reason: `a^T G` gives you a row vector and the 1-norm of a row vector is the maximum absolute entry, not the sum.

If you agree, we should double-check where we use this code, for example here:
https://github.com/JuliaReach/LazySets.jl/blob/05066786769525fe47fcfd521a3c9730a249c3a6/src/Interfaces/AbstractZonotope.jl#L239-L247

mforets: The link you provided is for the operator norm (`opnorm` in Base), but the computation that we require here is `||a^T G||_1` where the norm is the vector 1-norm, so it's `sum(abs, transpose(a) * G)`.

schillic: Okay so `||v||_1` and `||v^T||_1` are the same :+1:

mforets:
(build error is unrelated)

PR #2455, opened 2020-12-20
Faster support function for Ball2

### LINKED ISSUE #2454: Explicit support function for Ball2
The fallback implementation computes the support vector, but that is not necessary.

```julia
julia> function rho(d::AbstractVector, B::Ball2)
           return dot(d, B.center) + dot(d, d) * B.radius / norm(d, 2)
       end;

julia> @btime ρ($d, $B)
  63.156 ns (1 allocation: 96 bytes)
3.82842712474619

julia> @btime rho($d, $B)
  36.810 ns (0 allocations: 0 bytes)
3.82842712474619
```

mforets:
Closes #2454.

### DIFF
--- a/docs/src/lib/sets/Ball2.md
+++ b/docs/src/lib/sets/Ball2.md
@@ -6,6 +6,7 @@ CurrentModule = LazySets
 
 ```@docs
 Ball2
+ρ(::AbstractVector, ::Ball2)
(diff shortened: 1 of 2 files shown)


schillic (approved):
Even faster than my silly proposal :tada:

PR #2457, opened 2020-12-22
#2453 - Vertices list of IntervalArithmetic types

### LINKED ISSUE #2453: Vertices list of IA types
Add:

- `vertices_list(::IA.Interval)`
- `vertices_list(::IA.IntervalBox)`

mforets:
Closes #2453.

Notes:

- The method for `IntervalBox` can certainly be improved, but this is better than nothing (we can improve it as a follow-up PR).
- The method for `Interval` creates static arrays because that's consistent with the implementation of the interval box type (note that the conversion from `IntervalBox` to `Hyperrectangle` returns SArrays already).

### DIFF
--- a/src/Sets/Interval.jl
+++ b/src/Sets/Interval.jl
@@ -696,3 +696,19 @@ end
         return result
     end
 end
+
+function vertices_list(x::IntervalArithmetic.Interval{N}) where {N}
+    a = IntervalArithmetic.inf(x)
+    b = IntervalArithmetic.sup(x)
+    ST = IntervalArithmetic.SVector{1, N}
+    if _isapprox(a, b)
+        vlist = [ST(a)]
+    else
+        vlist = [ST(a), ST(b)]
+    end
+    return vlist
+end
+
+function vertices_list(H::IntervalArithmetic.IntervalBox)
+    return vertices_list(convert(Hyperrectangle, H))
+end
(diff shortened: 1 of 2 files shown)


schillic (commented):
I would prefer that `vertices_list(::LazySets.Interval)` and `vertices_list(::IntervalArithmetic.Interval)` return the same result. Can you redefine `vertices_list(::LazySets.Interval)` such that it falls back to this new method?

mforets:
It's as bit strange to return an `SArray` since `StaticArrays` is not a dependency of `LazySets` (here I used it through `IntervalArithmetic`). We use `Vector` by default in everywhere, eg.

```julia
julia> x = rand(Hyperrectangle, dim=1)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([1.9237336716037892], [0.14284009250436838])

julia> vertices_list(x)
 [2.0665737641081576]
 [1.7808935790994207]
```
The reason to return a vector of `SArray` for the IA.Interval case here was that it matches the multidim `IntervalBox`.

schillic:
Good point. But at least the option to return a single vertex if the end points are (almost) identical should be consistent I think.

mforets:
Good point; I can do that in a separate PR (https://github.com/JuliaReach/LazySets.jl/issues/2458).

schillic (approved):

PR #2465, opened 2020-12-28
#2464 - Specific singleton list plot

### LINKED ISSUE #2464: Plotting large samples is slow
I found that plotting samples of sets (eg. long lists of singletons) is very inefficient to the extent that it takes an extremely long time to plot only ~5000 points using the plot recipe for a list of singletons.
 
```julia
B = (1 .. 2) × (3 .. 4)
Bs = sample(convert(Hyperrectangle, B), 1000);

@time Plots.plot(Singleton.(Bs))
  0.681963 seconds (1.96 M allocations: 97.579 MiB, 3.95% gc time)

x = [p[1] for p in Bs];
y = [p[2] for p in Bs];
@time plot(x, y, seriestype=:scatter, lab="")
  0.002430 seconds (2.96 k allocations: 189.648 KiB)

```

mforets:
Closes #2464.

### DIFF
--- a/src/Plotting/plot_recipes.jl
+++ b/src/Plotting/plot_recipes.jl
@@ -199,7 +199,7 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
 end
 
 # recipe for vector of singletons
-@recipe function plot_list(list::Vector{SN}) where {N, SN<:AbstractSingleton{N}}
+@recipe function plot_list(list::AbstractVector{SN}) where {N, SN<:AbstractSingleton{N}}
 
     label --> DEFAULT_LABEL
     grid --> DEFAULT_GRID
--- a/src/Plotting/plot_recipes.jl
+++ b/src/Plotting/plot_recipes.jl
@@ -229,7 +229,7 @@ end
     _plot_singleton_list(list)
 end
 
-function _plot_singleton_list(list::Vector{SN}) where {N, SN<:AbstractSingleton{N}}
+function _plot_singleton_list(list)
     n = dim(first(list))
     if n == 1
         _plot_singleton_list_1D(list)
--- a/src/Plotting/plot_recipes.jl
+++ b/src/Plotting/plot_recipes.jl
@@ -382,8 +382,7 @@ julia> plot(Singleton([0.5, 1.0]))
         ylims --> lims[:y]
     end
 
-    p = element(S)
-    [p[1], p[2]]
+    plot_recipe(S, ε)
 end
 
 """
--- a/src/Plotting/plot_recipes.jl
+++ b/src/Plotting/plot_recipes.jl
@@ -241,7 +241,7 @@ function _plot_singleton_list(list)
     end
 end
 
-function _plot_singleton_list_1D(list::Vector{SN}) where {N, SN<:AbstractSingleton{N}}
+function _plot_singleton_list_1D(list::AbstractVector{SN}) where {N, SN<:AbstractSingleton{N}}
     m = length(list)
 ```suggestion
     x = Vector{N}(undef, m)
--- a/src/Plotting/plot_recipes.jl
+++ b/src/Plotting/plot_recipes.jl
@@ -198,6 +198,50 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
     end
 end
 
+# recipe for vector of singletons
+@recipe function plot_list(list::Vector{SN}) where {N, SN<:AbstractSingleton{N}}
+
+    label --> DEFAULT_LABEL
+    grid --> DEFAULT_GRID
+    if DEFAULT_ASPECT_RATIO != :none
+        aspect_ratio --> DEFAULT_ASPECT_RATIO
+    end
+    seriesalpha --> DEFAULT_ALPHA
+    seriescolor --> DEFAULT_COLOR
+    seriestype --> :scatter
+
+    _plot_singleton_list(list)
+end
+
+# plot recipe for the union of singletons
+@recipe function plot_list(X::UnionSetArray{N, SN}) where {N, SN<:AbstractSingleton{N}}
+
+    label --> DEFAULT_LABEL
+    grid --> DEFAULT_GRID
+    if DEFAULT_ASPECT_RATIO != :none
+        aspect_ratio --> DEFAULT_ASPECT_RATIO
+    end
+    seriesalpha --> DEFAULT_ALPHA
+    seriescolor --> DEFAULT_COLOR
+    seriestype --> :scatter
+
+    list = array(X)
+    _plot_singleton_list(list)
+end
+
+function _plot_singleton_list(list::Vector{SN}) where {N, SN<:AbstractSingleton{N}}
+    m = length(list)
+    x = Vector{N}()
+    y = Vector{N}()
+
+    for Xi in list
+        p = element(Xi)
+        push!(x, p[1])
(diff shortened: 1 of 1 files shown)


src/Plotting/plot_recipes.jl
```diff
@@ -313,7 +357,8 @@ julia> plot(Singleton([0.5, 1.0]))
         ylims --> lims[:y]
     end
 
-    plot_recipe(S, ε)
+    p = element(S)
+    [p[1], p[2]]
```
schillic, line 361: 1D plots are broken, and I think this call to `plot_recipe` was supposed to resolve that.
```julia
julia> B = (1 .. 2);
julia> Bs1 = LazySets.sample(LazySets.Interval(B), 1000);
julia> Plots.plot(Singleton.(Bs1))
ERROR: BoundsError: attempt to access 1-element Array{Float64,1} at index [2]
```

mforets: good catch; i had in mind 2d plots, but we can special case that.

schillic: This method does exactly that.

https://github.com/JuliaReach/LazySets.jl/blob/9d8a38773c568d7e2a29a21aff6d79794b5750f9/src/Interfaces/AbstractSingleton.jl#L369-L377

mforets: do you think we can in some way reuse that function?

this PR was to improve the plotting recipe of vectors of a large number of singletons? the main difference of this PR with the one in master is that we make a scatter plot and we treat all the singletons in the same loop. the performance is very good (same as using Plots directly).

schillic: I tried undoing this change and it gives the same result with exactly the same allocations, so I would say "yes" :smiley:
Was there a reason why you did not undo this change?

mforets: ah, ok i see now! yes, i'll undo this change. the reason was that it was very unexpected to see `plot_recipe(S, ε)` where i just expected `[p[1], p[2]]` so i thought this could be introducing additional allocations etc.

schillic: I agree that this is not very transparent :sweat_smile:

mforets: i mean, the epsilon stuff seemed complicated for only returning the array.

schillic: > and it gives the same result with exactly the same allocations

Well, that was not surprising because it does not trigger the discussed code anymore :roll_eyes:

src/Plotting/plot_recipes.jl
```diff
@@ -198,6 +198,75 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
...
+
+function _plot_singleton_list_1D(list::Vector{SN}) where {N, SN<:AbstractSingleton{N}}
+    m = length(list)
+    x = Vector{N}()
+    y = Vector{N}()
+
+    for Xi in list
+        p = element(Xi)
+        push!(x, p[1])
+        push!(y, zero(N))
```
schillic, line 252:
```suggestion
```suggestion
    x = Vector{N}(undef, m)
    y = zeros(N, m)

    @inbounds for (i, Xi) in enumerate(list)
        p = element(Xi)
        x[i] = p[1]
```

src/Plotting/plot_recipes.jl
```diff
@@ -198,6 +198,75 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
...
+    end
+    x, y
+end
+
+function _plot_singleton_list_2D(list::Vector{SN}) where {N, SN<:AbstractSingleton{N}}
```
schillic, line 257:
```suggestion
function _plot_singleton_list_2D(list::AbstractVector{SN}) where {N, SN<:AbstractSingleton{N}}
```

src/Plotting/plot_recipes.jl
```diff
@@ -198,6 +198,75 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
     end
 end
 
+# recipe for vector of singletons
+@recipe function plot_list(list::Vector{SN}) where {N, SN<:AbstractSingleton{N}}
```
schillic, line 202:
```suggestion
@recipe function plot_list(list::AbstractVector{SN}) where {N, SN<:AbstractSingleton{N}}
```

src/Plotting/plot_recipes.jl
```diff
@@ -198,6 +198,75 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
...
+    end
+    seriesalpha --> DEFAULT_ALPHA
+    seriescolor --> DEFAULT_COLOR
+    seriestype --> :scatter
+
+    list = array(X)
+    _plot_singleton_list(list)
+end
+
+function _plot_singleton_list(list::Vector{SN}) where {N, SN<:AbstractSingleton{N}}
```
schillic, line 232:
```suggestion
function _plot_singleton_list(list)
```

src/Plotting/plot_recipes.jl
```diff
@@ -198,6 +198,75 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
...
+        _plot_singleton_list_1D(list)
+    elseif n == 2
+        _plot_singleton_list_2D(list)
+    else
+        throw(ArgumentError("plotting a vector of singletons is only available for dimensions " *
+             "one or two, got dimension $n"))
+    end
+end
+
+function _plot_singleton_list_1D(list::Vector{SN}) where {N, SN<:AbstractSingleton{N}}
```
schillic, line 244:
```suggestion
function _plot_singleton_list_1D(list::AbstractVector{SN}) where {N, SN<:AbstractSingleton{N}}
```

src/Plotting/plot_recipes.jl
```diff
@@ -198,6 +198,75 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
...
+
+function _plot_singleton_list_2D(list::Vector{SN}) where {N, SN<:AbstractSingleton{N}}
+    m = length(list)
+    x = Vector{N}()
+    y = Vector{N}()
+
+    for Xi in list
+        p = element(Xi)
+        push!(x, p[1])
+        push!(y, p[2])
```
schillic, line 265:
```suggestion
    x = Vector{N}(undef, m)
    y = Vector{N}(undef, m)

    @inbounds for (i, Xi) in enumerate(list)
        p = element(Xi)
        x[i] = p[1]
        y[i] = p[2]
```

src/Plotting/plot_recipes.jl
```diff
@@ -198,6 +198,75 @@ julia> plot(Bs, 1e-2)  # faster but less accurate than the previous call
...
+        _plot_singleton_list_2D(list)
+    else
+        throw(ArgumentError("plotting a vector of singletons is only available for dimensions " *
+             "one or two, got dimension $n"))
+    end
+end
+
+function _plot_singleton_list_1D(list::AbstractVector{SN}) where {N, SN<:AbstractSingleton{N}}
+    m = length(list)
+```suggestion
```
schillic, line 246:
```suggestion
```

schillic (commented):
oops

mforets:
Thanks for all the suggestions :+1:  Let me know when it's ready to merge from your side.

schillic (approved):

PR #1213, opened 2019-03-07
Add halfspace and hpolyhedron concrete projection.

mforets:
**Note that** the new part cannot be used by `decompose`, since it needs four argument syntax 

```julia
    project(::HalfSpace{N},
            block::AbstractVector{Int},
            set_type::HalfSpace,
            [n]::Int=dim(H)
           ) where {N<:Real}
```

PR #2478, opened 2021-01-02
Generalize special case for concrete projection of polyhedra

schillic:
This PR generalizes the concrete projection of polyhedra based on the idea of the constrained dimensions (#1213).
As a bonus it also removes the boundedness check of the `HPolytope` in the end.

#### Idea

If a constraint's constrained dimensions all get projected away, the constraint becomes redundant and we can just ignore it.

#### Example

A 2D `Hyperrectangle` can be projected exactly to dimension 1 (`x`). This is still true if we consider the constraint representation. The constraints have the form `ax + 0y <= b` or `0x + ay <= b`. In the first case one can project away the redundant dimension `y`. In the second case one can drop the whole constraint. (Note that this is only correct because we assume that our half-spaces are feasible.)

#### Advantage

- The implementation is more efficient for a class of cases; in particular we do not require `Polyhedra`/`CDDLib` in this case.

```julia
julia> P = convert(HPolytope, rand(BallInf));

julia> @btime project($P, [1])  # master
  496.385 μs (1328 allocations: 56.42 KiB)
HPolytope{Float64,Array{Float64,1}}(HalfSpace{Float64,Array{Float64,1}}[
 HalfSpace{Float64,Array{Float64,1}}([1.0], -1.8694508399505545),
 HalfSpace{Float64,Array{Float64,1}}([-1.0], 3.714432065736666)])

julia> @btime project($P, [1])  # this branch
  694.282 ns (16 allocations: 1.23 KiB)
HPolytope{Float64,Array{Float64,1}}(HalfSpace{Float64,Array{Float64,1}}[
 HalfSpace{Float64,Array{Float64,1}}([1.0], -1.8694508399505545),
 HalfSpace{Float64,Array{Float64,1}}([-1.0], 3.714432065736666)])
```

#### Disadvantages

- There is a very minor overhead compared to the previous implementation for the previous special case because of additional checks.
- The implementation got more complicated.

### DIFF
--- a/src/Interfaces/AbstractPolyhedron_functions.jl
+++ b/src/Interfaces/AbstractPolyhedron_functions.jl
@@ -1056,32 +1056,41 @@ julia> project(P, [1, 2]) |> constraints_list
 ```
 """
 function project(P::AbstractPolyhedron{N}, block::AbstractVector{Int}) where {N}
+    function _check_constrained_dimensions(c::HalfSpace)
+        # 1 = constrained dimensions subset of block
+        # -1 = constrained dimensions disjoint from block
+        # 0 = mixed case
+        status = 0
+        for i in constrained_dimensions(c)
+            if i in block
+                # case 1: is a subset
+                if status == -1
+                    status = 0
+                    break
+                end
+                status = 1
+            else
+                # case 1: is disjoint
+                if status == 1
+                    status = 0
+                    break
+                end
+                status = -1
+            end
+        end
+        return status
+    end
+
     general_case = false
 
     # cheap case
     clist = nothing  # allocate later
     @inbounds for c in constraints(P)
-        subset_status = 0  # initially 0, set to 1 if subset and -1 if disjoint
-        for i in constrained_dimensions(c)
-            if i in block
-                # case 1: is a subset
-                if subset_status == -1
-                    general_case = true
-                    break
-                elseif subset_status == 0
-                    subset_status = 1
-                end
-            else
-                # case 1: is disjoint
-                if subset_status == 1
-                    general_case = true
-                    break
-                elseif subset_status == 0
-                    subset_status = -1
-                end
-            end
-        end
-        if subset_status == 1
+        status = _check_constrained_dimensions(c)
+        if status == 0
+            general_case = true
+            break
+        elseif status == 1
             # simple projection of half-space
             hs = HalfSpace(c.a[block], c.b)
             if clist == nothing
@@ -1089,6 +1098,8 @@ function project(P::AbstractPolyhedron{N}, block::AbstractVector{Int}) where {N}
             else
                 push!(clist, hs)
             end
+        elseif status == -1
+            # drop the constraint because it became redundant
         end
     end
 
@@ -1105,6 +1116,6 @@ function project(P::AbstractPolyhedron{N}, block::AbstractVector{Int}) where {N}
 end
 
 function _project_polyhedron(P::LazySet, block)
-    lm = project(P, block, nothing)
-    return constraints_list(lm)
+    projected = project(P, block)  # concrete projection of P
+    return constraints_list(projected)
 end
--- a/src/Interfaces/AbstractPolyhedron_functions.jl
+++ b/src/Interfaces/AbstractPolyhedron_functions.jl
@@ -184,7 +184,7 @@ Note that the result may be `true` even if the constraints are infeasible.
 For example, ``x ≤ 0 && x ≥ 1`` will return `true` without removing any
 constraint.
 To check if the constraints are infeasible, use
-`isempty(HPolyhedron(constraints)`.
+`isempty(HPolyhedron(constraints))`.
 
 ### Algorithm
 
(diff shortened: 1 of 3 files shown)


schillic:
The build error was unrelated. I try to fix it here but it is nondeterministic.

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -1052,14 +1056,53 @@ julia> project(P, [1, 2]) |> constraints_list
...
+            if clist == nothing
+                clist = [hs]  # get the right type of the constraints
+            else
+                push!(clist, hs)
+            end
+        end
+    end
+
+    # general case
+    if general_case
```
mforets, line 1096: can you outsource the general case to a separate function `_project_polyhedron` or sth?

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -1052,14 +1056,53 @@ julia> project(P, [1, 2]) |> constraints_list
...
 """
 function project(P::AbstractPolyhedron{N}, block::AbstractVector{Int}) where {N}
-    if constrained_dimensions(P) ⊆ block
-        clist = [HalfSpace(c.a[block], c.b) for c in constraints_list(P)]
-    else
+    general_case = false
+
+    # cheap case
+    clist = nothing  # allocate later
+    @inbounds for c in constraints(P)
```
mforets, line 1063: does the check `constrained_dimensions(P) ⊆ block` still apply? and if it is false => use the general case

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -1052,14 +1056,53 @@ julia> project(P, [1, 2]) |> constraints_list
...
+        subset_status = 0  # initially 0, set to 1 if subset and -1 if disjoint
+        for i in constrained_dimensions(c)
```
mforets, line 1065: can we make a small helper function for the loop in line 1065-1082 ? iiuc, this function receives a half-space `c` and a vector of integers `block` and returns two flags, `general_case` and `subset_status`

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -993,8 +993,12 @@ If `P` was bounded, the result is an `HPolytope`; otherwise the result is an
 
 ### Algorithm
 
-- If the unconstrained dimensions of `P` are a subset of the `block` variables,
-  each half-sace `c` of `P` is transformed to `HalfSpace(c.a[block], c.b)`.
+- We first try to exploit the special case where each of the constraints of `P`
```
mforets, line 996: i guess that the necessary `condition constrained_dimensions(P) ⊆ block` is easy to check before doing some work as in lines 1084 - 1090

src/Interfaces/AbstractPolyhedron_functions.jl
schillic: done

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -1052,14 +1056,55 @@ julia> project(P, [1, 2]) |> constraints_list
...
-        lm = linear_map(M, P)
-        clist = constraints_list(lm)
+         return HPolyhedron(clist)
     end
-    T = isbounded(P) ? HPolytope : HPolyhedron
-    return T(clist)
+end
+
+function _project_polyhedron(P::LazySet, block)
+    lm = project(P, block, nothing)
```
mforets, line 1108: i know this is an unrelated comment, but i find that calling again `project(P, block, nothing)` to be unclear as what is the code actually doing (the `nothing` is strange), while we could do `constraints_list(linear_map(projection_matrix(block, dim(P), N), P))` without needing to call again `project`.

src/Interfaces/AbstractPolyhedron_functions.jl
schillic: The old check is part of the new check. `constrained_dimensions(P)` goes through all constraints of `P` to collect the constrained dimensions in *all* constraints. The new check does not require the aggregated constrained dimensions. First doing the old check would essentially do the same work twice and does not simplify the code because we would still need to have the current code again for having the general case.

src/Interfaces/AbstractPolyhedron_functions.jl
schillic: I don't know. `project` gives you a concrete projection. I can remove the `nothing if you want, that part is optional. I prefer to reuse existing code rather than having a copy.

src/Interfaces/AbstractPolyhedron_functions.jl
schillic: done

mforets (approved):

src/Interfaces/AbstractPolyhedron_functions.jl
mforets: we get a stackoverflow by calling again `project`: https://github.com/JuliaReach/LazySets.jl/issues/2539

PR #2479, opened 2021-01-03
Fix bug in isdisjoint of two zonotopic sets

schillic:
This fixes a serious bug in #1269 (the result was inverted). The tests were also wrong :face_with_head_bandage:


mforets (approved):

schillic:
The build error is unrelated (`project` docs).

PR #2483, opened 2021-01-03
Fix docs (project methods)

schillic:
This fixes the docs error that sometimes happens since #2475.

Don't ask me why
```julia
 project(S::LazySet, ::AbstractVector{Int}, ::Type{var"#s27"} where var"#s27"<:LazySet, ::Int=dim(S))
```
works but
```julia
 project(S::LazySet, ::AbstractVector{Int}, ::Type{T} where T<:LazySet, ::Int=dim(S))
```
does not...

### DIFF
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -74,9 +74,9 @@ surface(::LazySet{N}) where {N}
 area(::LazySet{N}) where {N}
 concretize(::LazySet)
 complement(::LazySet)
-project(S::LazySet{N}, ::AbstractVector{Int}, ::Nothing, ::Int=dim(S)) where {N}
-project(S::LazySet, ::AbstractVector{Int}, ::Type{<:LazySet}, ::Int=dim(S))
-project(S::LazySet, ::AbstractVector{Int}, ::Pair{<:UnionAll, <:Real}, ::Int=dim(S))
+project(S::LazySet{N}, ::AbstractVector{Int}, ::Nothing=nothing, ::Int=dim(S)) where {N}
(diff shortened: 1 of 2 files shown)


docs/src/lib/interfaces.md
```diff
@@ -74,9 +74,9 @@ surface(::LazySet{N}) where {N}
 area(::LazySet{N}) where {N}
 concretize(::LazySet)
 complement(::LazySet)
-project(S::LazySet{N}, ::AbstractVector{Int}, ::Nothing, ::Int=dim(S)) where {N}
-project(S::LazySet, ::AbstractVector{Int}, ::Type{<:LazySet}, ::Int=dim(S))
-project(S::LazySet, ::AbstractVector{Int}, ::Pair{<:UnionAll, <:Real}, ::Int=dim(S))
+project(S::LazySet{N}, ::AbstractVector{Int}, ::Nothing=nothing, ::Int=dim(S)) where {N}
+project(S::LazySet, ::AbstractVector{Int}, ::Type{var"#s27"} where var"#s27"<:LazySet, ::Int=dim(S))
```
mforets, line 78: that hmm not robust? if we define a new variable in that test (or a new lazyset type?) that may change

schillic: What do you mean?

mforets: that `var"#s27"` refers to some internal id that may change.

schillic: Hm, but at least it works for now. I could not find a different way. What should we do?

mforets: ok

Issue #1599, opened 2019-08-13
Generic concrete minkowski sum


mforets:
By generics i mean function that dispatch on `LazySet`s. We can check if the argument is 2D then use the concrete minkowski sum in v-representation. It is more efficient than using the higher-dimensional approach (see #1597).

mforets:
There are different choices, this is one possibility:

- if the set is 2D
    - if the vertices_list is available => use `_minkowski_sum_vrep`
    - else if the constraints_list is available => use `_minkowski_sum_hrep`
    - else error
- else if the set is > 2D
    - if the constraints_list is available => use `_minkowski_sum_hrep`
    - else if the vertices_list is available  => use `_minkowski_sum_vrep`
    - else error

PR #2488, opened 2021-01-04
update concrete minkowski sum

mforets:
related to https://github.com/JuliaReach/LazySets.jl/issues/1599


src/ConcreteOperations/minkowski_sum.jl
```diff
@@ -389,9 +395,17 @@ function minkowski_sum(P1::VPolytope, P2::VPolytope;
 
     vlist1 = _vertices_list(P1, backend)
     vlist2 = _vertices_list(P2, backend)
+    Vout = _minkowski_sum_vrep_nd(vlist1, vlist2, apply_convex_hull=apply_convex_hull, backend=backend, solver=solver)
+    return VPolytope(Vout)
+end
+
+function _minkowski_sum_vrep_nd(vlist1::Vector{VT}, vlist2::Vector{VT};
```
schillic, line 402:
```suggestion
function _minkowski_sum_vrep_nd(vlist1::AbstractVector{VT}, vlist2::AbstractVector{VT};
```

schillic, line 402: What if the vectors have mixed types?

mforets: i think the `Vector` of vectors covers the common cases.. ?

mforets: i think that use case is quite uncommon and we can leave the conversion to the user. the new thing that this brings is that before we hardcoded to `Vector{N}` and now it uses `VT`. im not even sure if this makes sense in general because if `VT` is a SingleEntryVector, their sum need not be the of that type.

schillic: Does the implementation rely on anything that is specific to `Vector`? If not, I think it should not have this restriction.

schillic: A vertex of type `SingleEntryVector` would be strange. The only case I see is an `SVector` mixed with a normal `Vector`.
But you are right that this was not possible before as well, so this is already a step forward.

src/Sets/VPolygon.jl
```diff
@@ -301,57 +301,79 @@ const binary_or_brute_force = 10
 
 # return the index of a support vector of P along d from the list of vertices of P
 function _σ_helper(d::AbstractVector, P::VPolygon)
-    if length(P.vertices) > binary_or_brute_force
-        return _binary_support_vector(d, P)
+    vlist = P.vertices
+    return _σ_helper(d, vlist)
+end
+
+function _σ_helper(d::AbstractVector, vlist::Vector{VN}) where {N, VN<:AbstractVector{N}}
```
mforets, line 308: i had to add this function so that it is cleaner to reuse `_σ_helper` in [this line](https://github.com/JuliaReach/LazySets.jl/pull/2488/files#diff-10c1b027a265435b398c6151f8e72799995bcff09e9a72d9d818b12683df9e98R335).

src/ConcreteOperations/minkowski_sum.jl
```diff
@@ -318,17 +318,23 @@ given polygons. This algorithm assumes that the vertices of P and Q are sorted
...
     mQ = length(vlistQ)
 
     EAST = N[1, 0]
     ORIGIN = N[0, 0]
-    k = _σ_helper(EAST, P)
-    j = _σ_helper(EAST, Q)
-    R = Vector{Vector{N}}(undef, mP+mQ)
+    k = _σ_helper(EAST, vlistP)
+    j = _σ_helper(EAST, vlistQ)
+    R = Vector{VT}(undef, mP+mQ)
```
mforets, line 337: with the generalization `VT`, we can now use static arrays, eg.

```julia
julia> @btime minkowski_sum($a, $b);  # SArray
  1.002 μs (12 allocations: 1.78 KiB)

julia> @btime minkowski_sum($v, $q); # same with Vector
  3.523 μs (74 allocations: 7.03 KiB)
```

src/ConcreteOperations/minkowski_sum.jl
mforets: ok. i think the implementation doesn't rely specifically an outer `Vector`; i just assumed that it is the case because it is the usual case. but i can change it

src/ConcreteOperations/minkowski_sum.jl
```diff
@@ -400,12 +413,12 @@ function minkowski_sum(P1::VPolytope, P2::VPolytope;
     end
     if apply_convex_hull
         if backend == nothing
-            backend = default_polyhedra_backend(P1, N)
+            backend = default_polyhedra_backend(N)
```
mforets, line 416:
```suggestion
            backend = default_polyhedra_backend_nd(N)
```

mforets:
all checks have passed.. merge? @schillic

schillic:
> merge?

Okay, the only thing that could be generalized is the [restriction to `Vector`](https://github.com/JuliaReach/LazySets.jl/pull/2488#discussion_r551304126) in some places.

mforets:
:+1:  then i'll merge now

PR #2489, opened 2021-01-04
Increase timeout for docs build

schillic:
Recently we breach the 30 minutes :face_with_head_bandage:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,5 +37,5 @@ jobs:
       script:
         - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
                                                Pkg.instantiate();'
-        - travis_wait 30 julia --project=docs/ docs/make.jl
+        - travis_wait 40 julia --project=docs/ docs/make.jl
       after_success: skip
--- a/docs/src/man/lazy_intersections.md
+++ b/docs/src/man/lazy_intersections.md
@@ -122,13 +122,6 @@ but overapproximation with template directions can be used in any dimension.
 
 Let's time it!
 
-```@example ellipsoids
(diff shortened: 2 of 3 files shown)


mforets (approved):

schillic:
> Timeout (40 minutes) reached. Terminating "julia --project=docs/ docs/make.jl"

Wow, something seems to be broken... I'll try to build locally.

schillic:
My findings: The majority of the time is spent in compiling all the methods. Apart from that, the docs build very fast (7 seconds) with two exceptions:
* https://github.com/JuliaReach/LazySets.jl/blob/master/docs/src/man/reach_zonotopes_hybrid.md
    * We compute 4k sets and plot them (~125 seconds).
* https://github.com/JuliaReach/LazySets.jl/blob/master/docs/src/man/lazy_intersections.md
    * Here we use `@btime` several times (I stopped it after ~10 minutes). Is this a regression?

@mforets What do you think about replacing the `@btime` by ``@time` from the docs (only takes 1 second then) and using faster parameters in the reachability algorithm? I tried `δ = 0.01` (10x bigger) and that only takes 2 seconds with almost the same result.

schillic:
I pushed a proposal. Let's look at the build time.

mforets:
> @mforets What do you think about removing the @btime from the docs (only takes 1 second then) and using faster parameters in the reachability algorithm? I tried δ = 0.01 (10x bigger) and that only takes 2 seconds with almost the same result.

sounds good!

> Here we use @btime several times (I stopped it after ~10 minutes). Is this a regression?


idk, i can check too

schillic:
26 minutes :tada:

mforets (approved):

Issue #2310, opened 2020-08-04
Make ready for Julia v1.5


schillic:
[build log](https://travis-ci.org/github/JuliaReach/LazySets.jl/jobs/714930786)

* [x] The tests pass but there are warnings about global variables. (#2318)
* [x] The doctests fail because of a new output style for empty arrays. (#2504)
* [x] The [documentation of `isdisjoint`](https://github.com/JuliaReach/LazySets.jl/blob/bec8d2a4db6699262620ff359fabba70d255166a/src/ConcreteOperations/isdisjoint.jl#L54-L59) does not work. (#2504)
```julia
Warning: no docs found for 'isdisjoint' in `@docs` block in src/lib/binary_functions.md:23-46
```

mforets:
>  The doctests fail because of a new output style for empty arrays.

in v1.5 it is 

```julia
julia> Vector{Vector{Float64}}()
Array{Float64,1}[]
```

and in v1.6 it is 

```julia
julia> Vector{Vector{Float64}}()
Vector{Float64}[]
```

PR #2502, opened 2021-01-13
WIP- Add Julia v1.6 to tests

mforets:
this is to begin testing with the new julia version (no need that we merge this immediately)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,6 +1,6 @@
 environment:
   matrix:
-  - julia_version: 1.4
+  - julia_version: 1.5
 
 platform:
   #- x86 # 32-bit
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: julia-actions/setup-julia@latest
         with:
-          version: 1.3
+          version: 1.5
       - name: Pkg.add("CompatHelper")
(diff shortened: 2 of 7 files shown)


schillic:
I would be surprised if this worked, see #2310 ...

mforets:
i thought that doctests are not ran for all versions, only for v1.4

mforets:
i think we should just skip v1.5 :smile: and prepare everything for v1.6

schillic:
> i thought that doctests are not ran for all versions, only for v1.4

I think that is not the case :wink:
When I run tests locally in v1.5, they always fail in the end in the doctests.

mforets:
indeed, we have set it to behave like that :) https://github.com/JuliaReach/LazySets.jl/blob/master/test/runtests.jl#L29

mforets:
so i take back that "i think we should just skip v1.5 " -- but we will have to remove v1.4 and keep v1.5 (and do the same for the 1.5 to 1.6 transition), because the default printing of some types change between the different versions

schillic:
I am fine with not adapting the doctests to v1.5 and directly going to v1.6. We should only go this two-step way if the changes are incremental. The library itself runs fine in v1.5.

mforets:
it may take some time for the rest of the ecosystem to adapt (to the new Julia version), so i guess that addressing https://github.com/JuliaReach/LazySets.jl/issues/2310 is worthwhile, even if in a couple of weeks we may have to make some work again

Issue #2506, opened 2021-01-13
Convert HPolyhedron to HPolygon


mforets:
Eg.

```julia
R = HPolyhedron(rand(8, 2), rand(8))
convert(HPolygon, R)

MethodError: Cannot `convert` an object of type 
  HPolyhedron{Float64,Array{Float64,1}} to an object of type 
  HPolygon
Closest candidates are:
  convert(::Type{HPolygon}, !Matched::HPolygon) at /home/mforets/.julia/dev/LazySets/src/convert.jl:8
  convert(::Type{HPOLYGON1}, !Matched::HPOLYGON2) where {HPOLYGON1<:AbstractHPolygon, HPOLYGON2<:AbstractHPolygon} at /home/mforets/.julia/dev/LazySets/src/convert.jl:33
  convert(::Type{HPOLYGON}, !Matched::VPolygon) where HPOLYGON<:AbstractHPolygon at /home/mforets/.julia/dev/LazySets/src/convert.jl:56
  ...

Stacktrace:
 [1] top-level scope at In[18]:2
 [2] include_string(::Function, ::Module, ::String, ::String) at ./loading.jl:1091
```

fails -- we could easily write a fallback conversion method that calls `constraints_list` of the input set.

PR #2509, opened 2021-01-18
#2506 - Generic conversion to HPolygon

mforets:
Closes https://github.com/JuliaReach/LazySets.jl/issues/2506.

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -10,36 +10,6 @@ for T in subtypes(LazySet, true)
 end
 
 """
-    convert(::Type{HPOLYGON1},
-            P::HPOLYGON2) where {HPOLYGON1<:AbstractHPolygon,
-                                 HPOLYGON2<:AbstractHPolygon}
-
-Convert between polygon types in H-representation.
-
-### Input
-
-- `type` -- target type
-- `P`    -- source polygon
-
-### Output
-
-The polygon represented as the target type.
-
-### Notes
-
-We need the `Union` type for `HPOLYGON1` because the target type must be
-concrete.
-"""
-function convert(::Type{HPOLYGON1},
-                 P::HPOLYGON2) where {HPOLYGON1<:AbstractHPolygon,
-                                      HPOLYGON2<:AbstractHPolygon}
-    if P isa HPOLYGON1
-        return P
-    end
-    return HPOLYGON1(P.constraints)
-end
-
-"""
     convert(T::Type{HPOLYGON}, P::VPolygon) where {HPOLYGON<:AbstractHPolygon}
 
 Converts a polygon in vertex representation to a polygon in constraint representation.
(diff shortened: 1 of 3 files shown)


mforets:
the test is caused by removing the method `function convert(::Type{HPOLYGON1},	function convert(T::Type{HPOLYGON}, P::VPolygon) where {HPOLYGON<:AbstractHPolygon}
                 P::HPOLYGON2) where {HPOLYGON1<:AbstractHPolygon,	    return tohrep(P, T)
                                      HPOLYGON2<:AbstractHPolygon}`, because then there is no method to make the conversion `HPolygon{Float64,Array{Float64,1}}(::Array{HalfSpace{Float64,Array{Float64,1}},1})`.

we can add the method back or use a more general approach by defining a new `convert` method for `X` and `typeof(X)`

mforets:
actually that is fixed by creating the polygon with `basetype` ([this line](https://github.com/JuliaReach/LazySets.jl/pull/2509/files#diff-a72bb9800ba27210ca77c3215e0e82253c9d80789750f720fd45200f11fafb6bR50))

schillic:
```julia
julia> H = rand(HalfSpace)
HalfSpace{Float64,Array{Float64,1}}([0.11082914558894325, -0.0010776789901761029], -0.9539738894744267)

julia> convert(HPolygon, H)
HPolygon{Float64,Array{Float64,1}}(HalfSpace{Float64,Array{Float64,1}}
[HalfSpace{Float64,Array{Float64,1}}([0.11082914558894325, -0.0010776789901761029], -0.9539738894744267)])
```
Is it correct that we can create unbounded polygons this way?

mforets:
yes. we can add a boundedness check

schillic:
> we can add a boundedness check

I am okay with both ways because one can create `HPolygon()`. Just wanted to raise this point.

schillic (approved):

src/convert.jl
```diff
@@ -45,9 +47,12 @@ A polygon in constraint representation.
 
 We compute the list of constraints of `X`, then instantiate the polygon.
 """
-function convert(::Type{HPOLYGON}, X::LazySet) where {HPOLYGON<:AbstractHPolygon}
+function convert(::Type{HPOLYGON}, X::LazySet; check_boundedness::Bool=true) where {HPOLYGON<:AbstractHPolygon}
     @assert dim(X) == 2 "set must be two-dimensional for conversion, but it is of dimension $(dim(X))"
     PT = basetype(HPOLYGON)
+    if check_boundedness && !isbounded(X)
```
schillic, line 53: This check does not work because it uses the type assumption of `HPolygon`:
```julia
julia> H = HPolygon();

julia> addconstraint!(H, rand(HalfSpace));

julia> isbounded(H)
true
```

mforets: true, but that is a different issue? (that the construction of HPolygon does not check that it is bounded) i mean to say that if you do `convert(HPolygon, X::HPolygon)` it works well -- in fact i don't want it to make the `isbounded` check if you know it's a polygon :)

mforets: i mean we could do `isbounded(HPolyhedron(constraints_list(X)))` , and then create the polygon, but i think it's too complicated / inefficient for this function is just supposed to take a list of constraints an create a polygon.

mforets: i'm also fine to just keep the method as i had it initially (without the boundedness check)

schillic:
```suggestion
    if check_boundedness && !isbounded(X, false)
```
The second argument is `use_type_assumption`.

mforets: yes,  i think that `use_type_assumption = true` should be used in this conversion function.

schillic: I wanted to say

> Then the check is useless because it always returns `true` for `HPolygon`.

But now I saw that you execute it for `X`. All fine then :+1:

schillic (approved):

PR #2520, opened 2021-01-28
#2519 - Concrete projection of singleton

### LINKED ISSUE #2519: Concrete projection of singleton
```julia
ulia> X = Singleton(rand(2))
Singleton{Float64,Array{Float64,1}}([0.5754235916161452, 0.8115498983587521])

julia> project(X, 1:2)
Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}([0.5754235916161452, 0.8115498983587521], [0.0, 0.0])
```
should return a singleton.

SebastianGuadalupe:
Closes #2519.

### DIFF
--- a/src/Sets/Singleton.jl
+++ b/src/Sets/Singleton.jl
@@ -133,3 +133,22 @@ The `Singleton` that corresponds to the rectification of `S`.
 function rectify(S::Singleton)
     return Singleton(rectify(element(S)))
 end
+
+"""
+    project(S::Singleton, block::AbstractVector{Int})
+
+Concrete projection of a singleton.
+
+### Input
+
+- `S`        -- set
+- `block`    -- block structure, a vector with the dimensions of interest
+
+### Output
+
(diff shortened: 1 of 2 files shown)


src/Sets/Singleton.jl
```diff
@@ -133,3 +133,22 @@ The `Singleton` that corresponds to the rectification of `S`.
...
+
+"""
+    project(S::Singleton, block::AbstractVector{Int})
+
+Concrete projection of a singleton.
+
+### Input
+
+- `S`        -- set
+- `block`    -- block structure, a vector with the dimensions of interest
```
schillic, line 145:
```suggestion
- `S`     -- singleton
- `block` -- block structure, a vector with the dimensions of interest
```

schillic (approved):
You could generalize it to `AbstractSingleton`, but since the only other such set is `ZeroSet`, which could have another more specific solution, this is not necessary.

mforets (approved):

Issue #2532, opened 2021-01-31
isempty returns false for an empty set (precision error?)


Wei-TianHao:
The behavior is inconsistent in different functions.

```julia
s1 = HalfSpace([-1.0], -1.0000000000000002)
s2 = HalfSpace([1.0], 1.0)
b = HPolytope([s1,s2])
println(isempty(b))
println(LazySets.diameter(b))
```
false

an empty set does not have a radius

Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] radius(::EmptySet{Float64}, ::Float64) at /home/wth/.julia/packages/LazySets/T9kE6/src/Sets/EmptySet.jl:252
 [3] radius at /home/wth/.julia/packages/LazySets/T9kE6/src/Interfaces/LazySet.jl:354 [inlined]
 [4] diameter at /home/wth/.julia/packages/LazySets/T9kE6/src/Interfaces/LazySet.jl:378 [inlined] (repeats 2 times)
 [5] top-level scope at In[58]:5
 [6] include_string(::Function, ::Module, ::String, ::String) at ./loading.jl:1091

schillic:
> precision error?

Yes, the LP solver returns a solution in the following line:

https://github.com/JuliaReach/LazySets.jl/blob/b3f63f271aeb32eeecda34109f8f8f51b6c7c77a/src/Sets/HPolyhedron.jl#L513

It seems that the solver works until a precision of 1e-7.

```julia
julia> s1 = HalfSpace([-1.0], -1.00000009);
julia> b = HPolytope([s1,s2]);
julia> isempty(b)
false

julia> s1 = HalfSpace([-1.0], -1.0000001);
julia> b = HPolytope([s1,s2]);
julia> isempty(b)
true
```

Not sure what we could do there.

mforets:
It is not particular of this package that

```julia
juli> 1.0 ≈ 1.0000000000000002
true
```
but rather the fact that the difference of those numbers is smaller than the machine epsilon for double-precision floating-point:

```julia
eps(Float64)
2.220446049250313e-16
```
So to tell those numbers apart you have to choose a numeric representation different that `Float64` in the first place, the choose an appropriate algorithm and if needed set the LazySets tolerances which are on purpose higher than the machine epsilon, eg. `LazySets._rtol(Float64)` is by default 1.4901161193847656e-8
 but you can change it to another value with `LazySets.set_rtol(Float64, val)`.

mforets:
[...]

PR #2535, opened 2021-02-01
Fix isempty with Polyhedra backend

mforets:
See #2532.

### DIFF
--- a/src/Sets/HPolyhedron.jl
+++ b/src/Sets/HPolyhedron.jl
@@ -443,7 +443,7 @@ end
 
 """
    isempty(P::HPoly{N}, witness::Bool=false;
-           [use_polyhedra_interface]::Bool=false, [solver]=default_lp_solver(N),
+           [use_polyhedra_interface]::Bool=false, [solver]=nothing,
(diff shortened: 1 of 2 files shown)


src/Sets/HPolyhedron.jl
```diff
@@ -483,7 +483,7 @@ Otherwise, we set up the LP internally.
 function isempty(P::HPoly{N},
                  witness::Bool=false;
                  use_polyhedra_interface::Bool=false,
-                 solver=default_lp_solver(N),
+                 solver=nothing,
```
schillic, line 486: I guess the docs need to be adapted now.

schillic (approved):
Can you merge/rebase with master? I expect that this fixes the build error.

src/Sets/HPolyhedron.jl
mforets: done. on the other hand, having `(optional, default: `nothing`)` is sort of meaningless so i added a comment in that line

mforets:
the PR was approved and the build error in windows is unrelated ... merging !

PR #2545, opened 2021-02-04
Update Project.toml

mforets:


schillic:
Why the minor update? There was no breaking change.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "LazySets"
 uuid = "b4f0291d-fe17-52bc-9479-3d1a343d9043"
-version = "1.38.3"
+version = "1.39.0"
```
mforets, line 3:
```suggestion
version = "1.38.3"
```

schillic (approved):

mforets:
if we are strict to https://semver.org/ -- which we don't have to -- minor releases add new functionality (in a backwards compatible manner), while patch releases if for bug fixes. imo it makes sense to advance the minor release when there are "substantial" new features (even if none of them breaking previous code).

mforets:
probably that's what i had in mind -- in the series of v1.38.x releases we have substantially improved the way projections and some special types of cartesian products are handled (those that involve a 1 d set and a n dim set) with new methods depending on the n dim set.

schillic:
Here is what I remember about the policy we used so far:
* I would stick to the good practices of other packages. I think almost no package makes a _major_ release with each little _breaking change_. Instead they reserve _major_ versions for literally major changes in how things work.
* The good and bad thing in Julia is that you need upper bounds on dependencies. If `LazySets` had a new _minor_ release for each new _feature_, you would need to update your `Project` file quite often just to use these features (and of course you want to use them). I would be annoyed if the dependencies that we rely on required us to update their version (even if `CompatHelper` can do that automatically, but sometimes there are problems and it still takes build time with each PR) every week.
* For _breaking changes_ it makes sense to increment the _minor_ version at least so that you have to actively agree to use the newest version.

In the end it is just a number.

mforets:
are you sure you're not confusing with the behavior for packages with leading zeros (see docs [here](https://julialang.github.io/Pkg.jl/v1/compatibility/#compat-pre-1.0)), caret specifiers (default) with tilde specifiers? 
 (see docs [here](https://julialang.github.io/Pkg.jl/v1/compatibility/#Version-specifier-format)).

g. if a project has a `LazySets = "1.39"`, it is compatible with all versions `[1.39.0, 2.0.0)`.

mforets:
for example instead of https://github.com/JuliaReach/LazySets.jl/blob/master/Project.toml#L38 we can put `Plots = "1"`

Issue #2546, opened 2021-02-04
Uniform sampling from a convex polytope


schillic:
The only alternative to rejection sampling, which is slow for flat sets, [seems to be](https://math.stackexchange.com/questions/3137805/uniform-sampling-of-convex-polytopes-why-is-it-hard) triangulation (because it is known how to sample uniformly from triangles).

Issue #2536, opened 2021-02-01
very slow sampling from a 3-dimensional HPolytope (Julia 1.5.2)


ko56:
I have the following 3-dimensional polytope:

julia> GXent.GPolyTope
HPolytope{Float64,Array{Float64,1}}(HalfSpace{Float64,Array{Float64,1}}[HalfSpace{Float64,Array{Float64,1}}([1.0, 1.0, 0.0], 4.0), HalfSpace{Float64,Array{Float64,1}}([-1.0, -1.0, -0.0], -4.0), HalfSpace{Float64,Array{Float64,1}}([-1.0, 0.0, 0.0], -0.0), HalfSpace{Float64,Array{Float64,1}}([0.0, -1.0, 0.0], -0.0), HalfSpace{Float64,Array{Float64,1}}([0.0, 0.0, -1.0], -0.0), HalfSpace{Float64,Array{Float64,1}}([1.0, 0.0, 0.0], 10.0), HalfSpace{Float64,Array{Float64,1}}([0.0, 1.0, 0.0], 10.0), HalfSpace{Float64,Array{Float64,1}}([0.0, 0.0, 1.0], 10.0), HalfSpace{Float64,Array{Float64,1}}([0.0, 1.0, 1.0], 6.0)])

On a reasonably fast machine I see

julia> @time sample(GXent.GPolyTope)
  5.632507 seconds (30.91 M allocations: 3.224 GiB, 7.10% gc time)
3-element Array{Float64,1}:
 0.48320334527456144
 3.5167966552486147
 1.6576822969795617

julia> 

For comparison, sampling from a random 3-dimensional polytope is almost instantaneous.
Any help would be appreciated.

mforets:
Hi @ko56,

I think the underlying issue here is that the set is flat in some dimensions. I don't know if you are familiar with the fact that the default sampling, `sample(X, npoints)`, uses a method called [rejection sampling](https://en.wikipedia.org/wiki/Rejection_sampling). It is for that reason that random sampling from a flat set may take long time.

If you work only in three dimensions, it shouldn't be too hard to have a specialized method that checks if the set is flat on 2d projections and use a specialized method.

ko56:
Hi Marcelo,

Thanks for replying.  By "flat" do you mean that it has the x-, y-, and z-planes
as facets (first orthant)?
Yes, I know that the default sampler does rejection sampling, using a bounding
hyperrectangle.
[...]

PR #2547, opened 2021-02-04
Sampling from a convex polytope

schillic:
See https://github.com/JuliaReach/LazySets.jl/issues/2536#issuecomment-771679414.

Uniform sampling from a convex polytope turns out to be much more difficult (#2546).

### DIFF
--- a/src/Utils/samples.jl
+++ b/src/Utils/samples.jl
@@ -105,7 +105,6 @@ struct RejectionSampler{S<:LazySet, D} <: Sampler
     box_approx::Vector{D}
 end
 
-
 function RejectionSampler(X, distribution=DefaultUniform)
     B = box_approximation(X)
     box_approx = [distribution(low(B, i), high(B, i)) for i in 1:dim(B)]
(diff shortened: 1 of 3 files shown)


src/Utils/samples.jl
```diff
@@ -209,6 +235,31 @@ function _sample!(D::Vector{VN},
...
+                  rng::AbstractRNG=GLOBAL_RNG,
+                  seed::Union{Int, Nothing}=nothing
+                 ) where {N, VN<:AbstractVector{N}}
+    rng = reseed(rng, seed)
+    U = DefaultUniform(zero(N), one(N))
+    P = set(sampler)
+    vlist = vertices_list(P)
+    m = length(vlist)
+
+    @inbounds for i in 1:length(D)
```
mforets, line 249: The call to `rand` is not using `U` and the given `rng`. I propose the following loop. I made some tests and the runtime does not improve, but it allocate a factor of ~2x less because of the in-place updates of the random number generator. That is good because it handles better the case when the number of samples is very large.

I defined two new functions that dispatch on the uniform sampler; the first one is not used (but it is useful in general, let's add it), and the second one is the in-place version.

```julia
function Base.rand(rng::AbstractRNG, U::DefaultUniform, n::Int)
    [rand(rng, U) for i in 1:n]
end

function rand!(x, rng::AbstractRNG, U::DefaultUniform)
    @inbounds for i in eachindex(x)
        x[i] = rand(rng, U)
    end
end

function _sample!(....) 
   . . . .

    # combination coefficients
    r = Vector{N}(undef, m-1)

    @inbounds for i in 1:length(D)

        # get a list of m uniform numbers (https://cs.stackexchange.com/a/3229)
        # and compute the corresponding linear combination in-place
        rand!(r, rng, U)
        sort!(r)
        D[i] = r[1] * vlist[1]
        for j in 2:m-1
            α = r[j] - r[j-1]
            D[i] .+= α * vlist[j]
        end
        D[i] .+= (1 - r[m-1]) * vlist[m]
    end
    return D
end
```

schillic: Thanks, I added your suggestions.

mforets (approved):

PR #2549, opened 2021-02-08
Update Project.toml

mforets:


schillic (approved):
I am not sure this is what we want but I agree that it is equivalent :)

mforets:
yes, i removed the minor version annotations because it is equivalent, also i did this to double check my comments in https://github.com/JuliaReach/LazySets.jl/pull/2545#issuecomment-773989798

PR #2552, opened 2021-02-08
Better error message in minkowski_sum of V-reps

schillic:
In `master`:
```julia
julia> Z = minkowski_sum(X, Y)
ERROR: UndefVarError: default_polyhedra_backend_nd not defined
```
This branch:
```julia
julia> Z = minkowski_sum(X, Y)
ERROR: AssertionError: package 'Polyhedra' not loaded (it is required for executing `minkowski_sum`)
```
I am not 100% sure about the error message but I think it is better to have that.


mforets (approved):

mforets:
> I am not 100% sure about the error message but I think it is better to have that.

it may be required in 2 steps: (i) compute the vertices of X and Y, (ii) compute the convex hull of the result of X + Y

PR #2554, opened 2021-02-08
Implement Star set

mforets:
Implement the new set representation from https://github.com/JuliaReach/LazySets.jl/pull/2132; other methods to be added in subsequent PRs.

joint with @SebastianGuadalupe

PR #2557, opened 2021-02-08
#2556 - Support function and support vector of a star set

### LINKED ISSUE #2556: Support function and support vector of star
see [this comment](https://github.com/JuliaReach/LazySets.jl/pull/2132#issuecomment-736831208)

SebastianGuadalupe:
Requires ~~#2554.~~
Closes #2556.

### DIFF
--- a/src/Sets/Star.jl
+++ b/src/Sets/Star.jl
@@ -194,4 +194,18 @@ dim(X::Star) = length(X.c)
 # Support function and support vector
 # =====================================
 
-# TODO
+"""
+    ρ(ℓ, X::Star)
+Evaluate the support function of a star in a given direction.
+### Input
+- `ℓ` -- direction
+- `X` -- star set
+### Output
+The support function in the given direction.
+"""
+function ρ(ℓ, X::Star)
+    c = center(X)
+    V = basis(X)
+    P = predicate(X)
+    return dot(ℓ, c) + ρ(transpose(V) * ℓ, P)
+end
--- a/src/Sets/Star.jl
+++ b/src/Sets/Star.jl
@@ -214,3 +214,24 @@ function ρ(ℓ::AbstractVector, X::Star)
     P = predicate(X)
     return dot(ℓ, c) + ρ(transpose(V) * ℓ, P)
 end
+
+"""
+    σ(ℓ::AbstractVector, X::Star)
+
+Evaluate the support vector of a star in a given direction.
+
+### Input
+
+- `ℓ` -- direction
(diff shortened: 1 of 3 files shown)


src/Sets/Star.jl
```diff
@@ -0,0 +1,229 @@
...
+- `X` -- star set
+
+### Output
+
+The support function in the given direction.
+"""
+function ρ(ℓ, X::Star)
+    c = center(X)
+    P = predicate(X)
+    return dot(ℓ, c) + ρ(ℓ, P)
```
mforets, line 228: this should be: `dot(ℓ, c) + ρ(transpose(V) * ℓ, P)`

schillic:
The PR looks good, but I have a general concern about this `Star` type. From this PR I now understand that a `Star` as defined here is equivalent to an `AffineMap` of a polyhedron (which itself is equivalent to a polyhedron). So I would let it subtype `AbstractPolyhedron` instead. I am also not sure about just implementing the methods by creating an `AffineMap`. Can we even let `Star` just be an alias of `AffineMap`? Is there a reason for having this type? We can use an alias for the constructor, for printing, and for dispatch.

mforets:
> I am also not sure about just implementing the methods by creating an AffineMap. Can we even let Star just be an alias of AffineMap?

sounds good. 

>  Is there a reason for having this type?

this notion is presented in the references with the predicate being an arbitrary function (thus the "generalized star"), and the polyhedra predicate just a special case -- which turns out to be the implementation used in practice. i think it is relevant to have this type with docs and everything, but for the (polyhedral) star, using an alias seems like a good idea to not replicate the code.

schillic:
> So I would let it subtype `AbstractPolyhedron` instead.

`AbstractAffineMap` would also be a candidate. I would prefer to have several supertypes :sob:
But if we drop it anyway...

> i think it is relevant to have this type with docs and everything

You mean `AbstractStar`, right? That is fine (although I am not sure where we will use it :stuck_out_tongue:).

mforets:
i meant the user doing `?Star` gives the docstring now in `master` -- ultimately the `AffineMap` is an implementation detail.

mforets:
> You mean AbstractStar, right? That is fine (although I am not sure where we will use it stuck_out_tongue).

yeah, not for the new `Star`.. but i'd like to add a GeneralizedStar type which does not assume that the predicate is polyhedral.

mforets:
closing in favor of https://github.com/JuliaReach/LazySets.jl/pull/2578

PR #2548, opened 2021-02-05
WIP #2543 - Hit and run sampling of lazy sets

mforets:
Closes #2543.

Implements the hit and run algorithm. Still WIP (reasons marked with FIXME in the code) yet it works, see notebooks [1] and [2].

[1] https://github.com/mforets/escritoire/blob/master/2021/Week5/HitAndRun.ipynb
[2] https://github.com/mforets/escritoire/blob/master/2021/Week5/HitAndRun_Implementation.ipynb

Screenshot from 2021-02-05 02-51-33

PR #2563, opened 2021-02-09
Alternative polytope sampler

schillic:
Alternative to #2547 that seems more biased toward the edges. Related to #2536.

The idea is to start from the Chebyshev center and then walk toward random vertices. Crucially, some vertices can be skipped, which has the effect that the samples are not biased toward the center.

TODOs:
* [x] ~find a better name or~ merge with the existing `PolytopeSampler` via an option
* [x] docs
* [x] test

---

```julia
julia> using LazySets, Polyhedra, Plots
julia> X = rand(VPolygon);
julia> @time V = sample(X, 1000, sampler=LazySets.PolytopeSampler);
  0.000688 seconds (8.01 k allocations: 758.172 KiB)
julia> @time V2 = sample(X, 1000, sampler=LazySets.PolytopeSampler2);  # sampler from this branch
  1.107326 seconds (1.30 M allocations: 67.599 MiB)

julia> plot(X)
julia> scatter!([v[1] for v in V], [v[2] for v in V])
julia> scatter!([v[1] for v in V2], [v[2] for v in V2])
```

sampling

### DIFF
--- a/src/Utils/samples.jl
+++ b/src/Utils/samples.jl
@@ -118,7 +118,7 @@ end
 # ===================
 
 """
-    RejectionSampler{D} <: Sampler
+    RejectionSampler{D} <: AbstractSampler
 
 Type used for rejection sampling of an arbitrary set `X`.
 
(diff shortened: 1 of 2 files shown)


schillic:
Some new insights, visualized for the following polygon (sampling with the original version in this branch):



- First I tried to not skip vertices. The results are not good.



- Then I tried to permute the order of the vertices. In a `VPolygon` the vertices are sorted and hence the walk is not really random. Looks much better.



- With this new version I then tried again to not skip vertices. Basically no difference, so for the sake of simplicity I dropped the skipping.



- Finally I tried to not start from the Chebyshev center (which is relatively expensive) and instead start from some vertex. First I tried to start from the result of `an_element` (here: the vertex on the left, marked in green) and then I tried to start from a random vertex every time. Since the results are more or less the same, I decided to use the latter version for efficiency.




Finally here is an example of a line segment to check sampling from a flat set.

schillic:
I just tried to generalize the random walk by choosing `k` random vertices, so

- you do not have to walk to each vertex and
- you can walk to a vertex several times.

I tried with varying `k` and that does not seem to affect the quality of the sampling at all (`k = 5` and `k = 5000` look similar, see the plots below). So I decided to not do this change.

schillic:
I would go for the option to merge this approach with the existing `PolytopeSampler`. But for that we need a refactoring such that `sample` takes a sampler instance and not its type. I believe this was supposed to happen in #2548.

mforets:
> I believe this was supposed to happen in #2548.

:+1: i'll send it in a separate PR though

mforets (approved):

PR #2564, opened 2021-02-09
#674 - Remove redundant vertices from linear_map of V-reps

### LINKED ISSUE #674: Remove duplicate vertices from linear_map of V-reps
In #670 we added a `vertices_list` of `LinearMap` and `ExponentialMap`.
However, they can map previously distinct vertices to the same vertex, which results in duplicates.
Note that this is the case for both the concrete and the lazy map:

```julia
julia> A = [1. 0; 0 0]; X = BallInf([0., 0], 1.); L = A * X
LazySets.LinearMap{Float64,LazySets.BallInf{Float64},Float64,Array{Float64,2}}(
  [1.0 0.0; 0.0 0.0], LazySets.BallInf{Float64}([0.0, 0.0], 1.0))

julia> linear_map(A, X)
LazySets.VPolytope{Float64}(Array{Float64,1}[
  [1.0, 0.0], [-1.0, 0.0], [1.0, 0.0], [-1.0, 0.0]])

julia> vertices_list(linear_map(A, X))
4-element Array{Array{Float64,1},1}: [1.0, 0.0] [-1.0, 0.0] [1.0, 0.0] [-1.0, 0.0]
```

schillic:
Closes #674.

### DIFF
--- a/src/Interfaces/AbstractPolytope.jl
+++ b/src/Interfaces/AbstractPolytope.jl
@@ -112,7 +112,7 @@ end
 # given a polytope P, apply the linear map P to each vertex of P
 # it is assumed that the interface function `vertices_list(P)` is available
 @inline function _linear_map_vrep(M::AbstractMatrix, P::AbstractPolytope)
-    vertices = broadcast(v -> M * v, vertices_list(P))
+    vertices = broadcast(v -> M * v, vertices_list(P)) |> unique
     m = size(M, 1) # output dimension
     if m == 1
         # TODO: substitute with Interval(convex_hull(vertices)...): convex hull 1D
--- a/src/Interfaces/AbstractPolytope.jl
+++ b/src/Interfaces/AbstractPolytope.jl
@@ -111,16 +111,21 @@ end
 
 # given a polytope P, apply the linear map P to each vertex of P
 # it is assumed that the interface function `vertices_list(P)` is available
-@inline function _linear_map_vrep(M::AbstractMatrix, P::AbstractPolytope)
-    vertices = broadcast(v -> M * v, vertices_list(P)) |> unique
+@inline function _linear_map_vrep(M::AbstractMatrix, P::AbstractPolytope;
+                                  apply_convex_hull::Bool=false)
+    vlist = broadcast(v -> M * v, vertices_list(P))
+
     m = size(M, 1) # output dimension
     if m == 1
-        # TODO: substitute with Interval(convex_hull(vertices)...): convex hull 1D
-        return Interval(minimum(vertices)[1], maximum(vertices)[1])
+        # TODO: substitute with Interval(convex_hull(vlist)...): convex hull 1D
+        return Interval(minimum(vlist)[1], maximum(vlist)[1])
     elseif m == 2
-        return VPolygon(vertices)
+        return VPolygon(vlist)
     else
-        return VPolytope(vertices)
+        if apply_convex_hull
+            convex_hull!(vlist)
+        end
+        return VPolytope(vlist)
     end
 end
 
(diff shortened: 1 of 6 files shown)


src/Sets/VPolygon.jl
```diff
@@ -155,7 +155,7 @@ function linear_map(M::AbstractMatrix, P::VPolygon)
 end
 
 @inline function _linear_map_vrep(M::AbstractMatrix, P::VPolygon)
-    return broadcast(v -> M * v, vertices_list(P)) |> VPolygon
+    return broadcast(v -> M * v, vertices_list(P)) |> unique |> VPolygon
```
mforets, line 158: `... |> VPolygon` already runs a convex hull: this change is not needed afaict

src/Interfaces/AbstractPolytope.jl
```diff
@@ -112,7 +112,7 @@ end
 # given a polytope P, apply the linear map P to each vertex of P
 # it is assumed that the interface function `vertices_list(P)` is available
 @inline function _linear_map_vrep(M::AbstractMatrix, P::AbstractPolytope)
-    vertices = broadcast(v -> M * v, vertices_list(P))
+    vertices = broadcast(v -> M * v, vertices_list(P)) |> unique
```
mforets, line 115: i would rather optionally apply a convex hull on the function that calls this one, because:

- making it `false` by default, because in high dims we would need to run a possibly expensive convex hull
- it is more robust than `unique`
- the lower level `_linear_map_vrep` does one simple task (v -> Mv) keeping the simplification algorithm in a separate step

mforets (changes requested):

src/Interfaces/AbstractPolytope.jl
schillic: Okay about `convex_hull` instead of `unique`, but:
* Given the other comment about `VPolygon` I would pass the option to `_linear_map_vrep` because in the `VPolygon` version we need to pass that option to the constructor.
* We do a `convex_hull` call in `vertices_list` for the lazy `LinearMap` by default as well. Is it okay to have this different behavior?

src/Sets/VPolygon.jl
```diff
@@ -149,12 +149,15 @@ function remove_redundant_vertices(P::VPolygon;
     return remove_redundant_vertices!(copy(P), algorithm=algorithm)
 end
 
-function linear_map(M::AbstractMatrix, P::VPolygon)
+function linear_map(M::AbstractMatrix, P::VPolygon; convex_hull::Bool=false)
```
mforets, line 152: we have `prune`, `apply_convex_hull`, and now `convex_hull`  boolean options :smile:

src/Interfaces/AbstractPolytope.jl
mforets: > Given the other comment about VPolygon I would pass the option to _linear_map_vrep because in the VPolygon version we do not want to apply a convex hull again even if it is set to true.

ok, i see :+1: 

> We do a convex_hull call in vertices_list for the lazy LinearMap by default as well. Is it okay to have this different behavior?

i see, you mean `vertices_list(lm::LinearMap; prune::Bool=true)`. yeah i don't know, maybe that one should also be set to false..

src/Sets/VPolygon.jl
schillic: Oops, you are right. Which one do you prefer?

mforets: makes sense :+1:

mforets (approved):

schillic:
The build error looks unrelated (I opened #2568). @mforets Should I merge or wait?

mforets:
:+1:  let's merge this one...

i was aware of the new issue with MT , but haven't got time to get around it.

PR #2573, opened 2021-02-15
Use fallback _linear_map_vrep for VPolygon

### LINKED ISSUE #2567: linear_map applied to VPolygon returns a VPolygon even for non-square matrix
The problem is that we do not check the output dimension. But if the matrix is not square, the result is not 2D but we still return a `VPolygon`.
```julia
julia> A = rand(3, 2);
julia> P = VPolygon([[1.0, 1.0]])
julia> linear_map(A, P)
VPolygon{Float64,Array{Float64,1}}([[0.27955702718087605, 1.4851617424414982, 1.1596000041440733]])
```
https://github.com/JuliaReach/LazySets.jl/blob/8ce8680220ff815ec1cc56f77fe844eede33e0a8/src/Sets/VPolygon.jl#L152-L159

schillic:
Closes #2567.

### DIFF
--- a/src/Sets/VPolygon.jl
+++ b/src/Sets/VPolygon.jl
@@ -151,14 +151,7 @@ end
 
 function linear_map(M::AbstractMatrix, P::VPolygon; apply_convex_hull::Bool=false)
     @assert size(M, 2) == 2 "a linear map of size $(size(M)) cannot be applied to a set of dimension 2"
-    return _linear_map_vrep(M, P)
-end
-
-@inline function _linear_map_vrep(M::AbstractMatrix, P::VPolygon;
-                                  apply_convex_hull::Bool=false)
-    # note: `apply_convex_hull` is ignored because the VPolygon constructor
-    # always runs a convex-hull algorithm
-    return broadcast(v -> M * v, vertices_list(P)) |> VPolygon
+    return _linear_map_vrep(M, P, apply_convex_hull=apply_convex_hull)
 end
 
 # --- AbstractPolygon interface functions ---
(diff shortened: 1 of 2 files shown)


mforets:
there's a problem with removing the method, that the the output value can't longer be inferred by the compiler, try:

```julia
using LazySets, Test

julia> x = rand(VPolygon); m = rand(2, 2);

julia> linear_map(m, x)
VPolygon{Float64,Array{Float64,1}}([[-0.9954043596516247, 0.11053395161158013], [-0.06529937425556093, -0.009926410681658025], [1.1488421344198256, 0.55795301

53930504], [1.8984407274784327, 1.0124663945874632], [1.8115487950782836, 1.1123403187818237], [0.7741627387395416, 1.1711735235573724], [0.41589246467767604,
 0.9835706581210915]])

julia> @inferred linear_map(m, x)
ERROR: return type VPolygon{Float64,Array{Float64,1}} does not match inferred return type Union{Interval{Float64,IntervalArithmetic.Interval{Float64}}, VPolyg
on{Float64,Array{Float64,1}}, VPolytope{Float64,Array{Float64,1}}}
Stacktrace:
 [2] top-level scope at REPL[4]:1
```

i think we should merge the change in this PR, but discuss how to provide methods that are type stable to use in algorithms (eg. [here](https://github.com/JuliaReach/ReachabilityAnalysis.jl/blob/master/src/Algorithms/VREP/reach_homog.jl#L19))

schillic:
> the output value can't longer be inferred by the compiler

Yes, that is not possible because the output type now depends (and should depend) on the matrix dimension, which is normally not known from the type. This is also the behavior for other set types if you use the `"vrep"` algorithm. It should be type stable for static arrays, though (not checked).
The obvious alternative is to do what the `VPolytope` method does: always return a `VPolytope`.

> discuss how to provide methods that are type stable

It is not always possible to do that (for example `intersection`). In this case we could add a function `_linear_map_square_vrep`.

mforets (approved):

mforets:
indeed, i think that implementing internal (non-exported) methods that are type-stable is always a good idea.

PR #2577, opened 2021-02-15
#2187 - Sort constraints in HPolygon intersection

### LINKED ISSUE #2187: Error in concrete intersection of HPolygon

schillic:
Closes #2187.

The problem (see #2187's OP) is that `intersection` of two `HPolygon`s is supposed to combine the constraints in a sorted order but it does not. This PR proposes a quick fix: sort them in the end.
Ultimately the code should be fixed, but since it is complicated and the issue has been open for several months, I think we should first fix the bug and then propose an efficient fix later.

### DIFF
--- a/src/ConcreteOperations/intersection.jl
+++ b/src/ConcreteOperations/intersection.jl
@@ -586,7 +586,9 @@ function intersection(P1::AbstractHPolygon{N},
         deleteat!(c, length(c)-duplicates+1:length(c))
     end
 
-    P = HPolygon(c, sort_constraints=false)
+    # TODO see #2187: the above code does *not* sort the constraints correctly.
+    # after fixing the above code, we should pass sort_constraints=false again
+    P = HPolygon(c, sort_constraints=true)
     if prune
         remove_redundant_constraints!(P)
         if isempty(P)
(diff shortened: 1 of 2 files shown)


mforets (approved):

mforets:
about [this comment](https://github.com/JuliaReach/LazySets.jl/issues/2187#issuecomment-653788215) we should open a new issue (?) because sometimes it still gives an error.

Issue #1209, opened 2019-03-07
Add concrete projection of Cartesian sets


schillic:
The `project` method in the `Approximations` module uses a lazy `LinearMap` to project to suitable dimensions. For Cartesian sets, this is not necessary and the concrete projection can be easily constructed.
This issue proposes to add a new option for performing the projection in a concrete way if possible.

The detection of the property "Cartesian" is generally not easy. Below is a list of types where this is feasible with low efforts.

The following set types are Cartesian in any block structure by definition:
* [x] `AbstractHyperrectangle` (#2272)
* [x] `EmptySet`/`Universe` (#2476)

The following set types are Cartesian in a specific block structure by definition:
* [x] `CartesianProduct` (#2597)
* [ ] `CartesianProductArray` (#2598)

The following set types are not Cartesian but the projection is as simple as for `AbstractHyperrectangle` (in any block structure):
* [x] `BallX` for `X ∈ {1, 2, p}` (#2476)

The following set types can sometimes be projected easily if all dimensions that are projected out (or their complement) are unconstrained:
* [x] `HalfSpace` and `HPolyhedron`/`HPolytope` (#1213, #1957, #2196, #2478)
* [x] `Hyperplane` and `Line2D` (#2596)
* [ ] `Line` (#2598)

PR #2597, opened 2021-02-26
#1209 - Concrete projection of CartesianProduct

schillic:
See #1209.

### DIFF
--- a/src/LazyOperations/CartesianProduct.jl
+++ b/src/LazyOperations/CartesianProduct.jl
@@ -298,7 +298,7 @@ end
 function project(cp::CartesianProduct, block::AbstractVector{Int})
     res = _project_cp_same_block(cp, block)
     if res == nothing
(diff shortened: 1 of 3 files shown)


src/LazyOperations/CartesianProduct.jl
```diff
@@ -295,6 +295,26 @@ function concretize(cp::CartesianProduct)
     return cartesian_product(concretize(cp.X), concretize(cp.Y))
 end
 
+function project(cp::CartesianProduct, block::AbstractVector{Int})
+    res = _project_cp_same_block(cp, block)
+    if res == nothing
+        res = invoke(project, Tuple{LazySet, typeof(block)}, cp, block)
```
mforets, line 301: use an auxilary function?  `invoke`  makes it less clear the function that is actually used (it may have an impact in performance too?)

mforets (approved):
lgtm

mforets (approved):

PR #2598, opened 2021-02-27
#1209 - Concrete projection of Line & CartesianProductArray (and modification of CartesianProduct)

### LINKED ISSUE #1209: Add concrete projection of Cartesian sets

schillic:
Closes #1209.

### DIFF
--- a/src/LazyOperations/CartesianProduct.jl
+++ b/src/LazyOperations/CartesianProduct.jl
@@ -296,22 +296,21 @@ function concretize(cp::CartesianProduct)
 end
 
 function project(cp::CartesianProduct, block::AbstractVector{Int})
-    res = _project_cp_same_block(cp, block)
-    if res == nothing
-        res = _project_linear_map(cp, block)
-    end
-    return res
-end
-
-function _project_cp_same_block(cp, block)
-    min, max = extrema(block)
     n1 = dim(cp.X)
-    if max <= n1
+    if block[end] <= n1
+        # projection completely in the first block
         return project(cp.X, block)
-    elseif min > n1
+    elseif block[1] > n1
+        # projection completely in the second block
         return project(cp.Y, block .- n1)
-    else
-        return nothing
+    end
+    # projection is a new Cartesian product of the block-wise projections
+    for (i, bi) in enumerate(block)
+        if bi > n1
+            X = project(cp.X, block[1:i-1])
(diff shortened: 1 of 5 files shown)


src/LazyOperations/CartesianProduct.jl
```diff
@@ -296,22 +296,21 @@ function concretize(cp::CartesianProduct)
...
-        res = _project_linear_map(cp, block)
-    end
-    return res
-end
-
-function _project_cp_same_block(cp, block)
-    min, max = extrema(block)
     n1 = dim(cp.X)
-    if max <= n1
+    if block[end] <= n1
```
mforets, line 300: i dont know if we have here or in a more global description of the docs that `block` should be increasing integers.

mforets (approved):

src/LazyOperations/CartesianProduct.jl
schillic: True. In many other places we use `v[block]` for vector `v`, which gives wrong results if you break this assumption. For example:

```julia
julia> project(Singleton([1, 2, 3]), [3, 1])
Singleton{Int64,Array{Int64,1}}([3, 1])
```

PR #2599, opened 2021-02-27
#2047 - add helper function 'allequal'

### LINKED ISSUE #2047: Refactor same_constraints
move `same_constraints` (from `unit_Polygon.jl`) to `AbstractHPolygon.jl` or `HalfSpace.jl`. also, it uses `N<:Real` but doesn't make use of a tolerance.

schillic:
Closes #2047.

### DIFF
--- a/src/Arrays/vector_operations.jl
+++ b/src/Arrays/vector_operations.jl
@@ -378,3 +378,30 @@ A scalar representing ``‖ x - y ‖_p``.
 function distance(x::AbstractVector, y::AbstractVector, p::Real=2.0)
     return norm(x - y, p)
 end
+
+"""
+    allequal(x)
+
+Check whether all elements in a sequence are equal
+
+### Input
+
+- `x` -- sequence
+
+### Output
+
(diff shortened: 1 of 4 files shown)


src/Arrays/vector_operations.jl
```diff
@@ -378,3 +378,30 @@ A scalar representing ``‖ x - y ‖_p``.
...
+
+### Output
+
+`true` iff all elements in `x` are equal.
+
+### Notes
+
+The code is taken from [here](https://stackoverflow.com/a/47578613).
+"""
+function allequal(x)
```
mforets, line 399: `all(==(first(x)), x)` works too

mforets (approved):

src/Arrays/vector_operations.jl
schillic: Yes, in the link there was a benchmark and that was slower I think.

schillic: Yes, in the link there was a benchmark and that was slower I think.

PR #2578, opened 2021-02-15
Change Star constructor to be an alias for a specific AffineMap

mforets:
Closes #2556.

PR #2605, opened 2021-02-28
#1280 - Rotated hyperrectangle

### LINKED ISSUE #1280: Rotated hyperrectangle / orthogonal polytope
A rotated hyperrectangle is a normal hyperrectangle after applying a linear map (an alternative name is "orthogonal polytope"). It is thus a special case of a parallelotope or a zonotope (which is an affine map applied to a high-dimensional unit square) of order 1. The benefit over axis-aligned hyperrectangles is that the rotated versions are closed under linear maps. The benefit over zonotopes is that the support function is more efficient to evaluate.

It is probably easiest and most efficient to represent this type by storing the axis-aligned hyperrectangle and the matrix.

Maybe(!) a linear map of a hyperrectangle is equivalent to an affine map of the unit square. If so, an alternative is to store only the affine map, which might be more efficient in computations.

schillic:
Closes #1280.

* [x] ~This implementation is not closed under translation. Maybe add another field `offset`?~
* [x] tests
* [x] docs

The name `RotatedHyperrectangle` may be misleading. The matrix does not have to be a rotation matrix (not even invertible). I am open to suggestions.


mforets:
> The name RotatedHyperrectangle may be misleading. The matrix does not have to be a rotation matrix (not even invertible). I am open to suggestions.

The set is a special case of a lazy `LinearMap`. It a special case of a `Star`. In https://github.com/JuliaReach/LazySets.jl/pull/2554 we implemented the new star set type and in https://github.com/JuliaReach/LazySets.jl/pull/2578  we chose to use alias for the lazy type instead of defining a new type. Perhaps we should do the same with the `RotatedHyperrectangle`.

schillic:
I had thought about this as well. The benefit of having it as a separate type is the functionality that comes with inheriting from `AbstractZonotope`. And it can be used in all methods that take an `AbstractZonotope` as input (for example in reachability algorithms).

mforets (approved):

PR #2607, opened 2021-03-01
Accept kwargs in 'project' and pass to 'linear_map'

schillic:
Closes #2301 (I copied the idea from there and extended it to the current code base).

### DIFF
--- a/src/Approximations/template_directions.jl
+++ b/src/Approximations/template_directions.jl
@@ -102,7 +102,8 @@ isnormalized(::AD) where {AD<:AbstractDirections} = isnormalized(AD)
     project(S::LazySet,
             block::AbstractVector{Int},
             directions::Type{<:AbstractDirections},
-            [n]::Int
+            [n]::Int;
+            [kwargs...]
            )
 
 Project a high-dimensional set to a given block using template directions.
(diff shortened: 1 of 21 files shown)


mforets:
not sure about the approach of passing `kwargs...` to a function that actually ignores them. i think that smells like a design problem.

mforets:
here are two other approaches (which are not necessarily orthogonal):

- restrict the scope of `project` by simplifying the API and adding new functions with specialized use cases, eg. `project_template`, ..

- dispatch on "projection algorithms", that is `ConcreteProjection`, `TemplateProjection`, and such.

mforets:
in fact can we do:

```julia
function project(X::LazySet, block::AbstractVector{Int}; kwargs...)
    _project(X, block; kwargs...)
end

# we don't pass kwargs.. to the actual implementation
function _project(B::Ball1, block::AbstractVector{Int})
    return Ball1(B.center[block], B.radius)
end
```

schillic:
No, you cannot do that. If you pass kwargs, then dispatch will search for a method that takes them.

mforets:
ok. then i would suggest to merge this branch, because it adds a new feature (passing backends to the concrete projection), and review again when we have time.

>  i think that smells like a design problem.

one reason is that passing `kwargs...` is against clarity: not sure why/what does the function actually process by looking at the signature..

mforets (approved):

mforets:
of course there are cases in which passing `kwargs...` is extremely useful, but i'd expect that those are general/API functions and in most cases we pass a fixed number of args

schillic:
The intention of this PR was to close the other PR that started this work. I agree with you that it seems strange to pass kwargs everywhere. But I think you need to do this if you want full control (or set some global variables).

> review again when we have time

Can you open an issue, please? I am not sure what to write there.

PR #2608, opened 2021-03-01
#1926 - Add isboundedtype function

### LINKED ISSUE #1926: Add isboundedtype

schillic:
Closes #1926.
Closes #1979 (I copied the code for `isboundedtype(::Type{<:LazySet})` from there and extended it to the current code base).

### DIFF
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -12,6 +12,7 @@ export LazySet,
        diameter,
        an_element,
        isbounded,
+       isboundedtype,
        neutral,
        absorbing,
        tosimplehrep,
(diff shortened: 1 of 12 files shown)


src/Interfaces/LazySet.jl
```diff
@@ -248,6 +249,30 @@ Alias for the support vector σ.
...
+
+### Input
+
+- `LazySet` -- set type for dispatch
+
+### Output
+
+`true` if the set type only represents bounded sets.
+Note that some sets may still represent an unbounded set even though their type
+actually does not (example: [`HPolytope`](@ref)).
```
mforets, line 265:
```suggestion
actually does not (example: [`HPolytope`](@ref), because the constructor doesn't assert boundedness by default).
```

mforets (approved):

src/Interfaces/LazySet.jl
schillic: I will write `, because the construction with non-bounding linear constraints is allowed`.

mforets: :+1:  thanks

PR #2627, opened 2021-03-19
fix generator columns for the nonlinear terms in TM -> Zono conversion

mforets:
this is a fix to the fact that:

```julia
using TaylorSeries

x = set_variables("x", numvars=2, order=10)
2-element Array{TaylorN{Float64},1}:
  1.0 x₁ + 𝒪(‖x‖¹¹)
  1.0 x₂ + 𝒪(‖x‖¹¹)

p = (1 + x[1] - 2x[2])^2
 1.0 + 2.0 x₁ - 4.0 x₂ + 1.0 x₁² - 4.0 x₁ x₂ + 4.0 x₂² + 𝒪(‖x‖¹¹)

p - constant_term(p) - linear_polynomial(p)
 0.0 + 𝒪(‖x‖²)
```
which fails because the order of linear_polynomial is 1. hence we implement `_nonlinear_polynomial` that has the correct order. discussed on slack with @dpsanders

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -720,7 +720,7 @@ return quote
 using .TaylorModels: Taylor1, TaylorN, TaylorModelN, TaylorModel1,
                      polynomial, remainder, domain,
                      normalize_taylor, linear_polynomial,
-                     constant_term, evaluate, mid, get_numvars
+                     constant_term, evaluate, mid, get_numvars, HomogeneousPolynomial
 
 @inline function get_linear_coeffs(p::Taylor1)
     if p.order == 0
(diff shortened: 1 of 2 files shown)


src/Approximations/overapproximate.jl
```diff
@@ -970,16 +995,16 @@ We refer to the algorithm description for the univariate case.
...
-    gen_rem = Vector{T}(undef, m) # generators for the remainder
+        # preallocations
+        c = Vector{T}(undef, m) # center of the zonotope
+        gen_lin = Matrix{T}(undef, m, n) # generator of the linear part
+        gen_rem = Vector{T}(undef, m) # generators for the remainder
 
-    # compute overapproximation
-    return _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
+        # compute overapproximation
+        return _overapproximate_vTM_zonotope!(vTM, c, gen_lin, gen_rem)
```
schillic, line 1007: This indentation change is not correct.

src/Approximations/overapproximate.jl
```diff
@@ -847,7 +872,7 @@ julia> center(Z)
 3-element Array{Float64,1}:
   1.0
  -2.1
-  0.8999999999999999
+  2.4
```
schillic, line 875: So the previous example was broken. Do you know whether the new version is correct?

schillic (approved):

src/Approximations/overapproximate.jl
mforets: i don't know when did the result change.

Issue #1837, opened 2019-12-12
Generalize convexity in docstrings of lazy operations


mforets:
If Bloating applies to non-convex sets as well, the docstring should be generalized.

Related to https://github.com/JuliaReach/LazySets.jl/issues/1838.

- [x] Affine map (#2629)
- [x] Bloating (#2630)
- [x] Cached Minkowski Sum Array (#2631)
- [x] CartesianProduct (#2636)
- [x] CartesianProductArray (#2638)
- [x] Complement (#2639)
- [x] ConvexHull (#2640)
- [x] ConvexHullArray (#2677)
- [x] ExponentialMap (#2681)
- [x] Intersection (#2683)
- [x] IntersectionArray (#2685)
- [x] InverseLinearMap (@schillic: nothing to do)
- [x] LinearMap (#2690)
- [x] MinkowskiSum (#2697)
- [x] MinkowskiSumArray (#2697)
- [x] Rectification (#2696)
- [x] ResetMap (#2695)
- [x] SymmetricIntervalHull (#2694)
- [x] Translation (#2693)
- [x] UnionSet (#2692)
- [x] UnionSetArray (#2691)
- [x] For lazy operations that preserve convexity: add this information to the docstrings. (#2698)

schillic:
This applies to other lazy operations as well. Here is an example:

https://github.com/JuliaReach/LazySets.jl/blob/8f371b39480bf1251ee48c0b05c3594ad79643c6/src/LazyOperations/AffineMap.jl#L12-L21

mforets:
OK, we can do these in the same issue. I added an item list in my first comment.

schillic:
Currently subtypes of `LazySet` are convex. So these comments are correct in the current state and give you important information (namely that the operations preserve convexity). This information should be preserved in the documentation (possibly at a different place, e.g., under `### Notes`).

PR #2631, opened 2021-04-01
update cached MSA docstrings

mforets:
See #1837.

This file updates the Cached MinkowskiSumArray.

Unrelated: the struct `CachedPair` has an abstract field. We should change it to `Vector{N}` or add a new field.

### DIFF
--- a/src/LazyOperations/CachedMinkowskiSumArray.jl
+++ b/src/LazyOperations/CachedMinkowskiSumArray.jl
@@ -32,12 +32,12 @@ end
 """
     CachedMinkowskiSumArray{N, S<:LazySet{N}} <: LazySet{N}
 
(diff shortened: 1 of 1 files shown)


src/LazyOperations/CachedMinkowskiSumArray.jl
```diff
@@ -140,7 +139,7 @@ If the direction has norm zero, the result depends on the summand sets.
 
 The result is cached, i.e., any further query with the same direction runs in
 constant time.
-When sets are added to the caching Minkowski sum, the query is only performed
+When sets are added to the cached Minkowski sum, the query is only performed
```
mforets, line 142: in several places we had `caching` instead of `cached`. i think we should use only one name (?) so `cached` is better since it is part of the type's name.

schillic (approved):

PR #2642, opened 2021-04-02
#1453 - Remove redundant generators in constraints_list

### LINKED ISSUE #1453: Remove linear dependencies from zonotope's generators in constraints_list
This came up [in #1451](https://github.com/JuliaReach/LazySets.jl/pull/1451#issuecomment-503821023).
> So perhaps we can check if the matrix of generators is full rank, and if it is not, remove the linearly dependent generators with the QR factorization and apply the algorithm for the independent case.

schillic:
Closes #1453.

- I added `remove_redundant_generators(::AbstractZonotope)` because otherwise the call might fail. All non-`Zonotope` subtypes do not store the generators and hence the implementation is just the identity.
- I pulled up the handling of the 1D case because that is always more efficient. I also removed the `view` there because it is slower than just using `sum`. And I outsourced that code to a helper function.
- In contrast to the proposal in https://github.com/JuliaReach/LazySets.jl/issues/1453#issuecomment-812109639 I just perform the same checks again on the reduced zonotope. If that is not correct, I can change it.

I did not know what test to add because I could not find an error. Here is an example where the output changes.

```julia
julia> Z = Zonotope(rand(2), [2. 3 1; 0 0 1]);

julia> constraints_list(Z)  # master
6-element Vector{HalfSpace{Float64, Vector{Float64}}}:
 HalfSpace{Float64, Vector{Float64}}([0.0, -1.0], 0.6611828781793785)
 HalfSpace{Float64, Vector{Float64}}([-0.0, 1.0], 1.3388171218206215)
 HalfSpace{Float64, Vector{Float64}}([0.0, -1.0], 0.6611828781793785)  # duplicate
 HalfSpace{Float64, Vector{Float64}}([-0.0, 1.0], 1.3388171218206215)  # duplicate
 HalfSpace{Float64, Vector{Float64}}([0.7071067811865475, -0.7071067811865475], 3.975112612743048)
 HalfSpace{Float64, Vector{Float64}}([-0.7071067811865475, 0.7071067811865475], 3.0959551991224266)

julia> constraints_list(Z)  # this branch
4-element Vector{HalfSpace{Float64, Vector{Float64}}}:
 HalfSpace{Float64, Vector{Float64}}([0.0, -1.0], 0.6611828781793785)
 HalfSpace{Float64, Vector{Float64}}([-0.0, 1.0], 1.3388171218206215)
 HalfSpace{Float64, Vector{Float64}}([0.7071067811865475, -0.7071067811865475], 3.975112612743048)
 HalfSpace{Float64, Vector{Float64}}([-0.7071067811865475, 0.7071067811865475], 3.0959551991224266)
```

### DIFF
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -505,10 +505,12 @@ It first computes the vertices and then converts the corresponding polytope
 to constraint representation.
 """
 function constraints_list(Z::AbstractZonotope)
-    return _constraints_list_fallback(Z)
+    return _constraints_list_vrep(Z)
 end
 
-@inline _constraints_list_fallback(Z) = constraints_list(VPolytope(vertices_list(Z)))
+@inline function _constraints_list_vrep(Z::AbstractZonotope)
+    return constraints_list(VPolytope(vertices_list(Z)))
+end
 
 """
     constraints_list(Z::AbstractZonotope{N}; check_full_rank::Bool=true) where {N<:AbstractFloat}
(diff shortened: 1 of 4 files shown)


src/Interfaces/AbstractZonotope.jl
```diff
@@ -542,25 +544,38 @@ Reachable Sets of Hybrid Systems Using a Combination of Zonotopes and Polytopes.
...
-        return constraints
+    G = genmat(Z)
+    p = ngens(Z)
+
+    # check whether to use the fallback implementation
+    use_fallback = false
+    if p < n
+        # order < 1
+        use_fallback = true
+    elseif check_full_rank
```
mforets, line 565: The precondition for the algorithm in [Lemma 7, 1] is that the `p` generators should be "independent" (which obviously doesn't mean linear independence; what is required is that the denominator in `Ci+` is never zero).

It can happen that `p < n` but the `p` generators are independent.

At this point I think we should remove the `rank` criteria and instead apply `remove_redundant_generators` with an optional flag set to `true` by default (the flag we can could call it `prune`).

The suggestion to remove the rank criterion is because it helps only in a few cases and it's a bit complicated. Let `r = rank(G)`, if i'm not mistaken it should be like this:

- If `p <= n` then `r` is at most `p`. If `r = p` then for sure all the generators are independent. Otherwise we need to prune.
 
- If `p > n` then `r` is at most `n`. We can't conclude anything from rank criteria and always have to apply the prune method.

mforets (approved):

src/Interfaces/AbstractZonotope.jl
schillic: @mforets The test in `test/LazyOperations/ResetMap.jl:65` now fails, so the change seems to be incorrect. One of the sets does not use the vrep fallback now and the implementation computes an `HPolytope` with no constraints. I have not checked whether the computation has a bug or whether this input should use the vrep fallback.

schillic: This error was just a bug. When `remove_redundant_generators` could remove zero columns but nothing else, this fact was not realized.

schillic:
I fixed a bug and now this is ready. @mforets I think you already reviewed it. Should we merge?

mforets:
ok :+1:

schillic:
I just realized that the name `remove_redundant_constraints` may be misleading because it actually merges several constraints. I am still fine to keep the name because it sounds similar as `remove_redundant_constraints` and `remove_redundant_vertices` but I can also change it (to `simplify_generators`?). What do you think?

schillic:
Never mind, the method already existed before this PR. Then I will merge it now.

PR #2653, opened 2021-04-12
#2647 - Revise plot recipe for UnionSet

### LINKED ISSUE #2647: Revise plot recipe for UnionSet
Eg. 

```
julia> typeof(swept)
UnionSetArray{Float64, Zonotope{Float64, Vector{Float64}, Matrix{Float64}}}

julia> plot(swept; aspect_ratio=1.0, color=colorant"green", label="continuous over-approximated")
```
prints the label as many times as there are sets. Probably we need to review how are `UnionSetArray` plotting options handled.

schillic:
Closes #2647.

- The `UnionSet` and `UnionSetArray` recipe now just plots a sequence of sets with the same options (in particular the same color and a single label).
- There was a plotting argument `fast` but using it was not comfortable, so I turned it into a keyword argument.
- I also changed the name of the `fast` argument to `same_recipe` because that describes the effect better. That it is faster is a nice side effect but less important.

### DIFF
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -88,7 +88,7 @@ that the overapproximation using iterative refinement is available:
 ```@docs
 plot_recipe(::LazySet{N}, ::Any=zero(N)) where {N}
 RecipesBase.apply_recipe(::AbstractDict{Symbol,Any}, ::LazySet{N}, ::N=N(1e-3)) where {N}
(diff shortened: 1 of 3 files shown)


mforets (approved):

schillic:
Plotting flat sets with this `same_recipe` option results in black line segments. This is because we do not set the line color by default and this plot recipe does not call the special "flat set" recipe. I suggest that we set the line color to the same color.

mforets (approved):

PR #2669, opened 2021-04-21
WIP - Star set with a lazy intersection predicate

mforets:
This type is a variation of the star set where the predicate is a lazy intersection. The main advantage over `Star` is that we now have an `HPolyhedron` "buffer" in the predicate so we can accumulate results for intersections without having to compute over the "initial" zonotope (i'm assuming that the predicate is `P = X \cap Y`  where `X` is a zonotope and `Y` is an `HPolyhedron`).


mforets:
Closing as stale.

PR #2673, opened 2021-04-24
Update symbolic constructors signatures

mforets:


src/Sets/Hyperplane.jl
```diff
@@ -572,9 +572,7 @@ function Hyperplane(expr::Num, vars=_get_variables(expr); N::Type{<:Real}=Float6
     return Hyperplane(coeffs, β)
 end
 
-function Hyperplane(expr::Num, vars::NTuple{L, Union{<:Num, <:Vector{Num}}}; N::Type{<:Real}=Float64) where {L}
-    vars = _vec(vars)
-    return Hyperplane(expr, vars, N=N)
-end
+Hyperplane(expr::Num; N::Type{<:Real}=Float64) = Hyperplane(expr, _get_variables(expr); N=N)
+Hyperplane(expr::Num, vars; N::Type{<:Real}=Float64) = Hyperplane(expr, _vec(vars), N=N)
```
schillic, line 576: This is an infinite loop. Either you need to restrict the type of `vars`, like this:
```suggestion
Hyperplane(expr::Num, vars::???; N::Type{<:Real}=Float64) = Hyperplane(expr, _vec(vars), N=N)
```
Or, if you want to allow any input, you need to add another method where `vars` is restricted to the output of `_vec(vars)` such that the recursive call uses that method.

mforets: The `Hyperplane`'s signature works as intended: the signature of the main method now dispatches on `Vector{Num}`. Note that I made some modifications to `_vec`.

The reason for these different signatures is to cope with the different returns types of `@variables ...` (the changed in the latest Symbolics version; now returns a vector):

```julia
julia> using Symbolics

julia> typeof(@variables t)
Vector{Num} (alias for Array{Num, 1})

julia> typeof(@variables t x)
Vector{Num} (alias for Array{Num, 1})

julia> typeof(@variables t x[1:2])
Vector{Any} (alias for Array{Any, 1})
```

schillic: I see, there is another method that catches the output of `_vec(vars)` already (my second suggestion). I just saw the error and thought all cases had this problem, but it was only in `HPolytope` :+1:

src/Sets/Hyperplane.jl
```diff
@@ -556,7 +556,7 @@ Therefore, the order in which the variables appear in `vars` affects the final r
 Finally, the returned set is the hyperplane with normal vector `[a1, …, an]` and
 displacement `b`.
 """
-function Hyperplane(expr::Num, vars=_get_variables(expr); N::Type{<:Real}=Float64)
+function Hyperplane(expr::Num, vars::Vector{Num}; N::Type{<:Real}=Float64)
```
schillic, line 559: It can also be `AbstractVector` because the code is not specific to `Vector`.

schillic (approved):

PR #2687, opened 2021-05-14
Fix pol order

mforets:


schillic (commented):
I do not understand this. If `p.order <= 1`, then `return p` should be correct, no? And in the new tests I always get `p.order == 10`, so they should have taken the `else` case before.

schillic (approved):

PR #2690, opened 2021-05-15
#1837 - Remove convexity requirement in LinearMap

schillic:
See #1837.

I am not 100% sure whether the generalization of `σ`/`ρ` is correct. My intuitive argument is that a linear map preserves the general shape (it only rotates/resizes/stretches) and this should also hold for non-square matrices.


mforets (approved):

mforets:
i think it's correct. the property `<y, Mx> = <M^T y, x>` holds for non-square matrices `M` too.

Issue #1186, opened 2019-03-02
Reordering dimensions


schillic:
Given a set `X`, one may want to permute certain dimensions (say, dimension 1 becomes dimension 2, dimension 2 becomes dimension 3, dimension 3 becomes dimension 1).
```julia
julia> v = [1., 2., 3., 4.];

julia> p = [2, 3, 1, 4];  # permutation vector

julia> v[p]
4-element Array{Float64,1}:
 2.0
 3.0
 1.0
 4.0
```

We can define a concrete function (with a method for each set type) that performs this permutation (usually this means to apply the permutation to vectors).

We can also define a lazy wrapper that reorders inputs and outputs for common functions, e.g. of the support vector.

What would be a good name?

mforets:
This is a particular case of a `LinearMap`. Can we call it `PermuatationMap` (or `Permutation`)? compare to [permutation matrix](https://en.wikipedia.org/wiki/Permutation_matrix).

schillic:
I created #2707 for the concrete operation. So this issue considers the lazy operation only.

PR #2706, opened 2021-05-20
#1186 - permute for Singleton and Hyperrectangle

schillic:
See #1186.

### DIFF
--- a/src/Interfaces/LazySet.jl
+++ b/src/Interfaces/LazySet.jl
@@ -1514,6 +1514,7 @@ Permute the dimensions of a set according to a given permutation vector.
 
 ### Output
 
-A new equivalent set where the dimensions have been permuted.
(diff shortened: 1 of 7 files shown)


mforets (approved):

src/Interfaces/LazySet.jl
```diff
@@ -1500,3 +1501,19 @@ rationalize(::Type{T}, X::LazySet{N}; tol::Real=eps(N)) where {T<:Integer, N<:Ab
...
+Permute the dimensions of a set according to a given permutation vector.
+
+### Input
+
+- `X` -- set
+- `p` -- permutation vector
+
+### Output
+
+A new equivalent set where the dimensions have been permuted.
```
mforets, line 1517: I think that the resulting set is not "equivalent" in the sense of `isequivalent`, so should we use another word here?

PR #2743, opened 2021-06-16
Update .travis.yml

mforets:
i wasn't sure if we have to use all the `Pkg`. statements or if it's enough with the `include`.


schillic (approved):
There was again no Travis run. I suggest we merge and see :sweat_smile:

PR #2747, opened 2021-06-18
#2745 - Conversion from polyhedron to hyperplane

### LINKED ISSUE #2745: Conversion from polyhedron to hyperplane
Check that there are exactly two half-spaces and that they are "opposite."

schillic:
Closes #2745.

* [x] outsource check to a new exported `ishyperplanar` function


mforets (approved):
LGTM

Nitpicking here but `is_hyperplanar` reads better IMO. And we already have [`_is_hyperplane`](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Sets/Hyperplane.jl#L498) (and `_is_halfspace`).

PR #2759, opened 2021-07-01
Proposal for unbounded SIH

schillic:


mforets:
now that the issue that triggered this PR is closed, we could close this one too but i leave that for @schillic

PR #2769, opened 2021-07-06
Add Javis.jl as an optional dependency

mforets:
This PR adds [Javis.jl](https://wikunia.github.io/Javis.jl/dev/), which uses [Luxor.jl](https://github.com/JuliaGraphics/Luxor.jl) as the graphics backend, as an optional dependency to LazySets. Some boilerplate functions are added together with an `animate` method that receives an array of sets and produces an animated gif.

There are several things that could be improved, such as using the global `Javis.scale` method, and passing colors to fill the sets. I made some quick tests but it didn't work, so I would rather leave that for follow-up PRs.

Example:

```julia
using LazySets, Javis

B = BallInf(ones(2), 1.0)
E(r) = Ball2(zeros(2), r)
out = [B ⊕ E(ri) for ri in range(2.0, 0.0, length=50)]

animate(out, tmax=100) 
```

javis_fBeSJq3


mforets:
Another example:

```julia
using LazySets, Javis, ReachabilityAnalysis

B0 = Ball2([1.0, 0.0], 0.1)
X0 = overapproximate(B0, PolarDirections(25))
A = [0.0 1.0; -1.0 0.0]
prob = @ivp(x' = Ax, x(0) ∈ X0)

sol = solve(prob, T=10.0, alg=VREP(δ=0.05));
```

```julia
using Plots

Plots.plot(sol, vars=(1, 2), ratio=1.)
```
Screenshot from 2021-07-06 03-59-27

```julia
Javis.animate(set.(sol), tmax=200, scale=200.0)
```

javis_48SoEyZ

mforets:
this pr is currently blocked since Javis.jl can't install on julia v1.3. possible fix: skip installing this optional dependency for julia v1.3 using a custom installation script.

schillic:
We can remove it from `Project.toml` again and instead add it here:

https://github.com/JuliaReach/LazySets.jl/blob/35ef1b5ca64d6754b3fa8e10fa80bde1b47d4eb6/deps/build_private.jl#L12-L14

mforets:
thanks for fixing this :100: 
i was told by Javis devs that the approach used here does not scale well; instead of defining frame by frame, we should use the `change` function (discussion [here](https://julialang.zulipchat.com/#narrow/stream/253420-javis.2Ejl/topic/Quickly.20Generating.20100's.20of.20Points.20in.20Javis))

PR #2779, opened 2021-07-13
Add chebyshev_center for Interval

schillic:
`master`:
```julia
julia> c = chebyshev_center(Interval(1.0, 2.0))
ERROR: MethodError: no method matching sethrep!(::Polyhedra.Interval{Float64, StaticArrays.SVector{1, Float64}, StaticArrays.Size{(1,)}}, ::Polyhedra.Intersection{Float64, StaticArrays.SVector{1, Float64}, StaticArrays.Size{(1,)}})
Closest candidates are:
  sethrep!(::Any, ::Any, ::Any) at ~/.julia/packages/Polyhedra/ajeAR/src/defaultlibrary.jl:115
  sethrep!(::DefaultPolyhedron, ::HRepresentation) at ~/.julia/packages/Polyhedra/ajeAR/src/defaultlibrary.jl:120
  sethrep!(::DefaultPolyhedron, ::HRepresentation, ::Any) at ~/.julia/packages/Polyhedra/ajeAR/src/defaultlibrary.jl:120
```

### DIFF
--- a/src/Sets/Interval.jl
+++ b/src/Sets/Interval.jl
@@ -692,3 +692,10 @@ end
 function vertices_list(H::IntervalArithmetic.IntervalBox)
     return vertices_list(convert(Hyperrectangle, H))
 end
+
+function chebyshev_center(x::Interval{N}; compute_radius::Bool=false) where {N}
+    if compute_radius
+        return center(x), zero(N)
+    end
+    return center(x)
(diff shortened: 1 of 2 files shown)


mforets:
the error message seems like it is worth reporting to Polyhedra.jl

PR #2780, opened 2021-07-13
Add a useful fallback for the support vector of lazy intersections

mforets:


src/LazyOperations/Intersection.jl
```diff
@@ -224,7 +224,8 @@ direction.
 The support vector in the given direction.
 """
 function σ(d::AbstractVector, cap::Intersection)
-    error("the exact support vector of an intersection is not implemented")
+    X = concretize(cap)
```
schillic, line 227: Something like `σ(d, rand(Ball2) ∩ rand(Ball2))` would now give a cryptic error message. Consider this:
```suggestion
    if !applicable(intersection(cap.X, cap.Y))
        error("the exact support vector of an intersection is not implemented")
    end
    X = concretize(cap)
```

schillic (approved):
There are some more tests to be changed.

src/LazyOperations/Intersection.jl
mforets: hmm i prefer the error message with the full stacktrace than not implemented error. the former is more informative and often gives a hint of what needs to be implemented. yes, for newcomers they have to learn how to read stacktraces, but that's a separate issue.

```julia
julia> d = rand(2); σ(d, rand(Ball2) ∩ rand(Ball2))
ERROR: MethodError: no method matching intersection(::Ball2{Float64, Vector{Float64}}, ::Ball2{Float64, Vector{Float64}})
Closest candidates are:
  intersection(::LinearMap, ::LazySet) at /home/mforets/.julia/dev/LazySets/src/ConcreteOperations/intersection.jl:856
  intersection(::AbstractSingleton, ::LazySet) at /home/mforets/.julia/dev/LazySets/src/ConcreteOperations/intersection.jl:44
  intersection(::Universe, ::LazySet) at /home/mforets/.julia/dev/LazySets/src/ConcreteOperations/intersection.jl:884
  ...
Stacktrace:
 [1] concretize(cap::Intersection{Float64, Ball2{Float64, Vector{Float64}}, Ball2{Float64, Vector{Float64}}})
   @ LazySets ~/.julia/dev/LazySets/src/LazyOperations/Intersection.jl:725
 [2] σ(d::Vector{Float64}, cap::Intersection{Float64, Ball2{Float64, Vector{Float64}}, Ball2{Float64, Vector{Float64}}})
   @ LazySets ~/.julia/dev/LazySets/src/LazyOperations/Intersection.jl:227
 [3] top-level scope
   @ REPL[2]:1
```

mforets: another argument to use `X = concretize(cap)` directly is that `concretize` can specialize on its arguments; i wouldn't assume that it always falls back to the concrete intersection.

mforets:
> There are some more tests to be changed.

yes, thanks. the latests commit fixes them locally

schillic (approved):

PR #2809, opened 2021-08-04
#2741 - Efficient concrete Cartesian product of hyperrectangular sets and Universe

### LINKED ISSUE #2741: Efficient concrete Cartesian product of hyperrectangular sets and Universe
Hyperrectangular sets return constraints with `SingleEntryVector`, which other methods can specialize on. But the Cartesian product with a `Universe` uses the default method via `Polyhedra` conversion and hence does not return this vector type. The task is to define two new methods for the combination `AbstractHyperrectangle` and `Universe`.
```julia
julia> B = rand(BallInf)
BallInf{Float64, Vector{Float64}}([0.7932902532734889, -1.662932838563669], 0.5720352339334028)

julia> constraints_list(B)
4-element Vector{HalfSpace{Float64, LazySets.Arrays.SingleEntryVector{Float64}}}:
 HalfSpace{Float64, LazySets.Arrays.SingleEntryVector{Float64}}([1.0, 0.0], 1.3653254872068916)
 HalfSpace{Float64, LazySets.Arrays.SingleEntryVector{Float64}}([0.0, 1.0], -1.0908976046302663)
 HalfSpace{Float64, LazySets.Arrays.SingleEntryVector{Float64}}([-1.0, 0.0], -0.22125501934008607)
 HalfSpace{Float64, LazySets.Arrays.SingleEntryVector{Float64}}([0.0, -1.0], 2.2349680724970717)

julia> cartesian_product(B, Universe(1))
HPolyhedron{Float64, Vector{Float64}}(HalfSpace{Float64, Vector{Float64}}[HalfSpace{Float64, Vector{Float64}}([1.0, 0.0, 0.0], 1.3653254872068916), HalfSpace{Float64, Vector{Float64}}([0.0, 1.0, 0.0], -1.0908976046302663), HalfSpace{Float64, Vector{Float64}}([-1.0, 0.0, 0.0], -0.22125501934008607), HalfSpace{Float64, Vector{Float64}}([0.0, -1.0, 0.0], 2.2349680724970717)])
```

schillic:
Closes #2741.

### DIFF
--- a/src/ConcreteOperations/cartesian_product.jl
+++ b/src/ConcreteOperations/cartesian_product.jl
@@ -117,7 +117,7 @@ The polyhedron obtained by the concrete cartesian product of `P1` and `P2`.
 For further information on the supported backends see
 [Polyhedra's documentation](https://juliapolyhedra.github.io/).
 """
-function cartesian_product(P1::Union{HPoly, Universe}, P2::Union{HPoly, Universe}; backend=nothing)
+function cartesian_product(P1::HPoly, P2::HPoly; backend=nothing)
     require(:Polyhedra; fun_name="`cartesian_product")
 
     return _cartesian_product_hrep(P1, P2, backend1=backend, backend2=backend)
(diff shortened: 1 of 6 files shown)


src/ConcreteOperations/cartesian_product.jl
```diff
@@ -180,3 +180,11 @@ function cartesian_product(H1::AbstractHyperrectangle, H2::AbstractHyperrectangl
     r = vcat(radius_hyperrectangle(H1), radius_hyperrectangle(H2))
     return Hyperrectangle(c, r)
 end
+
+@commutative function cartesian_product(H::AbstractHyperrectangle{N},
+                                        U::Universe) where {N}
+    n = dim(H) + dim(U)
+    constraints = [HalfSpace(SingleEntryVector(c.a.i, n, c.a.v), c.b) for c in
```
mforets, line 187: this assumes that the `constraints_list` of an AbstractHyperrectangle *always* returns a SEV, which could not be the case in general. i see two alternatives:

- dispatch on the concrete type Hyperrectangle{..., SEV}, or
- (my preferred choice) use [iterated dispatch](https://docs.julialang.org/en/v1/manual/methods/#Iterated-dispatch) along these lines:

```julia
@commutative function cartesian_product(H::AbstractHyperrectangle{N},
                                        U::Universe) where {N}
    n = dim(H) + dim(U)
    clist = constraints_list(H)
    return _cartesian_product_hyperrectangle_universe(clist, U)
end

function _cartesian_product_hyperrectangle_universe(clist::Vector{SingleEntryVector{N}}, U) where {N}
    constraints = [HalfSpace(SingleEntryVector(c.a.i, n, c.a.v), c.b) for c in clist]
    return HPolyhedron(constraints)
end
```

schillic: I improved it quite a lot. Going down to the `HalfSpace` generalizes to arbitrary polyhedra. Before we went through `Polyhedra` in that case. Now we do not and if the polyhedron has sparse constraints, so will the result of `cartesian_product`.

The idea is to either append or prepend zeros to the normal direction, depending on whether the universe comes last or first.

mforets (approved):

PR #2836, opened 2021-09-13
Add boxdot as alias for symmetric_interval_hull

schillic:


src/Approximations/symmetric_interval_hull.jl
```diff
@@ -33,6 +35,13 @@ function symmetric_interval_hull(S::LazySet{N}) where {N}
     return Hyperrectangle(zeros(N, length(c)), abs.(c) .+ r)
 end
 
+"""
+    ⊡
+
+Alias for `symmetric_interval_hull`.
+"""
+const ⊡ = symmetric_interval_hull
```
mforets, line 43: oh, i thought we'd use this unicode alias for the symbolic one, given that we use \oplus, \cross etc for symbolics

schillic: Oops, you are right. I fixed it.

mforets (approved):

PR #2849, opened 2021-09-17
#2831 - Inclusion of interval in union of intervals

### LINKED ISSUE #2831: Inclusion of interval in union of intervals

schillic:
Closes #2831.

This PR also adds witness production for the interval-interval inclusion check.

The method checks inclusion in unions of arbitrary sets, as long as they can be converted to `Interval`s. That means it will fail for unions of non-convex sets.

### DIFF
--- a/src/ConcreteOperations/issubset.jl
+++ b/src/ConcreteOperations/issubset.jl
@@ -508,17 +508,12 @@ end
 
 """
     ⊆(x::Interval, y::Interval, [witness]::Bool=false)
-
 Check whether an interval is contained in another interval.
-
 ### Input
-
 - `x`       -- interval
 - `y`       -- interval
 - `witness` -- (optional, default: `false`) compute a witness if activated
-
 ### Output
-
 `true` iff ``x ⊆ y``.
 """
 function ⊆(x::Interval, y::Interval, witness::Bool=false)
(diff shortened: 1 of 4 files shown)


mforets:
ping @AnderGray i think we wanted to use this in an example

src/ConcreteOperations/issubset.jl
```diff
@@ -526,6 +526,46 @@ function ⊆(x::Interval, y::Interval, witness::Bool=false)
     return x.dat ⊆ y.dat
 end
 
+function ⊆(x::Interval, U::UnionSet, witness::Bool=false)
```
mforets, line 529: special implementation, after discussion:

```julia
"""
    issubset(x::Interval, U::UnionSet{N, IT, IT}) where {N, IT<:Interval{N}}

Return whether an interval is contained in the union of two intervals.

### Input

- `x` -- interval
- `U` -- union of two intervals

### Output

`true` if `x ⊆ U` and `false` otherwise.

### Algorithm

Let ``U = a ∪ b `` where ``a`` and ``b`` are intervals. Then
the set difference ``y = x \ a`` can have three results:

1. If it is empty, then ``x ⊆ a`` hence ``x ⊆ a ∪ b``.

2. If the result is an interval, we check whether ``y ⊆ b``.

3. If the result is a union of intervals, it suffices to check if ``x ⊆ b``
   (since ``b` is an interval), see diagram below.

```
x      -------
a        ***
y      ++   ++
b    @@@@@@@@@@
```
"""
function issubset(x::Interval, U::UnionSet{N, IT, IT}) where {N, IT<:Interval{N}}
    a = U.X
    b = U.Y
    aux = difference(x, a)
    isempty(aux) && return true

    if isa(aux, Interval)
        aux2 = difference(aux, b)
        return isempty(aux2)
    else # union
        return x ⊆ b
    end
end
```

src/ConcreteOperations/issubset.jl
```diff
@@ -526,6 +526,46 @@ function ⊆(x::Interval, y::Interval, witness::Bool=false)
...
+function ⊆(x::Interval, U::UnionSetArray, witness::Bool=false)
+    @assert dim(U) == 1 "an interval is incompatible with a set of dimension $(dim(U))"
+    V = array(U)
+    if !(V isa AbstractVector{<:Interval})
+        V = [convert(Interval, Y) for Y in V]
+    end
+    return _issubset_interval(x, array(U), witness)
+end
+
+function _issubset_interval(X::Interval{N}, intervals, witness) where {N}
```
mforets, line 547:
```suggestion
function _issubset_interval!(X::Interval{N}, intervals, witness) where {N}
```
(since `intervals` may be changed)

src/ConcreteOperations/issubset.jl
```diff
@@ -526,6 +526,46 @@ function ⊆(x::Interval, y::Interval, witness::Bool=false)
...
+    return _issubset_interval(x, V, witness)
+end
+
+function ⊆(x::Interval, U::UnionSetArray, witness::Bool=false)
+    @assert dim(U) == 1 "an interval is incompatible with a set of dimension $(dim(U))"
+    V = array(U)
+    if !(V isa AbstractVector{<:Interval})
+        V = [convert(Interval, Y) for Y in V]
+    end
+    return _issubset_interval(x, array(U), witness)
```
mforets, line 544:
```suggestion
    return _issubset_interval(x, V, witness)
```

src/ConcreteOperations/issubset.jl
```diff
@@ -526,6 +526,46 @@ function ⊆(x::Interval, y::Interval, witness::Bool=false)
...
+    if !(V isa AbstractVector{<:Interval})
+        V = [convert(Interval, Y) for Y in V]
+    end
+    return _issubset_interval(x, V, witness)
+end
+
+function ⊆(x::Interval, U::UnionSetArray, witness::Bool=false)
+    @assert dim(U) == 1 "an interval is incompatible with a set of dimension $(dim(U))"
+    V = array(U)
+    if !(V isa AbstractVector{<:Interval})
```
mforets, line 541: dispatch?

src/ConcreteOperations/issubset.jl
```diff
@@ -526,6 +526,46 @@ function ⊆(x::Interval, y::Interval, witness::Bool=false)
...
+    return _issubset_interval(x, array(U), witness)
+end
+
+function _issubset_interval(X::Interval{N}, intervals, witness) where {N}
+    # sort intervals by lower bound
+    sort!(intervals, lt=(x, y)->low(x, 1) <= low(y, 1))
+
+    # subtract intervals from X
+    for Y in intervals
+        if low(Y, 1) > low(X, 1)
```
mforets, line 553: we may also use `min` and `max`  for the 1d case

src/ConcreteOperations/issubset.jl
mforets: the other option is to special case the n-ary union method:

```julia
function _issubset2(x::Interval, U::UnionSet{N, IT, IT}) where {N, IT<:Interval{N}}
    a = U.X
    b = U.Y
    if min(a) < min(b)
        # a is on the left of b
        min(x) < min(a) && return false
        return difference(x, a) ⊆ b
    else 
        # a is on the right of b
        min(x) < min(b) && return false
        return difference(x, b) ⊆ a
    end
end
```

mforets (approved):

PR #2860, opened 2021-09-20
Fix ⊂ (for good)

schillic:
Fix for the fix in #2857.


schillic:
I am going to merge here to have a working `master` branch because the build error is unrelated. Locally the tests and docs work.

PR #1995, opened 2020-02-20
WIP #1954 - In-place concrete translation

mforets:
Closes #1954.

To be outsourced to smaller PRs:

* [x] balls
* [x] Ellipsoid
* [x] EmptySet
* [ ] HPolygon
* [ ] HalfSpace
* [ ] Hyperrectangle
* [x] Interval
* [ ] Line
* [ ] LineSegment
* [x] Singleton
* [x] Universe
* [x] VPolygon
* [x] VPolytope
* [ ] Zonotope

PR #2869, opened 2021-09-23
In-place translation (cont'd)

mforets:
see #1995

### DIFF
--- a/src/Sets/EmptySet.jl
+++ b/src/Sets/EmptySet.jl
@@ -340,8 +340,34 @@ Translate (i.e., shift) an empty set by a given vector.
 ### Output
 
 The empty set.
+
+### Notes
+
+See also [`translate!(::EmptySet, ::AbstractVector)`](@ref) for the in-place version.
 """
 function translate(∅::EmptySet, v::AbstractVector)
+    return ∅
+end
+
+"""
+    translate!(∅::EmptySet, v::AbstractVector)
+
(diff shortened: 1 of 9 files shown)


src/Sets/EmptySet.jl
```diff
@@ -340,8 +340,36 @@ Translate (i.e., shift) an empty set by a given vector.
 ### Output
 
 The empty set.
+
+### Notes
+
+See also [`translate!(::EmptySet, ::AbstractVector)`](@ref) for the in-place version.
 """
 function translate(∅::EmptySet, v::AbstractVector)
+    return translate!(∅, v)
```
mforets, line 349: obs: in this case, and also for the universe, i decided not to copy

schillic: I agree. But I think the documentation should not be the generic one as for the other types. Referring to `translate!` as the in-place version of `translate` does not make much sense to me here because they are identical.

schillic (approved):

src/Sets/EmptySet.jl
mforets: ok. here `translate!` is only included for completeness with respect to other methods, but it's really a no-op so i just remove the docstring altogether..

schillic (approved):

src/Sets/EmptySet.jl
schillic: You removed it from `translate` for `Universe` instead (hence the build failure).

PR #2871, opened 2021-09-23
In-place translation for VPoly

mforets:
see #1995


schillic:
As we suspected, this is a case where you need to do `deepcopy` or copy the vertices manually.

schillic (approved):

PR #2876, opened 2021-09-26
Check emptiness of HParallelotope in constructor

### LINKED ISSUE #2786: vertices_list of an empty HParallelotope is non-empty
```julia
julia> H = [1.0 1.0; 0.0 1.0];
julia> o = [2.5, 2.0, -4.0, -1.0];
julia> Pl = HParallelotope(H, o);  # this set is empty

julia> vertices_list(Pl)
4-element Vector{Vector{Float64}}:
 [2.0, 2.0]
 [0.5, 2.0]
 [1.5, 1.0]
 [3.0, 1.0]
```
~The question is whether we want to fix this because in a sense the set is illegal: we do not allow empty parallelotopes conceptually.~ We want to fix `vertices_list`.

schillic:
Closes #2786.

This is an alternative to #2875.

### DIFF
--- a/src/Sets/HParallelotope.jl
+++ b/src/Sets/HParallelotope.jl
@@ -58,9 +58,14 @@ struct HParallelotope{N, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}} <: Abstra
     offset::VN
 
     # default constructor with dimension check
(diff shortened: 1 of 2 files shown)


schillic:
We decided to choose #2875.

PR #2882, opened 2021-09-29
Update README.md

mforets:


schillic (approved):
Not sure how this looks like (what happened to the previews?) but :+1:

mforets:
> Not sure how this looks like (what happened to the previews?

:+1:  Have a look now.. I think that having this table helps those who are not familiar with badges and github.

PR #2885, opened 2021-10-07
Reduce alias methods

schillic:
This should fix the build error in #2884.


schillic:
The solution in #2887 was better.

PR #2888, opened 2021-10-09
New version of IntervalMatrices

schillic:


mforets:
> We support only the latest version of IA.

makes sense to me. but then we may have to drop the julia v1.3 support (which is ok IMO. related: "Almost everyone should be downloading and using the latest stable release of Julia. ...." in https://julialang.org/downloads/#download_julia )

PR #2896, opened 2021-11-02
New version of TaylorModels

schillic:


mforets (approved):

Project.toml
```diff
@@ -43,7 +43,7 @@ Reexport = "0.2, 1"
 Requires = "0.5, 1"
 StaticArrays = "0.12, 1"
 Symbolics = "0.1"
-TaylorModels = "0.0.1, 0.1, 0.2, 0.3, 0.4"
+TaylorModels = "0.0.1, 0.1, 0.2, 0.3, 0.4, 0.5"
```
mforets, line 46: should we remove those much older versions?

schillic: Since we have the old versions of IntervalArithmetic as well, I think not. It does not hurt to keep them.

PR #2907, opened 2021-11-27
#1312 - Add more exponentiation backends

### LINKED ISSUE #1312: Add more exponentiation backends
With https://github.com/JuliaReach/LazySets.jl/pull/1290 the functionality for lazy matrix exponential maps is optional and clears the way to use other backends.

Proposal: add support for choosing [ExponentialUtilities.jl](https://github.com/JuliaDiffEq/ExponentialUtilities.jl) to compute the matrix-exponential-vector product using Krylov.

In Julia v1.1.0-rc2.0:

```julia
julia> A = sprandn(100, 100, 0.1); v = rand(100);

julia> @btime expmv(1.0, A, v);  # Expokit.jl
  951.862 μs (74 allocations: 283.67 KiB)

julia> @btime expv(1.0, A, v);  # ExponentialUtilities.jl
  224.119 μs (1026 allocations: 205.09 KiB)
```

---

Another test on Julia v1.3

```julia
julia> A = sprandn(1000, 1000, 0.01); v = rand(1000);

julia> @btime expmv(1.0, $A, $v);
  2.122 ms (72 allocations: 523.77 KiB)

julia> @btime expv(1.0, $A, $v);
  794.993 μs (1013 allocations: 356.52 KiB)

julia> 2.1e-3 / 794e-6
2.644836272040302
```

schillic:
Closes #1312.

This is not exactly what was proposed in https://github.com/JuliaReach/LazySets.jl/issues/1312#issuecomment-975883251. Instead the backend is globally set and there is no additional `backend` kwarg. This is a bit simpler and in practice I expect that one uses only one backend for everything anyway. We can add this kwarg later if the use case arises.

```julia
julia> using LazySets, SparseArrays
julia> A = sprand(2, 2, 0.5);
julia> M = SparseMatrixExp(A);

julia> LazySets.exponential_backend
missing

julia> M  # the printing calls `get_column`
2×2 SparseMatrixExp{Float64, SparseMatrixCSC{Float64, Int64}}:
Error showing value of type SparseMatrixExp{Float64, SparseMatrixCSC{Float64, Int64}}:
ERROR: AssertionError: no package from '[:ExponentialUtilities, :Expokit]' loaded (one of them is required for executing `get_column`)

julia> using ExponentialUtilities  # load first package (automatically sets the default)

julia> LazySets.exponential_backend
Val{:ExponentialUtilities}()

julia> M
2×2 SparseMatrixExp{Float64, SparseMatrixCSC{Float64, Int64}}:
 1.33389   0.0
 0.300614  1.0

julia> using Expokit  # load other package (has no effect)

julia> LazySets.exponential_backend  # still uses ExponentialUtilities
Val{:ExponentialUtilities}()

julia> M
2×2 SparseMatrixExp{Float64, SparseMatrixCSC{Float64, Int64}}:
 1.33389   0.0
 0.300614  1.0

julia> LazySets.set_exponential_backend!(Expokit)  # use Expokit
Val{:Expokit}()

julia> M
2×2 SparseMatrixExp{Float64, SparseMatrixCSC{Float64, Int64}}:
 1.33389   0.0
 0.300614  1.0
```


src/Utils/matrix_exponential.jl
```diff
@@ -0,0 +1,61 @@
...
+function set_exponential_backend!(backend::Module)
+    global exponential_backend = Val(Symbol(backend))
+end
+
+function get_exponential_backend()
+    ismissing(exponential_backend) && error("the exponential backend is not loaded")
+    return exponential_backend
+end
+
+function _expmat(t, A, b)
```
mforets, line 19: packages call this `expmv` (or `expv`) where the `v` stands for the vector (in our notation, it is `b`)

then i would suggest to name this `_expv` or `_expmv`

src/Utils/matrix_exponential.jl
```diff
@@ -0,0 +1,61 @@
...
+return quote
+
+if ismissing(exponential_backend)
+    set_exponential_backend!(ExponentialUtilities)
+end
+
+end end  # quote / load_exponentialutilities
+
+
+function _expmat(::Val{:ExponentialUtilities}, t, A, b)
```
mforets, line 37: nice :+1:  this makes it very easy to add other exponentiation backends than the ones listed in this file; i mean, from the outside, as well as tweaking options (say, now i want to pass `expv` some extra options i can do it by overwriding `_expmat`)

i wonder if we need `SUPPORTED_EXPONENTIAL_PACKAGES` at all; i would rather remove (and also `require`) because if it was not loaded, a nice error message is shown anyways ([here](https://github.com/JuliaReach/LazySets.jl/pull/2907/files#diff-94a0f99d9daf3a4c7fe1a253635aebc5626457a3b734a92b6086c277375d3c2cR15))

src/LazyOperations/ExponentialMap.jl
```diff
@@ -305,17 +299,13 @@ If the direction has norm zero, the result depends on the wrapped set.
 
 If ``E = \\exp(M)⋅S``, where ``M`` is a matrix and ``S`` is a set, it
 follows that ``σ(d, E) = \\exp(M)⋅σ(\\exp(M)^T d, S)`` for any direction ``d``.
-
-We allow sparse direction vectors, but will convert them to dense vectors to be
-able to use `expmv`.
 """
 function σ(d::AbstractVector, em::ExponentialMap)
-    require(:Expokit; fun_name="σ")
+    require(SUPPORTED_EXPONENTIAL_PACKAGES; fun_name="σ")
```
mforets, line 304: here we may as well remove this `require` call, this this anyways fails when calling `_expmat` if no backend is loaded

(same in all other functions)

mforets (commented):
> This is a bit simpler and in practice I expect that one uses only one backend for everything anyway. We can add this kwarg later if the use case arises.

yes, i agree that for general usage it is convenient to set the backend once and for all

otoh, for development i find it convenient to be able to tweak the backend (and possibly some options, eg. either through a dedicated struct or through additional keywords) on a per-function basis

src/Utils/matrix_exponential.jl
schillic: We do not need it. But the generic error message does not give concrete instructions what to do, so you have to know which packages work. I think it is helpful for new users to see the list of packages that are automatically supported.

src/LazyOperations/ExponentialMap.jl
schillic: In fact we should remove these calls because otherwise you cannot add new packages from outside :+1:

mforets (approved):

PR #2920, opened 2021-12-13
#2825 - Update to new version of Symbolics

### LINKED ISSUE #2825: Update to new version of Symbolics
Changes are needed in `init_Symbolics.jl` to account for the new return type of `@variables`. The documentation in `tour.md` does not build with the new version.

We should also turn the related docs examples into doctests:

- `init_Symbolics.jl`
- `HalfSpace.jl`
- `HPolyhedron.jl`
- `HPolytope.jl`
- `Hyperplane.jl`

schillic:
Closes #2825.


src/Initialization/init_Symbolics.jl
```diff
@@ -1,50 +1,51 @@
...
+    ```
+    """
+    function _vec end
 
     # reduce for several variables e.g. when vars = @variables x[1:3] t
     _vec(vars::Vector{Any}) = reduce(vcat, vars)
     _vec(vars::Vector{Num}) = vars
+    _vec(vars::Vector{Symbolics.Arr{Num, 1}}) = reduce(vcat, vars)
     _vec(vars::Vector{Vector{Num}}) = reduce(vcat, vars)
+    _vec(vars::Vector{Real}) = reduce(vcat, vars)
```
mforets, line 48: since only the implementation for `::Vector{Num}` differs from the rest, i wonder if we should simplify this by doing:

```julia
_vec(vars::Vector) = reduce(vcat, vars) # fallback
_vec(vars::Vector{Num}) = vars # no-op
```

mforets (approved):

src/Initialization/init_Symbolics.jl
schillic: I do not understand the logic here, so I cannot say if that is correct. The catch-all will apply to other cases and that may be wrong. If you think it is correct, we can change it.

mforets (approved):

src/Initialization/init_Symbolics.jl
mforets: I guess so, but we can also simplify that part in the future.

PR #2925, opened 2022-01-29
Update Project.toml

schillic:
Bump for minor version because the lower bound for Julia was raised to v1.5.


schillic:
I would suggest to get #2927 in the release because it is related to the change.

mforets (approved):

PR #2923, opened 2022-01-28
Revamp optimization deps

mforets:

PR #2931, opened 2022-01-31
Bring back fixed test

schillic:
This test was temporarily deactivated in #2923.

I suggest to include a check based on the version of `GLPK` since at least for now some package(s) in my default Manifest hold(s) the new `GLPK` version with the fix back and hence tests would fail locally.


schillic:
```julia
ERROR: LoadError: ArgumentError: Package Pkg not found in current path:
- Run `import Pkg; Pkg.add("Pkg")` to install the Pkg package.
```
?

mforets (approved):

PR #2941, opened 2022-02-24
Faster low/high for Zonotope

schillic:
The `master` version computes the support function, which is already quite fast. The first version I tried is faster. The final version is even slightly faster than the other version because it avoids the `@view`.

```julia
function low2(Z::AbstractZonotope, i::Int)
    return center(Z, i) - sum(abs, @view genmat(Z)[i, :])
end

function low3(Z::AbstractZonotope, i::Int)
    G = genmat(Z)
    v = center(Z, i)
    for j in 1:ngens(Z)
        v -= abs(G[i, j])
    end
    return v
end

julia> Z = rand(Zonotope, dim=50, seed=0);

julia> @btime LazySets.low($Z, 1);  # master
  120.834 ns (0 allocations: 0 bytes)

julia> @btime LazySets.low2($Z, 1);
  61.188 ns (0 allocations: 0 bytes)

julia> @btime LazySets.low3($Z, 1);  # this branch
  55.798 ns (0 allocations: 0 bytes)
```


mforets:
> The master version computes the support function, which is already quite fast. The first version I tried is faster.

I think that, in general it depends on the cost of computing `genmat`. It is for free if the set is a `Zonotope`, but could otherwise be cheaper to compute support function instead.

mforets:
For the general (`AbstractZonotope`) case, I'm tempted to prefer the leaving the [default fallback](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Interfaces/LazySet.jl#L1451), but we can keep the faster version in this branch for the `Zonotope` case.

schillic:
Good point :+1:

```julia
julia> LazySets.subtypes(AbstractZonotope)
5-element Vector{Any}:
 AbstractHyperrectangle
 HParallelotope
 LineSegment
 RotatedHyperrectangle
 Zonotope


julia> L = rand(LineSegment, dim=2, seed=0);

julia> @btime LazySets.low($L, 1);  # master
  338.658 ns (7 allocations: 544 bytes)

julia> @btime LazySets.low2($L, 1);  # this branch
  453.561 ns (10 allocations: 768 bytes)


julia> H = rand(HParallelotope, dim=20, seed=0);

julia> @btime LazySets.low($H, 1);  # master
  767.961 μs (171 allocations: 172.12 KiB)

julia> @btime LazySets.low2($H, 1);  # this branch
  1.168 ms (305 allocations: 261.94 KiB)
```

mforets (approved):

PR #2946, opened 2022-03-04
Update Project.toml

mforets:


schillic:
I suggest to get #2942 into the release to propagate these new versions (if you agree with the changes there).

schillic (approved):

PR #2964, opened 2022-05-13
Add special cases for disjointness and inclusion involving universe

mforets:


src/ConcreteOperations/issubset.jl
```diff
@@ -1091,3 +1091,9 @@ function ⊆(Z::AbstractZonotope, H::AbstractHyperrectangle, witness::Bool=false
     end
     return true
 end
+
+function ⊆(Z::AbstractZonotope, C::CartesianProduct{N, <:LazySet, <:Universe}) where N
+    X = C.X
+    Zp = project(Z, 1:dim(X))
+    return ⊆(Zp, X)
+end
```
schillic, line 1099: Let's add the other case as well:

```suggestion
end

function ⊆(Z::AbstractZonotope, C::CartesianProduct{N, <:Universe, <:LazySet}) where N
    Y = C.Y
    Zp = project(Z, dim(C.X)+1:dim(C))
    return ⊆(Zp, Y)
end
```

schillic (approved):

src/ConcreteOperations/issubset.jl
mforets: This may introduce an ambiguity with the cartesian product of a pair of `Universe`.

schillic: True, it did. But the other methods also did :sweat_smile:

PR #2972, opened 2022-05-30
initial implementation of sparse polynomial zonotopes

lucaferranti:
Introduce SimpleSparsePolynomialZonotopes as described in Niklas thesis and some basic operations with it

### TO DO

- [ ] add tests
- [ ] better docstrings
- [ ] probably `overapproximate` should go somewhere else?


src/Sets/Interval.jl
```diff
@@ -709,3 +709,38 @@ function chebyshev_center(x::Interval{N}; compute_radius::Bool=false) where {N}
...
+- `n` -- integer
+
+### Output
+
+A non-rigorous approximation of `a^n`.
+
+### Notes
+
+For a rigorous approximation with correct rounding,
+use `a^n` from `IntervalArithmetic.jl`."""
```
mforets, line 730:
```suggestion
use `a^n` from `IntervalArithmetic.jl`.

Review after IntervalArithmetic.jl#388
"""
```

src/Sets/Interval.jl
```diff
@@ -709,3 +709,38 @@ function chebyshev_center(x::Interval{N}; compute_radius::Bool=false) where {N}
...
+    elseif isodd(n)
+        return IA.Interval(a.lo ^ n, a.hi ^ n)
+    else
+        lon = a.lo ^ n
+        hin = a.hi ^ n
+        if 0 ∈ a
+            return IA.Interval(zero(lon), max(lon, hin))
+        else
+            return IA.Interval(min(lon, hin), max(lon, hin))
+        end
```
schillic, line 744: In the `if` case, you could skip one of the power computations by only taking the power of the (absolute-wise) bigger value.

```suggestion
        if 0 ∈ a
            return IA.Interval(zero(lon), max(abs(lon), abs(hin)) ^ n)
        else
            lon = a.lo ^ n
            hin = a.hi ^ n
            return IA.Interval(min(lon, hin), max(lon, hin))
        end
```

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,105 @@
+"""
+    SimpleSparsePolynomialZonotope{N, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}, ME<:AbstractMatrix{<:Integer}} <: AbstractPolynomialZonotope{N}
+
+Type that represents a sparse polynomial zonotope.
+
+Given a constant offset ``c ∈ \\mathbb{R}^n``, a generator matrix ``G ∈ \\mathbb{R}^{n \times h}`` and an exponent matrix
+``E ∈ \\mathbb{N}^{q×h}_{≥0}``, a sparse polynomial zonotope ``\\mathcal{PZ} ⊂ \\mathbb{R}^n``.
```
schillic, line 7:
```suggestion
A sparse polynomial zonotope ``\\mathcal{PZ} ⊂ \\mathbb{R}^n`` is represented by
a constant offset ``c ∈ \\mathbb{R}^n``, a generator matrix ``G ∈ \\mathbb{R}^{n \times h}`` and an exponent matrix
``E ∈ \\mathbb{N}^{q×h}_{≥0}``.
```

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,105 @@
...
+
+### Fields
+
+- `c` -- constant offset vector
+- `G` -- generators matrix
+- `E` -- exponents matrix
```
schillic, line 13:
```suggestion
- `G` -- generator matrix
- `E` -- exponent matrix
```

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,105 @@
...
+"""
+struct SimpleSparsePolynomialZonotope{N, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}, ME<:AbstractMatrix{<:Integer}}
+    c::VN
+    G::MN
+    E::ME
+
+    function SimpleSparsePolynomialZonotope(c::VN, G::MN, E::ME) where {N<:Real, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}, ME<:AbstractMatrix{<:Integer}}
+        length(c) == size(G, 1) || throw(DimensionMismatch("c and G should have the same number of rows"))
+        size(G, 2) == size(E, 2) || throw(DimensionMismatch("G and E should have the same number of columns"))
+        all(>=(0), E) || throw(ArgumentError("E should have non-negative integers"))
```
schillic, line 28: Use `@assert` instead because we can then deactivate these checks.

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,105 @@
...
+const SSPZ = SimpleSparsePolynomialZonotope
+
+dim(p::SSPZ) = size(p.c, 1)
+ngens(p::SSPZ) = size(p.G, 2)
+nparams(p::SSPZ) = size(p.E, 1)
+order(p::SSPZ) = ngens(p) // dim(p)
+
+center(p::SSPZ) = p.c
+genmat(p::SSPZ) = p.G
+expmat(p::SSPZ) = p.E
```
schillic, line 43: I think `expmat` should be exported.

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,105 @@
...
+
+"""
+    overapproximate(p::SSPZ, ::Type{Zonotope}; nsdiv=1)
```
schillic, line 46: We usually use capital letters for one-letter sets.

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,105 @@
...
+
+Returns a zonotope containing ``p``.
+
+### Input
+
+- `p`     -- simple sparse polynomial zonotope
+- `nsdiv` -- (optional, default: `1`)
```
schillic, line 53:
```suggestion
- `nsdiv` -- (optional, default: `1`) size of uniform partitioning grid
```

src/Sets/SimpleSparsePolynomialZonotope.jl
schillic, line 46: These `overapproximate` methods can be moved to `Approximations/overapproximate.jl`.

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,105 @@
...
+        end
+    end
+    return Zonotope(cnew, Gnew)
+end
+
+"""
+    overapproximate(p::SSPZ, ::Type{Zonotope}, dom::IntervalBox)
+
+Compute the zonotope overapproximation of the given sparse polynomial zonotope over
+the parameter domain `dom`, which should be a subset of `[-1, 1]^q`, where `q = nparams(p)`.
```
schillic, line 80: Maybe add an `@assert` for "subset of `[-1, 1]^q`".

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,105 @@
...
+"""
+function overapproximate(p::SSPZ, ::Type{Zonotope}, dom::IntervalBox)
+    G = genmat(p)
+    E = expmat(p)
+    cnew = copy(center(p))
+    Gnew = similar(G)
+    @inbounds for (j, g) in enumerate(eachcol(G))
+        α = IA.Interval(1, 1)
+         for (i, vi) in enumerate(dom)
```
schillic, line 89:
```suggestion
        for (i, vi) in enumerate(dom)
```

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,105 @@
...
+        end
+        # α = mapreduce(x -> _fast_interval_pow(x[1],  x[2]), *, zip(dom, E[:, i])) # monomial value over the domain
+        m, r = midpoint_radius(α)
+        cnew .+= m * g
+        Gnew[:, j] .= abs.(g) * r
+    end
+    return Zonotope(cnew, Gnew)
+end
+
+function overapproximate(p::SSPZ, ::Type{UnionSetArray{Zonotope}}; nsdiv=100)
```
schillic, line 100: Add a kwarg `partition=nothing`.

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,105 @@
...
+- `c` -- constant offset vector
+- `G` -- generators matrix
+- `E` -- exponents matrix
+
+### Notes
+
+Sparse polynomial zonotopes were introduced in N. Kochdumper and M. Althoff.
+*Sparse Polynomial Zonotopes: A Novel Set Representation for Reachability Analysis*. Transactions on Automatic Control, 2021.
+"""
+struct SimpleSparsePolynomialZonotope{N, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}, ME<:AbstractMatrix{<:Integer}}
```
mforets, line 20: After rebasing, we can make this an `<: AbstractPolynomialZonotope{N}`.

src/Sets/SimpleSparsePolynomialZonotope.jl
lucaferranti: what's the style policy for this? Do you have all exports in the main file or each export in the file where the corresponding function is defined?

src/Sets/SimpleSparsePolynomialZonotope.jl
lucaferranti: so let's review this a second. Was the idea that `nsdiv` can be either one int or a tuple of ints and will indicate how many parts all (resp. each) dimension will be divided. While partition will specify how to divide each dimension (e..g allow non-uniform partitions? What would be a data structure to represent those?

src/Sets/SimpleSparsePolynomialZonotope.jl
schillic: We usually add it to the top of the same file in this package (and to the main package's file in most other packages I think).

src/Sets/SimpleSparsePolynomialZonotope.jl
schillic: Yes, the `Int` argument `4` would be a short-hand for `4 * ones(Int, n)` where `n` is the dimension. Does that answer the question?

lucaferranti: Yes, it answers the first question about `nsdiv`. Still not sure what to do with partition though... or is the idea to add the kwarg now and logic using it later?

mforets:
pls mark the PR as "ready for review" if that's the case

lucaferranti:
I still need to add a few tests (prob the examples in Niklas thesis) and I'm not sure about the partition thing (which can maybe be left for future pr?), otherwise it's close to ready :)

mforets:
> I still need to add a few tests (prob the examples in Niklas thesis) 

True; you can check those in https://github.com/JuliaReach/LazySets.jl/blob/master/test/Sets/DensePolynomialZonotope.jl as template. The polynomial zonotope from his slides should be fine.

> and I'm not sure about the partition thing 

Something like this?

```julia
function overapproximate(P::SSPZ, ::Type{UnionSetArray{Zonotope}}; nsdiv=100, partition=nothing)
    q = size(P.E, 1)
    dom = IntervalBox(IA.Interval(-1, 1), q)
    cells = mince(dom, isnothing(partition) ? fill(nsdiv, q) : partition)
    return UnionSetArray([overapproximate(P, Zonotope, c) for c in cells])
end
```

lucaferranti:
> Something like this?

doesn't this contradict what Christian answered above?

mforets (approved):

src/Approximations/overapproximate.jl
```diff
@@ -708,6 +709,78 @@ function overapproximate(cap::Intersection{N,
...
+"""
+    overapproximate(P::SimpleSparsePolynomialZonotope, ::Type{Zonotope}; nsdiv=1)
+
+Returns a zonotope containing ``P``.
+
+### Input
+
+- `P`     -- simple sparse polynomial zonotope
+- `nsdiv` -- (optional, default: `1`) size of uniform partitioning grid
+
```
schillic, line 726: Documentation for `partition` is missing.

src/Sets/Interval.jl
```diff
@@ -709,3 +709,41 @@ function chebyshev_center(x::Interval{N}; compute_radius::Bool=false) where {N}
...
+### Output
+
+A non-rigorous approximation of `a^n`.
+
+### Notes
+
+For a rigorous approximation with correct rounding,
+use `a^n` from `IntervalArithmetic.jl`.
+
+Review after IntervalArithmetic.jl#388
```
schillic, line 732: We should open an issue to not forget this.

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,100 @@
...
+
+const SSPZ = SimpleSparsePolynomialZonotope
+
+dim(P::SSPZ) = size(P.c, 1)
+ngens(P::SSPZ) = size(P.G, 2)
+
+"""
+    nparams(P::SimpleSparsePolynomialZonotope)
+
+Returns the number of variables used in the polynomial representation, this corresponds to
```
schillic, line 45:
```suggestion
Returns the number of variables used in the polynomial representation, which corresponds to
```

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,100 @@
...
+the number of rows in the exponents matrix.
+
+### Input
+
+- `P` -- simple sparse polynomial zonotope
+
+### Output
+
+number of variables in the polynomial representation of P.
```
schillic, line 54:
```suggestion
The number of variables in the polynomial representation of P.
```

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,100 @@
...
+Returns the matrix of exponents of the sparse polynomial zonotope. In the matrix, each
+row corresponds to a variable and each column to a monomial.
+
+### Input
+
+- `P` -- simple sparse polynomial zonotope
+
+### Output
+
+matrix of exponents, each column is a multidegree.
```
schillic, line 86:
```suggestion
The matrix of exponents, where each column is a multidegree.
```

test/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,19 @@
+@testset "Simple Sparse Polynomial Zonotope" begin
```
schillic, line 1: We usually put the `@testset` into `runtests.jl` and instead use a loop for different number types. Check one of the other test files.

Also, this file is not included in `runtests.jl` and hence these tests are not executed.

schillic (approved):
Looks good, only minor comments :+1:

docs/src/lib/sets/SimpleSparsePolynomialZonotope.md
```diff
@@ -0,0 +1,11 @@
+```@meta
+CurrentModule = LazySets
+```
+
+# [SimpleSparsePolynomialZonotope](@id def_SimpleSparsePolynomialZonotope)
+
+```@docs
+SimpleSparsePolynomialZonotope
+nparams
+expmat
```
mforets, line 10: The downside with only including docs for those methods and not for the others is that the page `SimpleSparsePolynomialZonotope` will not display most common methods that apply to it. There is no easy way to automate the behavior we'd like apart from adding those manually at the expense of some repetition (which is what we've currently been doing).

lucaferranti: So should I include also genmat order etc.?

schillic: Why do you get a repetition? You can just add the type annotation (see an example [here](https://github.com/JuliaReach/LazySets.jl/blob/master/docs/src/lib/sets/Zonotope.md)).

schillic (approved):

docs/src/lib/sets/SimpleSparsePolynomialZonotope.md
lucaferranti: I think Marcelo meant repetition as "you have to type it manually from every set". I don't mind the typing but fwiw I do think the current system (one docstring for each method even if they are all conceptually identical) makes docstrings very user-unfriendly if used from the REPL (which is where I use them generally, don't know if it's more common to look at docstrings from the docs or REPL)

schillic: > So should I include also genmat order etc.?

I would prefer that, yes. Otherwise there is no point in writing this documentation :)

mforets (approved):

PR #2976, opened 2022-06-10
Add plotting recipe for SPZ

### LINKED ISSUE #2966: Plotting mechanism for polynomial zonotopes
In principle the mechanism would be a replica of `overapproximate(X::TaylorModelReachSet, Zonotope; kwargs...)`, but without the time domain, ie. the relevant keywords are `dom` (to choose a subdomain of `[-1, 1]^n`, and `nsdiv`, `partition` to partition such domain. The result is either a zonotope or a union of zonotopes. We can have a similar method with `Hyperrectangle` as output.

cc: @lucaferranti

mforets:
Closes #2966.

### DIFF
--- a/src/Plotting/plot_recipes.jl
+++ b/src/Plotting/plot_recipes.jl
@@ -533,3 +533,16 @@ end
 
     return _plot_list_same_recipe(array(cup), ε)
 end
+
+@recipe function plot_polyzono(P::SimpleSparsePolynomialZonotope{N}, ε::N=N(PLOT_PRECISION); nsdiv=100, partition=nothing) where {N}
+    label --> DEFAULT_LABEL
+    grid --> DEFAULT_GRID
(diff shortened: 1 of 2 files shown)


schillic (approved):

lucaferranti (approved):
FYI, there's [this trick](https://discourse.julialang.org/t/how-to-test-plot-recipes/2648/6) to test plotting recipes. Probably not the most robust way, but can be useful to boost coverage (which is after all the main purpose of tests)

schillic:
> FYI, there's [this trick](https://discourse.julialang.org/t/how-to-test-plot-recipes/2648/6) to test plotting recipes. Probably not the most robust way, but can be useful to boost coverage (which is after all the main purpose of tests)

We do have such plot tests [here](https://github.com/JuliaReach/LazySets.jl/blob/master/test/Utils/plot.jl).

PR #2978, opened 2022-06-12
operations on simple sparse polynomial zonotopes

lucaferranti:
This PR adds the operations on SSPZ described in Niklas presentation. Currently I added those into the same file of SSPZ, but maybe we want to move those in the corresponding files in `concrete_operations`?

### TODO

- [x] ~~linear combination~~ in future PR
- [x] ~~quadratic map~~ in future PR
- [x] ~~convex hull~~ in future PR
- [x] tests

### DIFF
--- a/docs/src/lib/sets/SimpleSparsePolynomialZonotope.md
+++ b/docs/src/lib/sets/SimpleSparsePolynomialZonotope.md
@@ -13,4 +13,7 @@ dim(::SimpleSparsePolynomialZonotope)
 ngens(::SimpleSparsePolynomialZonotope)
 nparams(::SimpleSparsePolynomialZonotope)
 order(::SimpleSparsePolynomialZonotope)
+linear_map(::AbstractMatrix, ::SimpleSparsePolynomialZonotope)
+minkowski_sum(::SimpleSparsePolynomialZonotope, ::SimpleSparsePolynomialZonotope)
+cartesian_product(::SimpleSparsePolynomialZonotope, ::SimpleSparsePolynomialZonotope)
 ```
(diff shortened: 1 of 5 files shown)


lucaferranti:
> Looks good after moving the binary methods to the respective files 👍

among the functions implemented here, it seems only `cartesian_product` and `minkowski_sum` have a corresponding file in `ConcreteOperations`, should I create a new file for `linear_map` or leave it under `SimpleSparsePolynomialZonotope.jl` ?

schillic:
`linear_map` is not binary and hence we keep it in the same file.

lucaferranti:
moved the files. Out of curiosity, does LazySets rely on promotion mechanisms or how should it handles things like e.g.

```julia
minkowski_sum(::SimpleSparsePolynomialZonotope, ::Star)
```

schillic (approved):

schillic:
> moved the files. Out of curiosity, does LazySets rely on promotion mechanisms or how should it handles things like e.g.
> 
> ```julia
> minkowski_sum(::SimpleSparsePolynomialZonotope, ::Star)
> ```

That would require adding a specific method and is out of scope for now.

lucaferranti:
btw currently the following fails

```julia
linear_map(I, S)
```
where `S` is SSPZ, the reason for this is that `I` is uniform scaling and it's not a subtype of `AbstractMatrix`, that's arguably a pretty boring linear map to apply. Is it worth to change the signature for that?

schillic:
> Is it worth to change the signature for that?

I do not think so.

mforets (approved):

PR #2979, opened 2022-06-15
linear combination and convex hull on sspz

lucaferranti:


src/ConcreteOperations/convex_hull.jl
```diff
@@ -595,3 +595,38 @@ end
...
+"""
+    convex_hull(P::SimpleSparsePolynomialZonotope)
+
+computes the convex hull of the simple spare polynomial zonotope `P`.
+
+### Input
+
+- `P` -- simple sparse polynomial zonotopes
+
+### Outuput
```
schillic, line 626:
```suggestion
Computes the convex hull of the simple sparse polynomial zonotopes `P1` and `P2`.

### Input

- `P1` : simple sparse polynomial zonotopes
- `P2` : simple sparse polynomial zonotopes

### Output

Tightest convex simple sparse polynomial zonotope containing `P1` and `P2`.
"""
function convex_hull(P1::SimpleSparsePolynomialZonotope, P2::SimpleSparsePolynomialZonotope)
    return linear_combination(linear_combination(P1, P1), linear_combination(P2, P2))
end

"""
    convex_hull(P::SimpleSparsePolynomialZonotope)

Computes the convex hull of the simple sparse polynomial zonotope `P`.

### Input

- `P` -- simple sparse polynomial zonotopes

### Output
```

src/ConcreteOperations/linear_combination.jl
```diff
@@ -0,0 +1,29 @@
...
+
+### Input
+
+- `P1` -- simple sparse polynomial zonotope
+- `P2` -- simple sparse polynomial zonotope
+
+### Output
+
+linear combination of `P1` and `P2`.
+"""
```
schillic, line 16: I think this will be the only implementation of the function in this library, so maybe add the definition of the resulting set here. At least I would not know immediately what `linear_combination` means.

```suggestion
### Output

linear combination of `P1` and `P2`.

### Notes

...
"""
```

src/ConcreteOperations/linear_combination.jl
```diff
@@ -0,0 +1,29 @@
...
+function linear_combination(P1::SimpleSparsePolynomialZonotope, P2::SimpleSparsePolynomialZonotope)
+    c = 0.5 * (center(P1) + center(P2))
+    G = 0.5 * hcat(center(P1) - center(P2), genmat(P1), genmat(P1), genmat(P2), -genmat(P2))
+    E1, E2 = expmat(P1), expmat(P2)
+    et = promote_type(eltype(E1), eltype(E2))
+    E = hcat(vcat(zeros(et, nparams(P1) + nparams(P2)), one(et)),
+             vcat(E1, zeros(et, nparams(P2) + 1, ngens(P1))),
+             vcat(E2, zeros(et, nparams(P2), ngens(P1)), ones(et, 1, ngens(P1))),
+             vcat(zeros(et, nparams(P1), ngens(P2)), E2, zeros(et, 1, ngens(P2))),
+             vcat(zeros(et, nparams(P1), ngens(P2)), E2, ones(et, 1, ngens(P2))))
```
schillic, line 26: Hm, this does not exactly look "sparse" :sweat_smile:

lucaferranti: actually it seems that concatenating a sparse matrix with a dense matrix is "smart enough" to avoid converting to dense and storing zeros

```julia
julia> c = [2.0, 0.0]
2-element Vector{Float64}:
 2.0
 0.0

julia> G = [1 2;2 2.]
2×2 Matrix{Float64}:
 1.0  2.0
 2.0  2.0

julia> E = SparseMatrixCSC([1 4;1 2])
2×2 SparseMatrixCSC{Int64, Int64} with 4 stored entries:
 1  4
 1  2

julia> S = SimpleSparsePolynomialZonotope(c, G, E)
SimpleSparsePolynomialZonotope{Float64, Vector{Float64}, Matrix{Float64}, SparseMatrixCSC{Int64, Int64}}([2.0, 0.0], [1.0 2.0; 2.0 2.0], sparse([1, 2, 1, 2], [1, 1, 2, 2], [1, 1, 4, 2], 2, 2))

julia> P = linear_combination(S, S)
SimpleSparsePolynomialZonotope{Float64, Vector{Float64}, Matrix{Float64}, SparseMatrixCSC{Int64, Int64}}([2.0, 0.0], [0.0 0.5 … -0.5 -1.0; 0.0 1.0 … -1.0 -1.0], sparse([5, 1, 2, 1, 2, 1, 2, 5, 1, 2  …  3, 4, 3, 4, 3, 4, 5, 3, 4, 5], [1, 2, 2, 3, 3, 4, 4, 4, 5, 5  …  6, 6, 7, 7, 8, 8, 8, 9, 9, 9], [1, 1, 1, 4, 2, 1, 1, 1, 4, 2  …  1, 1, 4, 2, 1, 1, 1, 4, 2, 1], 5, 9))

julia> expmat(P)
5×9 SparseMatrixCSC{Int64, Int64} with 21 stored entries:
 ⋅  1  4  1  4  ⋅  ⋅  ⋅  ⋅
 ⋅  1  2  1  2  ⋅  ⋅  ⋅  ⋅
 ⋅  ⋅  ⋅  ⋅  ⋅  1  4  1  4
 ⋅  ⋅  ⋅  ⋅  ⋅  1  2  1  2
 1  ⋅  ⋅  1  1  ⋅  ⋅  1  1
```

but I definitely agree the code is unidiomatic at best, not sure what's the best way to write it. Maybe allocate the finale exponents matrix first and then fill the non-zero elements?

src/ConcreteOperations/linear_combination.jl
schillic: Of course add something for the `...` :wink:

src/ConcreteOperations/linear_combination.jl
schillic: I do not know. Typically you can outsource to a helper function and add a method for sparse matrices.

schillic: But then again, this is probably not so important. We can keep it for now. I was just concerned that you would get a dense matrix, but you showed that this does not happen.

src/ConcreteOperations/linear_combination.jl
lucaferranti: Oh, I didn't notice that 😄 I guess you mean to add the definition of linear combination of two sets, since it's not necessarily known to the user?

src/ConcreteOperations/linear_combination.jl
schillic: Just for the record, the implementation looks correct :+1:

schillic:
For the docs error you need to add the two methods [here](https://github.com/JuliaReach/LazySets.jl/blob/master/docs/src/lib/binary_functions.md#convex-hull) and add a new block for `linear_combination`.

schillic (approved):

mforets:
@lucaferranti  can you pls add some tests for the new functionality?

src/ConcreteOperations/linear_combination.jl
```diff
@@ -0,0 +1,37 @@
...
+### Output
+
+linear combination of `P1` and `P2`.
+
+### Notes
+
+The linear combination of two sets ``P₁`` and ``P₂`` is defined as
+
+```math
+\\{λp₁ + (1-λ)p₂ | p₁ ∈ P₁, p₂ ∈ P₂, λ ∈ [0, 1]\\}
```
mforets, line 22: This is correct but I would rather use the convention in https://github.com/JuliaReach/juliareach-days-3-reachathon/blob/master/Challenge_5/Challenge5_PolynomialZonotopes.pdf , so it's easier to read the implementation. Or you can just add a second sentence saying that by shifting the domain from `[0, 1]` to `[-1, 1]`, one gets `\\{1/2(1+λ)p₁ + 1/2(1-λ)p₂ | p₁ ∈ P₁, p₂ ∈ P₂, λ ∈ [-1, 1]\\}`

src/ConcreteOperations/linear_combination.jl
```diff
@@ -0,0 +1,37 @@
...
+```
+"""
+function linear_combination(P1::SimpleSparsePolynomialZonotope, P2::SimpleSparsePolynomialZonotope)
```
mforets, line 25: Can you define some auxiliary variables `c1 = center(P1)` and `c2 = center(P2)`, and same for the matrices? I think then the formulas become more readable, eg. we can write `G = 1/2 * [c1-c2  G1  G1  G2  -G2]` as in paper.

src/ConcreteOperations/linear_combination.jl
lucaferranti: The docstring is definitely not the place to do the math and derive one formulation from the other. I think the one I used is more common? (At least it's the one I've always encountered when talking about convex sets and linear combination) Anyway, I can change it

test/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -32,4 +32,25 @@ for N in [Float64, Float32, Rational{Int}]
     @test genmat(CPS) == [1 2 0 0;2 2 0 0;0 0 1 2;0 0 2 2.]
     @test expmat(CPS) == [1 4 0 0;1 2 0 0;0 0 1 4;0 0 1 2]
 
+    _c = N[2.0, 0.0]
```
schillic, line 35: It is okay now, but typically we do not use underscores for variables. Since we have the outer loops in the test files, there are usually no scoping problems.

schillic (approved):

src/ConcreteOperations/linear_combination.jl
```diff
@@ -0,0 +1,41 @@
...
+
+    c = 0.5 * (c1 + c2)
+    G = 0.5 * hcat(c1 - c2, G1, G1, G2, -G2)
+
+    et = promote_type(eltype(E1), eltype(E2))
+    E = hcat(vcat(zeros(et, nparams(P1) + nparams(P2)), one(et)),
+             vcat(E1, zeros(et, nparams(P2) + 1, ngens(P1))),
+             vcat(E2, zeros(et, nparams(P2), ngens(P1)), ones(et, 1, ngens(P1))),
+             vcat(zeros(et, nparams(P1), ngens(P2)), E2, zeros(et, 1, ngens(P2))),
+             vcat(zeros(et, nparams(P1), ngens(P2)), E2, ones(et, 1, ngens(P2))))
```
mforets, line 38: Nitpicking: we generally use `N` or `T` for numeric types.

```suggestion
    N = promote_type(eltype(E1), eltype(E2))
    E = hcat(vcat(zeros(N, nparams(P1) + nparams(P2)), one(N)),
             vcat(E1, zeros(N, nparams(P2) + 1, ngens(P1))),
             vcat(E2, zeros(N, nparams(P2), ngens(P1)), ones(N, 1, ngens(P1))),
             vcat(zeros(N, nparams(P1), ngens(P2)), E2, zeros(N 1, ngens(P2))),
             vcat(zeros(N, nparams(P1), ngens(P2)), E2, ones(N, 1, ngens(P2))))
```

src/ConcreteOperations/linear_combination.jl
mforets: In this case, for instance, we would keep the initial version that you had written (because it is the most common one) and add a little `# Notes` paragraph with the observation that implementation-wise we use the normalized interval representation.

src/ConcreteOperations/linear_combination.jl
```diff
@@ -0,0 +1,41 @@
...
+    E1, E2 = expmat(P1), expmat(P2)
+
+    c = 0.5 * (c1 + c2)
+    G = 0.5 * hcat(c1 - c2, G1, G1, G2, -G2)
+
+    N = promote_type(eltype(E1), eltype(E2))
+    E = hcat(vcat(zeros(N, nparams(P1) + nparams(P2)), one(N)),
+             vcat(E1, zeros(N, nparams(P2) + 1, ngens(P1))),
+             vcat(E2, zeros(N, nparams(P2), ngens(P1)), ones(N, 1, ngens(P1))),
+             vcat(zeros(N, nparams(P1), ngens(P2)), E2, zeros(N 1, ngens(P2))),
```
mforets, line 37:
```suggestion
             vcat(zeros(N, nparams(P1), ngens(P2)), E2, zeros(N, 1, ngens(P2))),
```

mforets (approved):

src/ConcreteOperations/linear_combination.jl
```diff
@@ -0,0 +1,41 @@
...
+    G1, G2 = genmat(P1), genmat(P2)
+    E1, E2 = expmat(P1), expmat(P2)
+
+    c = 0.5 * (c1 + c2)
+    G = 0.5 * hcat(c1 - c2, G1, G1, G2, -G2)
+
+    N = promote_type(eltype(E1), eltype(E2))
+    E = hcat(vcat(zeros(N, nparams(P1) + nparams(P2)), one(N)),
+             vcat(E1, zeros(N, nparams(P2) + 1, ngens(P1))),
+             vcat(E2, zeros(N, nparams(P2), ngens(P1)), ones(N, 1, ngens(P1))),
```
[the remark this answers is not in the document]
schillic: Maybe you fixed it locally, but this fix was not pushed.

PR #2991, opened 2022-07-08
add conversion from zonotope to sspz

lucaferranti:


src/convert.jl
```diff
@@ -1225,3 +1225,19 @@ function convert(::Type{Hyperplane}, P::HPolyhedron; skip_check::Bool=false)
     c1 = @inbounds first(constraints_list(P))
     return Hyperplane(c1.a, c1.b)
 end
+
+
+"""
+    convert(::Type{SimpleSparsePolynomialZonotope}, P::Zonotope)
+
+convert a zonotope to a simple sparse polynomial zonotope.
+
```
schillic, line 1234: Can you check that this source is correct?

```suggestion
Convert a zonotope to a simple sparse polynomial zonotope.

### Algorithm

This method implements Proposition 3 in [1].

[1] Kochdumper, Althoff. *Sparse polynomial zonotopes - a novel set
representation for reachability analysis*. 2021
```

src/convert.jl
```diff
@@ -1225,3 +1225,19 @@ function convert(::Type{Hyperplane}, P::HPolyhedron; skip_check::Bool=false)
...
+"""
+function convert(::Type{SimpleSparsePolynomialZonotope}, P::Zonotope)
+    c = center(P)
+    G = genmat(P)
+    n = ngens(P)
+    E = 1 * Matrix(I, n, n)
```
schillic, line 1240:
```suggestion
    E = Matrix(1 * I, n, n)
```
This is more efficient because it does not allocate a `Bool` matrix.

```julia
julia> @time 1 * Matrix(I, n, n)
  0.000013 seconds (2 allocations: 160 bytes)
2×2 Matrix{Int64}:
 1  0
 0  1

julia> @time Matrix(1 * I, n, n)
  0.000008 seconds (1 allocation: 96 bytes)
2×2 Matrix{Int64}:
 1  0
 0  1
``

**Review by schillic — APPROVED** [review]

File: src/convert.jl [hunk 3]
```diff
@@ -1225,3 +1225,25 @@ function convert(::Type{Hyperplane}, P::HPolyhedron; skip_check::Bool=false)
...
+Convert a zonotope to a simple sparse polynomial zonotope.
+
+### Algorithm
+
+This method implements Proposition 3 in [1].
+
+[1] Kochdumper, Althoff. *Sparse polynomial zonotopes - a novel set
+representation for reachability analysis*. 2021
+"""
+function convert(::Type{SimpleSparsePolynomialZonotope}, P::Zonotope)
```
mforets, line 1242: Let's dispatch on `P::AbstractZonotope` since they define `center`, `genmat` and `ngens` (should do if not). This way we can reuse this method for `Hyperrectangle` and other zonotopic sets.

schillic: I would also rename `P` to `Z`.

schillic (approved):

src/convert.jl
```diff
@@ -1225,3 +1225,25 @@ function convert(::Type{Hyperplane}, P::HPolyhedron; skip_check::Bool=false)
     c1 = @inbounds first(constraints_list(P))
     return Hyperplane(c1.a, c1.b)
 end
+
+
+"""
+    convert(::Type{SimpleSparsePolynomialZonotope}, P::Zonotope)
```
mforets, line 1231:
```suggestion
    convert(::Type{SimpleSparsePolynomialZonotope}, Z::Zonotope)
```

mforets (approved):

PR #2998, opened 2022-07-11
add docs preview

lucaferranti:


lucaferranti:
seems to work 😎

schillic (approved):
We have an almost identical script [here](https://github.com/JuliaReach/ReachabilityAnalysis.jl/blob/47ac9bba4d8dff02edffbb1cf7e6dc5d150d76d5/.github/workflows/clean-gh-pages.yml). The differences are that the script here first checks if the `previews` folder exists and also pushes in the same step (while the other script pushes in a separate step). I am not sure if this is necessary, but it looks reasonable to me.

schillic:
I had to allow force-push on the `gh-pages` branch.

PR #3007, opened 2022-07-15
check convexity in line segment

mforets:


schillic (approved):

lucaferranti (approved):

docs/src/lib/binary_functions.md
```diff
@@ -122,7 +122,7 @@ issubset
 ⊆(::AbstractSingleton, ::AbstractSingleton, ::Bool=false)
 ⊆(::Ball2, ::Ball2, ::Bool=false)
 ⊆(::Union{Ball2, Ballp}, ::AbstractSingleton, ::Bool=false)
-⊆(::LineSegment, ::ConvexSet, ::Bool=false)
+⊆(::LineSegment{N}, ::ST, ::Bool=false) where {ST<:LazySet{N}}
```
schillic, line 125:
```suggestion
⊆(::LineSegment{N}, ::ST, ::Bool=false) where {N, ST<:LazySet{N}}
```

lucaferranti: I think in this case it might be better
```julia
⊆(::LineSegment, ::ST, ::Bool=false) where {ST<:LazySet}
```

unused type parameters can make the function slower, see ⊆(::LineSegment{N}, ::ST, ::Bool=false) where {N, ST<:LazySet{N}}

lucaferranti: since the conversation was resolved about while I was writing and not sure how notifications work in that case, pinging @schillic @mforets to make sure the comment doesn't go unnoticed

schillic: Yes, I just wanted to fix the build. But now there is an ambiguity, so @mforets maybe you can remove the type parameter.

src/ConcreteOperations/issubset.jl
```diff
@@ -493,6 +493,11 @@ function ⊆(L::LineSegment, S::ConvexSet, witness::Bool=false)
     return _issubset_line_segment(L, S)
```
lucaferranti, line 493:
```suggestion
    return _issubset_line_segment(L, S, witness)
```

unrelated to the scope of this PR (and hence maybe can be left for a separated one) but currently `witness` is not passed to the inner function, I think it should be this way? (this also prob means the option with `witness=true` is not tested) Also this is type-unstable in the current form

lucaferranti:
a `@test_throw` for the fallback could be nice to have, otherwise LGTM

src/ConcreteOperations/issubset.jl
```diff
@@ -489,12 +489,15 @@ optionally compute a witness.
...
-function ⊆(L::LineSegment{N}, S::ST, witness::Bool=false) where {N, ST<:LazySet{N}}
-    isconvextype(ST) || throw(ArgumentError("this function requires the set type to be convex, but " *
-                                            "it is not the case for a $ST"))
+function ⊆(L::LineSegment, S::ConvexSet, witness::Bool=false)
     return _issubset_line_segment(L, S)
 end
 
+function ⊆(L::LineSegment, S::LazySet, witness::Bool=false)
+    throw(ArgumentError("this function requires the set type to be convex, but " *
+    "it is not the case for a $ST"))
```
schillic, line 498:
```suggestion
    return _issubset_line_segment(L, S, witness)
end

function ⊆(L::LineSegment, S::LazySet, witness::Bool=false)
    throw(ArgumentError("this function requires the set type to be convex, but " *
    "it is not the case for a $(typeof(S))"))
```

lucaferranti (approved):
(not sure if I count as reviewer anymore given my last change, but...) LGTM, a couple of things that could be added

1. A `test_throw` for the fallback case (arguably quite pedantic)
2. A test for the case `witness=true` (falls out of the original scope of the PR, but since the change was incorporated)

mforets:
> a couple of things that could be added

i opened https://github.com/JuliaReach/LazySets.jl/issues/3008

PR #3015, opened 2022-07-27
initial implementation of SparsePolynomialZonotope

lucaferranti:
- [x] Type and constructor
- [x] `dim`
- [x] `ngens` (two methods for dependent and independent generators and one for the total numbers)
- [x] `nparams`
- [x] `order`
- [x] `center`
- [x] `genmat` (similar to `ngens`)
- [x] `expmat`
- [x] `indexvector`
- [x] `linear_map`

In principle, the implementation of the methods are very similar to SSPZ (and indeed it could make sense to try to avoid boilerplate code as much as possible), so we could either implement eveyrthing here or in separate PRs


docs/src/lib/sets/SparsePolynomialZonotope.md
```diff
@@ -0,0 +1,21 @@
...
+SparsePolynomialZonotope
+center(::SparsePolynomialZonotope)
+dependent_genmat(::SparsePolynomialZonotope)
+independent_genmat(::SparsePolynomialZonotope)
+expmat(::SparsePolynomialZonotope)
+indexvector(P::SparsePolynomialZonotope)
+dim(::SparsePolynomialZonotope)
+ndependentgens(::SparsePolynomialZonotope)
+nindependentgens(::SparsePolynomialZonotope)
+ngens(::SparsePolynomialZonotope)
```
schillic, line 17: I wonder if we should have `genmat` as for `SimpleSparsePolynomialZonotope` and `AbstractZonotope`. Then if you `convert` such a set to an SPZ, the same function works.
As for `ngens` defined as the sum, this is indeed a generalization because there would be 0 independent generators, so the results coincide.

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,296 @@
+export SparsePolynomialZonotope, expmat, nparams, ndependentgens, nindependentgens, ngens,
+        dependent_genmat, independent_genmat, indexvector,
```
schillic, line 2:
```suggestion
       dependent_genmat, independent_genmat, indexvector,
```

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,296 @@
...
+    SparsePolynomialZonotope{N, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}, MNI<:AbstractVecOrMat{N}, ME<:AbstractMatrix{<:Integer}, VI<:AbstractVector{<:Integer}} <: AbstractPolynomialZonotope{N}
+
+Type that represents a sparse polynomial zonotope.
+
+A sparse polynomial zonotope ``\\mathcal{PZ} ⊂ \\mathbb{R}^n`` is represented by the set
+```math
+\\mathcal{PZ} = \\left\\{x \\in \\mathbb{R}^n : x = c + ∑ᵢ₌₁ʰ\\left(∏ₖ₌₁ᵖ α_k^{E_{k, i}} \\right)Gᵢ+∑ⱼ₌₁^qβⱼGIⱼ,~~ α_k ∈ [-1, 1]~~ ∀ i = 1,…,p, j=1,…,q \\right\\},
+```
+where ``c ∈ \\mathbb{R}^n`` is the offset vector (or center), ``G ∈ \\mathbb{R}^{n \\times h}`` is the dependent generator matrix with columns ``Gᵢ``,
+``GI ∈ \\mathbb{R}^{n×q}`` is the independent generator matrix and where ``E ∈ \\mathbb{N}^{p×h}_{≥0}`` is the exponent matrix with matrix elements ``E_{k, i}``.
```
schillic, line 15:
```suggestion
``GI ∈ \\mathbb{R}^{n×q}`` is the independent generator matrix and ``E ∈ \\mathbb{N}^{p×h}_{≥0}`` is the exponent matrix with matrix elements ``E_{k, i}``.
```

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,296 @@
...
+
+### Fields
+
+- `c` -- offset vector
+- `G` -- dependent generator matrix
+- `GI` -- independent generator matrix
+- `E` -- exponent matrix
+- `idx` -- identifiers vector
```
schillic, line 23:
```suggestion
- `c`   -- offset vector
- `G`   -- dependent generator matrix
- `GI`  -- independent generator matrix
- `E`   -- exponent matrix
- `idx` -- identifier vector
```

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,296 @@
...
+                                ME<:AbstractMatrix{<:Integer},
+                                VI<:AbstractVector{<:Integer}} <: AbstractPolynomialZonotope{N}
+    c::VN
+    G::MN
+    GI::MNI
+    E::ME
+    idx::VI
+
+    function SparsePolynomialZonotope(c::VN, G::MN, GI::MNI, E::ME, idx::VI) where {N<:Real, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}, MNI<:AbstractVecOrMat{N}, ME<:AbstractMatrix{<:Integer}, VI<:AbstractVector{<:Integer}}
+        @assert length(c) == size(G, 1) throw(DimensionMismatch("c and G should have the same number of rows"))
```
schillic, line 44: Cool, I did not know that you can throw other errors in assertions.

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,296 @@
...
+        @assert all(>(0), idx) throw(ArgumentError("identifiers in index vector must be positive integers"))
+        return new{N, VN, MN, MNI, ME, VI}(c, G, GI, E, idx)
+    end
+end
+
+function SparsePolynomialZonotope(c::AbstractVector, G::AbstractMatrix, GI::AbstractVecOrMat, E::AbstractMatrix{<:Integer})
+    n = size(E, 1)
+    idx = collect(1:n)
+    return SparsePolynomialZonotope(c, G, GI, E, idx)
+end
```
schillic, line 57:
```julia
julia> 1:4 isa AbstractVector{<:Integer}
true
```
Is there a reason to not use `1:n` directly?

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,296 @@
...
+
+### Input
+
+- `P` -- sparse polynomial zonotope
+
+### Output
+
+The ambient dimension of `P`.
+"""
+dim(P::SPZ) = size(P.c, 1)
```
schillic, line 75:
```suggestion
dim(P::SPZ) = length(P.c)
```

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,296 @@
...
+
+"""
+ndependentgens(P::SparsePolynomialZonotope)
```
schillic, line 78:
```suggestion
    ndependentgens(P::SparsePolynomialZonotope)
```

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,296 @@
...
+Return the number of dependent generators of `P`.
+
+### Input
+
+- `P` -- sparse polynomial zonotope
+"""
+ndependentgens(P::SPZ) = size(P.G, 2)
+
+"""
+nindependentgens(P::SparsePolynomialZonotope)
```
schillic, line 89:
```suggestion
    nindependentgens(P::SparsePolynomialZonotope)
```

src/Sets/SparsePolynomialZonotope.jl
schillic: The identifier vector is not explained in the text. You can copy the description from the getter.

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,296 @@
...
+indexvector(P::SPZ) = P.idx
+
+"""
+    linear_map(M::AbstractMatrix, P::SparsePolynomialZonotope)
+
+Apply the linear map `M` to the sparse polynomial zonotope `P`.
+
+### Input
+
+- `M` -- square matrix with size(M) == dim(P)
```
schillic, line 226:
```suggestion
- `M` -- square matrix with `size(M) == dim(P)`
```

test/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,31 @@
+for N in [Float64, Float32, Rational{Int}]
+
+    # example from Niklas thesis (page )
```
schillic, line 3: add page

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,296 @@
...
+function linear_map(M::AbstractMatrix, P::SPZ)
+    return SparsePolynomialZonotope(M * center(P),
+                                    M * dependent_genmat(P),
+                                    M * independent_genmat(P),
+                                    expmat(P),
+                                    indexvector(P)
+                                    )
+end
+
+# function quadratic_map(Q::Vector{MT}, S::SparsePolynomialZonotope) where {N, MT<:AbstractMatrix{N}}
```
schillic, line 242: There is some commented code at the end of the file. Is this intended?

schillic (commented):
LGTM!

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,296 @@
...
+        @assert size(G, 2) == size(E, 2) throw(DimensionMismatch("G and E should have the same number of columns"))
+        @assert all(>=(0), E) throw(ArgumentError("E should have non-negative integers"))
+        @assert all(>(0), idx) throw(ArgumentError("identifiers in index vector must be positive integers"))
+        return new{N, VN, MN, MNI, ME, VI}(c, G, GI, E, idx)
+    end
+end
+
+function SparsePolynomialZonotope(c::AbstractVector, G::AbstractMatrix, GI::AbstractVecOrMat, E::AbstractMatrix{<:Integer})
+    n = size(E, 1)
+    idx = collect(1:n)
```
lucaferranti, line 55:
```suggestion
```

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -0,0 +1,296 @@
...
+    return SparsePolynomialZonotope(c, G, GI, E, idx)
```
lucaferranti, line 56:
```suggestion
    return SparsePolynomialZonotope(c, G, GI, E, 1:n)
```

test/Sets/SparsePolynomialZonotope.jl
lucaferranti, line 3:
```suggestion
    # example from Niklas thesis (page 32)
```

src/Sets/SparsePolynomialZonotope.jl
lucaferranti: not really, updated

src/Sets/SparsePolynomialZonotope.jl
lucaferranti: I created this file from `SparsePolynomialZonotopes.jl` and then adapted. `quadratic_map` is not implemented for SPZ, but I just commented this out instead of deleting it to leave it as a starting point for the quadratic map, I can also delete it and just copy paste the base code when the time comes

docs/src/lib/sets/SparsePolynomialZonotope.md
lucaferranti: I also wondered whether we should define `genmat` for `SparsePolynomialZonotope`, I thought a logical definition would be to have 

```julia
genmat(P::SparsePolynomialZonotope) = hcat(P.G, P.GI)
```

but I'm not sure whether this would be useful.

Since `SparsePolynomialZonotope` is not a subset of `AbstractZonotope`, I think the conversion to SPZ would still work?

src/Sets/SparsePolynomialZonotope.jl
schillic: Okay, just wanted to make sure you did not forget something.

docs/src/lib/sets/SparsePolynomialZonotope.md
schillic: > Since `SparsePolynomialZonotope` is not a subset of `AbstractZonotope`, I think the conversion to SPZ would still work?

If anything, it would be the other way around.
This is not a problem, just slightly inconvenient that something does not work anymore that used to work before the conversion. I am also fine to not have `genmat` (or to add it later if we find it useful).

The concatenation would be consistent with `ngens`. I originally thought about just the matrix that is used when converting from a `Zonotope`, but I do not like the inconsistency with `ngens` then.

lucaferranti: added `genmat` to return the concatenation of the generator matrices

schillic: I was thinking about it again and see only now that there is probably no application of the concatenation, sorry :sweat:
Let's see what @mforets says.

schillic (approved):

schillic (approved):

mforets (approved):

schillic (approved):

PR #3027, opened 2022-08-06
add mergeID

lucaferranti:
this is probably a brute force implementation of the pseudocode in niklas thesis (page 35), not sure if one can squeeze some more performance... I doubt that would be the bottleneck in any application anyway

Another question, should we have also an in-place version?


src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -207,6 +207,43 @@ indexvector(P::SPZ) = P.idx
...
+two polynomial zonotopes with the same index vector
+"""
+function mergeID(P1::SparsePolynomialZonotope, P2::SparsePolynomialZonotope)
+
+    idx1 = indexvector(P1)
+    idx2 = indexvector(P2)
+    idx1 == idx2 && return (P1, P2)
+
+    H = setdiff(idx2, idx1)
+    idx_new = vcat(idx1, H)
```
schillic, line 233: I am not sure this is correct. It seems non-symmetric. Does the test still work in the other order?

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -207,6 +207,43 @@ indexvector(P::SPZ) = P.idx
...
+    @show idx_new
```
schillic, line 234: Remove this.

src/Sets/SparsePolynomialZonotope.jl
lucaferranti: I think it's symmetric up to permutation. That is the merged zonotopes will have the same parameters, but these will not necessarily ordered by index (i.e. the index vector is not ordered). If one also sorted the index vector, then it would be symmetric I think

src/Sets/SparsePolynomialZonotope.jl
lucaferranti, line 234:
```suggestion
```

lucaferranti: now you know how I actually debug code :) 😅

mforets: You should give `Infiltrator.jl` a try. There's no way back!

src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -207,6 +207,42 @@ indexvector(P::SPZ) = P.idx
...
+    H = setdiff(idx2, idx1)
+    idx_new = vcat(idx1, H)
+    E1 = vcat(expmat(P1), zeros(Int, length(H), ndependentgens(P1)))
+    E2 = zeros(Int, length(idx1) + length(H), ndependentgens(P2))
+    E2_old = expmat(P2)
+    @inbounds for (i, idx) in enumerate(idx_new)
+        j = findfirst(isequal(idx), idx2)
+        !isnothing(j) && (E2[i, :] .= E2_old[j, :])
+    end
+    P1_new = SparsePolynomialZonotope(center(P1), dependent_genmat(P1), independent_genmat(P1), E1, idx_new)
```
mforets, line 241: unrelated but i find it cleaner to read (also to find by `genmat[TAB]`
```julia
 SparsePolynomialZonotope(center(P1), genmat_dep(P1), genmat_indep(P1), E1, idx_new)
```
see https://github.com/JuliaReach/LazySets.jl/issues/3031

lucaferranti: should I fix that first and then rebase the open PRs or implement the renaming after these PRs are merged?

mforets: we can do that later

src/Sets/SparsePolynomialZonotope.jl
lucaferranti: at least looking at the algorithm description on page 35 of Niklas' thesis, the algorithm doesn't seem to be symmetric (because it doesn't fix the order)

src/Sets/SparsePolynomialZonotope.jl
schillic: This line is still there.

src/Sets/SparsePolynomialZonotope.jl
schillic: So when you do `mergeID(PZ2, PZ1)` in the test case, you get the IDs `[2, 4, 1, 3]`. Is this acceptable or should the indices be sorted?

src/Sets/SparsePolynomialZonotope.jl
lucaferranti, line 234:
```suggestion
```

lucaferranti: strange, I thought I had removed, now should be gone

src/Sets/SparsePolynomialZonotope.jl
lucaferranti: based on today's conversation, should we normalize the order?

lucaferranti: if in functions that require the same idvector (quadratic map, exact sum) you so something like

```julia
indexvector(PZ1) == indexvector(PZ2) || (PZ1, PZ2 = mergeID(PZ1, PZ2)
```

I think even without normalizing the order it would be correct, because calling `mergeID` would not change `PZ1, PZ2`, on the other hand, I don't see harm in normalizing the order either

schillic: I would not be surprised if some operations assume the same order after `mergeID`. So I would sort.

mforets:
What is the status of this PR?

schillic:
Closing because we probably do not need this.

PR #3028, opened 2022-08-07
add Zonotope enclosure of SPZ

lucaferranti:
add Zonotope enclosure of SPZ

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -715,7 +715,7 @@ end
 # ================================================
 
 #=
-Given center, (dependent) generator matrix and exponent matrix of a (simple) sparse polynomialzonotope,
+Given center, (dependent) generator matrix and exponent matrix of a (simple) sparse polynomial zonotope,
 compute thew new center and generator matrix of its zonotope overapproximation.
 =#
 function _zonotope_overapprox(c, G, E)
@@ -723,7 +723,7 @@ function _zonotope_overapprox(c, G, E)
     Gnew = copy(G)
     @inbounds for (j, g) in enumerate(eachcol(G))
         if all(iseven, E[:, j])
-            cnew .+= 0.5 * g
+            cnew .+= g / 2
             Gnew[:, j] ./= 2
         end
     end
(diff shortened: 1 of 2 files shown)


src/Approximations/overapproximate.jl
```diff
@@ -714,6 +714,22 @@ end
 # Overapproximation of Sparse Polynomial Zonotopes
 # ================================================
 
+#=
+Given center, (dependent) generator matrix and exponent matrix of a (simple) sparse polynomialzonotope,
```
schillic, line 718:
```suggestion
Given center, (dependent) generator matrix and exponent matrix of a (simple) sparse polynomial zonotope,
```

src/Approximations/overapproximate.jl
```diff
@@ -714,6 +714,22 @@ end
...
+compute thew new center and generator matrix of its zonotope overapproximation.
+=#
+function _zonotope_overapprox(c, G, E)
+    cnew = copy(c)
+    Gnew = copy(G)
```
schillic, line 723: I suggest to use an in-place method and pass copies to it, like this:
```suggestion
function _zonotope_overapprox!(c, G, E)
```

src/Approximations/overapproximate.jl
```diff
@@ -714,6 +714,22 @@ end
...
+    @inbounds for (j, g) in enumerate(eachcol(G))
+        if all(iseven, E[:, j])
+            cnew .+= 0.5 * g
+            Gnew[:, j] ./= 2
```
schillic, line 727: This is somewhat inconsistent. 
```suggestion
            cnew .+= g / 2
            Gnew[:, j] ./= 2
```

schillic (approved):

src/Approximations/overapproximate.jl
lucaferranti: do you mean to have both in-place and non-inplace helpers or just in-place helper and call `copy` in the exported `overapproximate` ?

schillic: Both are fine, but the latter is sufficient. It is only a suggestion, though.

lucaferranti: looking at how the `overapproximate` is written, maybe having both `_zonotope_overapprox!` and `_zonotope_overapprox` might be clearer?

schillic (approved):

mforets (approved):

PR #3037, opened 2022-08-16
add generalized quadratic map for SSPZ

lucaferranti:


schillic:
I think there is some confusion of branches. This should be #3028 and #3028 has the `exact_sum` on top. The quadratic map is missing.

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -256,7 +256,51 @@ function quadratic_map(Q::Vector{MT}, S::SimpleSparsePolynomialZonotope) where {
...
+
+### Input
+
+- `Q` -- vector of square matrices
+- `S1` -- simple sparse polynomial zonotope
+- `S2` -- simple sparse polynomial zonotope
+
+### Output
+
+The quadratic map of the given zonotope represented as a polynomial zonotope.
```
schillic, line 276:
```suggestion
The quadratic map of the given polynomial zonotopes represented as a simple
sparse polynomial zonotope.
```

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -256,7 +256,51 @@ function quadratic_map(Q::Vector{MT}, S::SimpleSparsePolynomialZonotope) where {
...
+	E2 = expmat(S2)
+
+    c = [dot(c1, Qi, c2) for Qi in Q]
+	Ghat1 = reduce(vcat, c2' * Qi' * G1 for Qi in Q)
+	Ghat2 = reduce(vcat, c1' * Qi * G2 for Qi in Q)
+    Gbar = [reduce(vcat, Gj' * Qi * G2 for Qi in Q) for Gj in eachcol(G1)]
+	Ebar = [E2 .+ E1j for E1j in eachcol(E1)]
+	G = hcat(Ghat1, Ghat2, Gbar...)
+	E = hcat(E1, E2, Ebar...)
+	return remove_redundant_generators(SimpleSparsePolynomialZonotope(c, G, E))
```
schillic, line 303: There are tabs here. Replace by four spaces.

schillic (commented):
LGTM

test/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -67,11 +67,19 @@ for N in [Float64, Float32, Rational{Int}]
...
+    @test expmat(q1) == [ 1  0  2  1;
+                          0  1  0  1]
+
+    q2 = quadratic_map(Q, S2, S2)
+    @test center(q2) ≈ [-0.24, -0.416]
+    @test genmat(q2) ≈ [-1.2   0.16    0.0  0.8;
+                         -2.72  0.192  -3.2  0.96]
+
+    @test expmat(q2) == [ 1  0  2  1;
+                          0  1  0  1]
```
schillic, line 82: If the tests are only written for `Float64`, they should be removed from the `for N in [Float64, Float32, Rational{Int}]` loop because that would just run the exact same tests three times. You can add a loop `for N in [Float64]` below and move the code there (the purpose of this loop is to avoid warnings about local/global variables from other test files).

schillic (commented):
There are inexactness errors. Not sure what to do about them.

test/Sets/SimpleSparsePolynomialZonotope.jl
lucaferranti: the SSPZ is created in different numerical precisions 

```julia
S2 = SimpleSparsePolynomialZonotope(N[0.2, -0.6], N[1 0;0 0.4], [1 0;0 1])
```

however the vector `Q` has only Float64 atm.

Changing the precision of the tests on the right hand side of the test (e.g. `@test center(q2) ≈ [-0.24, -0.416]`) fixes the tests, but maybe if feels a little like cheating, because the lhs of the test is always float64?) Changing the precision of also `Q` causes overflow error with `Rational{Int}`

lucaferranti: suggestions / comments on what to do?

schillic: I will play around with it and send a commit.

schillic: Done. The problem is that if you create an array with `3.2`, this will be a `Float64` and the `N` conversion happens only later. But `Float64` generally does not give you a "nice" rational:

```julia
julia> Rational{Int}(3.2)
3602879701896397//1125899906842624
```

So instead you need to create the numbers as `Rational`s and then convert to `N`.

schillic (approved):

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -256,7 +256,52 @@ function quadratic_map(Q::Vector{MT}, S::SimpleSparsePolynomialZonotope) where {
...
+    h = ngens(S1)
+    G1 = genmat(S1)
+    G2 = genmat(S2)
+    E1 = expmat(S1)
+    E2 = expmat(S2)
+
+    c = [dot(c1, Qi, c2) for Qi in Q]
+    Ghat1 = reduce(vcat, c2' * Qi' * G1 for Qi in Q)
+    Ghat2 = reduce(vcat, c1' * Qi * G2 for Qi in Q)
+    Gbar = [reduce(vcat, Gj' * Qi * G2 for Qi in Q) for Gj in eachcol(G1)]
```
mforets, line 300:
```suggestion
    Gbar = [reduce(vcat, gj' * Qi * G2 for Qi in Q) for gj in eachcol(G1)]
```
for consistency with lower case letters for vectors

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -256,7 +256,52 @@ function quadratic_map(Q::Vector{MT}, S::SimpleSparsePolynomialZonotope) where {
...
+    Ebar = [E2 .+ E1j for E1j in eachcol(E1)]
```
mforets, line 301: same here

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -256,7 +256,52 @@ function quadratic_map(Q::Vector{MT}, S::SimpleSparsePolynomialZonotope) where {
...
+    G = hcat(Ghat1, Ghat2, Gbar...)
```
mforets, line 302: would `reduce(hcat, ..)` not work too?

src/Sets/SimpleSparsePolynomialZonotope.jl
```diff
@@ -256,7 +256,52 @@ function quadratic_map(Q::Vector{MT}, S::SimpleSparsePolynomialZonotope) where {
...
     end
-    return SimpleSparsePolynomialZonotope(cnew, Gnew, Enew)
+    return remove_redundant_generators(SimpleSparsePolynomialZonotope(cnew, Gnew, Enew))
+end
+
+
+"""
+    quadratic_map(Q::Vector{MT}, S1::SimpleSparsePolynomialZonotope, S2::SimpleSparsePolynomialZonotope) where {N, MT<:AbstractMatrix{N}}
+
+Return the quadratic map of the given simple sparse polynomial zonotopes.
```
mforets, line 266: might be good to include the mathematical definition here

src/Sets/SimpleSparsePolynomialZonotope.jl
lucaferranti, line 301:
```suggestion
    Ebar = [E2 .+ e1j for e1j in eachcol(E1)]
```

src/Sets/SimpleSparsePolynomialZonotope.jl
lucaferranti: I think the following might work
```julia
Gbar = reduce(hcat, reduce(vcat, gj' * Qi * G2 for Qi in Q) for gj in eachcol(G1))
G = hcat(Ghat1, Ghat2, Gbar)
```

this way we would avoid allocations and splatting

schillic (approved):

mforets (approved):

PR #3045, opened 2022-08-20
Switch to StaticArraysCore and make it a direct dependency

schillic:
Since a few weeks, `Revise` did not work with `LazySets` anymore. I originally thought that this was a general problem, but it turns out to be related to `StaticArrays`. I am not sure what caused it. `StaticArrays` had outsourced some code to `StaticArraysCore`, but switching to that package instead does not fix the issue. The problem may be related to the fact that we load `StaticArrays` anyway through the hard dependency `IntervalArithmetic`.

So here I propose to make the package a direct dependency. `StaticArraysCore` is actually pretty lightweight and we load it anyway. Having `Revise` support is just too important.


mforets:
> So here I propose to make the package a direct dependency

Sounds good to me. Moreover, we should make internal LazySets functions more compatible with static arrays at some point. We have some open issues and I know some package had trouble too (see eg [this comment](https://github.com/microsoft/OpticSim.jl/blob/adc956e22512891d941a1c6ccfab85b41c5b2d8c/src/RepeatingStructures/Lattice.jl#L95)).

mforets (approved):

PR #3055, opened 2022-08-21
underapproximate with a Ball2

schillic:
The first commit fixes an inconsistency: `chebyshev_center` existed with kwarg `get_radius` and `compute_radius`. I went for the latter because that was used in two places.

```julia
julia> X = rand(VPolygon);
julia> B = underapproximate(X, Ball2);
```
ball2

### DIFF
--- a/src/Interfaces/AbstractPolyhedron_functions.jl
+++ b/src/Interfaces/AbstractPolyhedron_functions.jl
@@ -854,7 +854,7 @@ end
 
 """
     chebyshev_center(P::AbstractPolyhedron{N};
-                     [get_radius]::Bool=false,
+                     [compute_radius]::Bool=false,
                      [backend]=default_polyhedra_backend(P),
                      [solver]=default_lp_solver_polyhedra(N; presolve=true)
                      ) where {N}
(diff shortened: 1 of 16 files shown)


mforets (approved):

src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -887,7 +887,7 @@ The Chebyshev center is the center of a largest Euclidean ball enclosed by `P`.
 In general, the center of such a ball is not unique (but the radius is).
 """
 function chebyshev_center(P::AbstractPolyhedron{N};
-                          get_radius::Bool=false,
+                          compute_radius::Bool=false,
```
mforets, line 890: Looking at the implementation, the radius is always computed. I think it's cleaner to define `chebyshev_center` and `chebyshev_radius`, 

```julia
function chebyshev_center(...)
    c, _ = _chebyshev(P)
end
function chebyshev_radius(...)
    _, r = _chebyshev(P)
end
``` 
The function `underapproximate` can call `_chebyshev` directly.

mforets: Or otherwise make `chebyshev_center` always return center and radius tuple.

schillic: I agree and will make the proposed change.

schillic: I went for renaming to `chebyshev_center_radius` for clarity. We currently do not need `chebyshev_center` and `chebyshev_radius`, but this way we can easily add them later.

There was also a bug in the implementation for `Interval`: It returned the radius 0.

schillic: I rebased with `master`. The new changes are in the last commit.

mforets (approved):

PR #3057, opened 2022-08-22
add minkowski sum for sparse polynomial zonotope

lucaferranti:


src/Sets/SparsePolynomialZonotope.jl
```diff
@@ -318,6 +318,28 @@ end
 
 const ⊞ = exact_sum
 
+"""
+    minkowski_sum(P1::SparsePolynomialZonotope, P2::SparsePolynomialZonotope)
```
schillic, line 322: This method should go to `ConcreteOperations/minkowski_sum.jl`.

schillic: And I guess we could also move the `exact_sum` to a new file there. But for now we can also keep that because there is only one combination defined.

schillic: Ah, and the docs entry should also be moved to the `binary_functions.md` file.

lucaferranti: moved function and dosc entry. With a quick look, I think `SparsePolynomialZonotopes.md` now doesn't have misplaced docstrings. SSPZ has a few that should be moved to binary_functions.md, but that's prob for a separated PR

lucaferranti: > And I guess we could also move the exact_sum to a new file there. But for now we can also keep that because there is only one combination defined.

I think some previous discussion was around these lines, it should be in it's own file, but since it's defined for SPZ only, let's keep it there for now

schillic (approved):

mforets (approved):

PR #3081, opened 2022-09-11
Revise HalfSpace code

schillic:
The first commit also moves some functions in the docs and changes some signatures so they only occur once in the docs.
The second and third commit change the name of two helper functions and change their signatures to avoid extra allocations.

### DIFF
--- a/docs/src/lib/interfaces.md
+++ b/docs/src/lib/interfaces.md
@@ -228,6 +228,15 @@ vertices_list(::AbstractPolyhedron)
 project(::AbstractPolyhedron{N}, ::AbstractVector{Int}) where {N}
 ```
 
+Some common functions to work with linear constraints:
+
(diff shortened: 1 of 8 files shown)


mforets:
What motivates using `nothing` instead of `default_lp_solver(N)` etc? I find the latter more readable.

schillic:
Oops sorry, I merged the wrong PR ...

> What motivates using nothing instead of default_lp_solver(N) etc? I find the latter more readable.

This was to make the docs catch the right method. It did not work otherwise.

schillic:
Should I undo the merge?

mforets:
> This was to make the docs catch the right method

Ok, that's fine, so it's a workaround for our docs setup.

mforets:
> Should I undo the merge?

No

PR #3085, opened 2022-09-15
Revise HParallelotope code

schillic:
The only code change is to replace `hcat(E...)` by `reduce(hcat, E)`:

```julia
julia> E
3-element Vector{Vector{Float64}}:
 [0.8, 0.19999999999999996, 0.0]
 [0.85, 0.09999999999999998, 0.0]
 [0.85, 0.15000000000000002, -0.0]

julia> @time reduce(hcat, E)
  0.000007 seconds (1 allocation: 128 bytes)
3×3 Matrix{Float64}:
 0.8  0.85   0.85
 0.2  0.1    0.15
 0.0  0.0   -0.0

julia> @time hcat(E...)
  0.000009 seconds (2 allocations: 160 bytes)
3×3 Matrix{Float64}:
 0.8  0.85   0.85
 0.2  0.1    0.15
 0.0  0.0   -0.0
```


src/Sets/HParallelotope.jl
```diff
@@ -16,57 +16,65 @@ Type that represents a parallelotope in constraint form.
...
-            "of the directions matrix, but they are $(length(c)) and $(size(D)) dimensional respectively"
+        @assert length(c) == 2*checksquare(D) "the length of the offset " *
+            "vector should be twice the size of the directions matrix, " *
+            "but they have sizes $(length(c)) and $(size(D)) respectively"
         return new{N, VN, MN}(D, c)
     end
 end
 
 isoperationtype(::Type{<:HParallelotope}) = false
-isconvextype(::Type{<:HParallelotope}) = true
```
mforets, line 69: why is this removed? is it because it'll use some interface method?

mforets (approved):

src/Sets/HParallelotope.jl
schillic: Yes, I removed this from all concrete types so far if their supertype already defined it the same way.

PR #3086, opened 2022-09-18
Revise Line code

schillic:
Main changes:
- fix docs setup for `constraints_list`
- fix bug in constructor with `normalize` option (did not really normalize)
   - since `normalize` crashes for `Rational` numbers, I changed the default value to `false`
- fix bug in `linear_map` (could result in a line with a zero direction, thus erroring)

### DIFF
--- a/src/Sets/Line.jl
+++ b/src/Sets/Line.jl
@@ -97,24 +97,24 @@ function Line(a::AbstractVector{N}, b::N; normalize=false) where {N}
     @assert length(a) == 2 "expected a normal vector of length two, but it " *
         "is $(length(a))-dimensional"
 
-    got_horizontal = @inbounds iszero(a[1])
-    got_vertical = @inbounds iszero(a[2])
+    got_horizontal = iszero(a[1])
+    got_vertical = iszero(a[2])
 
     if got_horizontal && got_vertical
         throw(ArgumentError("the vector $a must be non-zero"))
     end
 
     if got_horizontal
-        α = b / @inbounds a[2]
+        α = b / a[2]
         p = [zero(N), α]
         q = [one(N), α]
     elseif got_vertical
-        β = b / @inbounds a[1]
+        β = b / a[1]
         p = [β, zero(N)]
         q = [β, one(N)]
     else
-        α = b / @inbounds a[2]
-        @inbounds μ = a[1] / a[2]
+        α = b / a[2]
+        μ = a[1] / a[2]
         p = [zero(N), α]
         q = [one(N), α - μ]
     end
--- a/src/Sets/Line.jl
+++ b/src/Sets/Line.jl
@@ -10,13 +10,13 @@ export Line,
 """
     Line{N, VN<:AbstractVector{N}} <: AbstractPolyhedron{N}
 
-Type that represents a line in the form
+Type that represents a line of the form
 
 ```math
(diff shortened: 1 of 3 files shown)


src/Sets/Line.jl
```diff
@@ -71,52 +70,51 @@ Given two points ``p ∈ \\mathbb{R}^n`` and ``q ∈ \\mathbb{R}^n``, the line t
...
-    got_horizontal = iszero(a[1])
-    got_vertical = iszero(a[2])
+    got_horizontal = @inbounds iszero(a[1])
+    got_vertical = @inbounds iszero(a[2])
 
     if got_horizontal && got_vertical
         throw(ArgumentError("the vector $a must be non-zero"))
     end
 
     if got_horizontal
```
mforets, line 107: Having lots of `@inbounds` doesn't read very nicely. Maybe define a block?

```julia
@inbounds begin
   ...
end
```

schillic: Do you mean these two lines or the whole method body?

mforets: As in

```julia
    @inbounds begin
    if got_horizontal
        α = b /  a[2]
        p = [zero(N), α]
        q = [one(N), α]
    elseif got_vertical
        β = b / a[1]
        p = [β, zero(N)]
        q = [β, one(N)]
    else
        α = b / a[2]
        μ = a[1] / a[2]
        p = [zero(N), α]
        q = [one(N), α - μ]
    end
    end
```

On second thought, I suggest to not add these `@inbounds`, at least not in combination with code that dispatches on `AbstractVector`, while still assuming that the indices start at 1. Here is an example:

```julia
julia> using OffsetArrays

julia> x = OffsetArray([1,2,3], -1:1)
3-element OffsetArray(::Vector{Int64}, -1:1) with eltype Int64 with indices -1:1:
 1
 2
 3

julia> x[1]
3

julia> x[-1]
1

julia> x[0]
2

julia> function g(x)
           x[3] - 1
       end
g (generic function with 1 method)

julia> g(x)
ERROR: BoundsError: attempt to access 3-element OffsetArray(::Vector{Int64}, -1:1) with eltype Int64 with indices -1:1 at index [3]
Stacktrace:
 [1] throw_boundserror(A::OffsetVector{Int64, Vector{Int64}}, I::Tuple{Int64})
   @ Base ./abstractarray.jl:691
 [2] checkbounds
   @ ./abstractarray.jl:656 [inlined]
 [3] getindex
   @ ~/.julia/packages/OffsetArrays/80Lkc/src/OffsetArrays.jl:435 [inlined]
 [4] g(x::OffsetVector{Int64, Vector{Int64}})
   @ Main ./REPL[6]:2
 [5] top-level scope
   @ REPL[7]:1
   
julia> function f(x)
           @inbounds x[3] - 1
       end
f (generic function with 1 method)

julia> f(x)
4824405775
```

schillic: But with this argument we should remove `@inbounds` everywhere, right?

The code accesses the array precisely at indices 1 and 2 anyway, so removing the `@inbounds` check does not magically make the code work. The assumption that vectors start at 1 is very common and used all over the library. I would consider such exotic vectors a user error because you probably cannot use the library in a useful way with them.

mforets (approved):

PR #3110, opened 2022-10-04
Revise helper functions and generalize to LazySet

schillic:
The first commit revises the file and generalizes to `LazySet` (this had some consequences on doctests).
The second commit reorders the functions in the docs.
The third commit moves some functions to their own files. (In fact I think the functions in `numbers.jl` could be outsourced to the Base package).


mforets:
> (In fact I think the functions in numbers.jl could be outsourced to the Base package).

we could do that, yes

mforets (approved):

PR #3111, opened 2022-10-04
Fix (de)activate_assertions

schillic:


mforets:
don't we have a test for this functionality in LazySets?

schillic:
Apparently not :(

PR #3212, opened 2022-12-06
Fix remove_redundant_constraints! output

schillic:
~The implementation could return `true` even if no constraint was removed.~

This was intended. The comment is a bit confusing, but the function should return `false` if and only if the constraints are infeasible.


schillic:
I should read the

PR #3213, opened 2022-12-06
#2426 - Fix binary search in addconstraint! for polygons

### LINKED ISSUE #2426: addconstraint! for HPolygon differs with linear_search option
The test code is here:

https://github.com/JuliaReach/LazySets.jl/blob/72ea657517e5d4df28ed87725443eea8f46237f9/test/unit_Polygon.jl#L455-L469

This random test errored here: https://travis-ci.org/github/JuliaReach/LazySets.jl/jobs/745365104#L767
`Evaluated: 3 == 4 == 4 == 4`
Unfortunately, the inputs are not printed. We need to test this with random inputs until we find a new failure.

[(fixed in #2584) Side note: One of the `length(po2.constraints)` in the `@test` should be `length(po1.constraints)`.]

schillic:
Closes #2426.

I think the problem was that our implementation of `<=` is not antisymmetric, so for very similar directions, the index was not unique. Now the method returns the smallest possible index in such cases.

### DIFF
--- a/src/Interfaces/AbstractHPolygon.jl
+++ b/src/Interfaces/AbstractHPolygon.jl
@@ -477,12 +477,11 @@ function addconstraint!(constraints::Vector{LC}, new_constraint::HalfSpace;
             end
         else
             # binary search
-            k = binary_search_constraints(
-                d, constraints, k, 1 + div(k, 2), choose_lower=true)
+            k = binary_search_constraints(d, constraints, choose_lower=true)
         end
     end
 
-    # here constraints[k] <= new_constraint <= constraints[(k%m)+1]
+    # here constraints[k].a <= new_constraint.a <= constraints[(k%m)+1].a
     if prune && m >= 2
         # check if new constraint is redundant
         k += 1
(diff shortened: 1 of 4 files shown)


src/Interfaces/AbstractHPolygon.jl
```diff
@@ -543,39 +541,49 @@ Perform a binary search in the constraints.
...
-            upper = k
+            upper = m
         end
-        k = lower + div(upper - lower, 2)
+        m = div(lower + upper, 2)
     end
+
+    # since `<=` is approximate, it can happen that x <= y and y <= x
+    # we want to return the smallest index, so (linearly) search to the left
+    while lower > 1 && d <= constraints[lower].a && constraints[lower].a <= d
```
mforets, line 577: can't we use `_isaprox` here?

mforets (approved):

src/Interfaces/AbstractHPolygon.jl
schillic: Here is a counterexample:

```julia
julia> u = ones(2);
julia> v = u .+ 1e-5;

julia> _isapprox(u, v)
false

julia> u <= v <= u
true
```

---

`u <= v` (defined for 2D vectors) falls back to `_leq_quadrant(u, v)`, which checks if the vectors are in different quadrants. If they are not, it returns `is_right_turn(u, v)`, which is `_geq(right_turn(u, v), 0)`. And `right_turn(u, v)` is `u[1] * v[2] - u[2] * v[1]`.

So `u <= v <= u` holds if `_geq(u[1] * v[2] - u[2] * v[1], 0) && _geq(v[1] * u[2] - v[2] * u[1], 0)`.

Not sure this helps (it did not for me) :sweat_smile:

PR #3219, opened 2022-12-16
#2786 - Add consistency check to HParallelotope constructor and use cheap isempty method

### LINKED ISSUE #2786: vertices_list of an empty HParallelotope is non-empty

schillic:
Closes #2786.

This PR adds a consistency check to the `HParallelotope` constructor. Thus we accept that `vertices_list` and `isempty` return wrong results for inconsistent sets.

### DIFF
--- a/src/Sets/HParallelotope.jl
+++ b/src/Sets/HParallelotope.jl
@@ -389,12 +389,11 @@ function rand(::Type{HParallelotope};
             offset[i] = -offset[i-dim] + abs(offset[i])
         end
 
-        P = HParallelotope(D, offset)
-
         # convert to polyhedron to check boundedness
-        Q = HPolyhedron(vcat(P.directions, -P.directions), P.offset)
+        clist = _constraints_list_hparallelotope(D, offset, N, typeof(offset))
+        Q = HPolyhedron(clist)
         if isbounded(Q)
-            return P
+            return P = HParallelotope(D, offset; check_consistency=false)
         end
         # set is unbounded; sample a new set in the next iteration
     end
--- a/src/Sets/HParallelotope.jl
+++ b/src/Sets/HParallelotope.jl
@@ -400,28 +400,6 @@ function rand(::Type{HParallelotope};
 end
 
 """
-    isempty(P::HParallelotope)
-
-Check whether a parallelotope in constraint representation is empty.
-
-### Input
-
-- `P` -- parallelotope in constraint representation
-
-### Output
-
-`true` iff the constraints are consistent/feasible.
-
-### Algorithm
-
-We call `isempty` for an `HPolyhedron`. We do not assume `HPolytope` because the
-set may be wrongly represented and in fact be unbounded.
-"""
-function isempty(P::HParallelotope)
-    return isempty(convert(HPolyhedron, P))
-end
-
-"""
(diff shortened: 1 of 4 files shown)


src/Sets/HParallelotope.jl
```diff
@@ -403,3 +404,25 @@ function volume(P::HParallelotope)
...
+### Output
+
+A list of the vertices.
+
+### Algorithm
+
+We call `vertices_list` for an `HPolyhedron`. We do not assume `HPolytope`
+because the set may be wrongly represented and in fact be unbounded.
+"""
+function vertices_list(P::HParallelotope)
```
mforets, line 426: Why is 

```julia
julia> H = [1.0 1.0; 0.0 1.0];
julia> o = [2.5, 2.0, -4.0, -1.0];
julia> Pl = HParallelotope(H, o);  # this set is empty

julia> vertices_list(Pl)
4-element Vector{Vector{Float64}}:
 [2.0, 2.0]
 [0.5, 2.0]
 [1.5, 1.0]
 [3.0, 1.0]
```

returning an incorrect result in the first place?

schillic: The code calls this method, which uses the generator representation:

https://github.com/JuliaReach/LazySets.jl/blob/b11803321cd10ce904198d8abcf19759f9257f1d/src/Sets/Zonotope.jl#L516-L528

It seems that the generator representation is incorrect for empty sets. But it is unclear what the generators should be. Maybe we should make `HParallelotope` not a subtype of `AbstractZonotope` anymore (zonotopes cannot be empty)? Or should we check for emptiness in the constructor (could be a user flag)?

mforets: Arguably the second option is better; or we can also check for emptiness of `HParallelotope` inside its `vertices_list` function, and if it is non-empty, then call the generator representation implementation.

schillic: Okay, I think I will go with the last option. So the plan is to change this PR to add an emptiness check to `vertices_list` and in the non-empty case fall back to the `AbstractZonotope` implementation as before.

schillic: I just realized that `isempty` also converts to `HPolyhedron`, so that would not be cheaper. And if the parallelotope is unbounded, the result would still be wrong (another ill definition one can unfortunately make).

Given that, I changed my mind and suggest again that we either add a check in the constructor for emptiness and boundedness (which we can make controllable by a flag) or remove the assumption that an `HParallelotope` is zonotopic.

mforets: Adding a check in the constructor sounds good to me.

mforets (approved):

PR #3227, opened 2023-01-20
Fix check for elimination algorithm type in _minkowski_sum_hrep

Zinoex:
This fixes two issues in `_minkowski_sum_hrep`:
1. EliminationAlgorithm results in an UndefVarError without the Polyhedra quantifier.
2. `algorithm` is a value, not a type. Hence `algorithm <: EliminationAlgorithm` results in a TypeError.


mforets (approved):

schillic (approved):
Thanks, I will add a test later.

schillic:
Thanks again @Zinoex! I added your name to the contributors for the release log. Let me know if you prefer to not appear there.

PR #3231, opened 2023-01-23
New algorithm to compute support vectors for HPoly

Zinoex:
New algorithm to compute support vectors for HPoly based on direction of normal vectors of halfspaces. The observed speed-up has been anywhere from 60x to 200x depending on the dimensionality. The limitation of this algorithm is that it only works if there are no redundant constraints.

It should be possible to achieve higher speed-ups by considering a spatial index such as an R-tree of the normal vectors projected onto the n-dimensional unit sphere, to search fewer constraints. A similar approach with spatial indexing could be applied to VPolytopes for all non-interior points to accelerate the computation of support vectors.

### DIFF
--- a/src/Sets/HPolyhedron.jl
+++ b/src/Sets/HPolyhedron.jl
@@ -165,11 +165,9 @@ entries.
 
 ### Algorithm
 
-This method applies either a LP-based approach or an approach based on direction of
-halfspaces. The default is to apply the LP-based approach. This is the safer algorithm
-as it works correctly for `HPoly` with redundant constraints, i.e. if `P` is empty or
-has redudant constraints, then the approach based on halfspace directions is not
-guaranteed to produce a correct result. 
+This method applies either an LP-based approach or an approach based on direction of
+half-spaces. The default is to apply the LP-based approach. This is the safer algorithm,
+as it works correctly for `HPoly` with redundant constraints or if `P` is empty.
 
 #### LP
 
@@ -182,17 +180,16 @@ resuling vector of `±Inf`.
 
 #### Halfspace normal vector directions
 
-Note: This algorithm only applies if the `P` contains no redundant constraints. If not,
-this algorithm is not guaranteed to produce a correct result.
+Note: This algorithm only applies if `P` contains no redundant constraints.
 
-This algorithm is solves the problem of finding support vectors by analyzing the normal
-vectors of halfspace constraints. The algorithm relies on the following three key insights:
+This algorithm analyzes the normal vectors of the half-space constraints and
+relies on the following three key insights:
 
 1. The support vector will always be a vertex or unbounded.
-2. The `dim(P)` constraints with the smallest angle to `d` will constraint the vertex ``x``
-   maximizing ``d^\\mathrm{T}x``.
-3. If the angle between `d` and the normal vector of a halfspace is greater than 
-   ``90^{\\circ}``, then that halfspace cannot constraint the optimal vertex ``x``.
+2. The `dim(P)` constraints with the smallest angle to `d` will constrain the
+   vertex ``x`` maximizing ``d^\\mathrm{T}x``.
+3. If the angle between `d` and the normal vector of a half-space is greater than 
+   ``90^{\\circ}``, then that half-space cannot constrain the optimal vertex ``x``.
 
 Following this train of thought, the algorithm finds the `dim(P)` constraints with the
 largest  ``\\cos(\\theta) = \\frac{\\langle a, d \\rangle}{\\lVert a \\rVert \\lVert d \\rVert}``
--- a/src/Sets/HPolyhedron.jl
+++ b/src/Sets/HPolyhedron.jl
@@ -178,7 +178,7 @@ infeasible, then the primal is unbounded meaning that the polyhedron is unbounde
 the given direction. An infeasibility certificate is used to find the sign of the
 resuling vector of `±Inf`.
 
-#### Halfspace normal vector directions
+#### Half-space normal vector directions
 
 Note: This algorithm only applies if `P` contains no redundant constraints.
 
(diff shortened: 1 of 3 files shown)


src/Sets/HPolyhedron.jl
```diff
@@ -130,21 +144,79 @@ direction.
...
 If a polyhedron is unbounded in the given direction, the result contains `±Inf`
 entries.
+
+### Algorithm
+
+This method applies either a LP-based approach or an approach based on direction of
+halfspaces. The default is to apply the LP-based approach. This is the safer algorithm
+as it works correctly for `HPoly` with redundant constraints, i.e. if `P` is empty or
+has redudant constraints, then the approach based on halfspace directions is not
+guaranteed to produce a correct result.
```
schillic, line 172:
```suggestion
This method applies either an LP-based approach or an approach based on direction of
halfspaces. The default is to apply the LP-based approach. This is the safer algorithm,
as it works correctly for `HPoly` with redundant constraints or if `P` is empty.
```

src/Sets/HPolyhedron.jl
```diff
@@ -130,21 +144,79 @@ direction.
...
+both the primal and dual are feasible, the optimal value of x is the support vector.
+If the primal is infeasible, then `P` is empty, and an error is thrown. If the dual is
+infeasible, then the primal is unbounded meaning that the polyhedron is unbounded in 
+the given direction. An infeasibility certificate is used to find the sign of the
+resuling vector of `±Inf`.
+
+#### Halfspace normal vector directions
+
+Note: This algorithm only applies if the `P` contains no redundant constraints. If not,
+this algorithm is not guaranteed to produce a correct result.
```
schillic, line 186:
```suggestion
Note: This algorithm only applies if `P` contains no redundant constraints.
```

src/Sets/HPolyhedron.jl
```diff
@@ -130,21 +144,79 @@ direction.
...
+
+This algorithm is solves the problem of finding support vectors by analyzing the normal
+vectors of halfspace constraints. The algorithm relies on the following three key insights:
```
schillic, line 189:
```suggestion
This algorithm analyzes the normal vectors of the half-space constraints and
relies on the following three key insights:
```

src/Sets/HPolyhedron.jl
```diff
@@ -130,21 +144,79 @@ direction.
...
+
+1. The support vector will always be a vertex or unbounded.
+2. The `dim(P)` constraints with the smallest angle to `d` will constraint the vertex ``x``
+   maximizing ``d^\\mathrm{T}x``.
```
schillic, line 193:
```suggestion
2. The `dim(P)` constraints with the smallest angle to `d` will constrain the
   vertex ``x`` maximizing ``d^\\mathrm{T}x``.
```

src/Sets/HPolyhedron.jl
```diff
@@ -130,21 +144,79 @@ direction.
...
+3. If the angle between `d` and the normal vector of a halfspace is greater than 
+   ``90^{\\circ}``, then that halfspace cannot constraint the optimal vertex ``x``.
```
schillic, line 195:
```suggestion
3. If the angle between `d` and the normal vector of a half-space is greater than 
   ``90^{\\circ}``, then that half-space cannot constrain the optimal vertex ``x``.
```

src/Sets/HPolyhedron.jl
```diff
@@ -130,21 +144,79 @@ direction.
...
+
+Following this train of thought, the algorithm finds the `dim(P)` constraints with the
+largest  ``\\cos(\\theta) = \\frac{\\langle a, d \\rangle}{\\lVert a \\rVert \\lVert d \\rVert}``
```
schillic, line 198: I have a mismatch between "The `dim(P)` constraints with the smallest angle" above and the "largest" here. Is one of them a typo?

schillic (commented):
I have yet to read the implementation, but here are already some minor suggestions for the documentation. From reading the description, the algorithm sounds solid, and your results look exciting.

src/Sets/HPolyhedron.jl
Zinoex:
```suggestion
This method applies either an LP-based approach or an approach based on direction of
half-spaces. The default is to apply the LP-based approach. This is the safer algorithm,
as it works correctly for `HPoly` with redundant constraints or if `P` is empty.
```

src/Sets/HPolyhedron.jl
Zinoex: No, this is not a typo. The smallest absolute angle correspond to the largest `cos(theta)`, i.e. the mapping from `abs(theta)` to `cos(theta)` for -pi <= theta <= pi is monotone, so by considering  `cos(theta)`  directly, we avoid taking the `arccos` and `abs`. Furthermore, `cos(theta) < 0` if and only if the angle is greater than 90 degrees, hence a normal vector satisfying `cos(theta) < 0` cannot bound `x` and we can filter it (to improve the speed of the topk via partial sorting).

schillic: Okay, do you mind adding a clarifying comment to avoid future confusion? Maybe just add "(which corresponds to the smallest absolute angle)" to this sentence. Or just copy what you wrote here.

Zinoex: The documentation has now been updated to explain the monotone mapping. I have also included an explanation of how the algorithm handles an unbounded support vector.

schillic (commented):
I have a few minor comments and one bigger question.

Is this idea really sound? Take a unit cube and ask for the support vector to the right (`d = [1, 0, 0]`). There are five constraints that are relevant (right, top, bottom, front, back). Your step in `if length(constraints) > dim(P)` will then decide to (randomly?) remove two of the four orthogonal constraints. Is it guaranteed that the resulting system is always feasible? Say you removed the front and back constraints. Then you want a vertex where the three constraints (right, top, bottom) meet, which does not exist.

```julia
julia> A = [1. 0 0; 0 1 0; 0 -1 0]; b = [1., 1, 1];

julia> A \ b
ERROR: LAPACKException(3)
```

Zinoex:
Thank you for your feedback, Christian. It made me realize that it is possible to constrain with `< dim(P)` constraints, which the algorithm currently does not handle. I believe that the key insights can still turn into an algorithm, which will be faster, but it needs a little more careful thought. Would you prefer I close this pull request while I rethink the idea and tinker around, or do you like it to remain open but pending?

schillic:
Both options are fine, but maybe close if you do not plan to work on it for several months. You can always refer back to this pull request later.

Another note (but I might misremember the algorithm): I think you cannot simply ignore constraints that are more than 90° away from the direction. Take this example:

triangle

```julia
julia> V = VPolygon([[1., 1], [1., -1], [-1., 0]]);

# same as HPolygon([HalfSpace([1., 0.], 1.), HalfSpace([-1., 2.], 1.), HalfSpace([-1., -2.], 1.)])

julia> σ([1., 0], V)
2-element Vector{Float64}:
  1.0
 -1.0
```
Any answer $[1, y]$ for $y ∈ [-1, 1]$ is correct, but if you only keep the constraint $x \le 1$, you can choose any $y$.

PR #3239, opened 2023-02-23
Fix #3238 -- MathOptInterface -> MOI

### LINKED ISSUE #3238: Depend on `MathOptInterface` directly instead of referencing the internal variable of `JuMP`
Currently, `MathOptInterface` is loaded through the `JuMP` module scope:

https://github.com/JuliaReach/LazySets.jl/blob/b11803321cd10ce904198d8abcf19759f9257f1d/src/Initialization/init_JuMP.jl#L1

The latest release of `JuMP` has renamed `MathOptInterface` internally to MOI. Thus, `LazySets` now fails to load.

I suggested in https://github.com/jump-dev/JuMP.jl/issues/3233 to keep an alias around until a breaking release is tagged. Nonetheless, it would be good to make the changes here now to depend on `MathOptInterface` directly instead of "stealing" it from `JuMP`.

lassepe:


lassepe:
As per the discussion in https://github.com/jump-dev/JuMP.jl/issues/3233, accessing MathOptInterface as `JuMP.MOI` variable is encouraged and also works on older JuMP version.

schillic (approved):
Thanks for the proposal! Looks good to me.

schillic:
The new release is out.
CC @BenCurran98

src/Initialization/init_JuMP.jl
```diff
@@ -1,34 +1,34 @@
-using JuMP.MathOptInterface: AbstractOptimizer, OptimizerWithAttributes
+using JuMP.MOI: AbstractOptimizer, OptimizerWithAttributes
 using JuMP: Model, @variable, @objective, @constraint, optimize!,
             termination_status, objective_value, value, primal_status
 
 # solver status
 function is_lp_optimal(status)
-    return status == JuMP.MathOptInterface.OPTIMAL
+    return status == JuMP.MOI.OPTIMAL
```
odow, line 7: This can be `JuMP.OPTIMAL`, and so forth

src/Initialization/init_JuMP.jl
```diff
@@ -1,34 +1,34 @@
-using JuMP.MathOptInterface: AbstractOptimizer, OptimizerWithAttributes
+using JuMP.MOI: AbstractOptimizer, OptimizerWithAttributes
```
odow, line 1: If you need these, you should probably instead declare a `Pkg` dependency on `MathOptInterface` explicitly? It's weird to import structs/methods from a dependency of a package.

mforets: thanks, done in https://github.com/JuliaReach/LazySets.jl/pull/3245#pullrequestreview-1346354090

PR #3241, opened 2023-02-26
`Pkg.checkout()` is not supported anymore

Zitzeronion:
In the **Updating** section the code snipped 
```julia
Pkg.checkout("LazySets")
```
Does not work anymore, it creates this error
```julia
Pkg.checkout("LazySets")
ERROR: UndefVarError: checkout not defined
Stacktrace:
 [1] getproperty(x::Module, f::Symbol)
   @ Base .\Base.jl:31
 [2] top-level scope
   @ REPL[7]:1
```

One possible solution is to change `Pkg.checkout("LazySets")` for `Pkg.update("LazySets")`

### DIFF
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -55,7 +55,7 @@ After working with `LazySets` for some time, you may want to get the newest
 version.
 For this you can use the following command (e.g., from the REPL):
 ```julia
-Pkg.checkout("LazySets")
+Pkg.update("LazySets")
 ```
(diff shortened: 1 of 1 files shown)


schillic:
Thanks! Maybe rather `Pkg.add`? I believe this is the successor of `Pkg.checkout`.

Zitzeronion:
> I would actually suggest to remove that whole paragraph.

Agree with that.

schillic:
I added #3243 for that.

mforets (approved):

Issue #3249, opened 2023-03-31
Generalize VPolytope container type


mforets:
This feature would be nice to have in applications such as https://github.com/ONSAS/ONSAS.jl/tree/main/src

cc @mvanzulli

PR #3250, opened 2023-03-31
Generalize outer array container of VPolytope

mforets:
Closes https://github.com/JuliaReach/LazySets.jl/issues/3249

### DIFF
--- a/src/Sets/VPolytope.jl
+++ b/src/Sets/VPolytope.jl
@@ -54,7 +54,7 @@ julia> P.vertices
 ```
 """
 struct VPolytope{N, VN<:AbstractVector{N}, VT<:AbstractVector{VN}} <: AbstractPolytope{N}
-    vertices::Vector{VT}
+    vertices::VT
 end
 
 isoperationtype(::Type{<:VPolytope}) = false
--- a/src/Sets/VPolytope.jl
+++ b/src/Sets/VPolytope.jl
@@ -54,7 +54,7 @@ julia> P.vertices
 ```
 """
 struct VPolytope{N, VN<:AbstractVector{N}, VT<:AbstractVector{VN}} <: AbstractPolytope{N}
-    vertices::Vector{VN}
+    vertices::Vector{VT}
 end
 
 isoperationtype(::Type{<:VPolytope}) = false
--- a/src/Sets/VPolytope.jl
+++ b/src/Sets/VPolytope.jl
@@ -9,7 +9,7 @@ export VPolytope,
        tovrep
 
 """
-    VPolytope{N, VN<:AbstractVector{N}} <: AbstractPolytope{N}
+    VPolytope{N, VN<:AbstractVector{N}, VT<:AbstractVector{VN}} <: AbstractPolytope{N}
 
 Type that represents a convex polytope in vertex representation.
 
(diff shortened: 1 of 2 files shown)


mforets:
Many autoformatter edits. I can turn that off for this case if it bothers for the review. See also https://github.com/JuliaReach/LazySets.jl/issues/3251

src/Sets/VPolytope.jl
```diff
@@ -187,7 +187,7 @@ Then we solve the following ``m``-dimensional linear program.
 ```
 """
 function ∈(x::AbstractVector{N}, P::VPolytope{N};
-           solver=default_lp_solver(N)) where {N}
+    solver=default_lp_solver(N)) where {N}
     vertices = P.vertices
```
schillic, line 191: The other formatter changes are fine, but I really do not like changing the indentation of function arguments to the same level as the body. It is totally unclear where the arguments end and the body starts.
(This type of change repeats below.)
Could you undo them? (Or, if easier, redo the changes without formatting.)

mforets: It's easy to redo changes without formatting, but let me do a quick check whether it is possible to turn off just the setting you've described.

mforets: Seems like it's doable, it has lots of options really > https://domluna.github.io/JuliaFormatter.jl/dev/
 
But let's discuss it properly in the context on the other task, we need to come up with a custom format that is suitable for the package.

test/Sets/Polytope.jl
```diff
@@ -228,6 +228,15 @@ for N in [Float64, Rational{Int}, Float32]
...
     @test permute(P, [3, 1, 2]) == Q
+
+    # construction with static arrays
+    vlist = [SVector{2}(N[0, 0]), SVector{2}(N[1, 0]), SVector{2}(N[0, 1])]
+    V = VPolytope(vlist)
+    @test vertices_list(V) == vlist
+    # check that the outer container can also be static
+    vlist = SVector{3}(SVector{2}(N[0, 0]), SVector{2}(N[1, 0]), SVector{2}(N[0, 1]))
+    V = VPolytope(vlist)
+    @test vertices_list(V) == vlist
```
schillic, line 239:
```suggestion
    @test ispermutation(vertices_list(V), vlist)
    # check that the outer container can also be static
    vlist = SVector{3}(SVector{2}(N[0, 0]), SVector{2}(N[1, 0]), SVector{2}(N[0, 1]))
    V = VPolytope(vlist)
    @test ispermutation(vertices_list(V), vlist)
```
I think this is "more correctly" what the specification says.
(If you wanted to test that the output is an `SVector`, you have to test that explicitly because I think `==` does not care.)

schillic (approved):

src/Sets/VPolytope.jl
```diff
@@ -53,8 +53,8 @@ julia> P.vertices
  [0, 0, 1]
 ```
 """
-struct VPolytope{N, VN<:AbstractVector{N}} <: AbstractPolytope{N}
-    vertices::Vector{VN}
+struct VPolytope{N, VN<:AbstractVector{N}, VT<:AbstractVector{VN}} <: AbstractPolytope{N}
+    vertices::Vector{VT}
```
schillic, line 57:
```suggestion
    vertices::VT
```

PR #3261, opened 2023-04-04
#3194 - Remove AbstractStar interface

### LINKED ISSUE #3194: AbstractStar is recognized as a concrete set type
```julia
julia> LazySets.AbstractStar ∈ LazySets.subtypes(LazySet, true)
true
```
The problem is that there is currently no set type implementing that interface.
One way to "fix" that is to add an implementation.

schillic:
Closes #3194.

Once we have an implementation, we can easily add it again. But for now this just breaks some assumptions.


mforets (approved):

mforets:
Why is `Star` not an implementation? 🤔 
I guess because it's ` <: AbstractPolyhedron`. But then why not `AbstractPolyhedron <: AbstractStar`?

schillic:
Right now we do not have any code for `AbstractStar`.

Issue #3259, opened 2023-04-02
Implement efficient point in tetrahedron check


mforets:
https://stackoverflow.com/questions/25179693/how-to-check-whether-the-point-is-in-the-tetrahedron-or-not

Requires https://github.com/JuliaReach/LazySets.jl/issues/3258

PR #3263, opened 2023-04-05
Add Tetrahedron set type

mforets:
Closes https://github.com/JuliaReach/LazySets.jl/issues/3258

### DIFF
--- a/src/Sets/Tetrahedron.jl
+++ b/src/Sets/Tetrahedron.jl
@@ -1,0 +1,117 @@
+export Tetrahedron
+
+"""
+    Tetrahedron{N, VN<:AbstractVector{N}, VT<:AbstractVector{VN}} <: AbstractPolytope{N}
+
+Type that represents a tetrahedron in vertex representation.
+
+### Fields
+
+- `vertices` -- list of vertices
+
+### Examples
+
+A tetrahedron can be constructed by passing the list of vertices.
+The following builds the tetrahedron with edge length 2 from the [wikipedia page Tetrahedron](https://en.wikipedia.org/wiki/Tetrahedron): 
+
+```jldoctest
+julia> vertices = [[1, 0, -1/sqrt(2)], [-1, 0, -1/sqrt(2)], [0, 1, 1/sqrt(2)], [0, -1, 1/sqrt(2)]];
+
+julia> T = Tetrahedron(vertices); T isa Tetrahedron
+true
+
+julia> zeros(3) ∈ T
+true
+
+julia> dim(T)
+3
+
+julia> σ(ones(3), T)
+3-element Vector{Float64}:
+ 0.0
+ 1.0
+ 0.7071067811865475
+```
+"""
+struct Tetrahedron{N,VN<:AbstractVector{N},VT<:AbstractVector{VN}} <: AbstractPolytope{N}
+    vertices::VT
+    # default constructor with length constraint
+    function Tetrahedron(vertices::VT) where {N,VN<:AbstractVector{N},VT<:AbstractVector{VN}}
+        # It is assumed (not checked) that each vertex has length 3.
+        @assert length(vertices) == 4 "a tetrahedron needs four vertices, got $(length(vertices))"
+        return new{N,VN,VT}(vertices)
+    end
+end
+
+isoperationtype(::Type{<:Tetrahedron}) = false
+
+# constructor with empty vertices list
+Tetrahedron{N}() where {N} = Tetrahedron(Vector{Vector{N}}())
+
+# constructor with no vertices of type Float64
+Tetrahedron() = Tetrahedron{Float64}()
+
+# constructor from rectangular matrix
+function Tetrahedron(vertices_matrix::MT) where {N,MT<:AbstractMatrix{N}}
+    vertices = [vertices_matrix[:, j] for j in axes(vertices_matrix, 2)]
+    return Tetrahedron(vertices)
+end
+
+function dim(::Tetrahedron)
+    return 3
+end
+
+"""
+    σ(d::AbstractVector, P::VPolytope)
+
+Return a support vector of a tetrahedron in a given direction.
+
+### Input
+
+- `d` -- direction
+- `P` -- tetrahedron
+
+### Output
+
+A support vector in the given direction.
+
+### Algorithm
+
+Currently falls back to the `VPolytope` implementation.
+"""
+function σ(d::AbstractVector, T::Tetrahedron)
+    return σ(d, convert(VPolytope, T))
+end
+
+"""
+    ∈(x::AbstractVector, T::Tetrahedron)
+
+Check whether a given point is contained in a tetrahedron.
+
+### Input
+
+- `x`      -- point/vector
+- `P`      -- tetrahedron in vertex representation
+
+### Output
+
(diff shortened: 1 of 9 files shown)


src/Sets/Tetrahedron.jl
```diff
@@ -0,0 +1,111 @@
...
+### Examples
+
+A tetrahedron can be constructed by passing the list of vertices.
+The following builds the tetrahedron with edge length 2 from the [wikipedia page Tetrahedron](https://en.wikipedia.org/wiki/Tetrahedron): 
+
+```jldoctest
+julia> vertices = [[1, 0, -1/sqrt(2)], [-1, 0, -1/sqrt(2)], [0, 1, 1/sqrt(2)], [0, -1, 1/sqrt(2)]];
+
+julia> T = Tetrahedron(vertices); T isa Tetrahedron
+true
```
schillic, line 21: I find writing two commands in one line hard to read, and `T isa Tetrahedron` is also not very surprising.

src/Sets/Tetrahedron.jl
```diff
@@ -0,0 +1,111 @@
...
+
+julia> zeros(3) ∈ T
+true
+
+julia> dim(T)
+3
```
schillic, line 27: Maybe flip the order because that motivates the `zeros(3)`:

```suggestion
julia> dim(T)
3

julia> zeros(3) ∈ T
true
```

src/Sets/Tetrahedron.jl
```diff
@@ -0,0 +1,111 @@
...
+
+# constructor with empty vertices list
+Tetrahedron{N}() where {N} = Tetrahedron(Vector{Vector{N}}())
+
+# constructor with no vertices of type Float64
+Tetrahedron() = Tetrahedron{Float64}()
+
+# constructor from rectangular matrix
+function Tetrahedron(vertices_matrix::MT) where {N,MT<:AbstractMatrix{N}}
+    vertices = [vertices_matrix[:, j] for j in axes(vertices_matrix, 2)]
```
schillic, line 50:
```suggestion
    vertices = [c for c in eachcol(vertices_matrix)]
```

```julia
julia> M = rand(2, 2)
2×2 Matrix{Float64}:
 0.118181  0.549152
 0.925909  0.79909

julia> @time [M[:, j] for j in axes(M, 2)]
  0.034929 seconds (43.36 k allocations: 2.315 MiB, 99.19% compilation time)
 [0.11818109598081683, 0.9259090178191179]
 [0.5491517086373754, 0.7990903995880422]

julia> @time [c for c in eachcol(M)]
  0.000011 seconds (3 allocations: 208 bytes)
2-element Vector{SubArray{Float64, 1, Matrix{Float64}, Tuple{Base.Slice{Base.OneTo{Int64}}, Int64}, true}}:
 [0.11818109598081683, 0.9259090178191179]
 [0.5491517086373754, 0.7990903995880422]
```

src/Sets/Tetrahedron.jl
```diff
@@ -0,0 +1,111 @@
...
+
+julia> σ(ones(3), T)
+3-element Vector{Float64}:
+ 0.0
+ 1.0
+ 0.7071067811865475
+```
+"""
+struct Tetrahedron{N,VN<:AbstractVector{N},VT<:AbstractVector{VN}} <: AbstractPolytope{N}
+    vertices::VT
```
schillic, line 37: I added a constructor with a quick sanity check.

```suggestion
    vertices::VT

    function Tetrahedron(vertices::AbstractVector)
        @assert length(vertices) == 4 "a tetrahedron requires four vertices"
        return new(vertices)
    end
```

Maybe we should also check for linear independence. The check can be made optional, and in general we can easily deactivate `@assert`s for benchmarks.

src/Sets/Tetrahedron.jl
```diff
@@ -0,0 +1,111 @@
...
+
+"""
+    ∈(x::AbstractVector, T::Tetrahedron)
+
+Check whether a given point is contained in a tetrahedron.
+
+### Input
+
+- `x`      -- point/vector
+- `P`      -- tetrahedron in vertex representation
```
schillic, line 88:
```suggestion
- `x` -- point/vector
- `P` -- tetrahedron in vertex representation
```

src/Sets/Tetrahedron.jl
```diff
@@ -0,0 +1,111 @@
...
+Currently falls back to the `VPolytope` implementation.
+TODO Implement efficient version, see LazySets#3259.
+"""
+function ∈(x::AbstractVector, T::Tetrahedron)
+    return x ∈ convert(VPolytope, T)
+end
+
+function rand(::Type{Tetrahedron}; N::Type{<:Real}=Float64, rng::AbstractRNG=GLOBAL_RNG, seed::Union{Int, Nothing}=nothing)
+    # Dummy result for tests.
+    vertices = [N[1, 0, -1/sqrt(2)], N[-1, 0, -1/sqrt(2)], N[0, 1, 1/sqrt(2)], N[0, -1, 1/sqrt(2)]];
```
schillic, line 105: Does the following work?
```suggestion
    P = rand(VPolytope; N=N, dim=3, rng=rng, seed=seed, num_vertices=4)
    vertices = P.vertices
```

schillic (approved):
Currently I do not see any benefit over using `VPolytope`.

mforets:
> Currently I do not see any benefit over using VPolytope.

you're right -- i plan to do https://github.com/JuliaReach/LazySets.jl/issues/3259 as the next step

src/Sets/Tetrahedron.jl
mforets: i'd leave it general for now, but can add a check that it is actually a tetrahedron  later.

src/Sets/Tetrahedron.jl
mforets: ok ill remove the `isa`

src/Sets/Tetrahedron.jl
mforets: right, good suggestion. it is unlikely a [*regular* tetrahedron](https://en.wikipedia.org/wiki/Tetrahedron), but it'll be a general one.

src/Sets/Tetrahedron.jl
mforets: that's due to measuring in global scope.

```julia
function Tetrahedron(vertices_matrix::MT) where {N,MT<:AbstractMatrix{N}}
    vertices = [vertices_matrix[:, j] for j in axes(vertices_matrix, 2)]
    return Tetrahedron(vertices)
end
```

```julia
julia> M = rand(3, 4);

julia> @btime LazySets.Tetrahedron($M);
  146.802 ns (5 allocations: 400 bytes)
```

though

```julia
function Tetrahedron2(vertices_matrix::MT) where {N,MT<:AbstractMatrix{N}}
    vertices = collect(eachcol(vertices_matrix))
    return Tetrahedron(vertices)
end
```

```julia
julia> @btime LazySets.Tetrahedron2($M);
  36.170 ns (1 allocation: 224 bytes)
```

mforets: the difference is due to `vertices_matrix[:, j]` allocates while `eachcol(..)` returns views

PR #3264, opened 2023-04-05
Revise interface tests

schillic:


mforets (approved):

test/Interfaces/LazySet.jl
```diff
@@ -1,6 +0,0 @@
-for ST in LazySets.subtypes(LazySet, true)
-    # check that certain methods are implemented
-    isconvextype(ST)
```
mforets, line 3: aren't these nice to keep?

schillic: I would say not, because they only test that the method does not crash. Since there are default implementations for `LazySet`, these are rather useless. I would add proper tests checking that the result is correct. But this has to be done individually for each set type.

PR #3267, opened 2023-04-06
Membership check for Tetrahedron

mforets:
Closes https://github.com/JuliaReach/LazySets.jl/issues/3259

### DIFF
--- a/src/Sets/Tetrahedron.jl
+++ b/src/Sets/Tetrahedron.jl
@@ -110,9 +110,7 @@ function ∈(x::AbstractVector, T::Tetrahedron)
            same_side(v[2], v[3], v[4], v[1], x)
 end
 
-"""
-Return `true` iff point `x` lies in the same half-space as `v4` with respect to the hyperplane determined by `v1`, `v2` and `v3`.
-"""
+# Return `true` iff point `x` lies in the same half-space as `v4` with respect to the hyperplane determined by `v1`, `v2` and `v3`.
 function same_side(v1, v2, v3, v4, x)
     normal = cross(v2 - v1, v3 - v1)
     dotv4 = dot(normal, v4 - v1)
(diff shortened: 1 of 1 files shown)


src/Sets/Tetrahedron.jl
```diff
@@ -97,11 +97,27 @@ Check whether a given point is contained in a tetrahedron.
...
+    return same_side(v[1], v[2], v[3], v[4], x) &&
+           same_side(v[4], v[1], v[2], v[3], x) &&
+           same_side(v[3], v[4], v[1], v[2], x) &&
+           same_side(v[2], v[3], v[4], v[1], x)
+end
+
+"""
+Return `true` iff point `x` lies in the same half-space as `v4` with respect to the hyperplane determined by `v1`, `v2` and `v3`.
+"""
+function same_side(v1, v2, v3, v4, x)
```
schillic, line 116: I think you need to use normal comments because `Documenter` may read this as a documentation (which is not included).

schillic (approved):

src/Sets/Tetrahedron.jl
```diff
@@ -97,11 +97,27 @@ Check whether a given point is contained in a tetrahedron.
...
+    v = T.vertices
+    return same_side(v[1], v[2], v[3], v[4], x) &&
+           same_side(v[4], v[1], v[2], v[3], x) &&
+           same_side(v[3], v[4], v[1], v[2], x) &&
+           same_side(v[2], v[3], v[4], v[1], x)
+end
+
+"""
+Return `true` iff point `x` lies in the same half-space as `v4` with respect to the hyperplane determined by `v1`, `v2` and `v3`.
+"""
```
mforets, line 115:
```suggestion
# Return `true` iff point `x` lies in the same half-space as `v4` with respect to the hyperplane determined by `v1`, `v2` and `v3`.
```

PR #3292, opened 2023-04-10
Outsource solver code to separate file

schillic:


src/Utils/solvers.jl
```diff
@@ -0,0 +1,27 @@
+# default LP solver for floating-point numbers
```
mforets, line 1: maybe rename the file to `lp_solvers`?

mforets (approved):

src/Utils/solvers.jl
schillic: Good point, I renamed it.

PR #3322, opened 2023-05-08
Fix linear_map with number

schillic:
This fixes a bug in #3284.


schillic:
I merge this and make a new release to fix the bug ASAP.

PR #3356, opened 2023-07-16
Switch from thread-local to task-local LP solvers

Zinoex:
As part of another project where we also use LP solvers in parallel, I've realized that it is better/more correct to use a task-local solver. The problem is that as tasks are executed, they may [migrate between threads](https://docs.julialang.org/en/v1/manual/multi-threading/#man-task-migration) (in particular when they yield). This means that an LP started in one task on say thread 1, might move to thread 2 following a yield and another task might start on thread 1. In this case, the second task interferes with the model used for the first task, as that was selected from thread 1. Instead, we should use task-local storage (so does the global RNGs, not sure why I read the older version of the RNG code where it was thread-local). With the current thread-local version, it is still correct if parallelized with `Threads.@threads :static for ...` - however, there might be performance gains with not using `:static`. I am also unsure whether the code yields at all, and thus might not perform task migration, but I would rather be on the safe side wrt. correctness. I expect the performance to be fairly similar to the thread-local version as `Threads.@threads` spawn `Threads.nthreads()` tasks and the entry point of the program is also wrapped in a single Task object.

### DIFF
--- a/src/Utils/lp_solvers.jl
+++ b/src/Utils/lp_solvers.jl
@@ -1,28 +1,8 @@
-const THREAD_FLOAT_LP_SOLVERs = JuMP.Model[]
-const THREAD_EXACT_LP_SOLVERs = JuMP.Model[]
+function default_lp_solver(::Type{T}) where {T}
+    key = task_local_lp_solver_key(T)
+    LP = get!(() -> JuMP.Model(default_lp_solver_factory(T)), task_local_storage(), key)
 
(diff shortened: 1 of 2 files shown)


schillic (approved):
Thanks, I guess I just believe you that this is the better implementation.

Zinoex:
In the vast majority of cases, the thread-local approach would be sufficient, but for the corner case of task migration, task-local is the right approach. This approach just mimics how the default RNG in the standard library works.

PR #3364, opened 2023-07-29
Better linear_map of AbstractZonotope for 1D output

schillic:
If the output dimension is 1, we may still want to return a `Zonotope` for type stability. But the result can have just one generator.

```julia
julia> @time linear_map(M, Z)  # master
  0.000015 seconds (3 allocations: 176 bytes)
Zonotope{Float64, Vector{Float64}, Matrix{Float64}}([0.6565135067379049], [0.06137160508809101 -1.0290195455026827 -0.29127884051790404])

julia> @time linear_map(M, Z)  # this branch
  0.000011 seconds (3 allocations: 160 bytes)
Zonotope{Float64, Vector{Float64}, Matrix{Float64}}([0.6565135067379049], [-1.2589267809324955;;])
```

### DIFF
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -306,6 +306,27 @@ function linear_map(M::AbstractMatrix, Z::AbstractZonotope)
     @assert dim(Z) == size(M, 2) "a linear map of size $(size(M)) cannot be " *
                                  "applied to a set of dimension $(dim(Z))"
 
+    if size(M, 1) == 1
+        # yields only one generator
+        return _linear_map_zonotope_1D(M, Z)
+    else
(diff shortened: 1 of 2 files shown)


src/Interfaces/AbstractZonotope.jl
```diff
@@ -306,6 +306,27 @@ function linear_map(M::AbstractMatrix, Z::AbstractZonotope)
...
                                  "applied to a set of dimension $(dim(Z))"
 
+    if size(M, 1) == 1
+        # yields only one generator
+        return _linear_map_zonotope_1D(M, Z)
+    else
+        return _linear_map_zonotope_nD(M, Z)
+    end
+end
+
```
mforets, line 316: maybe leave a comment here that this reduces the order of the zonotope

mforets (approved):

PR #3381, opened 2023-09-16
Bump compat for Documenter to 1

schillic:
This PR is equivalent to #3380. The problem is that `CompatHelper` builds always enforce to use the latest version of each dependency. We currently cannot use the latest version of `IntervalArithmetic` in the tests because `TaylorModels` holds it back, and hence the build in #3380 failed.


test/Project.toml
```diff
@@ -32,7 +32,7 @@ WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
 [compat]
 CDDLib = "0.6 - 0.9"
 Distributions = "0.19 - 0.25"
-Documenter = "0.23 - 0.27"
+Documenter = "0.23 - 0.27, 1"
```
mforets, line 35:
```suggestion
Documenter = "1"
```

docs/make.jl
```diff
@@ -80,7 +80,7 @@ makedocs(; sitename="LazySets.jl",
                                  "Parallel" => "lib/parallel.md"],
                 "About" => "about.md"],
          doctest=false,
-         strict=true)
+         pagesonly=true)
```
mforets, line 83:
```suggestion
         pagesonly=true,
         size_threshold = nothing)
```
?

PR #3427, opened 2024-01-13
spz docs nitpicks

mforets:
This PR introduces some minor updates to sparse polynomial zonotope docs.


mforets:
Some issue with julia v1.6 test. Shall we test with a more recent Julia version, such as 1.10 or 1.9 and drop older versions from CI?

schillic (approved):

schillic:
We should definitely test on the latest version, and we do (called "Julia 1"). Testing on 1.6 checks that we do not break the oldest supported version. This version unfortunately sometimes has multithreading issues, which are known and have been announced to not get fixed. You can ignore that run if you want, but I would like to keep it for now.

PR #3443, opened 2024-02-25
Replace Integer -> Int

schillic:


mforets (approved):
Not sure why we dont have a green pipeline (should we remove julia v1.6 tests?) but changes lgtm

PR #3447, opened 2024-02-26
`convert` from SparsePolynomialZonotope to Taylor model

schillic:
This branch is based on #3440 ~~and #3446~~.

One commit added convenience functions from `ReachabilityAnalysis` to construct important intervals (`zero_itv`/`sym_itv`). There are two follow-up tasks related to them.
- There was a ToDo to generalize them to other numeric types, so I did that. But that implies that currently every call creates a new instance. We could have a dictionary instead that stores these instances, or just store the `Float64` one because that is the one we would use in practice. But this can be done in a later step.
- We should then use these functions in `ReachabilityAnalysis`.


mforets (approved):

schillic:
@mforets do you also indirectly approve #3440, which this branch is based on?

PR #3484, opened 2024-04-06
Conversion from zonotopic set to Hyperrectangle

schillic:


mforets:
I see it can be useful syntax, but slightly confusing, given that only works for some but not (actually most) zonotopes? While in the other direction it always works. I'd rather not have this method, and in practice suggest using `overapproximate(Hyperrectangle, Z::Zonotope)`.

mforets (approved):

schillic:
Unfortunately, I do not remember why I suggested this PR. I would now agree to use `overapproximate(Hyperrectangle, Z::Zonotope)`. The implementation here is probably more efficient, but it is true that it only applies in rare cases. Nevertheless, Julia generally allows `convert` even if most inputs are invalid:

```julia
julia> convert(Int, 1.0)
1

julia> convert(Int, 1.1)
ERROR: InexactError: Int64(1.1)
```

In that light, I think we can add this method.

PR #3496, opened 2024-04-22
Use Unicode symbols in documentation

schillic:


schillic:
I merge this because it does not change any code and may require some rebasing in other PRs.

PR #3526, opened 2024-06-01
Support `is_interior_point` for mixed numeric types

schillic:


mforets (approved):

src/Interfaces/LazySet.jl
```diff
@@ -585,7 +585,17 @@ The default implementation determines `v ∈ interior(X)` with error tolerance
 `ε` by checking whether a `Ballp` of norm `p` with center `v` and radius `ε` is
 contained in `X`.
 """
-function is_interior_point(v::AbstractVector{N}, X::LazySet{N}; p=N(Inf), ε=_rtol(N)) where {N}
+function is_interior_point(v::AbstractVector, X::LazySet; kwargs...)
```
mforets, line 588: there's maybe an advantage of keeping `is_interior_point(v::AbstractVector{N}, X::LazySet{N}; p=N(Inf), ε=_rtol(N)) where {N}` and making `is_interior_point(v::AbstractVector, X::LazySet; kwargs...)` to dispatch on it?

the latter seems very generic (like `v` could hold any type, not just numbers)

mforets: the more general method could be written as `is_interior_point(v::AbstractVector{N}, X::LazySet{M}; kwargs...) where {N<:Number, M<:Number}`

schillic: I added another commit to address the comment.

PR #3530, opened 2024-06-16
Add tests for `API` module

schillic:
There are also some small revisions of `sample` and `is_interior_point`.


mforets (approved):

schillic:
The documentation fails, but this is unrelated (also fails on `master`). I merge here and then investigate on `master`.

PR #3533, opened 2024-06-18
Add more and better operations for `Interval`

schillic:


src/Sets/Interval/distance.jl
```diff
@@ -0,0 +1,7 @@
+function distance(X::Interval, Y::Interval; p::Real=2)
```
mforets, line 1: is it always independent of `p` ?

mforets (approved):
one comment left

src/Sets/Interval/distance.jl
schillic: Yes, all p-norms are equivalent in 1D.

PR #3551, opened 2024-07-05
Revise documentation of `Interval` operations

schillic:


mforets (approved):
removing docs for methods which are already documented somewhere else (for another set type or in some default), makes sense to me.

the cases for which it would a be a pity to remove are those that have an algorithmic description or a non-obvious input / output. but for those cases we can always keep the previous doc as code comment.

schillic:
> the cases for which it would a be a pity to remove are those that have an algorithmic description or a non-obvious input / output. but for those cases we can always keep the previous doc as code comment.

Yes, agreed. I think I always kept algorithm and other notes in this PR unless they were trivial.

PR #3553, opened 2024-07-07
Disallow `box_approximation` for unbounded sets

schillic:
```julia
julia> X = Line([0.0, 1.0], [-1.0, 1.0]);

julia> box_approximation(X)  # master
Hyperrectangle{Float64, Vector{Float64}, Vector{Float64}}([NaN, NaN], [Inf, Inf])

julia> box_approximation(X)  # this branch
ERROR: ArgumentError: the box approximation of an unbounded set is undefined
```


mforets (approved):
makes sense

PR #3592, opened 2024-07-17
Remove `RotatedHyperrectangle`

schillic:
This set type has never been used, so I suggest to remove it.
(We can always recover the files from the history if desired at a later point.)


mforets:
> Type that represents a hyperrectangle that is not necessarily axis aligned.

This is a special kind of zonotope, I wonder if there are some specific features or reasons why it was added (i dont remember).

schillic:
I added it back then and I think the motivation was a reachability algorithm for discrete systems where you start from a hyperrectangle and only apply `linear_map`s. But this is so specific that it is probably not worth keeping this type for it, even in this special case the savings are probably not significant, and we never actually used it. Here are the main differences:
* The `Zonotope` type implicitly uses the unit hypercube. Instead, a `RotatedHyperrectangle` uses an arbitrary hyperrectangle, but stores it explicitly. I think in most cases this additional storage is just overhead, but starting from a hyperrectangle is easier.
* The name `RotatedHyperrectangle` is very misleading. In fact, it can take the shape of any `Zonotope` because the matrix can be rectangular.
* Additionally, the `Zonotope` stores a translation vector, which the `RotatedHyperrectangle` does not. So for every `linear_map` operation you save a matrix-vector multiplication, but the translation operation is difficult instead (we do not even have it implemented).

In summary, I do not oppose keeping the `RotatedHyperrectangle` type, but since I was the one adding it and never used it, I am open to removing it. Reducing the code base has benefits as well.

mforets (approved):

PR #3609, opened 2024-07-23
#3495 - Avoid repeated creation of `zero_box`/`sym_box`

### LINKED ISSUE #3495: Use only one interval instance for `zero_itv` / `sym_itv`
In #3447 we added convenience functions from `ReachabilityAnalysis` to construct important intervals (`zero_itv`/`sym_itv`). They were generalized to other numeric types; hence currently every call creates a new instance. We should have a dictionary instead that stores these instances, or just store the `Float64` one because that is the one we would use in practice.

schillic:
Closes #3495.


mforets (approved):
Approving, but im not sure if this has any perf benefits? Not easy to measure in an actual example, as the difference might be quite small.

PR #3613, opened 2024-07-25
Merge metaprogramming loops

schillic:
The function `_issubset_in_emptyset` calls `isempty`. The removed loop used a shortcut for set types that are never empty, but the compiler figures that out and also skips `isempty`. So this was not really faster. Hence, I suggest to remove the loop to keep the codebase smaller.

### DIFF
--- a/src/ConcreteOperations/issubset.jl
+++ b/src/ConcreteOperations/issubset.jl
@@ -790,13 +790,8 @@ function _issubset_in_emptyset(X::LazySet, ∅::EmptySet, witness::Bool=false)
 end
 
 # disambiguations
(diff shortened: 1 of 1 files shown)


mforets (approved):

mforets:
> The removed loop used a shortcut for set types that are never empty, but the compiler figures that out and also skips isempty. 

that sounds quite advanced. how does it figure that out?

schillic:
> > The removed loop used a shortcut for set types that are never empty, but the compiler figures that out and also skips isempty.
> 
> that sounds quite advanced. how does it figure that out?

Simply because `isempty(X) = false` is constant, and then the `if isempty(X)` can be optimized away.

PR #3618, opened 2024-07-26
Split `convert.jl` into smaller files

schillic:


mforets (approved):
I see this and recall we discussed it already, so approving.

(But my personal choice would be not to fragment that much. We now end up with lots and lost of files that extend the same Julia method, with 50 loc or less, which seems kind of awkward.)

mforets:
The approach taken in which the file name corresponds to the target set is nice though.

PR #3623, opened 2024-07-26
Let `remove_redundant_constraints` return `nothing`

schillic:
Currently, `remove_redundant_constraints` returns an `EmptySet` if the constraints are contradicting. I think this is unexpected. Instead, I suggest to return `nothing`.

### DIFF
--- a/src/Interfaces/AbstractPolyhedron_functions.jl
+++ b/src/Interfaces/AbstractPolyhedron_functions.jl
@@ -235,7 +235,7 @@ Remove the redundant constraints of a given list of linear constraints.
                    linear program
 ### Output
 
-The list of constraints with the redundant ones removed, or an empty set if the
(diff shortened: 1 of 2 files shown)


src/Interfaces/AbstractPolyhedron_functions.jl
```diff
@@ -253,8 +253,7 @@ function remove_redundant_constraints(constraints::AbstractVector{S};
     if remove_redundant_constraints!(constraints_copy; backend=backend)
         return constraints_copy
     else  # the constraints are infeasible
-        N = eltype(first(constraints))
-        return EmptySet{N}(dim(constraints[1]))
+        return nothing
```
mforets, line 256: maybe lets instead return an array of the same type as `constraints` if it failed to find a reduction.

schillic: done

PR #3662, opened 2024-12-11
Add binary `convex_hull` method to `API` module

schillic:
As an alternative to this PR, I was wondering whether we should remove the binary `convex_hull` method from the library.

The convex hull is traditionally defined for a single set. Originally, there was no way to express unions of two sets, but now we can express that via a `UnionSet`. So it feels a bit arbitrary that we offer this binary method.

A similar argument goes for the lazy binary operation `ConvexHull`, which could be turned into a unary operation.

### DIFF
--- a/docs/src/lib/API.md
+++ b/docs/src/lib/API.md
@@ -84,6 +84,7 @@ translate!(::LazySet, ::AbstractVector)
 
 ```@docs
 cartesian_product(::LazySet, ::LazySet)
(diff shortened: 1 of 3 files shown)


mforets:
I think we should keep `convex_hull` / `ConvexHull` binary operations as they are common in reachability algorithms.

For the unary, lazy convex hull, we could define: `ConvexHull(X) = ConvexHull(X, X)`.

PR #3673, opened 2024-12-15
`EmptySet`: revise tests

schillic:
This branch is based on #3667 and #3672.
Only the last commit is new.

A test is added for each known operation, including assertion violations, proper comparison of numeric types, and commutativity tests.


schillic:
To move forward and not be blocked by #3667, I included the tests in #3672.

PR #3676, opened 2024-12-15
`AbstractHyperrectangle`: let `difference` return results of different types

schillic:
This is an optional PR. It arguably makes the result easier to understand if empty, but since it leads to type instability, I am not sure about it.


schillic:
We decided to close this one.

PR #3725, opened 2024-12-30
Add `isparametrictype` trait

schillic:
This function is used in one place to prevent bugs later when adding other parametric set types.
The function is not exported.

### DIFF
--- a/src/ConcreteOperations/exact_sum.jl
+++ b/src/ConcreteOperations/exact_sum.jl
@@ -1,12 +1,12 @@
-# the only parametric sets in this library are polynomial zonotopes;
-# hence, we compute the normal Minkowski sum for all other set combinations
+# for non-parametric set types, the exact sum coincides with the Minkowski sum
 function exact_sum(X::LazySet, Y::LazySet)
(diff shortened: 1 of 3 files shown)


src/ConcreteOperations/exact_sum.jl
```diff
@@ -1,12 +1,12 @@
-# the only parametric sets in this library are polynomial zonotopes;
-# hence, we compute the normal Minkowski sum for all other set combinations
+# for non-parametric set types, the exact sum coincides with the Minkowski sum
 function exact_sum(X::LazySet, Y::LazySet)
     @assert dim(X) == dim(Y) "the dimensions of the given sets should match, " *
                              "but they are $(dim(X)) and $(dim(Y)), respectively"
 
-    if X isa AbstractPolynomialZonotope || Y isa AbstractPolynomialZonotope
+    if isparametrictype(typeof(X)) || isparametrictype(typeof(Y))
```
mforets, line 6: what is a parametric type ?

schillic: Currently it is the same as a polynomial zonotope. It is the difference of the `exact_sum` to the `minkowski_sum`.

PR #3758, opened 2025-01-03
Inline one-line helper functions

schillic:
Argument: The helper functions are not used anywhere else and consist of only a single function call.


alecarraro (approved):
lgtm

PR #3768, opened 2025-01-04
Add missing documentation

schillic:


src/LazyOperations/MinkowskiSumArray.jl
```diff
@@ -24,6 +24,14 @@ struct MinkowskiSumArray{N,S<:LazySet{N}} <: LazySet{N}
     array::Vector{S}
 end
 
+"""
+    MinkowskiSum!(X, Y)
+
+Convenience function to compute the lazy Minkowski sum and modify
+`MinkowskiSumArray`s in-place.
+"""
+function MinkowskiSum! end
```
mforets, line 33: Is this new or it existed already and used elsewhere?

mforets (approved):

src/LazyOperations/MinkowskiSumArray.jl
schillic: It already existed, but I am not sure whether it is used anywhere. Now it is at least documented better.

https://github.com/JuliaReach/LazySets.jl/blob/bd58c7c749b4397544daa670e5e01a95638252ad/src/Utils/macros.jl#L246-L254

PR #3775, opened 2025-01-04
`Rectification`: add some `isconvextype` special cases

schillic:


schillic:
I merge this because it is open for more than 7 months and should be non-controversial.

PR #3776, opened 2025-01-04
Allow to construct H-rep from mixed numeric types

schillic:
Also fix an error message.

### DIFF
--- a/src/Convert/convert.jl
+++ b/src/Convert/convert.jl
@@ -80,7 +80,7 @@ for T in subtypes(AbstractHPolygon, true)
             @assert dim(X) == 2 "set must be two-dimensional for conversion, but it " *
                                 "has dimension $(dim(X))"
             if check_boundedness && !isbounded(X)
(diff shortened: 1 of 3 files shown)


src/Convert/convert.jl
```diff
@@ -80,7 +80,7 @@ for T in subtypes(AbstractHPolygon, true)
             @assert dim(X) == 2 "set must be two-dimensional for conversion, but it " *
                                 "has dimension $(dim(X))"
             if check_boundedness && !isbounded(X)
-                throw(ArgumentError("expected a bounded set for conversion to `$T`"))
+                throw(ArgumentError("expected a bounded set for conversion to `$($T)`"))
```
mforets, line 83: is this double `$` needed?

mforets (approved):

src/Convert/convert.jl
schillic: Yes, the outer `$` is string interpolation and the inner `$` is needed inside the `@eval` block.

PR #3777, opened 2025-01-04
SPZ overapproximation of CH(::SPZ,::SPZ)

schillic:


src/Approximations/overapproximate.jl
```diff
@@ -648,6 +648,45 @@ function load_paving_overapproximation()
...
+# Extended help
+
+    overapproximate(CH::ConvexHull{N,<:SparsePolynomialZonotope,<:SparsePolynomialZonotope},
+                    ::Type{<:SparsePolynomialZonotope}) where {N}
+
+### Algorithm
+
+This method implements Proposition 3.1.28 in [1].
+
+[1] Kochdumper, Niklas. *Extensions of polynomial zonotopes and their application
```
alecarraro: I checked the thesis, and the method seems correct. Only thing I notice is that in the other files the citations use `Documenter.jl`. So in this case the correct reference would be "[Kochdumper21a; Proposition 3.1.28](@citet)""

schillic: Yes, the new citations were probably added after this PR was opened. I had fixed it earlier today already, but good catch!

PR #3787, opened 2025-01-21
Add `distance` between point and set to `API` module 

schillic:


schillic:
Apparently I already did this in #3722...

Issue #3455, opened 2024-03-02
Convex hull algorithm in 2D sensitive to numeric precision


schillic:
The following happens because of the `-0.0`. Nevertheless, the result is unexpected.

```julia
julia> v = [[-0.0, 2], [2.0, 0], [1.0, 0], [0.0, 1], [0.0, 1.5]]
5-element Vector{Vector{Float64}}:
 [-0.0, 2.0]
 [2.0, 0.0]
 [1.0, 0.0]
 [0.0, 1.0]
 [0.0, 1.5]

julia> c = convex_hull(v)
4-element Vector{Vector{Float64}}:
 [-0.0, 2.0]
 [0.0, 1.5]
 [1.0, 0.0]
 [2.0, 0.0]

julia> plot([Singleton(x) for x in v], c=:blue, alpha=1)

julia> plot!([Singleton(x) for x in c], c=:red, alpha=1)

julia> plot!(VPolygon(v))
```

![convex_hull](https://github.com/JuliaReach/LazySets.jl/assets/9656686/062ebe3b-6e3b-435a-8803-429fec01b6d6)

schillic:
The problem is that `sort!` gives the wrong order due to the `-0.0`. It uses the order relation `<`, which ultimately calls the following `Base` function:

```julia
function cmp(A::AbstractVector, B::AbstractVector)
    for (a, b) in zip(A, B)
        if !isequal(a, b)
            return isless(a, b) ? -1 : 1
        end
    end
    return cmp(length(A), length(B))
end
```

Here, `isequal` distinguishes `0.0` and `-0.0` and thus `sort!` outputs `[-0.0, 2]` before `[2.0, 0]`. The remaining algorithm checks right turns, which are not precise enough to see the difference, and thus gives the wrong result.

I see two possible solutions:

1. Sanitize `-0.0` values to `0.0`.
2. Change the `sort!` order relation.

mforets:
> Change the sort! order relation.

maybe this relates to using a custom `<` comparison such as `_le` / `_leq` if that applies

schillic:
No, this is not allowed. From the current documentation of `sort!`:

```julia
The lt function must define a strict weak order, that is, it must be

    •  irreflexive: lt(x, x) always yields false,

    •  asymmetric: if lt(x, y) yields true then lt(y, x) yields false,

    •  transitive: lt(x, y) && lt(y, z) implies lt(x, z),

[...]

PR #3797, opened 2025-02-04
Fixes incorrect convex hull in #3455

crisiumnih:
- Replaces -0.0 with 0.0 in `!sort`
- Resolves incorrect convex hull results

Results - 
hello

### DIFF
--- a/src/ConcreteOperations/convex_hull.jl
+++ b/src/ConcreteOperations/convex_hull.jl
@@ -197,7 +197,7 @@ function _two_points_2d!(points)
 end
 
 function _three_points_2d!(points::AbstractVector{<:AbstractVector{N}}) where {N}
-    # Algorithmmonotone_chain!: the function takes three points and uses the formula from here:
+    # Algorithm: the function takes three points and uses the formula from here:
     # https://stackoverflow.com/a/2122620
     # to decide if the points are ordered in counter-clockwise order. The result
     # is saved in the 'turn' flag and returned in ccw fashion acting according
--- a/src/ConcreteOperations/convex_hull.jl
+++ b/src/ConcreteOperations/convex_hull.jl
@@ -197,7 +197,7 @@ function _two_points_2d!(points)
 end
 
 function _three_points_2d!(points::AbstractVector{<:AbstractVector{N}}) where {N}
-    # Algorithm: the function takes three points and uses the formula from here:
+    # Algorithmmonotone_chain!: the function takes three points and uses the formula from here:
     # https://stackoverflow.com/a/2122620
     # to decide if the points are ordered in counter-clockwise order. The result
     # is saved in the 'turn' flag and returned in ccw fashion acting according
@@ -458,7 +458,7 @@ function monotone_chain!(points::Vector{VN}; sort::Bool=true) where {N,VN<:Abstr
 
     if sort
         # sort the points lexicographically
-        sort!(points; by=x -> (x[1], x[2]))
+        sort!(points; by=x -> (x[1] == -0.0 ? 0.0 : x[1], x[2] == -0.0 ? 0.0 : x[2]))
     end
 
     # build lower hull


src/ConcreteOperations/convex_hull.jl
```diff
@@ -458,7 +458,7 @@ function monotone_chain!(points::Vector{VN}; sort::Bool=true) where {N,VN<:Abstr
 
     if sort
         # sort the points lexicographically
-        sort!(points; by=x -> (x[1], x[2]))
+        sort!(points; by=x -> (x[1] == -0.0 ? 0.0 : x[1], x[2] == -0.0 ? 0.0 : x[2]))
```
mforets, line 461: Could we not use one of the rules with tolerance defined in ReachabilityBase such as `_leq`  ?

schillic: No, this is not allowed. From the current documentation of `sort!`:

```julia
The lt function must define a strict weak order, that is, it must be

    •  irreflexive: lt(x, x) always yields false,

    •  asymmetric: if lt(x, y) yields true then lt(y, x) yields false,

    •  transitive: lt(x, y) && lt(y, z) implies lt(x, z),

    •  transitive in equivalence: !lt(x, y) && !lt(y, x) and !lt(y, z) && !lt(z, y) together imply !lt(x, z)
       && !lt(z, x). In words: if x and y are equivalent and y and z are equivalent then x and z must be
       equivalent.

  For example < is a valid lt function for Int values but ≤ is not: it violates irreflexivity. For Float64 values
  even < is invalid as it violates the fourth condition: 1.0 and NaN are equivalent and so are NaN and 2.0 but 1.0
  and 2.0 are not equivalent.
```

Since the issue seems to be restricted to `-0.0`, I think we should go with the sanitization option.

schillic: No, this is not allowed. From the current documentation of `sort!`:

```julia
The lt function must define a strict weak order, that is, it must be

    •  irreflexive: lt(x, x) always yields false,

    •  asymmetric: if lt(x, y) yields true then lt(y, x) yields false,

    •  transitive: lt(x, y) && lt(y, z) implies lt(x, z),

    •  transitive in equivalence: !lt(x, y) && !lt(y, x) and !lt(y, z) && !lt(z, y) together imply !lt(x, z)
       && !lt(z, x). In words: if x and y are equivalent and y and z are equivalent then x and z must be
       equivalent.

  For example < is a valid lt function for Int values but ≤ is not: it violates irreflexivity. For Float64 values
  even < is invalid as it violates the fourth condition: 1.0 and NaN are equivalent and so are NaN and 2.0 but 1.0
  and 2.0 are not equivalent.
```

Since the issue seems to be restricted to `-0.0`, I think we should go with the sanitization option.

schillic, line 461: Better use a helper function here:
```julia
_zero_abs(x::Number) = x

_zero_abs(x::AbstractFloat) = x == -0.0 ? zero(x) : x
```

schillic (commented):
Thanks! Please also add
- a comment why this sanitization is needed (you can refer to #3455),
- a test (use the one from #3455).

src/ConcreteOperations/convex_hull.jl
```diff
@@ -457,8 +462,9 @@ function monotone_chain!(points::Vector{VN}; sort::Bool=true) where {N,VN<:Abstr
     end
 
     if sort
+        # _zero_abs is necessary because floating-point arithmetic distinguishes between 0.0 and -0.0, which leads to incorrect sorting
```
schillic, line 465:
```suggestion
        # _zero_abs is necessary because floating-point arithmetic distinguishes
        # between 0.0 and -0.0, which leads to incorrect sorting (see #3455)
```

test/ConcreteOperations/convex_hull.jl
```diff
@@ -185,4 +185,9 @@ for N in [Float64, Rational{Int}]
     # binary convex hull with an empty set is identical to unary
     @test convex_hull(S, EmptySet{N}(1)) == convex_hull(EmptySet{N}(1), S) == S
     @test convex_hull(P, EmptySet{N}(2)) == convex_hull(EmptySet{N}(2), P) == Pc
+
+    v = [[-0.0, 2], [2.0, 0], [1.0, 0], [0.0, 1], [0.0, 1.5]]
+    v_ch = [[0.0, 1.0], [1.0, 0.0], [2.0, 0.0], [-0.0, 2.0]]
+
```
schillic, line 191: The numeric type `N` should be used instead of just `Float64` in the tests.
```suggestion
    # see #3455
    v = [N[-0.0, 2], N[0, 0], N[1, 0], N[0, 1], N[0, 3//2]]
    v_ch = [N[0, 1], N[1, 0], N[2, 0], N[-0.0, 2]]
```

src/ConcreteOperations/convex_hull.jl
```diff
@@ -443,6 +443,11 @@ construct the convex hull of a set of ``n`` points in the plane in
 ``O(n \\log n)`` time. For further details see
 [Monotone chain](https://en.wikibooks.org/wiki/Algorithm_Implementation/Geometry/Convex_hull/Monotone_chain)
 """
+
+_zero_abs(x::Number) = x
+
+_zero_abs(x::AbstractFloat) = x == -0.0 ? zero(x) : x
+
```
schillic, line 450: These need to be moved (because they now separate the docstring from the function below).

test/ConcreteOperations/convex_hull.jl
```diff
@@ -185,4 +185,9 @@ for N in [Float64, Rational{Int}]
     # binary convex hull with an empty set is identical to unary
     @test convex_hull(S, EmptySet{N}(1)) == convex_hull(EmptySet{N}(1), S) == S
     @test convex_hull(P, EmptySet{N}(2)) == convex_hull(EmptySet{N}(2), P) == Pc
+
+    # see #3455
+    v = [N[-0.0, 2], N[0, 0], N[1, 0], N[0, 1], N[0, 3//2]]
```
schillic, line 190: Sorry, I accidentally modified the example in my previous suggestion. This change will fix it.

```suggestion
    v = [N[-0.0, 2], N[2, 0], N[1, 0], N[0, 1], N[0, 3//2]]
```

schillic (approved):
Looks good after the fix!

schillic:
Thanks!

PR #3804, opened 2025-02-08
Rename and export `delaunay` and `triangulate`

schillic:
This PR renames `triangulate` to `triangulate_faces` and `delaunay` to `triangulate`, always exports these functions, and adds `triangulate` (new name) methods for `EmptySet` and `Universe` to get a useful error message.

#### Commits

- Commit 1 makes `delaunay` and `triangulate` a default (exported) function. The benefits are that there is now an error message informing which package to load and that it is possible to define methods even when `MiniQhull`/`Polyhedra` are not loaded.
- Commits 2 and 3 add `delaunay` methods for `EmptySet` and `Universe`.
~This PR is mildly breaking: when `using MiniQhull`, there is now a name clash with `delaunay`. `import MiniQhull` still works. I argue that most users do not need the `using` version.~
- Commit 4 renames `triangulate` to `triangulate_faces`.
- Commit 4 renames `delaunay` to `triangulate`.


mforets (approved):

schillic:
This needs to be reworked: the plan is to rename `triangulate` -> `triangulate_faces` and `delaunay` -> `triangulate`.

mforets (approved):

Issue #417, opened 2018-07-27
Alternative order reduction algorithm for zonotopes


schillic:
See [this paper](http://www.i6.in.tum.de/Main/Publications/Althoff2009b.pdf) for an alternative algorithm to reduce the order of a zonotope, including heuristics.

schillic:
Two more papers that are cited in the paper we got our implementation from:
* W. Kühn, Zonotope dynamics in numerical quality control, in Mathematical Visualization, H.-C. Hege, K. Polthier (Eds), pp 125-134, Springer, 1998.
* Combastel C., A state bounding observer based on zonotopes, in Proc. of European Control Conference, 2003.

mforets:
> Combastel C., A state bounding observer based on zonotopes, in Proc. of European Control Conference, 2003.

this one will be added in https://github.com/JuliaReach/LazySets.jl/pull/2220/files#diff-9e63c6c7780c2d50d0255d11bd5a17f1283f5f609b873bbb1bf94c0787cdbb5bR380

schillic:
[A comparison of several reduction techniques](https://www.researchgate.net/profile/Xuejiao-Yang-3/publication/327372765_A_comparison_of_zonotope_order_reduction_techniques/links/5c8943a292851c1df93fee6c/A-comparison-of-zonotope-order-reduction-techniques.pdf)

PR #3827, opened 2025-04-03
#417 - Addition of zonotope order reduction method

crisiumnih:
Implements method 4 of [A comparison of zonotope order reduction techniques](https://www.sciencedirect.com/science/article/abs/pii/S000510981830298X?via%3Dihub).


src/Interfaces/AbstractZonotope.jl
```diff
@@ -878,6 +892,170 @@ function reduce_order(Z::AbstractZonotope, r::Real,
     return Zonotope(c, Gred)
 end
```
schillic, line 894: You could split the previous method as follows:

```julia
function reduce_order(Z::AbstractZonotope, r::Real,
                      method::AbstractReductionMethod=GIR05())
    r >= 1 || throw(ArgumentError("the target order should be at least 1, " *
                                  "but it is $r"))
    n = dim(Z)
    p = ngens(Z)

    # if r is bigger than the order of Z => do not reduce
    (r * n >= p) && return Z
    return _reduce_order_zonotope_common(Z, r, n, p, method)
end

function _reduce_order_zonotope_common(Z, r, n, p, method::Union{ASB10,COMB03,GIR05})
    ...  # continue with old code
```

Then rename your new method signature to
```julia
function _reduce_order_zonotope_common(Z, r, n, p, method::JKS16)
```
and remove the common first lines.

Finally, you need to remove the docstring of the method. Move it to the docstring of the struct.

src/Interfaces/AbstractZonotope.jl
```diff
@@ -808,6 +808,19 @@ Zonotope order-reduction method from [AlthoffSB10](@citet).
 """
 struct ASB10 <: AbstractReductionMethod end
 
+"""
+    JKS16 <: AbstractReductionMethod
+
+Zonotope order-reduction method from [SCOTT2016126](@citet).
+"""
+struct JKS16 <: AbstractReductionMethod 
+
```
schillic, line 817:
```suggestion
```

src/Interfaces/AbstractZonotope.jl
crisiumnih: Done!

src/Interfaces/AbstractZonotope.jl
```diff
@@ -854,7 +860,8 @@ julia> subtypes(AbstractReductionMethod)
 See the documentation of each algorithm for references. These methods split the
 given zonotopic set `Z` into two zonotopes, `K` and `L`, where `K` contains the
 most "representative" generators and `L` contains the generators that are
-reduced, `Lred`, using a box overapproximation. We follow the notation from
+reduced, `Lred`, using a box overapproximation. This methodology vaires slightly
```
schillic, line 863:
```suggestion
reduced, `Lred`, using a box overapproximation. This methodology varies slightly 
```

Issue #3169, opened 2022-11-01
Outsource common code to implement concretize of lazy array operations


schillic:
Compare:

https://github.com/JuliaReach/LazySets.jl/blob/9fdf5e8ecc5f11c36715096c379a7916a88f155d/src/LazyOperations/MinkowskiSumArray.jl#L196-L208

https://github.com/JuliaReach/LazySets.jl/blob/9fdf5e8ecc5f11c36715096c379a7916a88f155d/src/LazyOperations/CartesianProductArray.jl#L658-L670

PR #3828, opened 2025-04-04
#3169 Outsource common code in concretize

alecarraro:
I created a helper function saved in the file ConcreteOperations/utils_concrete.jl to outsource the implementation of the concretize function in LazyOperations/MinkowskiSumArray and CartesianProductArray. The error message is now more generic


alecarraro:
@schillic I followed your indication to place the helper under the ConcreteOperations, however I had to make a new file to not "pollute" existing implementations. Also the error message is more generic but I could actually pass a string as an argument of  `_concretize_lazy_array` to make the error more specific.

PR #3831, opened 2025-04-06
#3169 - Outsource common `concretize` code of lazy array operations

### LINKED ISSUE #3169: Outsource common code to implement concretize of lazy array operations

alecarraro:
Closes #3169.
Refactored code for `concretize(X)`


src/LazyOperations/ConvexHullArray.jl
```diff
@@ -221,21 +221,12 @@ function ∈(x::AbstractVector, X::ConvexHullArray)
 end
 
 function concretize(cha::ConvexHullArray)
-    a = array(cha)
-    @assert !isempty(a) "an empty convex hull is not allowed"
```
schillic, line 225: This assertion should stay.

src/Interfaces/AbstractArraySet.jl
```diff
@@ -37,3 +37,15 @@ function flatten!(arr, X, bin_op)
     end
     return arr
 end
+
+function _concretize_lazy_array(container::T) where {T<:LazySet}
+    arr = array(container)
+    @assert !isempty(arr) "$(concrete_function(T)) not supported on empty array"
```
schillic, line 43:
```suggestion
    @assert !isempty(arr) "`concretize` not supported on empty array"
```

src/Interfaces/AbstractArraySet.jl
```diff
@@ -37,3 +37,15 @@ function flatten!(arr, X, bin_op)
...
+    
+    op = concrete_function(T)
+    X = concretize(first(arr))
+    @inbounds for Y in @view arr[2:end]
+        X = op(X, concretize(Y))
+    end
+    return X
+end
```
schillic, line 51:
```suggestion
end

```

src/LazyOperations/ConvexHullArray.jl
schillic: Could you re-add the assertion in this method? The reason is that there is a call to another function that does not check this condition. You can move it into the `if` block.

alecarraro: sorry , i oversaw your comment

schillic: But I think you want to move it into the `if` block...

src/Interfaces/AbstractArraySet.jl
```diff
@@ -37,3 +37,15 @@ function flatten!(arr, X, bin_op)
...
     return arr
 end
+
+function _concretize_lazy_array(container::T) where {T<:LazySet}
+    arr = array(container)
+    @assert !isempty(arr) "`concretize` not supported on empty array"
+    
+    op = concrete_function(T)
+    X = concretize(first(arr))
+    @inbounds for Y in @view arr[2:end]
```
schillic, line 47: Is there a benefit of rewriting the old loop to this version?

alecarraro: Tested on an Cartesian prod array of 100 random 2x2 zonotopes. For the old version
<pre> ```
BenchmarkTools.Trial: 3482 samples with 1 evaluation per sample.
 Range (min … max):  827.928 μs …   7.314 ms  ┊ GC (min … max):  0.00% … 66.20%
 Time  (median):       1.225 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):     1.425 ms ± 583.796 μs  ┊ GC (mean ± σ):  22.05% ± 20.81%

  ▃█▇▅▅▃▂▃▂▂▂▂▁▁▁▁▁▁   ▂▆▅▄▄▃▂▂▁▂▁   ▁       ▁                  ▁
  ████████████████████████████████████████████▇█▆█▇▄▇▄▅▅▆▆▅▅▄▆▇ █
  828 μs        Histogram: log(frequency) by time       3.17 ms <

 Memory estimate: 10.75 MiB, allocs estimate: 1158.
```</pre>
For the version using `@view`:
<pre> ``` 
BenchmarkTools.Trial: 4557 samples with 1 evaluation per sample.
Range (min … max): 754.289 μs … 4.856 ms ┊ GC (min … max): 0.00% … 50.95% 
Time (median): 1.149 ms ┊ GC (median): 26.47% 
Time (mean ± σ): 1.090 ms ± 296.663 μs ┊ GC (mean ± σ): 16.32% ± 13.70% 
▄█▄ ▁▅▅▂ ▃████▆▄▃▂▂▂▂▂▁▁▁▄████▆▄▄▃▃▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂ 
754 μs Histogram: frequency by time 2.14 ms < Memory estimate: 10.74 MiB, allocs estimate: 959. 
``` </pre>
It's not a massive win, but it reduces allocations and speed up a bit garbage collection. AFAIK it is basically a "free" optimization for large arrays

schillic: Okay, `for Y in @view arr[2:end]` seems to be equivalent to `for Y in Iterators.rest(arr, 2)` (at least I get the same allocations and timings).

I am very surprised that this saves so much. The previous code looked innocent.

schillic (approved):
Thanks!

PR #3832, opened 2025-04-07
#2768 - Outource common code in `BoxDiagDirections`

### LINKED ISSUE #2768: Shared code in DiagDirections and BoxDiagDirections
`DiagDirections` is more or less a copy of `BoxDiagDirections`. The latter should just iterate over a `DiagDirections` and a `BoxDirections` object.

alecarraro:
Closes #2768.

Refactored `BoxDiagDirections` to delegate iteration logic to combined `DiagDirections` and `BoxDirections` iterators

### DIFF
--- a/src/Approximations/template_directions.jl
+++ b/src/Approximations/template_directions.jl
@@ -576,19 +576,34 @@ LazySets.dim(bdd::BoxDiagDirections) = bdd.n
 isbounding(::Type{<:BoxDiagDirections}) = true
 isnormalized(::Type{<:BoxDiagDirections}) = false
 
-# delegator iterator 
+# initial iteratipn
 function Base.iterate(bdd::BoxDiagDirections{N,Vector{N}}) where {N}
-    return iterate(_boxdiag_iterator(bdd))
+    if bdd.n == 1
+        bd = BoxDirections{N,Vector{N}}(bdd.n)
+        return iterate(bd)
+    else
+        dd = DiagDirections{N}(bdd.n)
+        return iterate(dd)
+    end
 end
 
-function Base.iterate(bdd::BoxDiagDirections{N}, state) where {N}
-    return iterate(_boxdiag_iterator(bdd), state)
+# when the state is a vector we are in the diagonal phase
+function Base.iterate(bdd::BoxDiagDirections{N,Vector{N}}, state::Vector{N}) where {N}
+    # delegate to DiagDirections
+    dd = DiagDirections{N}(bdd.n)
+    result = iterate(dd, state)
+    if result === nothing
+        # switch to box directions
+        return bdd.n == 1 ? nothing : iterate(bdd, 1)
+    else
+        return result
+    end
 end
 
-function _boxdiag_iterator(bdd::BoxDiagDirections{N,VN}) where {N,VN}
-    dd = DiagDirections{N,VN}(bdd.n)
-    bd = BoxDirections{N,VN}(bdd.n)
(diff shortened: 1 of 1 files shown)


src/Approximations/template_directions.jl
```diff
@@ -576,34 +576,19 @@ LazySets.dim(bdd::BoxDiagDirections) = bdd.n
...
 end
 
-function Base.iterate(bdd::BoxDiagDirections{N,Vector{N}}, state::Int) where {N}
-    # continue with box directions
-    return iterate(BoxDirections{N,Vector{N}}(bdd.n), state)
+function _boxdiag_iterator(bdd::BoxDiagDirections{N,VN}) where {N,VN}
+    dd = DiagDirections{N,VN}(bdd.n)
+    bd = BoxDirections{N,VN}(bdd.n)
+    return Iterators.flatten((dd, bd))
 end
```
schillic, line 592: Is this efficient? I guess the allocation is minimal. But could you make a comparison with `@time` before and after this change?

schillic (commented):
Could you rebase the branch with `master` instead of having the changeset from #3831?

src/Approximations/template_directions.jl
alecarraro: I think you are right, it is slower. The difference becomes NOT negligible for larger $n$ (slower of a factor of 2 for n=8 on my laptop). I changed it, and now it should perform well and pass the test also for $n=1$

src/Approximations/template_directions.jl
```diff
@@ -576,34 +576,34 @@ LazySets.dim(bdd::BoxDiagDirections) = bdd.n
...
-            return nothing
-        else
-            # continue with box directions
-            return iterate(bdd, 1)
-        end
+# when the state is a vector we are in the diagonal phase
+function Base.iterate(bdd::BoxDiagDirections{N,Vector{N}}, state::Vector{N}) where {N}
+    # delegate to DiagDirections
+    dd = DiagDirections{N}(bdd.n)
+    result = iterate(dd, state)
```
schillic, line 594: This is expensive because it allocates a new object in every iteration.

src/Approximations/template_directions.jl
```diff
@@ -576,34 +576,34 @@ LazySets.dim(bdd::BoxDiagDirections) = bdd.n
...
+        return result
     end
 end
 
+# when the state is an integer we are in the box phase.
 function Base.iterate(bdd::BoxDiagDirections{N,Vector{N}}, state::Int) where {N}
-    # continue with box directions
-    return iterate(BoxDirections{N,Vector{N}}(bdd.n), state)
+    bd = BoxDirections{N,Vector{N}}(bdd.n)
+    return iterate(bd, state)
```
schillic, line 606: Same comment, but apparently this was already bad in the old implementation.

schillic (commented):
It would be better if you create the `DiagDirection` and `BoxDirections` only once and add them to the state (i.e., the state becomes a pair `(dir, oldstate)`).
Then you can add methods for dispatch that just call `iterate(dir, oldstate)`.

schillic (approved):
Very nice, this retains the old order and even has fewer allocations!

```julia
julia> @time collect(BoxDiagDirections(10));
  0.000182 seconds (3.14 k allocations: 187.836 KiB)  # master
  0.000177 seconds (2.09 k allocations: 155.211 KiB)  # this branch
```

schillic:
I had some suggestions but decided to merge here and added the suggestions separately (#3840) to speed things up.

Issue #2739, opened 2021-06-12
Concrete cartesian product in standard form


mforets:
If Z is an n-dimensional zonotope of order 2, we say that it is in standard form if its generators matrix is `G = [M D]` with M and D n x n matrices and D diagonal.

Create a function `cartesian_product_standard(X::Vector{Zonotope})` that returns the `n`-fold cartesian product in standard form, assuming all the zonotopes in `X` are in standard form. Bonus points if it can handle zonotopes (`X[i]`) of different dimensions.

schillic:
Other possibilities, ordered by effort:
* Add a new argument `normal_form::Bool` (or even `normal_form::FullDiagonal` if we want to add more normal forms later) to `cartesian_product`.
* Add a dispatch on `Zonotope`s whose generator matrix is given by a certain type (a block matrix `[M D]` where `D` is a `Diagonal`) to `cartesian_product`.
* Add a new set type `ZonotopeMD <: AbstractZonotope` which behaves like `Zonotope` but preserves the normal form.

PR #3837, opened 2025-04-16
#2739: added `ZonotopeMD` and support for Cartesian product in normal

alecarraro:
#2739 

> Add a new set type ZonotopeMD <: AbstractZonotope which behaves like Zonotope but preserves the normal form.

With reference to the issue above, I added a new `ZonotopeMD` type, a representation of zonotopes of order `k` in normal form. This type has 2 constructors for building `ZonotopeMD` from either a  `[M D]` generator matrix or from `M` and `d` separately.
 The specialized `cartesian_product` returns a `ZonotopeMD` in normal form. The `M'` block in the resulting generator `G' = [M' D']` , has 50% of 0 entries so it is stored as a sparse matrix for efficiency

### DIFF
--- a/src/Sets/Zonotope/ZonotopeMD/ZonotopeMD.jl
+++ b/src/Sets/Zonotope/ZonotopeMD/ZonotopeMD.jl
@@ -1,0 +1,97 @@
+"""
+    struct ZonotopeMD{N, VN<:AbstractVector{N}, MN<:AbstractMatrix{N}, DN<:AbstractVector{N}} <: AbstractZonotope{N}
+
+Type that represents a zonotope of order `k` in normal form.
+
+### Fields
+
+- `center::VN` — the center of the zonotope
+- `M::MN` — matrix of general (non-axis-aligned) generators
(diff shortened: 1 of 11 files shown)


src/Sets/Zonotope/ZonotopeMD/ZonotopeMD.jl
```diff
@@ -0,0 +1,97 @@
...
+or by passing `M` and a vector `d` separately.
+
+### Examples
+
+Constructing a zonotope in normal form from a center, general generator matrix `M`, and diagonal vector `d`:
+
+```jldoctest zonotopemd_label
+julia> c = [0.0, 0.0];
+julia> M = [1.0 2.0; 3.0 1.0];
+julia> d = [0.1, 0.2];
```
schillic, line 34: It seems that there must be empty lines after each command. At least that is how it is done [here](https://documenter.juliadocs.org/stable/man/doctests/#REPL-Examples).

schillic (approved):
I have a few suggestions, but I will rather merge and apply the changes separately.

PR #3847, opened 2025-05-27
add sampler for sparse polynomial zonotopes

alecarraro:
I added a new specialization of the `_add_generators` function to sample points on a sparse polynomial zonotopes


schillic (approved):
LGTM!

PR #3910, opened 2025-07-06
Better `isequivalent` for `AbstractSingleton` and `AbstractHyperrectangle`

schillic:


src/ConcreteOperations/isequivalent.jl
```diff
@@ -44,6 +44,17 @@ end
     return center(Z) == element(S) && iszero(genmat(Z))
 end
 
+@validate_commutative function isequivalent(S::AbstractSingleton, H::AbstractHyperrectangle)
+    @assert dim(S) == dim(H) "incompatible dimensions"
+
+    @inbounds for i in 1:dim(S)
+        if !(_isapprox(center(H, i), element(S, i)) && isapproxzero(radius_hyperrectangle(H, i)))
```
alecarraro, line 51: Since `isapproxzero` is not defined for `AbstractArray`s, I see why looping over each coordinate.
However I think it would be more elegant to add a dispatch for `isapproxzero` in `ReachabilityBase.Comparison`, so this could be written in one line.

The new dispatch could be used in a lot of functions in `LazySets`. To cite [one](https://github.com/JuliaReach/LazySets.jl/blob/4559c34648677d58de21858d8e136a2bfc50d512/src/Interfaces/AbstractSparsePolynomialZonotope.jl#L120C9-L120C30). This would also have way less overhead than calling `all(isapprox, gi)`

alecarraro: Besides this, for which we could open an issue later, it looks good to me.

schillic: You can always do `all(isapproxzero, x)` or `any(isapproxzero, x)` or `isapproxzero.(x)`.

In this case, the loop is on purpose. Normally, the answer is `false`, and you want to terminate as fast as possible. You are right that one could just write
```julia
_isapprox(center(H), element(S)) && all(isapproxzero, radius_hyperrectangle(H))
```
And for `Singleton` and `Hyperrectangle`, this would be equally fast I presume. But some subtypes of `AbstractSingleton` and `AbstractHyperrectangle` create new vectors for `center`/`radius_hyperrectangle`/`element`.

alecarraro (approved):

PR #3925, opened 2025-07-08
Add overapproximation of the l1 norm of a zonotope

alecarraro:
I added a function to efficiently compute an upper bound over the l1 norm of a zonotope

### DIFF
--- a/src/Approximations/overapproximate_norm.jl
+++ b/src/Approximations/overapproximate_norm.jl
@@ -1,0 +1,54 @@
+"""
+	overapproximate_norm(Z::AbstractZonotope, p::Real=1)
+
+Compute an upper bound on the ``\\ell_p`` norm over a zonotope `Z`.
+
+### Input
+- `Z::AbstractZonotope`-- A zonotope or subtype thereof.
+- `p::Real`-- The `ℓ_p` norm to overapproximate.
+
+### Output
+- `Float64`: An upper bound on ``\\max_{z ∈ Z} \\|z\\|_p``.
+
+"""
+function overapproximate_norm(Z::AbstractZonotope, p::Real = 1)
+	if p == 1
+		return _overapproximate_l1_norm(Z)
+	else
+		error("an overapproximation of the norm for this value of p=$p is not implemented")
+	end
+end
+
+"""
+	_overapproximate_l1_norm(Z::AbstractZonotope)
+
+Compute an upper bound on the ``\\ell_1`` norm over a zonotope `Z`.
+
+### Notes
+
+The problem ``\\max_{z ∈ Z} \\|z\\|_1`` is NP-hard in general. This function computes a convex relaxation
+using a formulation described in [Jordan2021](@citet), which fits a tight upper envelope to the 
+absolute value operator. For a zonotope `Z ∈ \\mathbb{R}^d` with `n` generators, this function
+has time complexity ``\\mathcal{O}(n ⋅ d)`` compared to ``\\mathcal{O}(2ⁿ ⋅ d)`` for the exact function.
+"""
+function _overapproximate_l1_norm(Z::AbstractZonotope)
+	H = box_approximation(Z)
+	ℓ = low(H)
+	u = high(H)
+
+	c = center(Z)
+	G = generators(Z)
+
+	S⁺ = ℓ .> 0
+	S⁻ = u .< 0
+	S = .!(S⁺ .| S⁻)
+
+	w = zeros(length(c))
+	w[S⁺] .= 1.0
+	w[S⁻] .= -1.0
+	@. w[S] = (u[S] + ℓ[S]) / (u[S] - ℓ[S])
+
+	const_term = sum(-2 .* u[S] .* ℓ[S] ./ (u[S] .- ℓ[S]))
+	linear_max = dot(w, c) + sum(abs.(G' * w))
+	return linear_max + const_term
+end
--- a/src/Approximations/overapproximate_norm.jl
+++ b/src/Approximations/overapproximate_norm.jl
@@ -27,10 +27,20 @@ Compute an upper bound on the ``\\ell_1`` norm over a zonotope `Z`.
 
 ### Notes
 
-The problem ``\\max_{z ∈ Z} \\|z\\|_1`` is NP-hard in general. This function computes a convex relaxation
-using a formulation described in [Jordan2021; Theorem 5](@citet), which fits a tight upper envelope to the 
-absolute value operator. For a zonotope ``Z ∈ ℝ^d`` with `n` generators, this function
-has time complexity ``\\mathcal{O}(n ⋅ d)`` compared to ``\\mathcal{O}(2ⁿ ⋅ d)`` for the exact function.
+The problem ``\\max_{z ∈ Z} \\|z\\|_1`` is NP-hard in general. This function computes a convex relaxation 
+by combining the MILP formulation described in [Jordan2021; Theorem 5](@citet) with the convex-hull construction
+in their supplement section §C.3.
+
+### Algorithm 
+
+We replace each coordinate's absolute value
+``
+|z_i|\\quad z_i ∈ [ℓ_i, u_i]
+``
+by its convex-hull secant upper envelope: the line through
+the two points ``((ℓ_i,|ℓ_i|)`` and ``((u_i,|u_i|)``.  This yields a
+linear-programming relaxation of complexity \\(O(n·d)\\), where `n` is the
+number of generators and `d` the ambient dimension.
 """
 function _overapproximate_l1_norm(Z::AbstractZonotope{N}) where {N}
     lb = low(Z)
--- a/src/Approximations/overapproximate_norm.jl
+++ b/src/Approximations/overapproximate_norm.jl
@@ -49,6 +49,7 @@ function _overapproximate_l1_norm(Z::AbstractZonotope{N}) where {N}
     @. w[S] = (ub[S] + lb[S]) / (ub[S] - lb[S])
 
     const_term = -2 * sum(ub[S] .* lb[S] ./ (ub[S] .- lb[S]))
+    # max_{x ∈ Z} wᵀx
     linear_max = dot(w, c) + sum(abs.(At_mul_B(G, w)))
     return linear_max + const_term
 end
--- a/src/Approximations/overapproximate_norm.jl
+++ b/src/Approximations/overapproximate_norm.jl
@@ -1,54 +1,54 @@
 """
-	overapproximate_norm(Z::AbstractZonotope, p::Real=1)
+    overapproximate_norm(Z::AbstractZonotope, p::Real=1)
 
 Compute an upper bound on the ``\\ell_p`` norm over a zonotope `Z`.
 
 ### Input
-- `Z::AbstractZonotope`-- A zonotope or subtype thereof.
-- `p::Real`-- The `ℓ_p` norm to overapproximate.
+
+- `Z` -- A zonotopic set
+- `p` -- The ``ℓ_p`` norm to overapproximate
 
 ### Output
-- `Float64`: An upper bound on ``\\max_{z ∈ Z} \\|z\\|_p``.
 
+- An upper bound on ``\\max_{z ∈ Z} \\|z\\|_p``.
 """
-function overapproximate_norm(Z::AbstractZonotope, p::Real = 1)
-	if p == 1
-		return _overapproximate_l1_norm(Z)
-	else
-		error("an overapproximation of the norm for this value of p=$p is not implemented")
-	end
+function overapproximate_norm(Z::AbstractZonotope, p::Real=1)
+    if p == 1
+        return _overapproximate_l1_norm(Z)
+    else
+        return norm(Z, p)
+    end
 end
 
 """
-	_overapproximate_l1_norm(Z::AbstractZonotope)
+    _overapproximate_l1_norm(Z::AbstractZonotope)
 
 Compute an upper bound on the ``\\ell_1`` norm over a zonotope `Z`.
 
 ### Notes
 
 The problem ``\\max_{z ∈ Z} \\|z\\|_1`` is NP-hard in general. This function computes a convex relaxation
-using a formulation described in [Jordan2021](@citet), which fits a tight upper envelope to the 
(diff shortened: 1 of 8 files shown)


src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+
+Compute an upper bound on the ``\\ell_p`` norm over a zonotope `Z`.
+
+### Input
+- `Z::AbstractZonotope`-- A zonotope or subtype thereof.
+- `p::Real`-- The ``ℓ_p`` norm to overapproximate.
+
+### Output
+- `Float64`: An upper bound on ``\\max_{z ∈ Z} \\|z\\|_p``.
+
```
schillic:
```suggestion
```suggestion
### Input

- `Z` -- A zonotopic set
- `p` -- The ``ℓ_p`` norm to overapproximate

### Output

- An upper bound on ``\\max_{z ∈ Z} \\|z\\|_p``.
```

src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+"""
+function overapproximate_norm(Z::AbstractZonotope, p::Real = 1)
+	if p == 1
```
schillic: 4 spaces for indentation

test/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,12 @@
+using LazySets, Test
+using LazySets.Approximations: overapproximate_norm
+
+let
+	for _ in 1:10
```
schillic: 4 spaces for indentation

src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+		return _overapproximate_l1_norm(Z)
+	else
+		error("an overapproximation of the norm for this value of p=$p is not implemented")
```
schillic: Alternatively you could just call `norm`?

src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+
+The problem ``\\max_{z ∈ Z} \\|z\\|_1`` is NP-hard in general. This function computes a convex relaxation
+using a formulation described in [Jordan2021](@citet), which fits a tight upper envelope to the 
+absolute value operator. For a zonotope ``Z ∈ ℝ^d`` with `n` generators, this function
+has time complexity ``\\mathcal{O}(n ⋅ d)`` compared to ``\\mathcal{O}(2ⁿ ⋅ d)`` for the exact function.
+"""
+function _overapproximate_l1_norm(Z::AbstractZonotope)
+	H = box_approximation(Z)
+	ℓ = low(H)
+	u = high(H)
```
schillic:
```suggestion
	ℓ = low(Z)
	u = high(Z)
```

src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+
+	c = center(Z)
+	G = genmat(Z)
+
+	S⁺ = ℓ .> 0
```
schillic: I find `ℓ` a somewhat confusing variable name in the context of a norm.

src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+Compute an upper bound on the ``\\ell_1`` norm over a zonotope `Z`.
+
+### Notes
+
+The problem ``\\max_{z ∈ Z} \\|z\\|_1`` is NP-hard in general. This function computes a convex relaxation
+using a formulation described in [Jordan2021](@citet), which fits a tight upper envelope to the 
+absolute value operator. For a zonotope ``Z ∈ ℝ^d`` with `n` generators, this function
+has time complexity ``\\mathcal{O}(n ⋅ d)`` compared to ``\\mathcal{O}(2ⁿ ⋅ d)`` for the exact function.
+"""
+function _overapproximate_l1_norm(Z::AbstractZonotope)
```
schillic:
```suggestion
function _overapproximate_l1_norm(Z::AbstractZonotope{N}) where {N}
```

src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+	u = high(H)
+
+	c = center(Z)
+	G = genmat(Z)
+
+	S⁺ = ℓ .> 0
+	S⁻ = u .< 0
+	S = .!(S⁺ .| S⁻)
+
+	w = zeros(length(c))
```
schillic:
```suggestion
	w = zeros(N, length(c))
```

src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+	w[S⁺] .= 1.0
+	w[S⁻] .= -1.0
+	@. w[S] = (u[S] + ℓ[S]) / (u[S] - ℓ[S])
+
+	const_term = sum(-2 .* u[S] .* ℓ[S] ./ (u[S] .- ℓ[S]))
```
schillic:
```suggestion
	const_term = -2 * sum(u[S] .* ℓ[S] ./ (u[S] .- ℓ[S]))
```

test/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,12 @@
...
+		Z = rand(Zonotope; dim = 10, num_generators = 10)
+		@test overapproximate_norm(Z, 1) ≥ norm(Z, 1)
+	end
+
+	Z = Zonotope([1., -1.], zeros(Float64, 2, 2))
+	@test norm(Z, 1) == 2
```
schillic:
```suggestion
```
No point in testing `norm` here I think.

src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+	linear_max = dot(w, c) + sum(abs.(G' * w))
```
schillic:
```suggestion
	linear_max = dot(w, c) + sum(abs.(At_mul_B(G, w)))
```
This is a helper function that can sometimes be faster.

src/Approximations/overapproximate_norm.jl
```diff
@@ -3,13 +3,15 @@
 
 Compute an upper bound on the ``\\ell_p`` norm over a zonotope `Z`.
 
+```suggestion
```
schillic:
```suggestion
```
Oops, my bad.

test/Approximations/overapproximate_norm.jl
alecarraro: It was supposed to be `overapproximate_norm`, I fixed it, thanks.

src/Approximations/overapproximate_norm.jl
alecarraro: I would keep the exact norm and the approximation separate , but maybe i could add in the error message to try using `norm` instead?

schillic: But the exact norm is a valid approximation. Why is an error better?

test/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,12 @@
+using LazySets, Test
+using LazySets.Approximations: overapproximate_norm
+
+let
+	for _ in 1:10
+		Z = rand(Zonotope; dim = 10, num_generators = 10)
+		@test overapproximate_norm(Z, 1) ≥ norm(Z, 1)
```
schillic: The tests take 8.5 seconds. Maybe reduce to fewer generators.

test/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,12 @@
...
+	end
+
+	Z = Zonotope([1., -1.], zeros(Float64, 2, 2))
```
schillic:
```suggestion
	Z = Zonotope([1., -1.], zeros(Float64, 2, 0))
```
Unless it was intended to have two zero generators...

alecarraro: When I wrote the test I was thinking about 2, but I guess there is no difference

src/Approximations/overapproximate_norm.jl
alecarraro: ok, you are right

src/Approximations/overapproximate_norm.jl
schillic: This is still relevant.

src/Approximations/overapproximate_norm.jl
schillic: This is still relevant.

alecarraro: I cannot fix it. everytime i change it to 4 and commit, it goes back to 8

schillic: That sounds wrong. Committing is a Git operation and should not modify files. We can have a look in the next dev call.

src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+
+"""
+	_overapproximate_l1_norm(Z::AbstractZonotope)
+
+Compute an upper bound on the ``\\ell_1`` norm over a zonotope `Z`.
+
+### Notes
+
+The problem ``\\max_{z ∈ Z} \\|z\\|_1`` is NP-hard in general. This function computes a convex relaxation
+using a formulation described in [Jordan2021](@citet), which fits a tight upper envelope to the
```
schillic:
```suggestion
using a formulation described in [Jordan2021; Theorem 5](@citet), which fits a tight upper envelope to the 
```

schillic (approved):
I still need to check the algorithm, but for now it LGTM.

src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+    c = center(Z)
+    G = genmat(Z)
+
+    S⁺ = lb .> 0
+    S⁻ = ub .< 0
+    S = .!(S⁺ .| S⁻)
+
+    w = zeros(N, length(c))
+    w[S⁺] .= 1.0
+    w[S⁻] .= -1.0
```
schillic:
```suggestion
    w[S⁺] .= N(1)
    w[S⁻] .= N(-1)
```

src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+    @. w[S] = (ub[S] + lb[S]) / (ub[S] - lb[S])
+
+    const_term = -2 * sum(ub[S] .* lb[S] ./ (ub[S] .- lb[S]))
+    linear_max = dot(w, c) + sum(abs.(At_mul_B(G, w)))
```
schillic, line 63:
```suggestion
    # max_{x ∈ Z} wᵀx
    linear_max = dot(w, c) + sum(abs.(At_mul_B(G, w)))
```

test/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,12 @@
...
+
+let
+    for _ in 1:5
+        Z = rand(Zonotope; dim=8, num_generators=5)
+        @test overapproximate_norm(Z, 1) ≥ norm(Z, 1)
+    end
+
+    Z = Zonotope([1.0, -1.0], zeros(Float64, 2, 0))
+    @test overapproximate_norm(Z, 1) == 2
+end
```
schillic:
```suggestion
for N in [Float64, Float32, Rational{Int}]
    for _ in 1:5
        Z = rand(Zonotope; N=N, dim=8, num_generators=5)
        @test overapproximate_norm(Z, 1) ≥ norm(Z, 1)
    end

    Z = Zonotope(N[1, -1], zeros(N, 2, 0))
    @test overapproximate_norm(Z, 1) == 2
end
```

src/Approximations/overapproximate_norm.jl
```diff
@@ -0,0 +1,54 @@
...
+    G = genmat(Z)
+
+    S⁺ = lb .> 0
+    S⁻ = ub .< 0
+    S = .!(S⁺ .| S⁻)
+
+    w = zeros(N, length(c))
+    w[S⁺] .= 1.0
+    w[S⁻] .= -1.0
+    @. w[S] = (ub[S] + lb[S]) / (ub[S] - lb[S])
```
schillic, line 59: I do not understand this. I would expect:
```suggestion
    @. w[S] = (2 * ub[S] - S) / (ub[S] - lb[S])
```

alecarraro: You are right, based on the paper , if you relax $$a$$ to be continuous in $$[0,1]$$, you compute the optimal $$a^\star$$ to maximise $$t$$ and then substitute it in either equation 27 or [28,](https://proceedings.mlr.press/v139/jordan21a/jordan21a-supp.pdf)  you obtain the linear constraint 
$$t \leq \frac{2u_i -1}{u_i-l_i} z - 2  \frac{2u_i l_i}{u_i -l_i}$$
However when I tested this methods, the results were quite bad, worse than simply using an approximation of the type 
 $$\max |z|_1 \leq | c+ \sum_i^n g_i|$$

The problem is that the resulting linear constraint is a line which intersects $$|z|$$ at one point and then diverges at the end of the interval $$[l, u]$$
If instead we compute the linear constraint $$t \leq mz + q$$ such that it is exact at $$l$$ and $$u$$ (the hull secant), the relaxation is tighter and this has slope $$m =  \frac{u-l}{u+l}$$

To show an example in the picture below I plotted the 2 resulting constraint, for the function $$|x|$$ in the region $$x \in [-2, 5]$$
comparison_relax

So actually in the code I used this modified version and the results were very good

test/Approximations/overapproximate_norm.jl
alecarraro: This test fail because `rand(Zonotope)` does not support rational  numbers AFAIK.

schillic: True, what about `for N in [Float64, Float32]`?

src/Approximations/overapproximate_norm.jl
schillic: Okay, then please add this to the documentation.

PR #3965, opened 2025-07-22
Add overapproximation methods for matrix zonotope maps

alecarraro:
In this PR, I included the following changes necessary to introduce the overapproximation methods for the `LinearMap` and `ExponentialMap`:

* Modified the definition and constructors of `LinearMap` to accept a union of `AbstractMatrix` and `MatrixZonotope`. Similarly, `ExponentialMap` now accepts a union of `AbstractMatrix` and `AbstractMatrixZonotope` (see below).

* Created an abstract type `AbstractMatrixZonotope` to ensure that the exponential map can accept both a single `MatrixZonotope` and products of them. The minimal interface so far includes only the `size` function which is necessary in all the subtypes.

* Implemented overapproximation methods for `LinearMap` for both `SparsePolynomialZonotope`s and simple `Zonotope`s.

* Implemented overapproximation methods for `ExponentialMap` for both `SparsePolynomialZonotope`s and simple `Zonotope`s, supporting both single and products of matrix zonotopes. A single function handles arbitrary numbers of `MatrixZonotope`s thanks to the logic being delegated to two helper functions.

* Introduced the `_taylor_expmap` function, which dispatches based on whether the input is a single `MatrixZonotope` or a product of matrix zonotopes wrapped in the `MatrixZonotopeProduct` type.

* Added the `MatrixZonotopeProduct` type, which stores an arbitrary number of `MatrixZonotope`s and includes methods to simplify the product, particularly in the case where a matrix zonotope has no generators. In such cases, the zonotope acts like a constant matrix and is "absorbed" by adjacent matrix zonotopes.

This extension to arbitrary-length products is motivated by multiple reasons:
First, it generalizes the implementation (which can later be used in the reachability algorithm) without increasing code complexity. No additional code is required when extending the Taylor expansion from two to more `MatrixZonotope` factors.

Second, it resolves edge cases that previously caused issues, particularly when a `MatrixZonotope` with no generators appeared in the middle of a product. Such cases previously led to segmentation faults in the overapproximation method. Now, these "degenerate" zonotopes are automatically handled by the `remove_redundant_factors` method. Since zonotopes with no generators behave like linear maps acting on the neighbouring factors, they are absorbed by adjacent factors in the product.

When the exponential is computed on a single `MatrixZonotope` with no generators, this special case is handled explicitl, because it corresponds to a constant matrix, its exponential is directly computable.

The `overapproximation_norm` function has also been extended to support `MatrixZonotopeProduct`s, which avoids code duplication and ensures consistency across overapproximation logic.

### DIFF
--- a/src/MatrixSets/scale.jl
+++ b/src/MatrixSets/scale.jl
@@ -9,10 +9,3 @@ function scale!(α::Real, MZ::MatrixZonotope)
     end
     return MZ
 end
-
-"""
-    *(a::Real, B::MatrixZonotope)
-
-Alias to scale a matrix zonotope.
-"""
-@commutative Base.:*(a::Real, B::MatrixZonotope) = scale(a, B)
--- a/docs/src/lib/approximations/overapproximate_expmap.md
+++ b/docs/src/lib/approximations/overapproximate_expmap.md
@@ -10,5 +10,6 @@ CurrentModule = LazySets.Approximations
 # Overapproximation
 
 ```@docs
-_taylor_expmap
+taylor_expmap_truncation
+taylor_expmap_remainder
 ```
--- a/src/Approximations/Approximations.jl
+++ b/src/Approximations/Approximations.jl
@@ -21,7 +21,9 @@ export approximate,
        SphericalDirections,
        CustomDirections,
        isbounding,
-       overapproximate_norm
+       overapproximate_norm,
+       taylor_expmap_truncation,
+       taylor_expmap_remainder
 
 using ..LazySets, ReachabilityBase.Arrays, Requires, LinearAlgebra, SparseArrays
 import IntervalArithmetic as IA
--- a/src/Approximations/overapproximate_expmap.jl
+++ b/src/Approximations/overapproximate_expmap.jl
@@ -8,8 +8,8 @@ function _compute_inner_powers(MZ::MatrixZonotope, P::S,
     
     @inbounds for i in 1:k
         invfact /= i
-        term = overapproximate(invfact * MZ * powers[i], T)
-        powers[i + 1] = remove_redundant_generators(term)
+        term = overapproximate(scale(invfact, MZ) * powers[i], T)
+        powers[i + 1] = term
     end
     return powers
 end
@@ -23,7 +23,7 @@ function _compute_outer_powers(MZ::MatrixZonotope, in_powers::Vector{S},
     @inbounds for i in 2:(k + 1)
         term = in_powers[i]
         for _ in 1:i
-            term = remove_redundant_generators(overapproximate(MZ * term, S))
+            term = overapproximate(MZ * term, S)
         end
         out_powers[i] = term
     end
@@ -31,7 +31,7 @@ function _compute_outer_powers(MZ::MatrixZonotope, in_powers::Vector{S},
 end
 
 """
-    _taylor_expmap(MZ::MatrixZonotope, P::S, k::Int) where S
+    taylor_expmap_truncation(MZ::MatrixZonotope, P::S, k::Int) where S
 
 Compute the k-th order truncated Taylor expansion of the exponential map of a matrix zonotope
 
@@ -53,7 +53,7 @@ This function computes the approximation:
 \\displaystyle\\boxplus_{i=0}^k \\frac{\\mathcal{A}^i }{i!} X
 ```
 """
-function _taylor_expmap(MZ::MatrixZonotope, P::S,
+function taylor_expmap_truncation(MZ::MatrixZonotope, P::S,
                         k::Int) where {S<:Union{SparsePolynomialZonotope,AbstractZonotope}}
     if isempty(generators(MZ))
         res = linear_map(exp(center(MZ)), P)
@@ -62,11 +62,11 @@ function _taylor_expmap(MZ::MatrixZonotope, P::S,
         res = reduce(exact_sum, inner)
     end
 
-    return remove_redundant_generators(res)
+    return res
 end
 
 """
-    _taylor_expmap(MZP::MatrixZonotopeProduct, P::S, k::Int) where S
+    taylor_expmap_truncation(MZP::MatrixZonotopeProduct, P::S, k::Int) where S
 
 Compute the k-th order truncated Taylor expansion of the exponential map of a matrix zonotope product
 
@@ -88,7 +88,7 @@ This function computes the approximation:
 \\displaystyle\\boxplus_{i=0}^k \\frac{\\mathcal{A}^i \\mathcal{B}^i}{i!} X
 ```
 """
-function _taylor_expmap(MZP::MatrixZonotopeProduct, P::S,
+function taylor_expmap_truncation(MZP::MatrixZonotopeProduct, P::S,
                         k::Int) where {S<:Union{SparsePolynomialZonotope,AbstractZonotope}}
     # inner powers on the last factor
     last_factor = factors(MZP)[end]
@@ -100,20 +100,49 @@ function _taylor_expmap(MZP::MatrixZonotopeProduct, P::S,
     end
 
     res = reduce(exact_sum, terms)
-    return remove_redundant_generators(res)
+    return res
 end
 
 function load_intervalmatrices_overapproximation_expmap()
     return quote
         using .IntervalMatrices: IntervalMatrix
 
+        """
+            taylor_expmap_remainder(P::S, matnorm::Real, k::Int) where S
+
+        Compute the Lagrange remainder term of the k-th order truncated Taylor expansion 
+        of the exponential map of a matrix zonotope applied to a zonotopic set.
+
+        ### Input
+
+        - `P` -- a zonotopic set
+        - `matnorm` -- an upper bound on the norm of the matrix zonotope
+        - `k` -- the order of the Taylor expansion
+
+        ### Output
+
+        A zonotope over-approximating the remainder term of the Taylor expansion.
+        """
+        function taylor_expmap_remainder(P::S, matnorm::Real, k::Int) where{S<:Union{SparsePolynomialZonotope,
+                                                                AbstractZonotope}}
+            n = dim(P)
+            N = eltype(P)
+
+            ϵ = matnorm / (k + 2)
+            ε = IntervalMatrix(fill(IA.interval(N(-1.0), N(1.0)), n, n))
+            ε *= matnorm^(k + 1) / (factorial(k + 1) * (1 - ϵ))
+
+            Zp = overapproximate(P, Zonotope)
+            rhs = overapproximate(ε * Zp, Zonotope)
+            return rhs
+        end
+
         function _overapproximate_emz(em::ExponentialMap{N,S,MAT},
                                       k::Int=2) where {N,S<:Union{SparsePolynomialZonotope,
                                                                 AbstractZonotope},
                                                        MAT<:AbstractMatrixZonotope{N}}
             MZP = matrix(em).M
-            P = set(em) #SPZ or Zonotope
-            n = size(MZP, 2)
+            P = set(em)
 
             matnorm = N(overapproximate_norm(MZP, Inf))
             ϵ = matnorm / (k + 2)
@@ -121,13 +150,9 @@ function load_intervalmatrices_overapproximation_expmap()
                 @warn "k should be chosen such that ϵ<1 " ϵ
             end
 
-            σ = _taylor_expmap(MZP, P, k)
-            ε = IntervalMatrix(fill(IA.interval(N(-1), N(1)), n, n))
-            ε *= matnorm^(k + 1) / (factorial(k + 1) * (1 - ϵ))
-
-            Zp = overapproximate(P, Zonotope)
-            rhs = overapproximate(ε * Zp, Zonotope)
-            P_approx = minkowski_sum(σ, rhs)
+            tayexp = _taylor_expmap(MZP, P, k)
+            lagrem = lagrange_remainder(P, matnorm, k)
+            P_approx = minkowski_sum(tayexp, lagrem)
 
             return remove_redundant_generators(P_approx)
         end


src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,267 @@
...
+
+    m, n = size(MZ)
+    w = ngens(MZ)
+    h = ngens_dep(P)
+    q = ngens_indep(P)
+
+    if n != dim(P)
+        throw(DimensionMismatch("incompatible dimensions:" *
+                                "size(MZ) = $(size(MZ)), dim(P) = $q"))
+    end
```
schillic: This is already checked in the `LinearMap` constructor, so you can assume that.

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,267 @@
...
+        G[:, h + i] = A * center(P)
+        G[:, (h + w + (i - 1) * h + 1):(h + w + i * h)] = A * genmat_dep(P)
+        Gi[:, (q * i + 1):(q * (i + 1))] = A * genmat_indep(P)
+    end
+
+    # compute exponent
+    Imat = Matrix{Int}(I, w, w)
+    Ê₁, Ê₂, idx = merge_id(indexvector(P), indexvector(MZ), expmat(P), Imat)
+    pₖ = size(Ê₁, 1)
+    E = Matrix{eltype(Ê₁)}(undef, pₖ, h + w + h * w)
```
schillic: Should `eltype(Ê₁)` not simply be `Int`?

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,267 @@
...
+    E[:, 1:h] = Ê₁
+    E[:, (h + 1):(h + w)] = Ê₂
+    @inbounds for l in 1:w
+        cstart = (h + w) + (l - 1) * h + 1
+        cend = (h + w) + l * h
+        E[:, cstart:cend] = Ê₂[:, l] * ones(1, h) .+ Ê₁
```
schillic: `ones(1, h)` is constant and could be defined outside the loop.

Actually, the following avoids the matrix allocation:
```julia
E[:, cstart:cend] = Ê₁
for j in cstart:cend
    E[:, j] += Ê₂[:, l]
end
```

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,267 @@
...
+### Input
+
+- `lm` -- a linear map of a sparse polynomial zonotope through a matrix zonotope
+
+### Output
+
+A sparse polynomial zonotope overapproximating the linear map.
+
+"""
+function overapproximate(lm::LinearMap{N,S,NM,MAT}) where {N,S<:SparsePolynomialZonotope{N},NM,
```
schillic: The `overapproximate` methods should receive the target type as second argument.

schillic (commented):
This is a long PR, so I give feedback in chunks. There are also several problems in the build process. Let me know if you need help with them.

src/LazyOperations/ExponentialMap.jl
```diff
@@ -252,9 +255,18 @@ julia> E * EmptySet(100)
...
+struct ExponentialMap{N,S<:LazySet{N},NM,
+                      MAT<:Union{SparseMatrixExp{NM},AbstractMatrixZonotope{NM}}} <:
+       AbstractAffineMap{N,S}
+    expmat::MAT
     X::S
+    function ExponentialMap(expmat::MAT,
+                            X::S) where {N,S<:LazySet{N},NM,
+                                         MAT<:Union{SparseMatrixExp{NM},AbstractMatrixZonotope{NM}}}
+        @assert dim(X) == size(expmat, 2) "an exponential map of size " *
+                                          "$(size(expmat)) cannot be applied to a set of dimension $(dim(Z))"
```
schillic:
```suggestion
                                          "$(size(expmat)) cannot be applied to a set of dimension $(dim(X))"
```

src/LazyOperations/LinearMap.jl
```diff
@@ -115,22 +115,22 @@ isconvextype(::Type{<:LinearMap{N,S}}) where {N,S} = isconvextype(S)
...
+    *(M::Union{AbstractMatrix, UniformScaling, AbstractVector, Real, MatrixZonotope},
       X::LazySet)
 ```
 
 Alias to create a `LinearMap` object.
 
 ### Input
 
-- `M` -- linear map
+- `M` -- linear map or matrix zonotope
```
schillic:
```suggestion
- `M` -- matrix or matrix zonotope
```

src/LazyOperations/ExponentialMap.jl
```diff
@@ -204,13 +204,16 @@ function get_rows(spmexp::SparseMatrixExp{N}, I::AbstractArray{Int};
...
+                      MAT<:Union{SparseMatrixExp{NM},AbstractMatrixZonotope{NM}}} <:
+       AbstractAffineMap{N,S}
 
 Type that represents the action of an exponential map on a set.
 
 ### Fields
 
-- `spmexp` -- sparse matrix exponential
+- `expmat` -- matrix exponential, either a concrete matrix exponential
+              or a matrix set represented by an `AbstractMatrixZonotope
```
schillic: I think `M` would be a better name because this is *not* the matrix exponential.
```suggestion
- `M` -- matrix or matrix zonotope
```

src/LazyOperations/LinearMap.jl
```diff
@@ -96,13 +96,13 @@ EmptySet{Int64}(3)
 ```
 """
 struct LinearMap{N,S<:LazySet{N},NM,
-                 MAT<:AbstractMatrix{NM}} <: AbstractAffineMap{N,S}
+                 MAT<:Union{AbstractMatrix{NM},MatrixZonotope{NM}}} <: AbstractAffineMap{N,S}
```
schillic: Is it intentional that `LinearMap` cannot receive a general `AbstractMatrixZonotope`?

src/MatrixSets/AbstractMatrixZonotope.jl
```diff
@@ -0,0 +1,25 @@
+"""
+    AbstractMatrixZonotope{N}
+
+Abstract supertype for all matrix zonotope representations, including standard (`MatrixZonotope`), products (`MatrixZonotopeProduct`), polynomial matrix zonotopes, and others.
```
schillic:
```suggestion
Abstract supertype for all matrix zonotope representations.
```

src/MatrixSets/scale.jl
```diff
@@ -9,3 +9,10 @@ function scale!(α::Real, MZ::MatrixZonotope)
     end
     return MZ
 end
+
+"""
+    *(A::Real, B::MatrixZonotope)
+
+Alias to scale a matrix zonotope.
+"""
+@commutative Base.:*(A::Real, B::MatrixZonotope) = scale(A, B)
```
schillic:
```suggestion
@commutative Base.:*(a::Real, B::MatrixZonotope) = scale(a, B)
```

src/LazyOperations/LinearMap.jl
alecarraro: It was. But good point, you made me realize I should generalize it. Since in Julia multiplication is left associative the product $$A\cdot B\cdot PZ$$ would give an error as it would first create a `MatrixZonotopeProduct`. I will add a dispatch on the `overapproximate` method for the MZP to perform reduction over the reversed factors

schillic: That can also be something for later.

src/Approximations/overapproximate_spz.jl
alecarraro: Thanks for the suggestion. I tried your version and some variants including using `repeat` to broadcast $$E_1$$ before the loop, but they all actually use more than double the allocations. The outer product must be trigger some very optimized BLAS call. Because
```julia
@inbounds for l in 1:w
        cstart = (h + w) + (l - 1) * h + 1
        cend = (h + w) + l * h
        E[:, cstart:cend]= Ê₁
        col = @view Ê₂[:, l]
        for j in cstart:cend
            E[:, j] .+= col 
        end
    end
```
takes `13.132 ms (2054 allocations: 49.43 MiB)`, while what I had before + preallocating the vector of ones as you suggested and adding `@views`
```julia
ones_h = ones(Int, 1, h)
  @inbounds for l in 1:w
      cstart = (h + w) + (l - 1) * h + 1
      cend = (h + w) + l * h
      col = @view Ê₂[:, l]
      E[:, cstart:cend] = col * ones_h .+ Ê₁
  end
   ```
    takes 12.499 ms (1003 allocations: 49.41 MiB)

Comment (schillic) [reply src/Approximations/overapproximate_spz.jl]: Interesting. Out of curiosity, could you also try:
```julia
    ones_h = ones(Int, 1, h)
    @inbounds for l in 1:w
        cstart = (h + w) + (l - 1) * h + 1
        cend = (h + w) + l * h
        E[:, cstart:cend] = Ê₁
        col = @view Ê₂[:, l]
        E[:, cstart:cend] += col * ones_h
    end
```

docs/src/lib/approximations/overapproximate_spz.md
```diff
@@ -10,10 +10,17 @@ CurrentModule = LazySets.Approximations
...
+overapproximate(lm::LinearMap{N,S,NM,MAT},
+                         ::Type{Zonotope}) where {N,S<:AbstractZonotope{N},NM,
+                                                  MAT<:MatrixZonotope{NM}}
+overapproximate(lm::LinearMap{N,S,NM,MAT},
+                    ::Type{U}) where {N, S<:AbstractZonotope{N}, NM,
+                                      MAT<:MatrixZonotopeProduct{NM},
+                                      U<:Union{Zonotope, SparsePolynomialZonotope}}
 _taylor_expmap
-overapproximate(::ExponentialMap, ::Int)
+verapproximate(em::ExponentialMap{N,S,NM,MAT}, ::Type{U}, k::Int)
```
schillic:
```suggestion
overapproximate(::ExponentialMap{N,S,NM,MAT}, ::Type{U}, k::Int)
```
But I think you need the `where {...}`...

src/LazyOperations/LinearMap.jl
alecarraro: added

docs/src/lib/approximations/overapproximate_spz.md
schillic: It would be much easier if you do not add the `overapproximate` methods at all. There is a generic `overapproximate` catch-all already.

src/Approximations/overapproximate_spz.jl
```diff
@@ -1,6 +1,8 @@
 """
-	overapproximate(lm::LinearMap{N,S,NM,MAT}) where {N,S<:SparsePolynomialZonotope{N},NM,
-                                                           MAT<:MatrixZonotope{NM}}
+	overapproximate(lm::LinearMap{N,S,NM,MAT},
+                         ::Type{SparsePolynomialZonotope}) where {N,S<:SparsePolynomialZonotope{N},
```
schillic: The types should have a `<:` in front so that you can also pass more concretely-typed types. Like this:
```suggestion
                         ::Type{<:SparsePolynomialZonotope}) where {N,S<:SparsePolynomialZonotope{N},
```

src/Approximations/overapproximate_spz.jl
alecarraro: I included it in a new commit,  however there is an issue.
As of now,  the type of the object returned by the overapproximation of the exponential map is simply the type of the set in the `ExponentialMap` object. I can think about 3 options to make the method more consistent:
1 add a warning/error in case a user sets, as a target,  a type different from the input, 
2 keep it as it is now, so the target type is only there to match `LazySets` interface, but it is not actually used anywhere in the function
3 convert/overapproximate the ouput of the function to match the target

src/Approximations/overapproximate_spz.jl
alecarraro: In the experiments I performed earlier I was  using a random SPZ with only the ambient dimension specified, so the number of generators were picked at random. I do not remember all the dimensions from before so I generated a new one and thus the memory usage and time are a bit different (but they were and are all tested on the same instance).

- 15.206 ms (853 allocations: 58.88 MiB) for the version I commited
- 15.375 ms (1003 allocations: 59.17 MiB) for your last suggestion

src/Approximations/overapproximate_spz.jl
schillic: I think the cleanest solution is option 4: have two methods:
```julia
function overapproximate(em::ExponentialMap{N,S,NM,MAT}, T::Type{<:SparsePolynomialZonotope}, k::Int=2) where {N,S<:SparsePolynomialZonotope,NM,MAT<:AbstractMatrixZonotope}
    return _overapproximate_emz(em, k)
end

function overapproximate(em::ExponentialMap{N,S,NM,MAT}, T::Type{<:Zonotope}, k::Int=2) where {N,S<:AbstractZonotope,NM,MAT<:AbstractMatrixZonotope}
    return _overapproximate_emz(em, k)
end

function _overapproximate_emz(em::ExponentialMap{N,S,NM,MAT}, k::Int=2) where {N,S,NM,MAT<:AbstractMatrixZonotope}
    ...  # insert old code here
end

File: src/Approximations/overapproximate_spz.jl (hunk 3 above)
Comment (schillic) [reply src/Approximations/overapproximate_spz.jl]: Thanks!
Very strange... but okay, let us keep your version.

File: src/Approximations/overapproximate_spz.jl [hunk 13]
```diff
@@ -0,0 +1,304 @@
+"""
+	overapproximate(lm::LinearMap{N,S,NM,MAT},
+                         ::Type{SparsePolynomialZonotope}) where {N,S<:SparsePolynomialZonotope{N},
+                                                                  NM,
+                                                                  MAT<:MatrixZonotope{NM}}
+
+Overapproximate the linear map of a sparse polynomial zonotope through a matrix zonotope,
+following Proposition 1 of [HuangLBS2025](@citet).
```
schillic:
```suggestion
following Proposition 2 of [HuangLBS2025](@citet).
```

src/MatrixSets/linear_map.jl
```diff
@@ -0,0 +1,37 @@
+"""
+    linear_map(M::AbstractMatrix, MZ::MatrixZonotope)
+
+Applies a linear transformation to a matrix zonotope from the left.
+
+### Input
+- `M` -- a linear map / matrix
+- `MZ`` -- a matrix zonotope
```
schillic:
```suggestion
- `MZ` -- a matrix zonotope
```

src/MatrixSets/linear_map.jl
```diff
@@ -0,0 +1,37 @@
...
+
+### Output
+
+A matrix zonotope with transformed center and generators
+"""
+function linear_map(M::AbstractMatrix, MZ::MatrixZonotope)
```
schillic, line 14: Should this not just be `*`?

src/MatrixSets/linear_map.jl
```diff
@@ -0,0 +1,37 @@
...
+end
+
+"""
+    linear_map(M::AbstractMatrix, MZ::MatrixZonotope)
+
+Applies a linear transformation to a matrix zonotope from the right.
+
+### Input
+- `M` -- a linear map / matrix
+- `MZ`` -- a matrix zonotope
```
schillic:
```suggestion
- `MZ` -- a matrix zonotope
```

src/MatrixSets/scale.jl
```diff
@@ -9,3 +9,10 @@ function scale!(α::Real, MZ::MatrixZonotope)
     end
     return MZ
 end
+
+"""
+    *(A::Real, B::MatrixZonotope)
```
schillic:
```suggestion
    *(a::Real, B::MatrixZonotope)
```

src/LazyOperations/ExponentialMap.jl
alecarraro: > I think `M` would be a better name because this is _not_ the matrix exponential.

I agree that it causes confusion. Right now the `ExponentialMap` accepts either a `SparseMatrixExp` which is indeed the matrix exponential or  a matrix zonotope which is the *argument* of a matrix zonotope exponential. 
The matrix in the `SparseMatrixExp` is already named `M`, so i think it would cause some confusion to rename also that field `M`.

I think it would be better to make a `MatrixZonotopeExp` type which stores the matrix zonotope `A` in $$exp(A)$$ and is equipped with a couple of methods like `size`.
In this way we remove the ambiguity and we stay consistent with the way users defines an exponential map for both matrices and matrix zonotopes.

src/MatrixSets/AbstractMatrixZonotope.jl
```diff
@@ -0,0 +1,25 @@
...
+"""
+function Base.size(::AbstractMatrixZonotope) end
+
+"""
+    size(::AbstractMatrixZonotope, ::Int)
+
+Return the dimensions of a matrix zonotope along a specified axis.  
+
+"""
+function Base.size(::AbstractMatrixZonotope, ::Int) end
```
schillic: Can be written as one docstring. I adapted it from `Base.size`.

I even think that `size` does not require a docstring at all, so you might as well remove these methods altogether.

```suggestion
    size(::AbstractMatrixZonotope, [dim])

Return a tuple containing the dimensions of a matrix zonotope. Optionally you can specify a dimension to just get the length of that dimension.
"""
function Base.size(::AbstractMatrixZonotope) end
```

src/LazyOperations/ExponentialMap.jl
schillic: > I think it would be better to make a `MatrixZonotopeExp` type which stores the matrix zonotope `A` in $$exp(A)$$

Let us discuss this in the next call.

src/LazyOperations/ExponentialMap.jl
```diff
@@ -252,9 +255,18 @@ julia> E * EmptySet(100)
...
+       AbstractAffineMap{N,S}
+    expmat::MAT
     X::S
+    function ExponentialMap(expmat::MAT,
+                            X::S) where {N,S<:LazySet{N},NM,
+                                         MAT<:Union{SparseMatrixExp{NM},AbstractMatrixZonotope{NM}}}
+        @assert dim(X) == size(expmat, 2) "an exponential map of size " *
+                                          "$(size(expmat)) cannot be applied to a set of dimension $(dim(X))"
+        return new{N,S,NM,MAT}(expmat, X)
+    end
```
schillic:
```suggestion
    function ExponentialMap(expmat::MAT,
                            X::S) where {N,S<:LazySet{N},NM,
                                         MAT<:Union{SparseMatrixExp{NM},AbstractMatrixZonotope{NM}}}
        @assert dim(X) == size(expmat, 2) "an exponential map of size $(size(expmat)) cannot " *
                                          "be applied to a set of dimension $(dim(X))"
        @assert size(expmat, 1) == size(expmat, 2) "the matrix exponential requires a square " *
                                                   "matrix, but it has size $(size(expmat))"
        return new{N,S,NM,MAT}(expmat, X)
    end
```

src/Approximations/overapproximate_norm.jl
```diff
@@ -103,3 +103,21 @@ function overapproximate_norm(MZ::MatrixZonotope, p::Real=Inf)
...
+
+"""
+    overapproximate_norm(MZ::MatrixZonotopeProduct, [p]::Real=1)
+
+Compute an upper bound on the ``p``-norm of a product of matrix zonotopes.
+
+### Input
+
+- `MZP` -- Matrix zonotope product
+- `p` -- (optional, default: `1`) norm
```
schillic:
```suggestion
    overapproximate_norm(MZ::MatrixZonotopeProduct, [p]::Real=Inf)

Compute an upper bound on the ``p``-norm of a product of matrix zonotopes.

### Input

- `MZP` -- Matrix zonotope product
- `p` -- (optional, default: `Inf`) norm
```

src/Approximations/overapproximate_spz.jl
schillic: Proposition 2 relies on Proposition 1. It would be better to implement Proposition 1 first instead of duplicating the code.

src/MatrixSets/MatrixZonotopeProduct.jl
```diff
@@ -0,0 +1,126 @@
...
+
+- `factors` -- a vector of matrix zonotopes `[A1, A2, ..., An]`
+
+### Notes
+
+Mathematically, this represents the set:
+
+```math
+\\mathcal{C} = \\{ A_1 A_2 \\dots A_n ~|~ A_i \\in \\mathcal{A}_i \\}
+
```
schillic:
````suggestion
```
````

File: src/MatrixSets/MatrixZonotopeProduct.jl [hunk 22]
```diff
@@ -0,0 +1,126 @@
...
+        return new{N,T}(factors)
+    end
+end
+
+MatrixZonotopeProduct(ms::MatrixZonotope{N,S}...) where {N,S} = MatrixZonotopeProduct(collect(ms))
+
+"""
+    *(A::MatrixZonotope{N,S}, B::MatrixZonotope{N,S}) where {N,S}
+
+Appends matrix zonotope B to an existing matrix zonotope product P.
```
schillic:
```suggestion
    *(A::MatrixZonotope{N,S}, B::MatrixZonotope{N,S}) where {N,S}

Combines two matrix zonotopes to a matrix zonotope product.
```

schillic: Again I do not think that `*` requires a docstring.

src/MatrixSets/MatrixZonotopeProduct.jl
```diff
@@ -0,0 +1,126 @@
...
+Returns a simplified `MatrixZonotopeProduct` where generator-free factors are absorbed
+into adjacent factors via linear maps.
+
+### Input
+
+- `MZP` -- matrix zonotope product.
+
+### Output
+
+A matrix zonotope with redundant constant factors removed.
```
schillic:
```suggestion
A matrix zonotope product with redundant constant factors removed.
```

src/MatrixSets/linear_map.jl
alecarraro: Im not sure I understand what you mean. This is not the same as the `*` as the alias for `LinearMap` for `MatrixZonotope` with a (SP)Zonotope. This is conceptually the same as a linear map on a `LazySet`, but on a matrix zonotope, not a matrix zonotope applied to a `LazySet`.

docs/src/lib/approximations/overapproximate_spz.md
alecarraro: I do not understand how to do it. I tried so many different things, it never renders the way I want it. 
My idea was to show the specific `overapproximate` dispatch for the matrix zonotope maps, would the generic method show them all?

src/MatrixSets/linear_map.jl
schillic: It is multiplication of a matrix with a matrix set. This is basically a matrix multiplication, just generalized to sets.

docs/src/lib/approximations/overapproximate_spz.md
schillic: Yes, every method called `overapproximate` should be shown there. So you could just remove all these methods I think.

src/MatrixSets/MatrixZonotopeProduct.jl
```diff
@@ -0,0 +1,107 @@
...
+    end
+end
+
+MatrixZonotopeProduct(ms::MatrixZonotope...) = MatrixZonotopeProduct(collect(ms))
+
+"""
+    *(A::MatrixZonotope{N,S}, B::MatrixZonotope{N,S}) where {N,S}
+
+Alias to create a `MatrixZonotopeProduct` object.
+"""
```
schillic:
```suggestion
```

docs/src/lib/matrixsets/MatrixZonotope.md
```diff
@@ -29,3 +42,35 @@ Undocumented implementations:
...
+CurrentModule = LazySets.MatrixZonotopeModule
+```
+
+# [MatrixZonotopeProduct](@id def_MatrixZonotopeProduct)
+```@docs
+MatrixZonotopeProduct
+*(::MatrixZonotope{N,S}, ::MatrixZonotope{N,S}) where {N,S}
+*(::MatrixZonotope{N,S}, ::MatrixZonotopeProduct{N,S}) where {N,S}
+*(::MatrixZonotopeProduct{N,S}, ::MatrixZonotope{N,S}) where {N,S}
+*(::MatrixZonotopeProduct{N,S}, ::MatrixZonotopeProduct{N,S}) where {N,S}
```
schillic:
```suggestion
```

src/MatrixSets/MatrixZonotopeProduct.jl
```diff
@@ -0,0 +1,107 @@
...
+Base.:*(A::MatrixZonotope, B::MatrixZonotope) = MatrixZonotopeProduct([A, B])
+Base.:*(P::MatrixZonotopeProduct, B::MatrixZonotope) = MatrixZonotopeProduct(vcat(P.factors, B))
+Base.:*(A::MatrixZonotope, P::MatrixZonotopeProduct) =MatrixZonotopeProduct(vcat(A, P.factors))
```
schillic:
```suggestion
Base.:*(A::MatrixZonotope, P::MatrixZonotopeProduct) = MatrixZonotopeProduct(vcat(A, P.factors))
```

src/MatrixSets/MatrixZonotopeProduct.jl
```diff
@@ -0,0 +1,107 @@
...
+    factors_ = factors(MZP)
+    gens = MatrixZonotope{eltype(factors_[1]), typeof(factors_[1].A0)}[]
+
+    i = 1
+    while i < length(factors_)
+        MZ = factors_[i]
+        next = factors_[i + 1]
+
+        if isempty(generators(MZ))
+            push!(gens, linear_map(center(MZ), next))
```
schillic:
```suggestion

        if isempty(generators(MZ))
            push!(gens, linear_map(center(MZ), factors_[i + 1]))
```

src/MatrixSets/MatrixZonotopeProduct.jl
```diff
@@ -0,0 +1,107 @@
...
+
+### Input
+
+- `MZP` -- matrix zonotope product.
+
+### Output
+
+A matrix zonotope product with redundant constant factors removed.
+"""
+function remove_redundant_factors(MZP::MatrixZonotopeProduct)
```
schillic, line 73: Is this really something of relevance? Essentially you have a product of a MZ with a standard matrix, and for that you already have a function.

Also, instead of this function, you could consider resolving this case at construction time of the MZP.

schillic: I also think this implementation may not work if you have two consecutive matrices.

src/MatrixSets/MatrixZonotopeProduct.jl
```diff
@@ -0,0 +1,107 @@
...
+        if isempty(generators(MZ))
+            push!(gens, linear_map(center(MZ), next))
+            i += 2  # skip next since it's been merged
+        else
+            push!(gens, MZ)
+            i += 1
+        end
+    end
+
+    # in the the last element apply linear map to the left
```
schillic:
```suggestion
    # in the last element apply linear map to the left
```

src/MatrixSets/MatrixZonotopeProduct.jl
alecarraro: > Is this really something of relevance? Essentially you have a product of a MZ with a standard matrix, and for that you already have a function.

I have a reason, but we can discuss it tomorow during the call.
> I also think this implementation may not work if you have two consecutive matrices

Ah yes, you are right, good catch!

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,358 @@
...
+    return SparsePolynomialZonotope(c, G, Gi, E, idx)
+end
+
+"""
+	function overapproximate(lm::LinearMap{N,S,NM,MAT},
+                         ::Type{Zonotope}) where {N,S<:AbstractZonotope{N},NM,
+                                                  MAT<:MatrixZonotope{NM}}
+
+Overapproximate the linear map of a zonotope through a matrix zonotope,
+following a modification of Proposition 1 of [HuangLBS2025](@citet).
```
schillic:
```suggestion
	function overapproximate(lm::LinearMap{N,S,NM,MAT},
                         ::Type{Zonotope}) where {N,S<:AbstractZonotope{N},NM,
                                                  MAT<:MatrixZonotope{NM}}

Overapproximate the linear map of a zonotope through a matrix zonotope,
following a modification of Proposition 2 of [HuangLBS2025](@citet).
```
Right?

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,358 @@
...
+
+    c = center(MZ) * center(Z)
+
+    # compute matrix of dependent generators
+    G = Matrix{T}(undef, m, h + w + h * w)
+    G[:, 1:h] = center(MZ) * genmat(Z)
+    @inbounds for (i, A) in enumerate(generators(MZ))
+        G[:, h + i] = A * center(Z)
+        G[:, (h + w + (i - 1) * h + 1):(h + w + i * h)] = A * genmat(Z)
+    end
```
schillic: I assumed that you want to imitate the code for the `GI` matrix of independent generators. Why the dependent ones?

alecarraro: typo

schillic: It is not about the name but about what the code is doing. I think it is different.

alecarraro: Do you mean that given $MZ = A^{(0)} + \sum_i \rho_i *A^{(i)}$ and $Z = \braket{c, G}$  it should be: 
```math
MZ \cdot Z = \braket{\underbrace{A^{(0)}c + \sum_i A^{(i)}c }_{\text{center}} , [A^{(0)}G ... A^{(k)}G]}
```
?

schillic: Maybe. I meant that I expected that you copy the code structure for the independent generators from the other method:

```julia
    Gi = Matrix{T}(undef, m, q * (w + 1))
    Gi[:, 1:q] = center(MZ) * genmat_indep(P)
    @inbounds for (i, A) in enumerate(generators(MZ))
        Gi[:, (q * i + 1):(q * (i + 1))] = A * genmat_indep(P)
    end
```

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,358 @@
...
+                         ::Type{<:Zonotope}) where {N,S<:AbstractZonotope{N},NM,
+                                                    MAT<:MatrixZonotopeProduct{NM}}
+
+Overapproximate the linear map of a zonotope through a product of matrix zonotopes,
+by recursively applying the overapproximation rule from the inside out.
+
+### Input
+
+- `lm` -- a linear map of a zonotope through a `MatrixZonotopeProduct`
+- `U` -- the target overapproximation type
```
schillic:
```suggestion
- `Zonotope` -- target type
```

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,358 @@
...
+                         ::Type{Zonotope}) where {N,S<:AbstractZonotope{N},NM,
+                                                  MAT<:MatrixZonotope{NM}}
+
+Overapproximate the linear map of a zonotope through a matrix zonotope,
+following a modification of Proposition 1 of [HuangLBS2025](@citet).
+
+### Input
+
+- `lm` -- a linear map of a zonotope through a matrix zonotope
+
```
schillic:
```suggestion
- `lm` -- a linear map of a zonotope through a matrix zonotope
- `Zonotope` -- target type

```

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,358 @@
...
+                                                                  NM,
+                                                                  MAT<:MatrixZonotope{NM}}
+
+Overapproximate the linear map of a sparse polynomial zonotope through a matrix zonotope,
+following Proposition 2 of [HuangLBS2025](@citet).
+
+### Input
+
+- `lm` -- a linear map of a sparse polynomial zonotope through a matrix zonotope
+
```
schillic:
```suggestion
- `lm` -- a linear map of a sparse polynomial zonotope through a matrix zonotope
- `SparsePolynomialZonotope` -- target type

```

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,358 @@
...
+                                                                    NM,
+                                                                    MAT<:MatrixZonotopeProduct{NM}}
+
+Overapproximate the linear map of a sparse polynomial zonotope through a product of matrix zonotopes,
+by recursively applying the overapproximation rule from the inside out.
+
+### Input
+
+- `lm` -- a linear map of a sparse polynomial zonotope through a `MatrixZonotopeProduct`
+- `U` -- the target overapproximation type
```
schillic:
```suggestion
- `SparsePolynomialZonotope` -- target type
```

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,358 @@
...
+    return Zonotope(c, G)
+end
+
+function _overapproximate_lmzp(lm::LinearMap{N,S,NM,MAT}) where {N,S<:AbstractZonotope{N},NM,
+                                                                 MAT<:MatrixZonotopeProduct{NM}}
+    MZs = factors(matrix(lm))
+    P = set(lm)
+
+    # apply overapproximation from innermost to outermost
+    reduced = foldr((A, acc) -> overapproximate(A * acc, U), MZP_factors; init=P)
```
schillic: `U` is not defined here. I suggest to call it `T` instead.
This shows that this method is currectly not tested.

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,358 @@
...
+        end
+        out_powers[i] = term
+    end
+    return out_powers
+end
+
+"""
+    _taylor_expmap(MZ::MatrixZonotope, P::S, k::Int) where S
+
+Compute the k-th order truncated Taylor expansion of the the exponential mao of a matrix zonotope
```
schillic:
```suggestion
Compute the k-th order truncated Taylor expansion of the exponential map of a matrix zonotope
```

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,358 @@
...
+                        k::Int) where {S<:Union{SparsePolynomialZonotope,AbstractZonotope}}
+    # avoid seg fault
+    if isempty(generators(MZ))
+        σ = linear_map(exp(center(MZ)), P)
+        return remove_redundant_generators(σ)
+    end
+
+    inner = _compute_inner_powers(MZ, P, k)
+    σ = reduce(exact_sum, inner)
+    return remove_redundant_generators(σ)
```
schillic:
```suggestion
    if isempty(generators(MZ))
        res = linear_map(exp(center(MZ)), P)
    else
        inner = _compute_inner_powers(MZ, P, k)
        res = reduce(exact_sum, inner)
    end
    return remove_redundant_generators(res)
```

src/Approximations/overapproximate_spz.jl
```diff
@@ -0,0 +1,358 @@
...
+
+```math
+\\displaystyle\\boxplus_{i=0}^k \\frac{\\mathcal{A}^i }{i!} X
+```
+"""
+function _taylor_expmap(MZ::MatrixZonotope, P::S,
+                        k::Int) where {S<:Union{SparsePolynomialZonotope,AbstractZonotope}}
+    # avoid seg fault
+    if isempty(generators(MZ))
+        σ = linear_map(exp(center(MZ)), P)
```
schillic: I am not sure that `linear_map` is the best thing here. Why not `overapproximate`? That would also avoid evaluating the matrix exponential directly.

src/Approximations/overapproximate_spz.jl
alecarraro: Sorry, this method is not tested yet indeed. I added it as an answer to another review. It is on my TODO list.
I will convert this PR to a draft

schillic:
This PR is getting too large for reviewing. Some of the changes can be pulled out into smaller PRs with standalone chunks.
You can keep this PR open, but maybe mark it as a draft for now.

src/Approximations/overapproximate_spz.jl
alecarraro: Using `exp` is more accurate and faster than using `overapproximate` with only 2-3 terms and for small ambient dimension. For example for a SPZ in 2d, the results are:
  -Linear map: 1.082 μs (31 allocations: 1.62 KiB)
  -overapproximate: 5.559 μs (305 allocations: 16.84 KiB)
gg
The green one is using `overapproximate`.
However with increasing dimensions, the 3 terms taylor expansion gets faster (despite allocating much more memory)
expmap_time_vs_dim_log
expmap_allocs_vs_dim_log

schillic: Maybe it could be a kwarg option? You can set the default to whatever you prefer.

Issue #3971, opened 2025-07-28
Make `overapproximate` more robust


alecarraro:
The function 

https://github.com/JuliaReach/LazySets.jl/blob/458fb29b2efe9303faff4accabe68708247dd4bb/src/Approximations/overapproximate.jl#L364-L371

causes an error when the number of parameters of the SPZ is 0, since it is tries to create an interval box with 0 dimensions.

PR #3972, opened 2025-07-28
#3971 Make `overapproximate` more robust 

alecarraro:
[#3971 ](https://github.com/JuliaReach/LazySets.jl/issues/3971)


schillic (approved):
It would be better if there was a `convert` implementation, but this is fine.

PR #3975, opened 2025-07-28
Add `overapproximate` for `LinearMap` between MZ and SPZ

alecarraro:
This PR implements proposition 2 in [LKB23](https://arxiv.org/pdf/2406.11056)

### DIFF
--- a/src/Approximations/overapproximate.jl
+++ b/src/Approximations/overapproximate.jl
@@ -849,7 +849,7 @@ function overapproximate(lm::LinearMap{N,S,NM,MAT},
     Z = Zonotope(zeros(T, n), genmat_indep(P))
     Zi = overapproximate(MZ * Z, Zonotope)
 
-    return SparsePolynomialZonotope(center(Pd), genmat_dep(Pd), genmat(Z), expmat(Pd), indexvector(Pd))
+    return SparsePolynomialZonotope(center(Pd), genmat_dep(Pd), genmat(Zi), expmat(Pd), indexvector(Pd))
 end
 
 """


src/Approximations/overapproximate.jl
```diff
@@ -810,3 +810,70 @@ function overapproximate(cap::Intersection{N,<:AbstractPolyhedron,<:AbstractPoly
...
+
+"""
+function overapproximate(lm::LinearMap{N,S,NM,MAT},
+                         ::Type{<:SparsePolynomialZonotope}) where {N,
+                                                                    S<:SparsePolynomialZonotope{N},
+                                                                    NM,MAT<:MatrixZonotope{NM}}
+    P = set(lm)
+    MZ = matrix(lm)
+    n = dim(P)
+    T = promote_type(eltype(MZ), eltype(P))
```
schillic: Shorter:
```suggestion
    T = promote_type(N, NM)
```

src/Approximations/overapproximate.jl
```diff
@@ -810,3 +810,70 @@ function overapproximate(cap::Intersection{N,<:AbstractPolyhedron,<:AbstractPoly
...
+    Pd = linear_map(MZ, SP)
+
+    if ngens_indep(P) == 0
+        return Pd # fallback to linear_map
+    end
+
+    Z = Zonotope(zeros(T, n), genmat_indep(P))
+    Zi = overapproximate(MZ * Z, Zonotope)
+
+    return minkowski_sum(Pd, Zi)
```
schillic: You can avoid some allocations, and also preserve the parameters:
```suggestion
    return SparsePolynomialZonotope(center(Pd), genmat_dep(Pd), genmat(Z), expmat(Pd), indexvector(Pd))
```

test/Approximations/overapproximate.jl
```diff
@@ -166,6 +166,39 @@ for N in @tN([Float64, Float32, Rational{Int}])
     MZP = MZ2 * MZ
     res2 = overapproximate(MZP * Z, Zonotope)
     @test res2 == overapproximate(MZ2 * res, Zonotope)
+
+    # overapproximate the lm of a matrix zonotope with a SPZ
+    P = SparsePolynomialZonotope(N[1, -1], N[1 1; 0 -1], hcat(N[0, 1]), [2 1; 0 1; 1 0], [1, 2, 3])
+    res = overapproximate(MZ * P, SparsePolynomialZonotope)
+    @test center(res) == [0, -2]
+    @test genmat_dep(res) == hcat(N[1 0; -1 -2], N[1, -1], N[1 1; 1 -1])
+    @test genmat_indep(res) == hcat([1, 1], [0, 2], [0, 0])
```
schillic: This is not needed here, but for clarity that this is not an integer matrix...
```suggestion
    @test genmat_indep(res) == hcat(N[1, 1], N[0, 2], N[0, 0])
```

test/Approximations/overapproximate.jl
```diff
@@ -166,6 +166,39 @@ for N in @tN([Float64, Float32, Rational{Int}])
...
+    MZ = MatrixZonotope(N[1 1; -1 1], Vector{Matrix{N}}())
+    res = overapproximate(MZ * P, SparsePolynomialZonotope)
+    @test res == linear_map(N[1 1; -1 1], P)
+
+    #case: id_mz ≠ id_spz
+    MZ = MatrixZonotope(N[1 1; -1 1], [N[1 0; 1 2]], [4])
+    res = overapproximate(MZ * P, SparsePolynomialZonotope)
+    @test center(res) == [0, -2]
+    @test genmat_dep(res) == hcat(N[1 0; -1 -2], N[1, -1], N[1 1; 1 -1])
+    @test genmat_indep(res) == hcat([1, 1], [0, 2], [0, 0])
```
schillic:
```suggestion
    @test genmat_indep(res) == hcat(N[1, 1], N[0, 2], N[0, 0])
```

src/Approximations/overapproximate.jl
schillic: Maybe it should even be tested that the `idx` vector is correct?

test/Approximations/overapproximate.jl
alecarraro: > This is not needed here, but for clarity that this is not an integer matrix...

oh yes, i forgot the `N`

src/Approximations/overapproximate.jl
alecarraro: I will add a test now

alecarraro: Actually I am not sure it is necessary, since the `indexvector` is the one computed by the `linear_map  function
So in the test I should rewrite 
```julia
res = overapproximate(MZ * P, SparsePolynomialZonotope)
SP = SparsePolynomialZonotope(center(P), genmat_dep(P), zeros(N, n, 0), expmat(P), indexvector(P))
Pd = linear_map(MZ, SP)
...
@test indexvector(PD) == indexvector(res) 
```
Which boils down to either verify that the constructor of SPZ works correctly or alternatively that the `linear_map` produces the  id correctly, but they should already be tested

schillic: The old version did not return the correct `idx` in all cases. It returned `1:k` because the current implementation of `minkowski_sum` does that (we should fix that!). Here this is now fixed with my suggestion (which avoids the call to `minkowski_sum`), so I suggest to test that there is no regression about that.

schillic: > the current implementation of `minkowski_sum` does that (we should fix that!)

I created #3976

alecarraro: Until we find a solution on how to manage the IDs (see my reply on [#3976](https://github.com/JuliaReach/LazySets.jl/issues/3976) I would keep it like it is. 
If someone needs to work with the IDs (i.e for reachability) , you would have to write a separate function to handle the ID anyway. For example for the reachability algorithm you need to ensure that the resulting SPZ has an unique ID globally, considering all the `SPZ` instances generated in one session.

schillic: The paper clearly says to use the index vector $$\hat{id}$$ of the PZ:

Screenshot_20250729_155557

alecarraro: Once the other PR is merged, is it better to keep it as it is now or going back to `minkowski_sum`?
I think that `minkowski_sum` would be clearer but both are fine to me.

schillic: It was fine as it was.

PR #3979, opened 2025-07-29
Add `overapproximate` for exponential map for SPZ and Zonotope

alecarraro:
Implement proposition  of [LKB23](https://arxiv.org/pdf/2406.11056l)

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -12,6 +12,7 @@ makedocs(; sitename="LazySets.jl",
                                 assets=["assets/aligned.css", "assets/citations.css"],
                                 size_threshold_warn=150 * 2^10),
          pagesonly=true,
+         checkdocs = :exports,
          plugins=[bib],
          pages=[
                 #
@@ -138,7 +139,8 @@ makedocs(; sitename="LazySets.jl",
                                                    "lib/approximations/approximate.md",
                                                    "lib/approximations/decompose.md",
                                                    "lib/approximations/hausdorff_distance.md",
-                                                   "lib/approximations/overapproximate_norm.md"
+                                                   "lib/approximations/overapproximate_norm.md",
+                                                   "lib/approximations/overapproximate_expmap.md",
                                                    #
                                                    ],
                               "MatrixSets" => [
--- a/docs/src/lib/approximations/overapproximate_expmap.md
+++ b/docs/src/lib/approximations/overapproximate_expmap.md
@@ -1,0 +1,16 @@
+
+```@contents
+Pages = ["overapproximate_expmap.md"]
+Depth = 3
+```
+
+```@meta
+CurrentModule = LazySets.Approximations
+```
+
+# Overapproximation
+
+```@docs
+taylor_expmap_truncation
+taylor_expmap_remainder
+```
--- a/src/Approximations/Approximations.jl
+++ b/src/Approximations/Approximations.jl
@@ -21,7 +21,9 @@ export approximate,
        SphericalDirections,
        CustomDirections,
        isbounding,
-       overapproximate_norm
+       overapproximate_norm,
+       taylor_expmap_truncation,
+       taylor_expmap_remainder
 
 using ..LazySets, ReachabilityBase.Arrays, Requires, LinearAlgebra, SparseArrays
 import IntervalArithmetic as IA
@@ -49,6 +51,7 @@ include("approximate.jl")
 include("decompositions.jl")
 include("hausdorff_distance.jl")
 include("overapproximate_norm.jl")
+include("overapproximate_expmap.jl")
 include("init.jl")
 
 end # module
--- a/src/Approximations/overapproximate_expmap.jl
+++ b/src/Approximations/overapproximate_expmap.jl
@@ -178,7 +178,7 @@ function load_intervalmatrices_overapproximation_expmap()
             end
 
             tayexp = taylor_expmap_truncation(MZP, P, k)
-            lagrem = lagrange_remainder(P, matnorm, k)
+            lagrem = taylor_expmap_remainder(P, matnorm, k)
             P_approx = minkowski_sum(tayexp, lagrem)
 
             return remove_redundant_generators(P_approx)
--- a/src/Approximations/overapproximate_expmap.jl
+++ b/src/Approximations/overapproximate_expmap.jl
@@ -1,0 +1,233 @@
+"""
+    _compute_inner_powers(A::MatrixZonotope, P::S,
+                               k::Int) where {S<:Union{SparsePolynomialZonotope,
+                                                       AbstractZonotope}}
+    
+Compute the first `k` inner powers of the matrix zonotope `A` applied to the set `P`.
+
+This function returns a vector of overapproximated sets of the form:
+
+```math
+    (1/n!) * \\mathcal{A}^n * P
+```
+for `n = 0` to `k`.
+"""
+function _compute_inner_powers(A::MatrixZonotope, P::S,
+                               k::Int) where {S<:Union{SparsePolynomialZonotope,
+                                                       AbstractZonotope}}
+    T = typeof(P)
+    powers = Vector{T}(undef, k + 1)
+    invfact = 1
+    powers[1] = P
+
+    @inbounds for n in 1:k
+        invfact /= n
+        term = overapproximate(scale(invfact, A) * powers[n], T)
+        powers[n + 1] = term
+    end
+    return powers
+end
+
+"""
+    _compute_outer_powers(B::MatrixZonotope, in_powers::Vector{S},
+                               k::Int) where {S<:Union{SparsePolynomialZonotope,
+                                                       AbstractZonotope}}
+
+Apply `B` repeatedly to each element in `in_powers`, approximating:
+
+```math
+B^n * (A^n * P)
+```
+
+for `n = 0` to `k`.
+"""
+function _compute_outer_powers(B::MatrixZonotope, in_powers::Vector{S},
+                               k::Int) where {S<:Union{SparsePolynomialZonotope,
+                                                       AbstractZonotope}}
+    out_powers = similar(in_powers)
+    out_powers[1] = in_powers[1]
+
+    @inbounds for n in 2:(k + 1)
+        term = in_powers[n]
+        for _ in 1:n
+            term = overapproximate(B * term, S)
+        end
+        out_powers[n] = term
+    end
+    return out_powers
+end
+
+"""
+    taylor_expmap_truncation(A::MatrixZonotope, P::S, k::Int) where S
+
+Compute the k-th order truncated Taylor expansion of the exponential map of a matrix zonotope
+
+### Input
+
+- `A` -- a matrix zonotope
+- `P` -- a zonotopic set
+- `k` -- the order of the Taylor expansion
+
+### Output
+
+A zonotopic set representing the k-th order truncated Taylor expansion.
+
+### Algorithm
+
+This function computes the approximation:
+
+```math
+\\displaystyle\\boxplus_{i=0}^k \\frac{\\mathcal{A}^i }{i!} X
+```
+"""
+function taylor_expmap_truncation(A::MatrixZonotope, P::S,
+                        k::Int) where {S<:Union{SparsePolynomialZonotope,AbstractZonotope}}
+    if isempty(generators(A))
+        res = linear_map(exp(center(A)), P)
+    else
+        inner = _compute_inner_powers(A, P, k)
+        res = reduce(exact_sum, inner)
+    end
+
+    return res
+end
+
+"""
+    taylor_expmap_truncation(MZP::MatrixZonotopeProduct, P::S, k::Int) where S
+
+Compute the k-th order truncated Taylor expansion of the exponential map of a matrix zonotope product
+
+### Input
+
+- `MZP` -- a matrix zonotope product
+- `P` -- a zonotopic set
+- `k` -- the order of the Taylor expansion
+
+### Output
+
+A zonotopic set representing the k-th order truncated Taylor expansion.
+
+### Algorithm
+
+This function computes the approximation:
+
+```math
+\\displaystyle\\boxplus_{i=0}^k \\frac{\\mathcal{A}^i \\mathcal{B}^i}{i!} X
+```
+"""
+function taylor_expmap_truncation(MZP::MatrixZonotopeProduct, P::S,
+                        k::Int) where {S<:Union{SparsePolynomialZonotope,AbstractZonotope}}
+    # inner powers on the last factor
+    last_factor = factors(MZP)[end]
+    terms = _compute_inner_powers(last_factor, P, k)
+
+    # propagate outwards through the preceding factors
+    for A in reverse(view(factors(MZP), 1:(nfactors(MZP) - 1)))
+        terms = _compute_outer_powers(A, terms, k)
+    end
+
+    res = reduce(exact_sum, terms)
+    return res
+end
+
+function load_intervalmatrices_overapproximation_expmap()
+    return quote
+        using .IntervalMatrices: IntervalMatrix
+
+        """
+            taylor_expmap_remainder(P::S, matnorm::Real, k::Int) where S
+
+        Compute the Lagrange remainder term of the k-th order truncated Taylor expansion 
+        of the exponential map of a matrix zonotope applied to a zonotopic set.
+
+        ### Input
+
+        - `P` -- a zonotopic set
+        - `matnorm` -- an upper bound on the norm of the matrix zonotope
+        - `k` -- the order of the Taylor expansion
+
+        ### Output
+
+        A zonotope over-approximating the remainder term of the Taylor expansion.
+        """
+        function taylor_expmap_remainder(P::S, matnorm::Real, k::Int) where{S<:Union{SparsePolynomialZonotope,
+                                                                AbstractZonotope}}
+            n = dim(P)
+            N = eltype(P)
+
+            ϵ = matnorm / (k + 2)
+            ε = IntervalMatrix(fill(IA.interval(N(-1.0), N(1.0)), n, n))
+            ε *= matnorm^(k + 1) / (factorial(k + 1) * (1 - ϵ))
+
+            Zp = overapproximate(P, Zonotope)
+            rhs = overapproximate(ε * Zp, Zonotope)
+            return rhs
+        end
+
+        function _overapproximate_emz(em::ExponentialMap{N,S,MAT},
+                                      k::Int=2) where {N,S<:Union{SparsePolynomialZonotope,
+                                                                AbstractZonotope},
+                                                       MAT<:AbstractMatrixZonotope{N}}
+            MZP = matrix(em).M
+            P = set(em)
+
+            matnorm = N(overapproximate_norm(MZP, Inf))
+            ϵ = matnorm / (k + 2)
+            if ϵ > 1
+                @warn "k should be chosen such that ϵ<1 " ϵ
+            end
+
+            tayexp = taylor_expmap_truncation(MZP, P, k)
+            lagrem = lagrange_remainder(P, matnorm, k)
+            P_approx = minkowski_sum(tayexp, lagrem)
+
+            return remove_redundant_generators(P_approx)
+        end
+
+        """
+        	overapproximate(em::ExponentialMap{N,S,MAT},
+                                 T::Type{<:SparsePolynomialZonotope},
+                                 k::Int=2) where {N,S<:SparsePolynomialZonotope,
+                                                  MAT<:AbstractMatrixZonotope{N}}
+
+        Overapproximate the exponential map of a sparse polynomial zonotope through a composition of matrix 
+        zonotopes, following Proposition 1 of [HuangLBS2025](@citet).
+
+        ### Input
+
+        - `em` -- an expontial map of a sparse polynomial zonotope through a product of matrix zonotopes
+
+        ### Output
+
+        A sparse polynomial zonotope overapproximating the exponential map.
+        """
+        function overapproximate(em::ExponentialMap{N,S,MAT},
+                                 T::Type{<:SparsePolynomialZonotope},
+                                 k::Int=2) where {N,S<:SparsePolynomialZonotope,
+                                                  MAT<:AbstractMatrixZonotope{N}}
+            return _overapproximate_emz(em, k)
+        end
+
+        """
+        	overapproximate(em::ExponentialMap{N,S,MAT}, T::Type{<:Zonotope},
+                                 k::Int=2) where {N,S<:AbstractZonotope,
+                                                  MAT<:AbstractMatrixZonotope{N}}
+
+        Overapproximate the exponential map of a zonotope through a composition of matrix 
+        zonotopes, following Proposition 1 of [HuangLBS2025](@citet).
+
+        ### Input
+
+        - `em` -- an expontial map of a zonotope through a product of matrix zonotopes
+
+        ### Output
+
+        A zonotope overapproximating the exponential map.
+        """
+        function overapproximate(em::ExponentialMap{N,S,MAT}, T::Type{<:Zonotope},
+                                 k::Int=2) where {N,S<:AbstractZonotope,
+                                                  MAT<:AbstractMatrixZonotope{N}}
+            return _overapproximate_emz(em, k)
+        end
+    end
+end
--- a/test/Approximations/overapproximate_expmap.jl
+++ b/test/Approximations/overapproximate_expmap.jl
@@ -1,0 +1,45 @@
+
+using Test, LazySets, LinearAlgebra
+using IntervalMatrices: IntervalMatrix
+using RangeEnclosures
+using IntervalArithmetic: Interval
+
+for N in [Float32, Float64, Rational{Int}]
+    MZ = MatrixZonotope(N[1 1; -1 1], [N[1 0; 1 2]])
+    P = SparsePolynomialZonotope(N[1, -1], N[1 1; 0 -1], hcat(N[0, 1]), [2 1; 0 1; 1 0], [1, 2, 3])
+
+    # general test
+    mzexp = MatrixZonotopeExp(MZ)
+    em = ExponentialMap(mzexp, P)
+    Pex = overapproximate(em, SparsePolynomialZonotope, 2)
+    @test isa(Pex, SparsePolynomialZonotope)
+    @test size(center(Pex), 1) == size(center(P), 1)
+
+    # expectation: exp(0)*P = I*P = P
+    ZM = MatrixZonotope(zeros(N, 2, 2), Vector{Matrix{N}}())
+    mzexp = MatrixZonotopeExp(ZM)
+    em = ExponentialMap(mzexp, P)
+    Pex = overapproximate(em, SparsePolynomialZonotope, 2)
+    @test center(Pex) == center(P)
+    @test genmat_dep(Pex) == genmat_dep(P)
+    @test genmat_indep(Pex) == genmat_indep(P)
+    @test expmat(Pex) == expmat(P)
+end
+
+for N in [Float32, Float64]
+    P = SparsePolynomialZonotope(N[1, -1], N[1 1; 0 -1], hcat(N[0, 1]), [2 1; 0 1; 1 0], [1, 2, 3])
+
+    MZ = MatrixZonotope(N[1 -1; -1 2], [N[1.001 -0.999; -0.999 2.005]])
+    mzexp =MatrixZonotopeExp(MZ)
+    em_spz = ExponentialMap(mzexp, P)
+    Pex = overapproximate(em_spz,SparsePolynomialZonotope, 3)
+
+    Z = overapproximate(convert(SimpleSparsePolynomialZonotope, P), Zonotope)
+    em_z = ExponentialMap(mzexp, Z)
+    Zex = overapproximate(em_z, Zonotope, 3)
+
+    # samples from the SPZ and check if the points are contained in the overapproximation
+    sampler = LazySets.PolynomialZonotopeSampler()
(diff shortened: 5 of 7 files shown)


docs/make.jl
```diff
@@ -12,6 +12,7 @@ makedocs(; sitename="LazySets.jl",
                                 assets=["assets/aligned.css", "assets/citations.css"],
                                 size_threshold_warn=150 * 2^10),
          pagesonly=true,
+         checkdocs = :exports,
```
schillic: Is this needed? What is the motivation?

src/Approximations/Approximations.jl
```diff
@@ -21,7 +21,9 @@ export approximate,
        SphericalDirections,
        CustomDirections,
        isbounding,
-       overapproximate_norm
+       overapproximate_norm,
+       taylor_expmap_truncation,
+       taylor_expmap_remainder
```
schillic: I do not think these functions should be part of the API.

test/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,45 @@
+
+using Test, LazySets, LinearAlgebra
+using IntervalMatrices: IntervalMatrix
+using RangeEnclosures
+using IntervalArithmetic: Interval
```
schillic: This line does not seem to be needed.

docs/make.jl
alecarraro: I added docstrings for the internal function  for clarity and i was trying to avoid getting an error from Documenter for having docstrings not referenced but it didnt work

test/Approximations/overapproximate_expmap.jl
alecarraro: Yes I removed it already, sorry

src/Approximations/Approximations.jl
alecarraro: image
For the particular solution in the reachability algorithm I need to compute these expressions, which can be easily computed using these 2 functions. Thats why I also have the matrix norm as an argument for the `taylor_expmap_remainder` function

schillic: That is all fine. I just do not think they should be exported from `LazySets`. Do you think they should be?

docs/make.jl
alecarraro: The extra docstrings are for `_compute_inner_powers` and `_compute_outer_powers`

schillic: We sometimes have internal functions documented. You can just add them if that resolves the problem.

docs/src/lib/approximations/overapproximate_expmap.md
```diff
@@ -0,0 +1,16 @@
...
+
+```@meta
+CurrentModule = LazySets.Approximations
+```
+
+# Overapproximation
+
+```@docs
+taylor_expmap_truncation
+taylor_expmap_remainder
```
schillic, line 15: Maybe you need to add `LazySets.Approximations.` in front. Not sure why because you exported them. I am just guessing.

src/Approximations/Approximations.jl
alecarraro: ah no its ok. I can just call them using `Approximations.taylor...`. I do not think they have any other use than in that algorithm

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,233 @@
...
+        end
+
+        """
+        	overapproximate(em::ExponentialMap{N,S,MAT},
+                                 T::Type{<:SparsePolynomialZonotope},
+                                 k::Int=2) where {N,S<:SparsePolynomialZonotope,
+                                                  MAT<:AbstractMatrixZonotope{N}}
+
+        Overapproximate the exponential map of a sparse polynomial zonotope through a composition of matrix 
+        zonotopes, following Proposition 1 of [HuangLBS2025](@citet).
```
schillic:
```suggestion
        zonotopes, following Proposition 3 of [HuangLBS2025](@citet).
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,233 @@
...
+            return _overapproximate_emz(em, k)
+        end
+
+        """
+        	overapproximate(em::ExponentialMap{N,S,MAT}, T::Type{<:Zonotope},
+                                 k::Int=2) where {N,S<:AbstractZonotope,
+                                                  MAT<:AbstractMatrixZonotope{N}}
+
+        Overapproximate the exponential map of a zonotope through a composition of matrix 
+        zonotopes, following Proposition 1 of [HuangLBS2025](@citet).
```
schillic:
```suggestion
        zonotopes, following Proposition 3 of [HuangLBS2025](@citet).
```

test/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,44 @@
...
+
+for N in @tN([Float32, Float64, Rational{Int}])
+    MZ = MatrixZonotope(N[1 1; -1 1], [N[1 0; 1 2]])
+    P = SparsePolynomialZonotope(N[1, -1], N[1 1; 0 -1], hcat(N[0, 1]), [2 1; 0 1; 1 0], [1, 2, 3])
+
+    # general test
+    mzexp = MatrixZonotopeExp(MZ)
+    em = ExponentialMap(mzexp, P)
+    Pex = overapproximate(em, SparsePolynomialZonotope, 2)
+    @test isa(Pex, SparsePolynomialZonotope)
```
schillic: I find it more readable to use infix notation for `isa`.
```suggestion
    @test Pex isa SparsePolynomialZonotope{N}
```

test/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,44 @@
...
+    @test size(center(Pex), 1) == size(center(P), 1)
```
schillic:
```suggestion
    @test dim(Pex) == dim(P)
```
I guess this is what this means.

test/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,44 @@
...
+
+    # expectation: exp(0)*P = I*P = P
+    ZM = MatrixZonotope(zeros(N, 2, 2), Vector{Matrix{N}}())
+    mzexp = MatrixZonotopeExp(ZM)
+    em = ExponentialMap(mzexp, P)
+    Pex = overapproximate(em, SparsePolynomialZonotope, 2)
+    @test center(Pex) == center(P)
+    @test genmat_dep(Pex) == genmat_dep(P)
+    @test genmat_indep(Pex) == genmat_indep(P)
+    @test expmat(Pex) == expmat(P)
```
schillic: I think you also want to preserve the IDs, so you can just check for equality of the two SPZs. Note that this will fail now, but it should be fixed.
```suggestion
    @test Pex == P
```

test/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,44 @@
...
+    # general test
+    mzexp = MatrixZonotopeExp(MZ)
+    em = ExponentialMap(mzexp, P)
+    Pex = overapproximate(em, SparsePolynomialZonotope, 2)
+    @test isa(Pex, SparsePolynomialZonotope)
+    @test size(center(Pex), 1) == size(center(P), 1)
+end
+
+for N in @tN([Float32, Float64])
+    P = SparsePolynomialZonotope(N[1, -1], N[1 1; 0 -1], hcat(N[0, 1]), [2 1; 0 1; 1 0], [1, 2, 3])
```
schillic: The following change will uncover a problem: the IDs are not preserved...
```suggestion
    P = SparsePolynomialZonotope(N[1, -1], N[1 1; 0 -1], hcat(N[0, 1]), [2 1; 0 1; 1 0], [1, 3, 5])
```

test/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,44 @@
...
+    Pex = overapproximate(em, SparsePolynomialZonotope, 2)
+    @test center(Pex) == center(P)
+    @test genmat_dep(Pex) == genmat_dep(P)
+    @test genmat_indep(Pex) == genmat_indep(P)
+    @test expmat(Pex) == expmat(P)
+
+    MZ = MatrixZonotope(N[1 -1; -1 2], [N[1.001 -0.999; -0.999 2.005]])
+    mzexp =MatrixZonotopeExp(MZ)
+    em_spz = ExponentialMap(mzexp, P)
+    Pex = overapproximate(em_spz,SparsePolynomialZonotope, 3)
```
schillic:
```suggestion
    Pex = overapproximate(em_spz,SparsePolynomialZonotope, 5)
```
I get here:
```julia
┌ Warning: k should be chosen such that ϵ<1 
│   ϵ = 1.2008
```

test/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,44 @@
...
+
+    Z = overapproximate(convert(SimpleSparsePolynomialZonotope, P), Zonotope)
+    em_z = ExponentialMap(mzexp, Z)
+    Zex = overapproximate(em_z, Zonotope, 3)
```
schillic:
```suggestion
    Zex = overapproximate(em_z, Zonotope, 5)
```
Same as above...

test/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,44 @@
...
+
+    # samples from the SPZ and check if the points are contained in the overapproximation
+    sampler = LazySets.PolynomialZonotopeSampler()
+    pts = sample(Pex, 100; sampler=sampler)
```
schillic: 10 should be sufficient.
```suggestion
    pts = sample(Pex, 10; sampler=sampler)
```

test/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,44 @@
...
+    @test isa(Pex, SparsePolynomialZonotope)
+    @test size(center(Pex), 1) == size(center(P), 1)
+end
+
+for N in @tN([Float32, Float64])
+    P = SparsePolynomialZonotope(N[1, -1], N[1 1; 0 -1], hcat(N[0, 1]), [2 1; 0 1; 1 0], [1, 2, 3])
+
+    # expectation: exp(0)*P = I*P = P
+    ZM = MatrixZonotope(zeros(N, 2, 2), Vector{Matrix{N}}())
+    mzexp = MatrixZonotopeExp(ZM)
```
schillic:
```suggestion
    MZ = MatrixZonotope(zeros(N, 2, 2), Vector{Matrix{N}}())
    mzexp = MatrixZonotopeExp(MZ)
```

test/Approximations/overapproximate_expmap.jl
schillic: I think I found the reason and fixed it in #3980.

alecarraro: oh wow well spotted, i gave it for granted that it would preserve it

test/Approximations/overapproximate_expmap.jl
alecarraro: I kept 3 because higher `k`s are very expensive and the actual value of $\epsilon$ is often much smaller due to the overapproximation of the matrix zonotope norm , so chances are that actually $\epsilon <1$

schillic: Okay, but are the results valid? Could you try whether the exact norm computation yields a value below 1 (and if so write a comment to the test)?

alecarraro: I am thinking about adding, as a keyword argument , the variable `matnorm`. If nothing is provided it computes the norm overapproximation, else the argument.
The reason is that in the reachability algorithm you call this function at each iteration, so it might be convenient to just precompute the exact norm once

docs/src/lib/approximations/overapproximate_expmap.md
alecarraro: The problem is only with `taylor_expmap_remainder` which is defined in the `load_intervalmatrices_overapproximation_expmap()` method

test/Approximations/overapproximate_expmap.jl
schillic: Yes, very good!

docs/src/lib/approximations/overapproximate_expmap.md
schillic: Ah, good point. That helped me understand the problem. You need to add `IntervalMatrices` here:

https://github.com/JuliaReach/LazySets.jl/blob/49ac692ca81165ba9f92a5790ed93305d143b12b/docs/make.jl#L2

alecarraro:
> I tried out a simple example where the SPZ is just a zonotope and the MZ is just a matrix. The overapproximation is not enormously bigger, and it gets quite tight when raising κ to 5. This gives some confidence in the correctness.

thats very nice!

docs/make.jl
schillic: ~So I guess you can undo this change now. At least I do not see why we would want it.~

Never mind, you already did!

test/Approximations/overapproximate_expmap.jl
schillic: Could you merge `master` in here so we can try it?

alecarraro: sure

alecarraro: I did a mistake, one second and I ll do it
EDIT: nvm, it should be good

alecarraro:
In my environment the documentation compiles correctly. It says that `intervalMatrices` is not installed. I can try to check how send a PR to change Github Workflows

docs/src/lib/approximations/overapproximate_expmap.md
schillic: Ah, now you need to add `IntervalMatrices` to the `docs/Project.toml` file as well. You can copy the two lines containing `IntervalMatrices` from `test/Project.toml`.

test/Approximations/overapproximate_expmap.jl
schillic: So it seems to still not hold with the current implementation. Could you try to investigate where in the algorithm it loses the `idx`? At least from the algorithm it should return the same IDs I think.

alecarraro: I fixed it, it was the `remove_redundant_generator` function

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+                                      k::Int,
+                                      matnorm::Real) where {N,
+                                                            S<:Union{SparsePolynomialZonotope,
+                                                                     AbstractZonotope},
+                                                            MAT<:AbstractMatrixZonotope{N}}
+            MZP = matrix(em).M
+            P = set(em)
+
+            ϵ = matnorm / (k + 2)
+            if ϵ > 1
```
schillic: This is important because for `ϵ = 1` you have a division by 0.
```suggestion
            if ϵ >= 1
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+    res = reduce(exact_sum, terms)
+    return res
+end
+
+function load_intervalmatrices_overapproximation_expmap()
+    return quote
+        using .IntervalMatrices: IntervalMatrix
+
+        """
+            taylor_expmap_remainder(P::S, matnorm::Real, k::Int) where S
```
schillic: I suggest to only allow an `AbstractZonotope` here and adapt the code (→ several other code suggestions).
```suggestion
            taylor_expmap_remainder(Z::AbstractZonotope, matnorm::Real, k::Int)
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+        - `matnorm` -- an upper bound on the norm of the matrix zonotope
+        - `k` -- the order of the Taylor expansion
+
+        ### Output
+
+        A zonotope over-approximating the remainder term of the Taylor expansion.
+        """
+        function taylor_expmap_remainder(P::S, matnorm::Real,
+                                         k::Int) where {S<:Union{SparsePolynomialZonotope,
+                                                                 AbstractZonotope}}
```
schillic:
```suggestion
        function taylor_expmap_remainder(Z::AbstractZonotope{N}, matnorm::Real, Int) where {N}
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+            n = dim(P)
+            N = eltype(P)
+
+            ϵ = matnorm / (k + 2)
+            ε = IntervalMatrix(fill(IA.interval(N(-1.0), N(1.0)), n, n))
+            ε *= matnorm^(k + 1) / (factorial(k + 1) * (1 - ϵ))
+
+            Zp = overapproximate(P, Zonotope)
+            rhs = overapproximate(ε * Zp, Zonotope)
+            return rhs
```
schillic: I suggest to use a capital letter for a matrix.
```suggestion
            E = IntervalMatrix(fill(IA.interval(N(-1.0), N(1.0)), n, n))
            E *= matnorm^(k + 1) / (factorial(k + 1) * (1 - ϵ))
            res = overapproximate(E * Z, Zonotope)
            return res
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+            MZP = matrix(em).M
+            P = set(em)
+
+            ϵ = matnorm / (k + 2)
+            if ϵ > 1
+                @warn "k should be chosen such that ϵ<1 " ϵ
+            end
+
+            tayexp = taylor_expmap_truncation(MZP, P, k)
+            lagrem = taylor_expmap_remainder(P, matnorm, k)
```
schillic:
```suggestion
            Z = P isa AbstractZonotope ? P : overapproximate(P, Zonotope)
            lagrem = taylor_expmap_remainder(Z, matnorm, k)
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+end
+
+function load_intervalmatrices_overapproximation_expmap()
+    return quote
+        using .IntervalMatrices: IntervalMatrix
+
+        """
+            taylor_expmap_remainder(P::S, matnorm::Real, k::Int) where S
+
+        Compute the Lagrange remainder term of the k-th order truncated Taylor expansion
```
schillic:
```suggestion
        Overapproximate the Lagrange remainder term of the k-th order truncated Taylor expansion 
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+    else
+        inner = _compute_inner_powers(A, P, k)
+        res = reduce(exact_sum, inner)
+    end
+
+    return res
+end
+
+"""
+    taylor_expmap_truncation(MZP::MatrixZonotopeProduct, P::S, k::Int) where S
```
schillic:
```suggestion
    taylor_expmap_truncation(MZP::MatrixZonotopeProduct, P::S, k::Int)
        where {S<:Union{SparsePolynomialZonotope,AbstractZonotope}}
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+
+Compute the k-th order truncated Taylor expansion of the exponential map of a matrix zonotope product
+
+### Input
+
+- `MZP` -- a matrix zonotope product
+- `P` -- a zonotopic set
```
schillic:
```suggestion
- `P` -- a (potentially polynomial) zonotopic set
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+- `k` -- the order of the Taylor expansion
+
+### Output
+
+A zonotopic set representing the k-th order truncated Taylor expansion.
```
schillic:
```suggestion
A (polynomial) zonotopic set representing the k-th order truncated Taylor expansion.
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+
+### Algorithm
+
+This function computes the approximation:
+
+```math
+\\displaystyle\\boxplus_{i=0}^k \\frac{\\mathcal{A}^i \\mathcal{B}^i}{i!} X
```
schillic:
```suggestion
\\displaystyle\\boxplus_{i=0}^k \\frac{\\mathcal{A}^i \\mathcal{B}^i}{i!} P
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+        for _ in 1:n
+            term = overapproximate(B * term, S)
+        end
+        out_powers[n] = term
+    end
+    return out_powers
+end
+
+"""
+    taylor_expmap_truncation(A::MatrixZonotope, P::S, k::Int) where S
```
schillic:
```suggestion
    taylor_expmap_truncation(A::MatrixZonotope, P::S, k::Int)
        where {S<:Union{SparsePolynomialZonotope,AbstractZonotope}}
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+
+### Input
+
+- `A` -- a matrix zonotope
+- `P` -- a zonotopic set
+- `k` -- the order of the Taylor expansion
+
+### Output
+
+A zonotopic set representing the k-th order truncated Taylor expansion.
```
schillic:
```suggestion
- `P` -- a (potentially polynomial) zonotopic set
- `k` -- the order of the Taylor expansion

### Output

A (polynomial) zonotopic set representing the k-th order truncated Taylor expansion.
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+This function computes the approximation:
+
+```math
+\\displaystyle\\boxplus_{i=0}^k \\frac{\\mathcal{A}^i }{i!} X
+```
+"""
+function taylor_expmap_truncation(A::MatrixZonotope, P::S,
+                        k::Int) where {S<:Union{SparsePolynomialZonotope,AbstractZonotope}}
+    if isempty(generators(A))
+        res = linear_map(exp(center(A)), P)
```
schillic:
```suggestion
        res = exponential_map(center(A), P)
```
This is probably doing the same but can in principle sometimes be more efficient.

schillic: When you are in this case, you do not need the Lagrange remainder, right? So maybe it is worth catching this case earlier in the call stack?

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
+"""
+    _compute_inner_powers(A::MatrixZonotope, P::S,
+                               k::Int) where {S<:Union{SparsePolynomialZonotope,
+                                                       AbstractZonotope}}
+    
+Compute the first `k` inner powers of the matrix zonotope `A` applied to the set `P`.
```
schillic: Does "inner" have any meaning?

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+                               k::Int) where {S<:Union{SparsePolynomialZonotope,
+                                                       AbstractZonotope}}
+    T = typeof(P)
+    powers = Vector{T}(undef, k + 1)
+    invfact = 1
+    powers[1] = P
+
+    @inbounds for n in 1:k
+        invfact /= n
+        term = overapproximate(scale(invfact, A) * powers[n], T)
```
schillic: `T == S`
```suggestion
    powers = Vector{S}(undef, k + 1)
    invfact = 1
    powers[1] = P

    @inbounds for n in 1:k
        invfact /= n
        term = overapproximate(scale(invfact, A) * powers[n], S)
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+"""
+function _compute_inner_powers(A::MatrixZonotope, P::S,
+                               k::Int) where {S<:Union{SparsePolynomialZonotope,
+                                                       AbstractZonotope}}
+    T = typeof(P)
+    powers = Vector{T}(undef, k + 1)
+    invfact = 1
+    powers[1] = P
+
+    @inbounds for n in 1:k
```
schillic: We usually use `n` for the dimension and not as index.

src/Approximations/overapproximate_expmap.jl
schillic: I think instead of by `invfact` you should only scale by `1/n`. You can see this by unrolling the sequence you compute:

```julia
invfact = 1
p[1] = P

invfact = 1
p[2] = oa(A*invfact * p[1]) = oa(A/1 * P)

invfact = 1/2
p[3] = oa(A*invfact * p[2]) ~ oa(A/2 * A/1 * P)

invfact = (1/2) / 3 = 1/6
p[4] = oa(A*invfact * p[3]) ~ oa(A/6 * A/2 * A/1 * P)  # should only be A/3 * ...
```

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+
+"""
+    _compute_outer_powers(B::MatrixZonotope, in_powers::Vector{S},
+                               k::Int) where {S<:Union{SparsePolynomialZonotope,
+                                                       AbstractZonotope}}
+
+Apply `B` repeatedly to each element in `in_powers`, approximating:
+
+```math
+B^n * (A^n * P)
```
schillic: I find this confusing. Should `A` and `B` not be swapped?

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,249 @@
...
+function _compute_outer_powers(B::MatrixZonotope, in_powers::Vector{S},
+                               k::Int) where {S<:Union{SparsePolynomialZonotope,
+                                                       AbstractZonotope}}
+    out_powers = similar(in_powers)
+    out_powers[1] = in_powers[1]
+
+    @inbounds for n in 2:(k + 1)
+        term = in_powers[n]
+        for _ in 1:n
+            term = overapproximate(B * term, S)
```
schillic: Should this not be running once shorter (since `n = i+1`)?
```suggestion
        for _ in 1:(n-1)
            term = overapproximate(B * term, S)
```

src/Approximations/overapproximate_expmap.jl
alecarraro: yes , the docsrting is confusing. I will change it

src/Approximations/overapproximate_expmap.jl
alecarraro: you are right, Thanks for all the feedback, I will fix everything in the afternoon

src/Approximations/overapproximate_expmap.jl
alecarraro: the distinction is that in the sum $\sum (ABCD)^i *P$ this function only computes the "innermost" or ("rightmost") product. Maybe there is a better name

src/Approximations/overapproximate_expmap.jl
alecarraro: yes good point

src/Approximations/overapproximate_expmap.jl
schillic: 1. I think it only makes sense if you have a matrix zonotope *product* (which is not mentioned).
2. If you drop "inner" the meaning is preserved (I think) but it avoids the possible confusion.

src/Approximations/overapproximate_expmap.jl
schillic: Okay, when you change the docstring, I would also rename the variables in the code.

src/Approximations/overapproximate_expmap.jl
```diff
@@ -0,0 +1,247 @@
...
+        - `matnorm` -- an upper bound on the norm of the matrix zonotope
+        - `k` -- the order of the Taylor expansion
+
+        ### Output
+
+        A zonotope over-approximating the remainder term of the Taylor expansion.
+        """
+        function taylor_expmap_remainder(Z::AbstractZonotope{N}, matnorm::Real, Int) where {N}
+            n = dim(P)
+            N = eltype(P)
```
schillic:
```suggestion
```
redundant now
EDIT: In fact, it leads to an error...

src/Approximations/overapproximate_expmap.jl
alecarraro: Sure

src/Approximations/overapproximate_expmap.jl
alecarraro: but then would you keep `_compute_outer_powers` the same or I should also change that

schillic: I would literally just remove this one word in the docstring. The function names can stay.
I read this as a technical term and then did not know that that means.

schillic:
> I tried out a simple example where the SPZ is just a zonotope and the MZ is just a matrix. The overapproximation is not enormously bigger, and it gets quite tight when raising κ to 5. This gives some confidence in the correctness.

With the new code changes, the sets are actually identical already for κ = 2.

I was first surprised because I expected the approximation to get coarser with the fixes. But then I realized that this happens because you now handle this case exactly. So the example above is not a good benchmark anymore :sweat_smile:

schillic:
@alecarraro Is this ready to be merged?

alecarraro:
> @alecarraro Is this ready to be merged?

Yes, I implemented all the feedback from your reviews. I tested the reachability algorithm to test a real use case of these methods and it works .

PR #3998, opened 2025-08-09
Optimize `remove_redundant_generators`

alecarraro:
The current implementation of `remove_redundant_generators` performs a quadratic search which is very slow for large number of generators.

I implemented an optimized version which performs the following 3 steps to remove the generators from a matrix of size $$m \times n$$:

1. Normalize each generator to unit length $$\mathcal{O}(m \times n)$$
2. Sort all the generators. I think the complexity is something in the order of $$\mathcal{O}(n \log(n))$$
3. Traverse the sorted generators matrix  and combine equal generators
So the complexity is dominated by the sorting and it should be lower than the complexity of the current nested loop.

With a zonotope of dimension 100, 10 generators of which 9 linearly dependent, the performance is as follows:
 optimized: 5.122 μs (65 allocations: 28.02 KiB) 
 current: 3.150 μs (58 allocations: 19.00 KiB)  <==
 
Dimension 100, 100 generators of which 90 linearly dependent:
 optimized: 64.236 μs (621 allocations: 320.98 KiB) <==
 current: 60.563 μs (1392 allocations: 594.12 KiB) 

Dimension 100, 1000 generators of which 900 linearly dependent:
 optimized: 691.195 μs (6282 allocations: 3.25 MiB) <==
 current: 3.976 ms (103814 allocations: 45.62 MiB) 

For the most use cases we expect the number of generators to be at least as high or much higher than the dimension of the zonotope. In this scenario the newer version performs many times better both in terms of time and allocations.

EDIT: I had to change some of the tests and the doctests since the new version produces a generator matrix with a different ordering of the columns ( and also sign). However the results are equivalent

### DIFF
--- a/src/Interfaces/AbstractZonotope.jl
+++ b/src/Interfaces/AbstractZonotope.jl
@@ -859,9 +859,9 @@ function remove_redundant_generators(G::AbstractMatrix)
     cur_sum = norms[this]  # accumulated norm of identical cols
 
     # traverse the sorted matrix and compare cols
-    @inbounds for i in 2:p
+    @inbounds for j in 2:p
         prev = this
-        this = ord[i]
+        this = ord[j]
 
         if _isapprox(view(Gnorm, :, prev), view(Gnorm, :, this))
             cur_sum += norms[this]


alecarraro:
Code to test the performance
```julia
#generate random zonotope and add 10 duplicate generators for each column of G
Z = rand(Zonotope, dim=100, num_generators= 100) 
G = reduce(hcat, (repeat(genmat(Z)[:, i], 1, 10) for i in 1:size(genmat(Z), 2)))

#shuffle columns
perm = randperm(size(genmat(Z),2)) 
Base.permutecols!(G, randperm!(perm))

#multiply each column by a random factor
D = Diagonal(randn(size(G, 1))) #
G = D * G

Z = Zonotope(center(Z), G)
@btime remove_redundant_generators(Z)
println(ngens(remove_redundant_generators(Z))) 
```

alecarraro:
I have to fix some edge cases

src/Sets/Zonotope/remove_redundant_generators.jl
```diff
@@ -5,9 +5,9 @@
 
 ### Algorithm
 
-For each generator ``g_j`` that has not been checked yet, we find all other
-generators that are linearly dependent with ``g_j``.
-Then we combine those generators into a single generator.
+This method normalizes each generator to have unit norm, sorts the normalized 
+generators lexicographicaly, and then traverses them once to merge consecutive 
+generators that are approximately collinear by summing them into a single generator.
```
schillic: We can keep the old explanation because it is still valid in one case.

```suggestion
This method normalizes each generator to have unit norm, sorts the normalized 
generators lexicographicaly, and then traverses them once to merge consecutive 
generators that are approximately collinear by summing them into a single generator.

For `Rational` inputs, we use a less efficient implementation: For each generator ``g_j`` that has
not been checked yet, we find all other generators that are linearly dependent with ``g_j``.
Then we combine those generators into a single generator.

```

test/Approximations/overapproximate.jl
```diff
@@ -531,8 +531,7 @@ for N in [Float64]
         local p₂ = x₃ - x₁
         local vTM = [TaylorModels.TaylorModelN(pi, I, x0, D) for pi in [p₁, p₂]]
         Z1 = overapproximate(vTM, Zonotope)
-        @test center(Z1) == N[3, -2.5]
-        @test Matrix(genmat(Z1)) == N[1 -1 0; -1 0 0.5]
+        @test isequivalent(Z1, Zonotope(N[3, -2.5], N[0.0 1.0 1.0; 0.5 -1.0 -0.0]))
```
schillic:
```suggestion
        @test isequivalent(Z1, Zonotope(N[3, -2.5], N[0 1 1; 0.5 -1 0])) 
```

src/Sets/Zonotope/remove_redundant_generators.jl
```diff
@@ -5,9 +5,9 @@
...
 
 For one-dimensional zonotopes we use a more efficient implementation where we
 just take the absolute sum of all generators.
```
schillic, line 17:
```suggestion
just take the absolute sum of all generators.

### Notes

The implementation may change the order of the non-removed generators.
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -819,6 +818,62 @@ function remove_redundant_generators(G::AbstractMatrix)
...
+
+        # flip the vector if the first element is negative
+        if !isnothing(idx) && unit_col[idx] < 0
+            Gnorm[:, j] = -unit_col
+        else
+            Gnorm[:, j] = unit_col
+        end
+    end
+
+    # sort column in ascedning order
```
schillic:
```suggestion
    # sort column in ascending order
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -819,6 +818,62 @@ function remove_redundant_generators(G::AbstractMatrix)
...
+    ord = sortperm(eachcol(Gnorm))
+    merged = Vector{Vector{eltype(G)}}()
+    
+    cur_dir = view(Gnorm, :, ord[1])
+    cur_sum = norms[ord[1]] # accumulated norm of identical cols
+
+     # traverse the sorted matrix and compare cols
```
schillic:
```suggestion
    # traverse the sorted matrix and compare cols
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -819,6 +818,62 @@ function remove_redundant_generators(G::AbstractMatrix)
...
+    @inbounds for i in 2:p
+        prev = ord[i-1]
+        this = ord[i]
+
+        if isapprox(view(Gnorm, :, prev), view(Gnorm, :, this))
```
schillic: Not 100% sure this works, but if it does, this is preferred (because then the user can tweak the precision).
```suggestion
        if _isapprox(view(Gnorm, :, prev), view(Gnorm, :, this))
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -819,6 +818,62 @@ function remove_redundant_generators(G::AbstractMatrix)
...
+            Gnorm[:, j] = unit_col
+        end
+    end
+
+    # sort column in ascedning order
+    ord = sortperm(eachcol(Gnorm))
+    merged = Vector{Vector{eltype(G)}}()
+    
+    cur_dir = view(Gnorm, :, ord[1])
+    cur_sum = norms[ord[1]] # accumulated norm of identical cols
```
schillic:
```suggestion
    this = ord[1]
    cur_dir = view(Gnorm, :, this)
    cur_sum = norms[this]  # accumulated norm of identical cols
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -819,6 +818,62 @@ function remove_redundant_generators(G::AbstractMatrix)
...
+
+     # traverse the sorted matrix and compare cols
+    @inbounds for i in 2:p
+        prev = ord[i-1]
+        this = ord[i]
```
schillic:
```suggestion
        prev = this
        this = ord[i]
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -819,6 +818,62 @@ function remove_redundant_generators(G::AbstractMatrix)
...
+            cur_sum += norms[this]
+        else
+            push!(merged, cur_dir .* cur_sum) # merge multiples
+
+            cur_dir = view(Gnorm, :, this)
+            cur_sum = norms[this]
+        end
+    end
+
+    push!(merged, cur_dir .* cur_sum)
```
schillic, line 876: This should not always be applied. Only if the `then` branch was executed last. You can move it into the `then` branch as follows:
```julia
if _isapprox(view(Gnorm, :, prev), view(Gnorm, :, this))
    cur_sum += norms[this]
    if i == p
        push!(merged, cur_dir .* cur_sum)
    end
else
    ...
end
```

src/Interfaces/AbstractZonotope.jl
```diff
@@ -819,6 +818,62 @@ function remove_redundant_generators(G::AbstractMatrix)
...
+
+    # sort column in ascedning order
+    ord = sortperm(eachcol(Gnorm))
+    merged = Vector{Vector{eltype(G)}}()
+    
+    cur_dir = view(Gnorm, :, ord[1])
+    cur_sum = norms[ord[1]] # accumulated norm of identical cols
+
+     # traverse the sorted matrix and compare cols
+    @inbounds for i in 2:p
```
schillic: I would use `j` for column iterators.

src/Interfaces/AbstractZonotope.jl
```diff
@@ -819,6 +818,62 @@ function remove_redundant_generators(G::AbstractMatrix)
...
+        # flip the vector if the first element is negative
+        if !isnothing(idx) && unit_col[idx] < 0
+            Gnorm[:, j] = -unit_col
+        else
+            Gnorm[:, j] = unit_col
+        end
+    end
+
+    # sort column in ascedning order
+    ord = sortperm(eachcol(Gnorm))
```
schillic: Here is a suggestion:
```suggestion
    cols = eachcol(Gnorm)
    if VERSION < v"1.9"
        cols = collect(cols)
    end
    ord = sortperm(cols)
```

src/Interfaces/AbstractZonotope.jl
alecarraro: why?
At `i==p` we have 2 cases:
1.  the `then` branch is executed. In this case we clearly merge the last "x" vectors. 
2.  the `else` branch is executed. In this case, we merge the vectors up to the `p-1` -th while the  `p`-th vector is NOT redundant. So we take it as the new `cur_dir` and it also must be merged.

I think it is needed in both cases
Changing it make the tests fail.

schillic: You are right. It always `push`es the previous entries in the `else` branch.

Could you add two tests for these cases (column `p` enters `then` resp. `else` branch) [here](https://github.com/JuliaReach/LazySets.jl/blob/2cbb65aeff80adca71709963d076b4efdc9c3910/test/Interfaces/AbstractZonotope.jl#L9)? Maybe you need a separate loop to exclude `Rational`s (`for N in @tN([Float64, Float32])`).

alecarraro: Sure, done.

PR #3999, opened 2025-08-09
Add operations on matrix zonotopes

alecarraro:
In this PR i add:
- a `convert` method from matrix zonotopes to zonotope
- a definition of `order` for matrix zonotopes
- `remove_redundant_generators` and `reduce_order` for matrix zonotopes
- `minkowski_sum` for matrix zonotopes
- apply `remove_redundant_generators` to the overapproximation of the matrix zonotope product


docs/src/lib/matrixsets/MatrixZonotope.md
```diff
@@ -49,6 +50,13 @@ CurrentModule = LazySets.MatrixZonotopeModule
 Undocumented implementations:
 * [`size`](@ref size(::AbstractMatrixZonotope))
 
+```@meta
+CurrentModule = LazySets
+```
+* [`remove_redundant_generators`](@ref remove_redundant_generators(::AbstractZonotope))
+* [`minkowski_sum`](@ref minkowski_sum(::LazySet, LazySet))
+* [`reduce_order`](@ref reduce_order(::AbstractZonotope, ::Real, ::AbstractReductionMethod=GIR05()))
```
schillic: I think these functions should be documented for matrix sets.

src/Convert/Zonotope.jl
```diff
@@ -249,11 +249,35 @@ Convert a Cartesian product array of zonotopic sets to a zonotope.
 
 ### Output
 
-A zonotope with sparse matrix representation.
+A zonotope.
+
+### Algorithm
+
+The method vectorizes the center and the generators of the matrix zonotope.
```
schillic: I think you edited the wrong docstring here. This is an unrelated method.

src/Convert/Zonotope.jl
```diff
@@ -249,11 +249,35 @@ Convert a Cartesian product array of zonotopic sets to a zonotope.
...
+### Input
+
+- `Zonotope` -- target type
+- `MZ`       -- a matrix zonotope
+
+### Output
+
+A zonotope with sparse matrix representation.
+"""
+function convert(::Type{Zonotope}, MZ::MatrixZonotope)
```
schillic: I think the name `convert` is not good. An MZ lives in $$n \times n$$ dimensions and a zonotope lives in $$n$$ dimensions. While you could technically `convert` back, you need to remember the original dimensions for that, so it loses information. You also cannot `convert` a `Matrix` to a `Vector`.
Could `vec` be a good name? For that, you would need to `import Base.vec`. Then this code should be moved to the `MatrixSets` module.

src/MatrixSets/MatrixZonotopeModule.jl
```diff
@@ -4,10 +4,11 @@ using Reexport
 
 using Random: AbstractRNG, GLOBAL_RNG
 using ReachabilityBase.Distribution: reseed!
-using ..LazySets: Zonotope
+using ..LazySets: Zonotope, genmat
```
schillic:
```suggestion
using ..LazySets: Zonotope, AbstractReductionMethod, genmat
```

src/MatrixSets/MatrixZonotopeModule.jl
```diff
@@ -4,10 +4,11 @@ using Reexport
...
 
-@reexport import ..API: center, scale, scale!, rand, norm, linear_map
-@reexport import ..LazySets: generators, ngens
+@reexport import ..API: center, scale, scale!, rand, norm, linear_map, minkowski_sum
+@reexport import ..LazySets: generators, ngens, order, remove_redundant_generators,
+                             reduce_order, AbstractReductionMethod
```
schillic:
```suggestion
                             reduce_order
```

src/MatrixSets/reduce_order.jl
```diff
@@ -0,0 +1,15 @@
...
+    if order(A) <= r
+        return A
+    end
+
+    Z = convert(Zonotope, A)
+    Zred = reduce_order(Z, r, method)
+
+    # reshape to matrix zonotope
+    gens = [Matrix(reshape(col, size(A))) for col in eachcol(genmat(Zred))]
+    return MatrixZonotope(center(A), gens)
```
schillic: It could be a good idea to wrap these two lines in a function (because it is a pattern that repeats also in `remove_redundant_generators`). That would also make the code more readable. Maybe `_mat` could be a name (in analogy to `vec`); for now this can be an internal function. The function would take as arguments an `AbstractZonotope`, `dims`, and an optional argument for the `center` (and otherwise reshape the zonotope's center).

test/MatrixSets/MatrixZonotope.jl
```diff
@@ -120,6 +120,23 @@ for N in @tN([Float64, Float32, Rational{Int}])
     @test expMZ.M == A_mz
     @test size(expMZ) == size(A_mz)
     @test_throws AssertionError MatrixZonotopeExp(B_mz)
+
+    # conversion
+    c = N[1 0; 0 3]
+    gens = [N[1 -1; 0 2]]
+    MZ = MatrixZonotope(c, gens)
+    Z = convert(Zonotope, MZ)
+    @test Z == Zonotope(N[1, 0, 0, 3], hcat(N[1, 0, -1, 2]))
```
schillic, line 130:
```suggestion
    @test Z == Zonotope(c, hcat(N[1, 0, -1, 2]))
```

test/MatrixSets/MatrixZonotope.jl
```diff
@@ -120,6 +120,23 @@ for N in @tN([Float64, Float32, Rational{Int}])
...
+
+    # remove redundant generators 
+    MZ2 = MatrixZonotope(c, [N[1 4; 0 -2], N[-1 1; 0 -1], N[1 -1; 0 1]])
+    MZred = remove_redundant_generators(MZ2)
+    @test ngens(MZred) == 2
```
schillic, line 140: Now we have a computational way to compare matrix zonotopes!
(But it may generally be too expensive. Here it takes 0.002232 seconds.)
```suggestion
    MZ2 = MatrixZonotope(c, [N[1 4; 0 -2], N[-1 1; 0 -1], N[1 -1; 0 1]])
    MZred = remove_redundant_generators(MZ2)
    @test ngens(MZred) == 2
    @static if isdefined(@__MODULE__, :Polyhedra)
        @test isequivalent(vec(MZ2), vec(MZred))
    end
```

schillic (commented):
There are some things to do here. Most importantly, I suggest to rename `convert`.

test/MatrixSets/MatrixZonotope.jl
```diff
@@ -120,6 +120,23 @@ for N in @tN([Float64, Float32, Rational{Int}])
     @test expMZ.M == A_mz
     @test size(expMZ) == size(A_mz)
     @test_throws AssertionError MatrixZonotopeExp(B_mz)
+
```
schillic, line 124: Please also add tests for `order` and `reduce_order`.

src/MatrixSets/reduce_order.jl
alecarraro: Good point, In fact I used it also in other PRs and it was annoying me to duplicate the code ahah. 
What do you think about one of these 2 names for the conversion from and to a matrix zonotope?

1.  vectorize / matrixize 
2. flatten / lift

If you prefer to stick to functions from base then i guess `vec`/`reshape` could work

schillic: I do not have a strong opinion, and I think none of the options fully convinces me. Since I think these should be internal functions for now, at the end of the day, the names should not matter too much.

- In general, I think it is good to reuse `Base` functions if their meaning is preserved / they are generalized (here: `Base.vec` vectorizes a matrix (set) to a vector (set)) because it reduces the amount of function names you need to remember and thus makes code more readable.
- I like `vectorize`, but is that not exactly what `Base.vec` stands for?
- `matrixize` sounds clumsy but would fit together with `vectorize`.
- `flatten`/`lift` are good, but we already have `flatten` for something else (flattening nested array sets) and I think that would be confusing.
- `reshape` actually fits. It has the small downside that one may expect that the target dimension can be different from a matrix, and I do not think we plan to have general array sets.
- A single function in the spirit of `convert` but with a different name, like `transform`, could also work. But it requires an extra target argument and I guess here we really know the target type.

cc @mforets

test/MatrixSets/MatrixZonotope.jl
alecarraro: `isequivalent` for zonotopes is extremely expensive since it checks the inclusion of the zonotopes. It does not even work with few hundred of generators due to overflow of the binomial.
Extracting the logic from the `remove_redundant_generators` function I can make a dispatch of `isequivalent` for zonotopes which is very cheap, and then a dispatch for matrix zonotopes

schillic: I know, but for this example we can add the test. And I was only talking about this test :)

> Extracting the logic from the `remove_redundant_generators` function I can make a dispatch of `isequivalent` for zonotopes which is very cheap

Hm, is it that easy? Maybe you are right...
I suggest to open an issue for now.

src/MatrixSets/reduce_order.jl
alecarraro: > I  like vectorize, but is that not exactly what `Base.vec` stands for?

Yes, but there is no function called `vectorize` in base, so i think it is not confusing.
Since you have no strong preference, until further discussion, I would choose `vectorize` and `matrixize`. I agree that they are quite ugly but they are very unambiguous in my opinion.

schillic: Sure, as long as we do not export them, we can go with these names without further discussion :+1:

test/MatrixSets/MatrixZonotope.jl
alecarraro: `c` is a matrix here

src/Convert/Zonotope.jl
alecarraro: you are right, thanks

alecarraro:
I applied all the suggestions and made `minkowski_sum` more robus. I do not understand why the tests fail here on GitHub. I ran all the tests locally and they all pass

src/MatrixSets/minkowski_sum.jl
```diff
@@ -0,0 +1,26 @@
...
+"""
+function minkowski_sum(A::MatrixZonotope, B::MatrixZonotope)
+    @assert size(A) == size(B) "cannot sum a matrix zonotope of size $(size(A)) " *
+                               "with one of size $(size(B)) "
+
+    c = center(A) + center(B)
+
+    gensA = [convert(typeof(c), Ai) for Ai in generators(A)]
+    gensB = [convert(typeof(c), Bi) for Bi in generators(B)]
+    gens = vcat(gensA, gensB)
```
schillic: If the types agree, the `convert` is not needed, and this is the most common case. Maybe you can guard this by `if eltype(A) == eltype(B)` and in that case let `gensA = generators(A)` etc.

src/MatrixSets/reduce_order.jl
```diff
@@ -0,0 +1,43 @@
+"""
+    reduce_order(MZ::MatrixZonotope, r::Real,
+                 [method]::AbstractReductionMethod=GIR05())
+
+Reduce the order of a zonotopic set by overapproximating with a zonotope with
+fewer generators.
```
schillic:
```suggestion
Reduce the order of a matrix zonotope by overapproximating with a matrix zonotope with
fewer generators.
```

src/MatrixSets/reduce_order.jl
```diff
@@ -0,0 +1,43 @@
...
+    if order(MZ) <= r
+        return MZ
+    end
+
+    Z = vectorize(MZ)
+    Zred = reduce_order(Z, r, method)
+
+    # reshape to matrix zonotope
+    dim = size(MZ)
+    return matrixize(Zred, dim)
```
schillic:
```suggestion
    return matrixize(Zred, size(MZ))
```

src/MatrixSets/remove_redundant_generators.jl
```diff
@@ -0,0 +1,31 @@
...
+# Extended help
+
+    remove_redundant_generators(Z::Zonotope)
+"""
+function remove_redundant_generators(MZ::MatrixZonotope)
+    Z = vectorize(MZ)
+    Zred = remove_redundant_generators(Z)
+
+    dim = size(MZ)
+    return matrixize(Zred, dim)
```
schillic:
```suggestion
    return matrixize(Zred, size(MZ))
```

src/MatrixSets/reshape.jl
```diff
@@ -0,0 +1,40 @@
...
+function vectorize(MZ::MatrixZonotope)
+    c = vec(center(MZ))
+    G = hcat([vec(Ai) for Ai in generators(MZ)]...)
+    return Zonotope(c, G)
+end
+
+"""
+    matrixize(Z::Zonotope, dims::Tuple{Int,Int})
+
+Reshape a zonotope zonotope to transform it into a matrix zonotope.
```
schillic:
```suggestion
Reshape a zonotope to transform it into a matrix zonotope.
```

src/MatrixSets/reshape.jl
```diff
@@ -0,0 +1,40 @@
...
+
+### Input
+
+- `Z` -- a zonotope
```
schillic:
```suggestion
- `Z`    -- a zonotope
- `dims` -- target dimensions of the matrix zonotope
```

test/MatrixSets/MatrixZonotope.jl
```diff
@@ -120,9 +121,41 @@ for N in @tN([Float64, Float32, Rational{Int}])
...
 
 for N in @tN([Float64, Float32])
-    MZ = rand(MatrixZonotope; N=N)
+    MZ = rand(MatrixZonotope, N=N, dim=(2,2), num_generators=8)
     @test MZ isa MatrixZonotope{N}
+
+    # reduce order
+    MZred = reduce_order(MZ, 1)
+    @test order(MZred) ≤ 1
+    @test ngens(MZred) == 4
```
schillic: This is not needed.
```suggestion
```

test/MatrixSets/MatrixZonotope.jl
```diff
@@ -120,9 +121,41 @@ for N in @tN([Float64, Float32, Rational{Int}])
...
+
+    # order
+    @test order(MZ) == 1//4 && order(MZ) == order(Z)
+
+    # remove redundant generators 
+    MZ2 = MatrixZonotope(c, [N[1 4; 0 -2], N[-1 1; 0 -1], N[1 -1; 0 1]])
+    MZred = remove_redundant_generators(MZ2)
+    @static if isdefined(@__MODULE__, :Polyhedra)
+        @test isequivalent(vectorize(MZ2), vectorize(MZred))
+    end
```
schillic: You are not testing that the function has any effect. I think you should also check that the generators have been reduced.

src/MatrixSets/minkowski_sum.jl
alecarraro: > Maybe you can guard this by `if eltype(A) == eltype(B)`

 Then I would rather change the signature to 
```julia
function minkowski_sum(A::MatrixZonotope{N, MN}, B::MatrixZonotope{N, MN}) where {N, MN}
```
But i thought it would be more general to accept two different concrete numeric type and let Julia promote them (if needed).

test/MatrixSets/MatrixZonotope.jl
alecarraro: oh, you are right

schillic:
> I do not understand why the tests fail here on GitHub. I ran all the tests locally and they all pass

I also just tried and could not reproduce this either. Maybe print `N` before this line so we know at least for which `N` if fails.

src/MatrixSets/minkowski_sum.jl
schillic: Yes, but if you want to use dispatch, you need to create two methods. `eltype` can be inlined by the compiler, so it should be equally efficient. If you think writing two methods is more elegant, go for it.

test/MatrixSets/MatrixZonotope.jl
```diff
@@ -120,9 +121,41 @@ for N in @tN([Float64, Float32, Rational{Int}])
...
+
+    # order
+    @test order(MZ) == 1//4 && order(MZ) == order(Z)
+
+    # remove redundant generators 
+    MZ2 = MatrixZonotope(c, [N[1 4; 0 -2], N[-1 1; 0 -1], N[1 -1; 0 1]])
+    MZred = remove_redundant_generators(MZ2)
+    @test ngens(MZred) == 2
+    @static if isdefined(@__MODULE__, :Polyhedra)
+        @test isequivalent(vectorize(MZ2), vectorize(MZred)) == true
```
schillic: This is redundant (`X == true` is the same as `X`).
```suggestion
        @test isequivalent(vectorize(MZ2), vectorize(MZred))
```

schillic: Let's do some remote debugging:
```suggestion
        @show vectorize(MZ2)
        @show vectorize(MZred)
        @test isequivalent(vectorize(MZ2), vectorize(MZred)) == true
```

schillic (approved):
The changes look good, but we need to fix this strange error...

test/MatrixSets/MatrixZonotope.jl
alecarraro: I think it has something to do with `matrixize` and `vectorize`. 
Before I had those tests on line 128
```julia
MZ = MatrixZonotope(c, gens)
    Z = vectorize(MZ)
    @test Z == Zonotope(N[1, 0, 0, 3], hcat(N[1, 0, -1, 2]))
    MZ2 = matrixize(Z, (2, 2))
    @test MZ == MZ2
```
But they failed even though `MZ` and `MZ2` look identical when printed out.
I had to change to 
```julia
MZ = MatrixZonotope(c, gens)
    Z = vectorize(MZ)
    @test Z == Zonotope(N[1, 0, 0, 3], hcat(N[1, 0, -1, 2]))
    MZ2 = matrixize(Z, (2, 2))
    @test isapprox(center(MZ), center(MZ2))
    @test isapprox(generators(MZ), generators(MZ2))
```
for it to work

alecarraro: but it is very strange since `vectorize` and `matrixize` should just reshape the arrays

schillic:
```julia
vectorize(MZ2) = Zonotope{Float32, Vector{Float32}, Matrix{Float32}}(Float32[1.0, 0.0, 0.0, 3.0], Float32[1.0 -1.0 1.0; 0.0 0.0 0.0; 4.0 1.0 -1.0; -2.0 -1.0 1.0])
vectorize(MZred) = Zonotope{Float32, Vector{Float32}, Matrix{Float32}}(Float32[1.0, 0.0, 0.0, 3.0], Float32[1.0 1.9999999; 0.0 -0.0; 4.0 -1.9999999; -2.0 1.9999999])
```
So the result seems to be rounded. Very interesting because it does not happen for me, and I use the same Julia version.
So I guess the `isequivalent` implementation is not robust to minor imprecision. In that case you should just check something simpler and add a note so we do not forget why it is like that (also to the other place you mentioned, please).

schillic: Oh, and remove the `@show`s again.

schillic: Actually, for `Float64` you can keep the test!

test/MatrixSets/MatrixZonotope.jl
```diff
@@ -120,9 +121,44 @@ for N in @tN([Float64, Float32, Rational{Int}])
...
+
+    # vectorize and matrixize
+    c = N[1 0; 0 3]
+    gens = [N[1 -1; 0 2]]
+    MZ = MatrixZonotope(c, gens)
+    Z = vectorize(MZ)
+    @test Z == Zonotope(N[1, 0, 0, 3], hcat(N[1, 0, -1, 2]))
+    MZ2 = matrixize(Z, (2, 2))
+    @test isapprox(center(MZ), center(MZ2))
+    @test isapprox(generators(MZ), generators(MZ2))
```
schillic:
```suggestion
    # floating-point errors for Float32 (at least on the CI platform)
    # and `isequivalent` is not robust to minor imprecisions 
    if N == Float32
        @test isapprox(center(MZ), center(MZ2))
        @test isapprox(generators(MZ), generators(MZ2))
    else
        @static if isdefined(@__MODULE__, :Polyhedra)
            @test isequivalent(vectorize(MZ), vectorize(MZ2))
        end
    end
```

test/MatrixSets/MatrixZonotope.jl
```diff
@@ -120,9 +121,44 @@ for N in @tN([Float64, Float32, Rational{Int}])
...
+
+    # order
+    @test order(MZ) == 1//4 && order(MZ) == order(Z)
+
+    # remove redundant generators 
+    MZ2 = MatrixZonotope(c, [N[1 4; 0 -2], N[-1 1; 0 -1], N[1 -1; 0 1]])
+    MZred = remove_redundant_generators(MZ2)
+    @test ngens(MZred) == 2
+    # `remove_redundant_generators`` introduces floating point errors for Float32
```
schillic:
```suggestion
    # `remove_redundant_generators` introduces floating-point errors for Float32
    # (at least on the CI platform)
```

test/MatrixSets/MatrixZonotope.jl
alecarraro: > also to the other place you mentioned, please

The other place is fine actually. I just realized that `==` is not overloaded to check equality of `MatrizZonotope`s, and there is no rounding. 

I added the comments for the `isequivalent` case instead

test/MatrixSets/MatrixZonotope.jl
```diff
@@ -120,9 +121,53 @@ for N in @tN([Float64, Float32, Rational{Int}])
...
+    # floating-point errors for Float32 (at least on the CI platform)
+    # and `isequivalent` is not robust to minor imprecisions 
+    if N == Float32
+        @test isapprox(center(MZ), center(MZ2))
+        @test isapprox(generators(MZ), generators(MZ2))
+    else
+        @static if isdefined(@__MODULE__, :Polyhedra)
+            @test isequivalent(vectorize(MZ), vectorize(MZ2))
+        end
+    end
```
schillic:
```suggestion
    @test isequivalent(vectorize(MZ), vectorize(MZ2))
```

test/MatrixSets/MatrixZonotope.jl
schillic: Ah, but then you should not have accepted my proposal. I'll make a counter-proposal now.

schillic (approved):

PR #4001, opened 2025-08-10
Add `remove_small_generators` for Zonotope and matrix zonotope

alecarraro:


docs/src/lib/matrixsets/MatrixZonotope.md
```diff
@@ -45,10 +45,13 @@ Undocumented implementations:
...
 ```
-
-Undocumented implementations:
 * [`size`](@ref size(::AbstractMatrixZonotope))
 
+```@meta
+CurrentModule = LazySets
+```
+* [`remove_small_generators](@ref remove_small_generators(::AbstractZonotope, ::Real; ::Real))
+
```
schillic, line 54: You need to switch back.
````suggestion

```@meta
CurrentModule = LazySets.MatrixZonotopeModule
```

````

File: src/Interfaces/AbstractZonotope.jl [hunk 2]
```diff
@@ -884,6 +885,31 @@ function _remove_redundant_generators_1d(G::AbstractMatrix)
     return hcat(g)
 end
 
+"""
+    remove_small_generators(Z::AbstractZonotope, tol::Real=1e-4; p::Real=Inf)
```
schillic, line 889:
```suggestion
    remove_small_generators(Z::AbstractZonotope, [tol]::Real=1e-4, [p]::Real=Inf)
```
At least that is the current signature.

src/Interfaces/AbstractZonotope.jl
```diff
@@ -884,6 +885,31 @@ function _remove_redundant_generators_1d(G::AbstractMatrix)
...
+    
+Remove generators with norm smaller than a given tolerance.
+
+# Input
+
+- `Z` -- an `AbstractZonotope`
+- `tol` -- (default: `1e-4`) threshold below which a generator column is considered negligible
```
schillic, line 896: I think it would be better to use `_ztol(eltype(Z))`.

src/MatrixSets/remove_small_generators.jl
```diff
@@ -0,0 +1,8 @@
+function remove_small_generators(A::MatrixZonotope, tol::Real=1e-4; p::Real=Inf)
+    Z = convert(Zonotope, A)
+    Zred = remove_small_generators(Z, tol; p=p)
+
+    # reshape to matrix zonotope
+    gens = [Matrix(reshape(col, size(A))) for col in eachcol(genmat(Zred))]
+    return MatrixZonotope(center(A), gens)
```
schillic, line 7: As discussed in the other PR, this could be refactored.

src/MatrixSets/remove_small_generators.jl
```diff
@@ -0,0 +1,8 @@
+function remove_small_generators(A::MatrixZonotope, tol::Real=1e-4; p::Real=Inf)
```
schillic, line 1: I think this method requires a separate docstring. The reason is that the `p`-norm used is not clear. In this case, it is not the [induced matrix norm](https://en.wikipedia.org/wiki/Matrix_norm#Matrix_norms_induced_by_vector_p-norms) but the [entry-wise matrix norm](https://en.wikipedia.org/wiki/Matrix_norm#%22Entry-wise%22_matrix_norms).

test/Interfaces/AbstractZonotope.jl
```diff
@@ -22,4 +22,7 @@ for N in @tN([Float64, Float32])
 
     H = N[0 0 1; 1 1 0] # equal after sorting
     @test remove_redundant_generators(H) == N[0 1; 2 0]
+
+    G2 = N[0 1 1e-4 2; 0 0 0 0]
+    @test remove_small_generators(G2) == N[1 2; 0 0]
```
schillic, line 27: There should also be a test for a `Zonotope` (in the `Sets/Zonotope.jl` file).

src/Interfaces/AbstractZonotope.jl
```diff
@@ -884,6 +885,31 @@ function _remove_redundant_generators_1d(G::AbstractMatrix)
...
+- `p`  -- (optional, default: `Inf`) norm
+
+# Output
+
+A zonotope containing only the generators whose ``p``-norm is at least `tol`
+"""
+function remove_small_generators(Z::AbstractZonotope, tol::Real=1e-4, p::Real=Inf)
```
schillic, line 903: Maybe it is better to just implement this method for `Zonotope`. I know that `remove_redundant_generators` is implemented for `AbstractZonotope`, but I am not so happy with that anymore.

schillic (commented):
I do not expect that we will want to use this function in the application because it is unsound (as mentioned before).

src/MatrixSets/remove_small_generators.jl
alecarraro: Good catch, thanks

schillic:
> But if we do not plan to use it, is there even a point to implement it?

No, that is why I was surprised.

>  I thought in the end we agreed that it is, at least for now, a necessary (temporary) solution

No, we said that we have to get an overapproximation by imitating what `reduce_order` does.

alecarraro:
> No, we said that we have to get an overapproximation by imitating what `reduce_order` does.
Ok, for now I will convert it to a draft PR, and I will just use it in my branch to experiment with the reachability algorithm. 
In the next call we can discuss the correct implementation. Since it is not clear to me yet how this should work

schillic:
Yes, I promised I would add a method, but I did not have time yet.

alecarraro:
> Yes, I promised I would add a method, but I did not have time yet.

Don't worry... You already do a lot to review all these PR.

PR #4096, opened 2025-11-28
Fix conversion between  type `IntervalMatrix` and `MatrixZonotope`

alecarraro:


test/MatrixSets/MatrixZonotope.jl
```diff
@@ -128,15 +128,33 @@ for N in @tN([Float64, Float32, Rational{Int}])
     @static if isdefined(@__MODULE__, :IntervalMatrices)
         using IntervalMatrices: IntervalMatrix
         IM = IntervalMatrix([interval(-N(1.1), -N(0.9)) interval(-N(4.1), -N(3.9));
-                             interval(N(3.9), N(4.1)) interval(-N(1.1), -N(0.9))])
+                            interval( N(3.9),  N(4.1)) interval(-N(1.1), -N(0.9))])
```
schillic:
```suggestion
                             interval(N(3.9), N(4.1)) interval(-N(1.1), -N(0.9))])
```

test/MatrixSets/MatrixZonotope.jl
```diff
@@ -128,15 +128,33 @@ for N in @tN([Float64, Float32, Rational{Int}])
...
+            @test isapprox(
+                center(MZ),
+                N[-1.0 -4.0;
+                4.0 -1.0]
+            )
+
+            expected = [N[0.1 0.0; 0.0 0.0],
+                N[0.0 0.0; 0.1 0.0], 
+                N[0.0 0.1; 0.0 0.0], 
+                N[0.0 0.0; 0.0 0.1]]
```
schillic:
```suggestion
                N[-1 -4;
                  4 -1]
            )

            expected = [N[1//10 0; 0 0],
                N[0 0; 1//10 0], 
                N[0 1//10; 0 0], 
                N[0 0; 0 1//10]]
```

src/MatrixSets/convert.jl
```diff
@@ -28,13 +28,31 @@ function load_intervalmatrices_conversion()
...
+            center = mid(IM)
+
+            # radius matrix gives the half-widths
+            halfIM = radius(IM)
+
+            gens = Vector{Matrix{T}}(undef, n * m)
+            idx = 1
+            @inbounds for j in 1:m
+                for i in 1:n
+                    G = zeros(T, n, m)
```
schillic: Usually `m` is the number of rows and `n` is the number of columns.
And usually we use `N` instead of `T`.

```suggestion
        function Base.convert(::Type{MatrixZonotope}, IM::IntervalMatrix{N}) where {N}
            # TODO change to sparse matrix -> problem: all the generators of a MZ should have the same type
            m, n = size(IM)

            # center must be the midpoint matrix
            center = mid(IM)

            # radius matrix gives the half-widths
            halfIM = radius(IM)

            gens = Vector{Matrix{N}}(undef, m * n)
            idx = 1
            @inbounds for j in 1:n
                for i in 1:m
                    G = zeros(N, m, n)
```

src/MatrixSets/convert.jl
```diff
@@ -28,13 +28,31 @@ function load_intervalmatrices_conversion()
...
+                    G[i, j] = halfIM[i, j]
```
schillic: If `iszero(halfIM[i, j])`, you should not store the generator.

test/MatrixSets/MatrixZonotope.jl
```diff
@@ -128,15 +128,33 @@ for N in @tN([Float64, Float32, Rational{Int}])
...
+            @test isapprox(generators(MZ), expected)
         end
```
schillic, line 158: Please add a test with a zero interval, e.g., `IntervalMatrix([interval(0) interval(N(-1), N(2)); interval(0), interval(0)])`.
```suggestion
        end
```

src/MatrixSets/convert.jl
alecarraro: good point, i will update it

src/MatrixSets/convert.jl
```diff
@@ -31,25 +31,19 @@ function load_intervalmatrices_conversion()
...
-                for i in 1:m
-                    G = zeros(N, m, n)
-                    G[i, j] = halfIM[i, j]
-                    gens[idx] = G
-                    idx += 1
-                end
+            gens = Vector{Matrix{N}}()
+            sizehint!(gens, m*n)
+
+            @inbounds for (I, val) in pairs(halfIM)
```
schillic: Out of curiosity: is there a reason to use `pairs` instead of `enumerate`?

alecarraro: `enumerate`  returns an iterator `(i, val)` where `i` is a positive integer. `pairs` is more general and by default it returns the `IndexCartesian` if an array is multidimensional.
If I had written `enumerate`  I would have had to write something sloppy like:
```julia
col = div(i, m)
row = mod(i, n)
G[row, col] = ...
```
to compute the cartesian index myself
https://docs.julialang.org/en/v1/base/iterators/#Base.Iterators.enumerate
https://docs.julialang.org/en/v1/base/collections/#Base.pairs

schillic: Julia allows to access multi-dimensional arrays with linear indices.

```julia
julia> M = Matrix(undef, 2, 3)
2×3 Matrix{Any}:
 #undef  #undef  #undef
 #undef  #undef  #undef

julia> v = 1:6
1:6

julia> for (i, val) in enumerate(v)
           M[i] = val
       end

julia> M
2×3 Matrix{Any}:
 1  3  5
 2  4  6

julia> for (i, val) in pairs(v)
           M[i] = val
       end

julia> M
2×3 Matrix{Any}:
 1  3  5
 2  4  6
```

When I `@time` this, I get 19 allocations for `enumerate` and 25 allocations for `pairs` (because it needs to create those `IndexCartesian` objects).

alecarraro: Good to know! I will change it then

PR #4174, opened 2026-01-17
Refactor matrix zonotopes overapproximations

alecarraro:
Currently, the overapproximation methods targeting `MatrixZonotope` are distributed across multiple files. I believe a better design would be to introduce a new file, `overapproximate_matrixzonotope`, dedicated specifically to the overapproximation of sets with target `MatrixZonotope`. This included both the overapproximation of  `MatrixZonotopeProduct` and the `MatrixZonotopeExponential` types.

The idea is to clearly separate two different concerns:

1. Overapproximating the matrix zonotope exponential.   That is, approximating `exp(A)` itself as a new `MatrixZonotope`.

2. Overapproximating the action of the exponential on a set.  That is, approximating `exp(A) * X` as a new set of type `typeof(X)`.

Under this refactoring the new `overapproximate_matrixzonotope` file will handle the first concern, while the existing `overapproximate_expmap` will contain logic related to the second concern.

This change represents the first step (of two) toward a larger refactor. In the next step,  I will introduce a more flexible implementation of the overapproximation of `exp(A) * X`. The new design will split this overapproximation in more general intermediate blocks, which can be reused to compute integrals of the form $\int e^{At} X$ in `ReachabilotyAnalysis.jl`

### DIFF
--- a/src/Approximations/overapproximate_matrixzonotope.jl
+++ b/src/Approximations/overapproximate_matrixzonotope.jl
@@ -85,7 +85,6 @@ function load_intervalmatrices_overapproximation_matrixzonotope()
     end
 end
 
-
 """
     overapproximate(MZP::MatrixZonotopeProduct{N,S},
                          ::Type{<:MatrixZonotope}) where {N,S<:AbstractMatrix{N}}
--- a/test/Approximations/overapproximate_expmap.jl
+++ b/test/Approximations/overapproximate_expmap.jl
@@ -50,5 +50,14 @@ for N in @tN([Float32, Float64])
         sampler = LazySets.PolynomialZonotopeSampler()
         pts = sample(Pex, 10; sampler=sampler)
         @test all(p ∈ Zex for p in pts)
+
+        @static if isdefined(@__MODULE__, :ExponentialUtilities) || isdefined(@__MODULE__, :Expokit)
+            # matrix
+            mzexp = SparseMatrixExp(sparse(M))
+            em = ExponentialMap(mzexp, P)
+            MPex = overapproximate(em, SparsePolynomialZonotope, 2)
+            MZex = overapproximate(MPex, Zonotope)
+            @test MZex ⊆ Zex
+        end
     end
 end
--- a/test/Approximations/overapproximate_matrixzonotope.jl
+++ b/test/Approximations/overapproximate_matrixzonotope.jl
@@ -65,19 +65,18 @@ for N in @tN([Float32, Float64])
         res2 = overapproximate(expA, MatrixZonotope, 20)
         @test isapprox(center(res), center(res2))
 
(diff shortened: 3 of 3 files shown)


src/Approximations/overapproximate_matrixzonotope.jl
```diff
@@ -0,0 +1,146 @@
...
+            res = minkowski_sum(W, E)
+
+            return remove_redundant_generators(res; tol=tol)
+        end
+
+        # TODO avoid `convert`
+        function LazySets.minkowski_sum(MZ::MatrixZonotope, IM::IntervalMatrix)
+            return minkowski_sum(MZ, convert(MatrixZonotope, IM))
+        end
+
```
schillic, line 64: This method should not be in that file. I know it was in a similar file before, but we could fix that now.

If it is too much work, please open an issue at least.

test/Approximations/overapproximate_expmap.jl
```diff
@@ -51,43 +50,5 @@ for N in @tN([Float32, Float64])
...
-        @test isapprox(center(res), center(res2))
-
-        @static if isdefined(@__MODULE__, :ExponentialUtilities) || isdefined(@__MODULE__, :Expokit)
-            # matrix
-            mzexp = SparseMatrixExp(sparse(M))
-            em = ExponentialMap(mzexp, P)
-            MPex = overapproximate(em, SparsePolynomialZonotope, 2)
-            MZex = overapproximate(MPex, Zonotope)
-            @test MZex ⊆ Zex
-        end
```
schillic, line 78: This test should not have been removed. It is unrelated to `MatrixZonotope`.

test/Approximations/overapproximate_matrixzonotope.jl
```diff
@@ -0,0 +1,96 @@
...
+        MPex = overapproximate(em, SparsePolynomialZonotope, 2)
+        MZex = overapproximate(MPex, Zonotope)
+
+        Z = overapproximate(convert(SimpleSparsePolynomialZonotope, P), Zonotope)
+        em_z = ExponentialMap(mzexp, Z)
+        Zex = overapproximate(em_z, Zonotope, 5)
+
+        @test MZex ⊆ Zex
+
+        end
```
schillic: The indentation should be fixed.

src/Approximations/overapproximate_matrixzonotope.jl
alecarraro: I can do it, But I'd rather do it today or tomorrow after merging this PR

test/Approximations/overapproximate_expmap.jl
alecarraro: You are right

PR #4187, opened 2026-02-14
Refactor `overapproximate_expmap`

alecarraro:
This is the follow up PR to the refactoring started in [#4174](https://github.com/JuliaReach/LazySets.jl/pull/4174). Additions:

- Renamed the internal routines to reflect a more general purpose. The previous naming was tied to the exponential map, while the implementation actually evaluates a truncated operator power series. The same operations can be reused to construct other operators (for example the integral $\int_{t_0}^{t_f} exp(At)Xdt$, required for computation in `ReachabilityAnalysis.jl`
-  Replaced the "outer-power" expansion with a Horner-style evaluation for the factored operators. Instead of explicitly propagating each power separately, the coefficients are computed once and the remaining factors are folded using Horner’s method, which improves both the computational complexity and numerical stability. The result is guaranteed to be the equivalent since the `overapproximation(A * X, S)` of the linear map `A*X` distributes over the `exact_sum` operation. In particular , denote by $\bigcirc$ the `exact_sum` and let $\phi$ denote the overapproximation operator. We have that 
$$\phi_A (X \bigcirc Y) = \phi_A (X )  \bigcirc \phi_A ( Y)$$
 If needed I can explain more in details. Empirically this is confirmed by the tests

The number of calls to `overapproximate` reduces from $\mathcal{O}(k^2m)$ to \mathcal{O}(km)$, where $k$ is the taylor order and $m$ is the number of factors in the exponent of exponential map `exp(A*B*C...)*X` and

## File rename
I also changed the name of the file from `overapproximate_expmap` to `overapproximate_expmap_mz` since other approximations of the exponential map for "scalar" matrices can be found in other parts of the libraries. However I look for suggestions for a better name

### DIFF
--- a/src/Approximations/overapproximate_expmap_mz.jl
+++ b/src/Approximations/overapproximate_expmap_mz.jl
@@ -87,10 +87,10 @@ end
                          k::Int=2;
                          matnorm::Union{Real,Nothing}=nothing) where {N, S,
                                                                       P<:Union{SparsePolynomialZonotope,
-                                                                               AbstractZonotope},
+                                                                               Zonotope},
                                                                       MAT<:AbstractMatrixZonotope{N}}
 
-Overapproximate the exponential map of a set `X` of type `SparsePolynomialZonotope` or `AbstractZonotope`
+Overapproximate the exponential map of a set `X` of type `SparsePolynomialZonotope` or `Zonotope`
 through a composition of matrix zonotopes, following Proposition 3 of [HuangLBS2025](@citet).
 
 ### Input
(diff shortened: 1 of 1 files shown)


test/Approximations/overapproximate_expmap_mz.jl
```diff
@@ -1,4 +1,4 @@
-using LazySets, Test, SparseArrays
+using LazySets, Test
```
schillic, line 1: Please keep this. There is a call to `sparse` in this file.

alecarraro:
> This is a large diff to review. I presume that this is mainly a file and function rename. Could you merge the commits into one? Git should then be able to detect this and show the small diff. Sometimes the Git diff is not smart. In that case you can make a first commit where you only rename the file and then apply the edits in a second commit. Then I can look only at the second commit.

Ok, I can do it. But before doing that, can we agree on a name for the file? Because if we just keep `overapproximate_expmap` with no mention of `MatrixZonotope` then the PR boils down to just changing the signature/name of two functions and improving the logic in one 

EDIT: I think `overapproximate_expmap` could also work, it is just a bit ambiguous since basically it only pertains the exponential map of a matrix zonotope

schillic:
Sure, feel free to close.

PR #4188, opened 2026-02-16
Add `scale!` for `SPZ`

alecarraro:
This PR introduces an allocation free version of `scale`, which is very useful in reachabilty analysis where this function is called a lot of times

### DIFF
--- a/test/Sets/SparsePolynomialZonotope.jl
+++ b/test/Sets/SparsePolynomialZonotope.jl
@@ -144,8 +144,8 @@ for N in @tN([Float64, Float32, Rational{Int}])
     @test expmat(PZscaled) == expmat(PZ)
     @test PZscaled == linear_map(α * Matrix{N}(I, 2, 2), PZ)
 
-    PZscaled_inp = scale!(α, PZ)
-    @test PZscaled_inp === PZ
+    scale!(α, PZ)
+    @test PZ == PZscaled
 
     # remove_redundant_generators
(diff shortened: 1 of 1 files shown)


src/Sets/SparsePolynomialZonotope/scale.jl
```diff
@@ -1,7 +1,10 @@
...
+    return P
+end
+
 function scale(α::Real, P::SparsePolynomialZonotope)
-    return SparsePolynomialZonotope(α * center(P),
-                                    α * genmat_dep(P),
-                                    α * genmat_indep(P),
-                                    expmat(P),
-                                    indexvector(P))
+    return scale!(α, copy(P))
```
schillic, line 9: This changes the behavior: now the `expmat` and `indexvector` are also copied, while before they were shared. I am fine with this change but wanted to note it in case this comes up in the future.

test/Sets/SparsePolynomialZonotope.jl
```diff
@@ -144,6 +144,9 @@ for N in @tN([Float64, Float32, Rational{Int}])
     @test expmat(PZscaled) == expmat(PZ)
     @test PZscaled == linear_map(α * Matrix{N}(I, 2, 2), PZ)
 
+    PZscaled_inp = scale!(α, PZ)
+    @test PZscaled_inp === PZ
```
schillic: I think testing for the right return value is not critical. More importantly, this test does not check for the correct modification.

```suggestion
    scale!(α, PZ)
    @test PZ == PZscaled
```

PR #4190, opened 2026-02-20
Refactor `overapproximate_expmap` to use horner's method 

alecarraro:
The goal of this PR is renamed the methods to reflect their generality, without explicitely tying them only to the approximation of `exp(A)*S`. Also the evaluation of the polynomial in `_compute_outer_powers` is replaced by a more efficient method `_horner_series` ( the same idea is also used in `IntervalMatrices` to evaluate exponential of interval matrices)
- Rename the function `_compute_inner_powers` as `_compute_series_coefficient`, and rename some of its internal variable  for readability. I removed the docstring since it was too specific for the helper function
- The `_compute_outer_powers` function was replaced by a function to evaluate polynomials using horner's method. The results are identical, but it improves computational efficiency for a large number of factors
- The function `taylor_expmap_truncation` was renamed `_truncated_operator_series` (logic is the same)
- The function `taylor_expmap_remainder` was renamed  `_operator_series_remainder`

The rest is basically identical


src/Approximations/overapproximate_expmap.jl
```diff
@@ -1,252 +1,134 @@
...
-                               k::Int) where {S<:Union{SparsePolynomialZonotope,
-                                                       AbstractZonotope}}
-    powers = Vector{S}(undef, k + 1)
-    powers[1] = P
+    Bc = copy(B)
 
     @inbounds for i in 1:k
-        term = overapproximate(scale(1 / i, B) * powers[i], S)
-        powers[i + 1] = term
+        term = overapproximate(scale!(1 / i, Bc) * term, S)
```
schillic, line 11: The change from `scale(1 / i, B)` to `scale!(1 / i, Bc)` is not equivalent. Now you get:

1. `1 / 1 * Bc`
2. `1 / 2 * 1 / 1 * Bc`
3. `1 / 3 * 1 / 2 * 1 / 1 * Bc`

Previously you got:

1. `1 / 1 * Bc`
2. `1 / 2 * Bc`
3. `1 / 3 * Bc`

So there should be a difference for `k > 2`. Was this intended?

src/Approximations/overapproximate_expmap.jl
```diff
@@ -1,252 +1,134 @@
...
         using .IntervalMatrices: IntervalMatrix
-        """
-            taylor_expmap_remainder(Z::AbstractZonotope{N}, matnorm::Real, Int) where {N}
-
-        Overapproximate the Lagrange remainder term of the k-th order truncated Taylor expansion
-        of the exponential map of a matrix zonotope applied to a zonotopic set.
 
-        ### Input
+        function _operator_series_remainder(Z::AbstractZonotope{N},
+                                            _matnorm::Real,
```
schillic, line 55: Is there a good reason to name an argument with a leading `_`? We usually use this for internal functions.

src/Approximations/overapproximate_expmap.jl
```diff
@@ -1,252 +1,134 @@
...
-    out_powers = similar(in_powers)
-    out_powers[1] = in_powers[1]
-
-    @inbounds for i in 2:(k + 1)
-        term = in_powers[i]
-        for _ in 1:(i - 1)
-            term = overapproximate(A * term, S)
-        end
-        out_powers[i] = term
+@inline function _horner_series(A::MatrixZonotope,
```
schillic, line 17: I have a hard time understanding how this implementation relates to the previous one. There is an `exact_sum` whereas there was no sum before. Also, the output order seems to be different. Another reason to add documentation what this method is supposed to do.

src/Approximations/overapproximate_expmap.jl
```diff
@@ -1,252 +1,134 @@
...
-    # inner powers on the last factor
-    last_factor = factors(MZP)[end]
-    terms = _compute_inner_powers(last_factor, P, k)
-
-    # propagate outwards through the preceding factors
-    for A in reverse(view(factors(MZP), 1:(nfactors(MZP) - 1)))
-        terms = _compute_outer_powers(A, terms, k)
+    # fold the previous factors with Horner's method
+    for A in reverse(view(facts, 1:(nfactors(MZP) - 1)))
+        coeffs = [_horner_series(A, coeffs)]
```
schillic, line 44: This is very confusing. After the first iteration, `coeffs` is a vector of length 1. But then, calling `_horner_series` essentially does nothing.

schillic (commented):
I think some of the old docstrings were useful. Now there is really no explanation at all, and I doubt that we will remember what `_truncated_operator_series` does in a year. It does not need to end up as a docstring, but maybe you can keep some information as line comments:
```julia
# k-th order truncated Taylor expansion of the exponential map of a matrix zonotope
# \\boxplus_{i=0}^k \\frac{\\mathcal{A}^i}{i!} X
```
Similarly for some of the other helper functions.

PR #4347, opened 2026-07-14
Sampler for UnionSetArray

giacomobais:
Sampling from a UnionSetArray used the rejection sampler, and when it failed it fell back to the random walk sampler, which assumed a convex sets and sampled incorrectly. The new sampler samples from each set in the UnionSetArray individually. To avoid imbalanced sampling, the number of samples each set gets is weighted according to its volume, approximated by the box approximation of the set. If some sets in the UnionSetArray are overlapping, there may be some overrepresentation in the overlapping regions.


src/Utils/samples.jl
```diff
@@ -725,3 +726,46 @@ function load_Distributions_sample()
...
+"""
+struct UnionSetArraySampler <: AbstractSampler
+end
+
+function sample!(D::Vector{VN}, S::UnionSetArray, sampler::UnionSetArraySampler;
+                    rng::AbstractRNG=GLOBAL_RNG,
+                    seed::Union{Int,Nothing}=nothing) where {N,VN<:AbstractVector{N}}
+    num_samples = length(D)
+    # estimate size of each set in the union to make sampling more uniform
+    size_sets = [prod(high(box_approximation(set)) .- low(box_approximation(set))) for set in S]
```
schillic: This does not work if one of the sets is unbounded. I think this feature should either be dropped or done after checking `isbounded`.

src/Utils/samples.jl
```diff
@@ -725,3 +726,46 @@ function load_Distributions_sample()
...
+
+Type used for sampling from a union of sets.
+
+### Notes
+
+The sampling individually samples from each set in the union and tries to make the sampling more uniform by estimating the size of each set in the union through a box approximation.
+Overlaps between sets may result in overrepresentation in the regions of overlap.
+
+"""
+struct UnionSetArraySampler <: AbstractSampler
```
schillic: You need to add `UnionSetArraySampler` to docs/src/lib/utils.md.

schillic: Also, can we call it `UnionSampler`? It is shorter and we can later also extend it to the binary `UnionSet`.

schillic (commented):
Generally looks good, thanks!

src/Utils/samples.jl
```diff
@@ -728,34 +728,56 @@ function load_Distributions_sample()
...
+        # normalize to get weights
+        total_size = sum(size_sets)
+        weights = [size / total_size for size in size_sets]
+        # calculate each share
+        shares = [floor(Int, w * num_samples) for w in weights]
+        # we rounded down so make sure to sum up to num_samples
+        leftover = num_samples - sum(shares)
+        for i in 1:leftover
+            shares[i] += 1
+        end
```
schillic: This part is literally the same, so it could be pulled out.

schillic (approved):

schillic (approved):

PR #4368, opened 2026-08-22
Relax overapproximate condition

mforets:
Relax a condition which was otherwise failing for me locally:

```
using LazySets, LazySets.Approximations: OctDirections
X = Ball1(zeros(Float64, 2), 1.0)
Y = remove_zero_generators(overapproximate(X, Zonotope, OctDirections{Float64}(2)))
expected = Zonotope(Float64[0, 0], [0.5 0.5; 0.5 -0.5])

Y == expected   # false — center is [-5.07e-18, -5.07e-18], not exactly [0,0]
Y ≈  expected   # true  — robust to LP solver's floating-point noise
```


schillic (approved):
I am fine with this, but I cannot reproduce - I get the exact result. Maybe some difference with the Julia or package version...

PR #4373, opened 2026-09-07
Fix support vector if empty polygon

### LINKED ISSUE #3297: Support vector of empty HPolygon returns a value
```julia
julia> H = rand(HPolygon)
HPolygon{Float64, Vector{Float64}}(HalfSpace{Float64, Vector{Float64}}[HalfSpace{Float64, Vector{Float64}}
([-0.5141890373325597, 1.3756952703693164], -1.6693690811412345), HalfSpace{Float64, Vector{Float64}}
([-0.2103110083196491, -0.10916173981133498], 0.15288914577433016), HalfSpace{Float64, Vector{Float64}}
([-0.12400282059206358, -0.39972504077185755], 0.5778234315855568), HalfSpace{Float64, Vector{Float64}}
([0.5318388889703878, -0.6570085018357579], 1.264270898534744), HalfSpace{Float64, Vector{Float64}}
([0.3166639772738846, -0.20979998795036603], 0.5629649078533848)])

julia> addconstraint!(H, HalfSpace([0., -1.], 0.5));

julia> isempty(H)
true

julia> @time σ([1., 0], H)
2-element Vector{Float64}:
  1.2943597768037873
 -0.729684469493759
```

mforets:
Closes #3297


src/Sets/HPolygon/support_vector.jl
```diff
@@ -32,6 +32,12 @@ search by default.
     linear_search = true  # TODO fix binary search and then remove this line
     n = length(P.constraints)
     @assert n > 0 "the polygon has no constraints"
+    # `an_element` intersects the first two (angularly sorted) constraints, which is a
+    # genuine vertex iff `P` is non-empty; cheaper than `isempty`, which enumerates all
+    # vertices
+    if n >= 2 && an_element(P) ∉ P
```
schillic, line 38:
```suggestion
    # `an_element` intersects the first two (angularly sorted) constraints, which is a
    # genuine vertex iff `P` is non-empty; cheaper than `isempty`, which enumerates all
    # vertices
    if n >= 2 && isempty(P)
```

I do not follow this reasoning. But if `an_element(P) ∉ P` is indeed equivalent to `isempty` but cheaper, it should become the new implementation of `isempty` instead. But that should be explored in a separate PR.

This fix also seems very expensive (requires an `isempty` check for every support-vector query).

schillic (changes requested):

